PR #5, opened 2015-03-01
add react-in-jsx-scope rule

glenjamin:
Hope this makes sense - I've included tests & docs as well.

See https://github.com/eslint/eslint/issues/1905 for some additional context.

I'm planning to follow up with a variant on `no-unused` that knows about JSX using `React.createElement`.

Sidenote: should there be a test which verifies all rules are exported in `index.js` ?


glenjamin:
Whoops, missed a test case for nested scopes - will add one to bump the coverage back up.

yannickcr:
Thanks for contributing! Looks good to me.

> should there be a test which verifies all rules are exported in index.js ?

There is not, but i should definitely add one.

PR #6, opened 2015-03-01
Add jsx-uses-react rule

glenjamin:
See eslint/eslint#1905 for additional context.

This rule somewhat unusual, in that it modifies the behaviour of a core eslint rule.

I'm not hugely keen on this, but it does work. It is possible for the rule to be broken by changes in eslint core.

The alternative would be to copy-paste the original rule and modify it.
That would require users to disable the core rule and replace it with this one - and then any improvements to the core rule would have to be tracked here and updated.

Of those two choices, I think the first one is the most practical, even if it breaks the usual clean separation of rules.

### DIFF
--- a/docs/rules/jsx-uses-react.md
+++ b/docs/rules/jsx-uses-react.md
@@ -1,0 +1,29 @@
+# Make JSX count towards use of a declared variable (jsx-uses-react)
+
+JSX expands to a call to `React.createElement`, a file which includes `React`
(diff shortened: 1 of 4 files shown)


yannickcr:
Thanks!

This is a tricky case to check but I am agree with your approach to use the `eslintJSXUsed` flag used by the `no-unused-vars` rule.

Like you said it breaks the separation of rules but I can't really see any other way to check this.

PR #17, opened 2015-03-08
Bump eslint version 0.15.0 -> 0.16.0

lencioni:
This new version of eslint contains many improvements, including better
support for ES6 features. The full changelog can be found at:

  https://github.com/eslint/eslint/blob/master/CHANGELOG.md

With this change, I ran into an issue that was causing the tests to
pass. Lots of errors like:

> 10:12  error  "fs" used outside of binding context  block-scoped-var

It appears that this is an open issue with 0.16.0 that can be worked
around by setting `globalReturn` to `false` in the `.eslintrc` file.
More information:

  https://github.com/eslint/eslint/issues/1969

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,6 +3,7 @@
         "node": true
     },
     ecmaFeatures: {
+        globalReturn: false,
         jsx: true
     },
     "globals": {
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "bugs": "https://github.com/yannickcr/eslint-plugin-react/issues",
   "devDependencies": {
     "coveralls": "2.11.2",
-    "eslint": "0.15.0",
+    "eslint": "0.16.0",
     "eslint-tester": "0.6.0",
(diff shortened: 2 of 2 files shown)


ivanoats:
I think we need to wait for https://github.com/eslint/eslint/pull/1951 , https://github.com/eslint/eslint/issues/1911 , and https://github.com/yannickcr/eslint-plugin-react/issues/18 to close , there are quite a few bugs and changes for JSX in ESLint 0.16.x now.

lencioni:
It also looks like the `globalReturn` problem was fixed, so that should probably be removed from this anyway. Want me to push a new version?

yannickcr:
Bumped directly to ESLint 0.17.0 with some fixes for the [new JSX/React handling](http://eslint.org/blog/2015/03/eslint-0.17.0-released/#changes-to-jsx/react-handling).

But thanks for your contribution (and sorry for being so late to respond :cry: ).

PR #20, opened 2015-03-15
Sort rules alphabetically in README.md

lencioni:
Sorting these alphabetically makes it easier to find what you are
looking for, and provides a simple and clear order for new items to be
added to the list.


yannickcr:
Merged, thanks.

PR #22, opened 2015-03-17
Allow this.getState references (not calls) in lifecycle methods.

benmosher:
I think this should not be considered an error/warning by this rule: (my actual code follows)
```	
componentDidMount() {
	this.handleStoreEvent = _.compose(this.setState, getStoreState).bind(this);
	dependencies.add(this.handleStoreEvent);
},
```
While I don't know that this is great code per se, it is essentially binding `setState` to a list of Flux stores via a helper (`dependencies`). This seems to me like a valid use of `setState` within `componentDidMount`.

I refactored both rules pertaining to `setState` to respond to `CallExpression` nodes instead of `MemberExpression`, and it returns if the `callee` is not a `MemberExpression`. The original logic follows thereafter.

I can imagine also potentially making this reduction in strength an option, but I believe the behavior here captures the spirit of the rule, and thus an option should not be necessary. 

Submitted for your consideration. 👍

### DIFF
--- a/lib/rules/no-did-mount-set-state.js
+++ b/lib/rules/no-did-mount-set-state.js
@@ -16,16 +16,20 @@ module.exports = function(context) {
 
   return {
 
-    'MemberExpression': function(node) {
-      if (node.object.type !== 'ThisExpression' || node.property.name !== 'setState') {
+    'CallExpression': function(node) {
+      var callee = node.callee;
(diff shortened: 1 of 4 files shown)


yannickcr:
Looks good to me

And I think this is the way this rule should work, so an option is not necessary.

Thanks for your contribution :)

PR #36, opened 2015-03-29
Fix prop-types to check for computed prop names.

burnnat:
Previously, accessing props dynamically with an expression like `this.props[propName]` would result in a linting error like:
> 'propName' is missing in props validation

However "propName" is simply the name of a variable, not the actual property name. Since there's no easy way to resolve the real property name for dynamic accesses like this, this modification simply ignores such cases.


burnnat:
Looks like the Travis failure is unrelated to the code change - seems to be a connection issue, maybe because of the Github DDOS.

yannickcr:
Merged, thanks.

PR #42, opened 2015-03-30
Handle spread in jsx-sort-props

zertosh:


zertosh:
@yannickcr This is causing the plugin to throw

yannickcr:
Thanks, I totally forgot to test the spread attribute.

zertosh:
I was thinking about this a little more and there is another subtly that comes with the spread attribute. The spread attribute can be in any position, and that affects the ultimate value of the props passed. So does it make sense to enforce alpha order for all the attributes or to reset it after a spread? E.g.:

```jsx
// this would warn but it shouldn't because I might want my
// "onClick" to override the one in "this.props" - but not
// the "onLoad".
<MyComponent
  className="cls"
  onLoad={this.onLoad}
  {...this.props}
  onClick={this.onClick} />
```

What do you think?

yannickcr:
Good catch, I think resetting the alpha order after a spread attribute is a good way to handle this case.

zertosh:
See https://github.com/yannickcr/eslint-plugin-react/pull/47

lucasmotta:
@zertosh @yannickcr I know I'm late to this party, but shouldn't this be an option too?
I'm working on a pull-request related to this rule and I just wanted to ask first why this isn't optional.

For me the spread operator should be at the end, like on this example:
https://goo.gl/z9QEsU

So I would suggest adding a rule like: `resetOnSpread` and set its default value to `true`, so it has backwards compatibility.

What do you guys think? :)

lucasmotta:
And btw, the feature I'm working on is to add an option for shorthand properties at the beginning, like this:
```jsx
<Component x z a="a" b="b" /> // valid
<Component z x a="a" b="b" /> // invalid, should be in alphabetical order
<Component x z b="b" a="a" /> // invalid, should be in alphabetical order
<Component a="a" b="b" x z /> // invalid, shorthand props should be at the start 
```

PR #51, opened 2015-04-05
Enable jsx-uses-vars rule to WARN by default

### LINKED ISSUE #49: Rule 'jsx-uses-vars': why not on by default?
Thanks for this plugin! Very cool.

I'm thinking about adopting this for my default setup across the org, but just wondering why `jsx-uses-vars` wouldn't be turned on by default (maybe to 'warn')? I guess the real question is: what's the use case for _not_ wanting to mark them used?

Also, I'm totally willing to PR a change, but just wanted to get your feedback first. And if the consensus is no rule should be on by default, can we invert the language of the rule to something like `jsx-ignore-var-usage`?

banderson:
From original issue: "just wondering why jsx-uses-vars wouldn't be turned on by default (maybe to 'warn')? I guess the real question is: what's the use case for not wanting to mark them used?"

Fixes #49

### DIFF
--- a/index.js
+++ b/index.js
@@ -27,7 +27,7 @@ module.exports = {
     'no-did-mount-set-state': 0,
     'no-did-update-set-state': 0,
     'react-in-jsx-scope': 0,
-    'jsx-uses-vars': 0,
+    'jsx-uses-vars': 1,
     'jsx-no-undef': 0,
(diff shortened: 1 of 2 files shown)


yannickcr:
Thanks!

PR #53, opened 2015-04-06
fix describing comment for hasSpreadOperator() method

AlexKVal:
It seems it had been copy-pasted and then forgot.


yannickcr:
Thanks!

PR #69, opened 2015-04-17
Only mark component types as used, not prop names.

burnnat:
Previously, all JSX identifiers were marked as "used", even for prop names. This is an issue in cases where component props happened by chance to have the same name as a local variable, like this:

        var App;
        var foo;
        React.render(<App foo="bar"/>);

Here, the "foo" prop prevents the `foo` variable from being marked as unused and ESLint gives no warnings for the `no-unused-vars` rule, even though `foo` is not actually referenced anywhere.

This change updates the `jsx-uses-vars` rule to only mark variables as used if they occur in a `JSXOpeningElement` node (thus excluding prop names, as they exist in `JSXAttribute` nodes instead).

### DIFF
--- a/lib/rules/jsx-uses-vars.js
+++ b/lib/rules/jsx-uses-vars.js
@@ -20,8 +20,8 @@ module.exports = function(context) {
       variableUtil.markVariableAsUsed(context, node.expression.name);
     },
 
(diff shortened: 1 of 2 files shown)


yannickcr:
Nice catch! Thanks!

PR #82, opened 2015-05-26
adding support for setting the jsx pragma value via config

### LINKED ISSUE #80: Allow setting jsx pragma value in eslintrc
I'm using `jsxPragma: 'dom'` with my babel config. (so I don't need to manually add that pragma to each file) However, now my `dom` var appears unused, and there doesn't appear to be any config to set this value in the `.eslintrc`.

I'm not using React, but since the jsx pragma itself is supported for what appears to be other non-React users, I'm hoping this can be done. :)

dominicbarnes:
This fixes #80 

Instead of taking the approach mentioned there, this turned out to be far less complicated, it basically adds a second config option to the `react/jsx-uses-react` rule:

```js
[ 2, 'dom' ]
```

This causes `dom` to be used in place of `React` as the used variable. I've also included a test and updated the docs for the aforementioned rule.

Let me know if there's anything else we need to do to get this merged! :D

### DIFF
--- a/docs/rules/jsx-uses-react.md
+++ b/docs/rules/jsx-uses-react.md
@@ -7,6 +7,7 @@ If you are using the @jsx pragma this rule will mark the designated variable and
 
 This rule has no effect if the `no-unused-vars` rule is not enabled.
 
+
 ## Rule Details
 
 The following patterns are considered warnings:
@@ -38,6 +39,27 @@ var Foo = require('foo');
 
 var Hello = <div>Hello {this.props.name}</div>;
 ```
+
+
+## Rule Options
+
+```js
+...
+"jsx-uses-react": [<enabled>, { "pragma": <string> }]
+...
+```
+
+### `pragma`
+
+As an alternative to specifying the above pragma in each source file, you can specify
+this configuration option:
+
+```js
+var Foo = require('Foo');
+
+var Hello = <div>Hello {this.props.name}</div>;
+```
+
 
 ## When Not To Use It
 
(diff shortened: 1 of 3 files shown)


dominicbarnes:
Although outside the scope of this PR, the rule name `jsx-uses-react` is a confusing name considering that it simultaneously supports the JSX pragma **and** this new config.

I was thinking it could be named `jsx-uses-var` instead, although that could just be confused with the _very_ similarly-named `jsx-uses-vars`. (then there's backward-compatibility to consider)

yannickcr:
Thanks :)

But actually I'd prefer to keep the `[1, {"pragma": "dom"}]` format to be coherent with the other rules. Also, it allows us to add other options in the future without breaking anything.

docs/rules/jsx-uses-react.md
```diff
@@ -39,6 +39,19 @@ var Foo = require('foo');
...
+```js
+[2, 'Foo']
+```
+
+```js
+var Foo = require('Foo');
+
+var Hello = <div>Hello {this.props.name}</div>;
+```
+
```
yannickcr: Great!

Just some little things: can you make a Rule Options section and use the same format as the other rules for the config file example ?

You can see an example here: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-sort-prop-types.md#rule-options

dominicbarnes:
I'm done making those changes! Let me know if there are any other changes to make

yannickcr:
LGTM.

Thanks!

PR #91, opened 2015-05-29
Suggest `PropTypes.node` for children

matthewwithanm:
Just a small doc update to suggest using `node` instead of `any` for `children`. Related to #7.


yannickcr:
Thanks!

PR #115, opened 2015-06-16
check jsxPragma in eslintrc settings

NickStefan:
I'm using a wrapper around React called 'dom', and all of the react plugin linting trips up because it can't find the word 'React'. By setting a jsxPragma for your whole project, in the eslintrc settings, the linter becomes more flexible. 

The settings object gets passed into the utilities already, so I merely added a flexbilble jsxPragma setting in `isComponentDefinition`. It defaults to 'React' if there is no settings.jsxPragma .

- adding a JSX pragma config for every rule in react plugin is not ideal
- allow the settings config in eslintrc to define the jsxPragma.

### DIFF
--- a/lib/util/component.js
+++ b/lib/util/component.js
@@ -41,7 +41,8 @@ function isReactComponent(context, node) {
  * @param {ASTNode} node The AST node being checked.
  * @returns {Boolean} True the node is a component definition, false if not.
  */
-function isComponentDefinition(node) {
+function isComponentDefinition(context, node) {
+  var jsxPragma = context.settings.jsxPragma || 'React';
   var isES6Component = node.type === 'ClassDeclaration';
   var isES5Component = Boolean(
     node.type === 'ObjectExpression' &&
@@ -49,7 +50,7 @@ function isComponentDefinition(node) {
     node.parent.callee &&
     node.parent.callee.object &&
     node.parent.callee.property &&
-    node.parent.callee.object.name === 'React' &&
+    node.parent.callee.object.name === jsxPragma &&
     node.parent.callee.property.name === 'createClass'
   );
   return isES5Component || isES6Component;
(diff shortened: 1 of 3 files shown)


NickStefan:
Im noticing now that in babel, the jsxPragma option takes a string that would look like: 
```
jsxPragma : 'dom.createElement' 
```
However, in this pull request I assumed one would set the eslint config settings as:
```
jsxPragma: 'dom'
```
I gravitated there as it was more direct information in solving the problems specific to me (linter not recognizing my class definitions). However, I can see the value in keeping it consistent. If people would prefer 'dom.createElement' as the jsxPragma setting, then I can definitely edit this PR.

lib/util/component.js
```diff
@@ -41,7 +41,8 @@ function isReactComponent(context, node) {
  * @param {ASTNode} node The AST node being checked.
  * @returns {Boolean} True the node is a component definition, false if not.
  */
-function isComponentDefinition(node) {
+function isComponentDefinition(context, node) {
```
Cellule: Since you've changed the signature of the function, you should update its uses.
- https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/display-name.js#L124
- https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/prop-types.js#L528

NickStefan:
Updated the uses :)

yannickcr:
Thanks, and sorry for the delay.

Since I refactored the `Components` class and since some other rules where relying on a `pragma` option, I have merged your changes in 4cb2e9efe851283925b4dff3f1226e3a551e9d7b and I updated the other rules to use the shared settings.

Now all rules relying on component detection (`display-name`, `prop-types`, etc.) or on the `React` variable (`jsx-uses-react`, `react-in-jsx-scope`, etc.) are using the shared settings.

PR #117, opened 2015-06-17
Allow blocking of '.jsx' extensions with `require()`

scothis:
Created a new rule, 'require-extension', which inspects the required
module ID and warns if the file extension is .jsx.

The particular disallowed file extensions are configurable, defaulting to
just '.jsx'. If you wanted to also disallow '.js', the rule can be
configured as:

    "rules": {
      "react/require-extension": [1, { "extensions": [".js", ".jsx"] }],
    }

Note: this rule does not prevent required files from containing forbidden
extensions, it merely prevents the extension from being included in the
`require()` statement.

This rule is inspired by the Airbnb react style guide:
https://github.com/airbnb/javascript/tree/master/react#naming


scothis:
This is my first eslint rule. Any and all feedback is most welcome.

I intend to continue creating rules to enforce the Airbnb style guide.

yannickcr:
Accepted, thanks!

PR #119, opened 2015-06-18
fix detection of missing displayName and propTypes when ecmaFeatures.jsx is false

rpl:
This PR introduces a small tweak in the ```react/display-name``` and ```react/prop-types``` custom eslint rules which adds to the detected React components any class created using the ```React.createClass``` helper when there isn't any jsx syntax in the render method.

Adding the above components definition object expressions to the component list is enough to detect missing ```displayName``` and ```propTypes``` properties as expected.

### DIFF
--- a/lib/rules/display-name.js
+++ b/lib/rules/display-name.js
@@ -117,6 +117,12 @@ module.exports = function(context) {
         }
         markDisplayNameAsDeclared(node);
       });
+
+      if (componentUtil.isComponentDefinition(node)) {
+        componentList.set(context, node, {
+          isReactComponent: true
+        });
+      }
     },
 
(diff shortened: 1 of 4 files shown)


rpl:
@yannickcr here is a small tweak I'm using to introduce your awesome React-related custom eslint rules on one of the projects I'm working on (which is firebug/rdp-inspector, pull request firebug/rdp-inspector#48) which doesn't use jsx syntaxes in the render method and the React components with missing displayName weren't detected as expected.

This is my first experience on an eslint plugin, let me know if this fix is ok or if it needs any more tweaks.

yannickcr:
Great! Thanks!

PR #123, opened 2015-06-19
`jsx-sort-prop-types` support for keys as string

Cellule:
While investigating another bug, I stumbled onto this, which was easier to fix, so here it is.
Added support for prop type identified by strings in rule `jsx-sort-prop-types`


AlexKVal:
:+1:

PR #124, opened 2015-06-19
Fix `prop-types` desctructuring with properties as string.

### LINKED ISSUE #118: 'prop-types' rule doesn't recognize 'this-format' properties name
I wasn't been able to find out such issue. Sorry if I missed one.

```js
propTypes: {
  ...
  'aria-controls': React.PropTypes.string
},

let {
  ...
  'aria-controls': ariaControls,  // <- error 'undefined' is missing in props validation  react/prop-types
  ...props } = this.props; // eslint-disable-line object-shorthand
```
```
error  'undefined' is missing in props validation  react/prop-types
```

Cellule:
Fixes #118 
I saw that destructuring were not correctly detected when the first key was a string and also those were not handled properly.

### DIFF
--- a/lib/rules/prop-types.js
+++ b/lib/rules/prop-types.js
@@ -164,6 +164,16 @@ module.exports = function(context) {
     return tokens.length && tokens[0].value === '...';
   }
 
+  function getKeyValue(node) {
+    var key = node.key;
+    if (key) {
+      if (key.type === 'Identifier') {
+        return key.name;
+      }
+      return key.value;
+    }
+  }
+
   /**
    * Iterates through a properties node, like a customized forEach.
    * @param {Object[]} properties Array of properties to iterate.
@@ -174,11 +184,10 @@ module.exports = function(context) {
     if (properties.length && typeof fn === 'function') {
       for (var i = 0, j = properties.length; i < j; i++) {
         var node = properties[i];
-        var key = node.key;
-        var keyName = key.type === 'Identifier' ? key.name : key.value;
+        var key = getKeyValue(node);
 
         var value = node.value;
-        fn(keyName, value);
+        fn(key, value);
       }
     }
   }
@@ -331,7 +340,7 @@ module.exports = function(context) {
     } else if (
       node.parent.parent.declarations &&
       node.parent.parent.declarations[0].id.properties &&
-      node.parent.parent.declarations[0].id.properties[0].key.name
+      getKeyValue(node.parent.parent.declarations[0].id.properties[0])
     ) {
       type = 'destructuring';
     }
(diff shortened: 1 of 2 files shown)


tests/lib/rules/prop-types.js
```diff
@@ -415,6 +415,23 @@ eslintTester.addRuleTest('lib/rules/prop-types', {
         classes: true,
         jsx: true
       }
+    }, {
+      code: [
+        'class Hello extends React.Component {',
+        '  render() {',
+        '    var { ',
+        '      propX,',
+        '      "aria-controls": ariaControls, ',
```
AlexKVal: :+1: :smile:

AlexKVal:
Very nice ! :+1: :tada: :rainbow: 
Thank you :cherries:

lib/rules/prop-types.js
```diff
@@ -164,6 +164,16 @@ module.exports = function(context) {
...
   }
 
+  function getKeyValue(node) {
+    var key = node.key;
+    if (key) {
+      if (key.type === 'Identifier') {
+        return key.name;
+      }
+      return key.value;
+    }
```
AlexKVal: why not 
```js
return (key.type === 'Identifier') ? key.name : key.value;
```

here https://github.com/yannickcr/eslint-plugin-react/pull/123/files#diff-6149d4075a6f3125f18ce8c87bb253cbR39 you've done it like that:
```js
return node.key.type === 'Identifier' ? node.key.name : node.key.value;
```

Of course it is just a matter of taste :cherries:

Cellule: The difference is that in `jsx-sort-prop-types` it only checks the propTypes declaration and not its uses.
I was thinking (though I didn't implement it yet) that when checking the uses, you could have the case `const aria = this.props["aria-controls"];`. Right now, this is ignored by the rule because it's a "computed" property. However, I could easily inspect that property, realize it's a literal string value and add the check (there are other implications which is why I haven't done it yet, mostly with case `this.props["some.thing"]` and `name.split(".")` in the rule).
Therefore, I left it with that structure to make it easy to add more checks.
Note that it's possible, but make no sense to do `propType: {["aria-controls"]: React.PropTypes.string};` So they are deliberately ignored in the other rule.

yannickcr: Yep, there is really a lot of different ways to declare/use propTypes, it become hard to check all of them.

yannickcr:
Awesome! Thanks!

lib/rules/prop-types.js
AlexKVal: Thank you for the explanation :cherries: :smiley:

PR #138, opened 2015-07-01
Prevent usage of dangerous JSX properties (no-danger)

scothis:
Dangerous properties in React are those whose behavior is known to be a
common source of application vulnerabilities. The properties names
clearly indicate they are dangerous and should be avoided unless great
care is taken.

See https://facebook.github.io/react/tips/dangerously-set-inner-html.html

```jsx
var Hello = <div dangerouslySetInnerHTML={{ __html: "Hello World" }}></div>;
```

The following patterns are not considered warnings:

```jsx
var Hello = <div>Hello World</div>;
```


tmcw:
@yannickcr any work needed to get this ready to merge?

yannickcr:
@scothis Very useful rule. Thanks!

@tmcw Sorry for the delay, I was pretty busy last days.

PR #147, opened 2015-07-13
Add no-danger to README

lencioni:
This rule was added by 0c132f5559 but was not added to the README, so it
was difficult to discover.


yannickcr:
Thanks!

PR #150, opened 2015-07-14
Changed ignore check to use correct key name

CalebMorris:
Use case of failure:

eslintrc: 
```
"react/prop-types": [ 2, { ignore : [ 'get' ]} ],
```
test.js:
``` javascript
propTypes : {
  user : React.PropTypes.shape({
    id : React.PropTypes.node.isRequired,
  }).isRequired,
},

getInitialState() {
  console.log(this.props.user.get('foo'));
  return {};
}
```

despite `get` being an ignored attribute, it throws an error.


CalebMorris:
Need to fix tests

Issue #148, opened 2015-07-14
react/prop-types thinks built-in methods should be defined in the propTypes when using oneOfType


silvenon:
```js
import React from 'react/addons';

export default React.createClass({
  propTypes: {
    data: React.PropTypes.oneOfType([
      React.PropTypes.object,
      React.PropTypes.array
    ]).isRequired
  },

  render() {
    let content;

    if (this.props.data.length) {
      content = this.props.data.map((pair, i) => (
        // ..
      ));
    }

    return content;
  }
});

```

```
error  'data.length' is missing in props validation  react/prop-types
error  'data.map' is missing in props validation     react/prop-types
```

PR #151, opened 2015-07-14
Changed oneOfType to actually keep the child types

CalebMorris:
Should complete #148 
Failure case:

eslintrc
``` json
"react/prop-types": [ 2 ],
```

sample.jsx
``` javascript
{
  propTypes : {
    testType : React.PropTypes.oneOfType([
      React.PropTypes.string,
      React.PropTypes.array,
    ]),
  },

  getInitialState() {
    console.log(this.props.testType.length);
    return {};
  }
}
```

Fails on `length`


CalebMorris:
Need to add tests

PR #153, opened 2015-07-15
Fix/type of child types

CalebMorris:
Should complete #148 
Failure case:

eslintrc
``` json
"react/prop-types": [ 2 ],
```

sample.jsx
``` javascript
{
  propTypes : {
    testType : React.PropTypes.oneOfType([
      React.PropTypes.string,
      React.PropTypes.array,
    ]),
  },

  getInitialState() {
    console.log(this.props.testType.length);
    return {};
  }
}
```

Fails on `length`


yannickcr:
Great! Thanks!

Issue #145, opened 2015-07-10
PropType validation errors when using custom validators


HurricaneJames:
This just started happening when I updated recently, I think I was on v2.2 or 2.3 previously. I believe the issue is related to #134. 

````
var React = require('react')
  , ImmutablePropTypes = require('react-immutable-proptypes');

var MyComponent = React.createClass({
  displayName: 'MyComponent',
  propTypes: {
    immutableObject: ImmutablePropTypes.shape({
      xyz: React.PropTypes.string
    })
  },
  render: function() {
    return <div className={this.props.immutableObject.get('xyz')} />;
  }
});

module.exports = MyComponent;
````

eslint-react-plugin will complain that `'immutableObject.get' is missing in props validation (react/prop-types)`.

As you can see, I'm using a custom validator for immutable.js classes, where the default React.PropTypes.shape does not work. I know a lot of people use the [react-immutable-proptypes library](https://www.npmjs.com/package/react-immutable-proptypes), so this is going to affect quite a few people if they are also trying to lint with the latest eslint.

I like where object shape validation is going, but it should probably have a separate flag to enable/disable that feature or limit it to props defined with React.PropTypes.shape. I do not wish to turn off validator linting completely, but at the same time it is completely broken when using a custom validator.

yannickcr:
I don't have any experience with immutable.js or react-immutable-proptypes, but I can see that the problem here come from the fact we only allow the usage of the properties defined in the `shape` object.

Do you have some other cases where the propTypes validation conflict with ImmutablePropTypes ?

I'm thinking to allow all methods on `Map.prototype` on the `shape` objects, do you think it can solve your problem ?

HurricaneJames:
That will fix the immediate problem, but I think it is still problematic for anybody who is writing custom validators. Also, it seems somehow wrong to impose functionality from a separate third party library on a React linter. I assume the problem is because it is named `shape` and there is no way to know if it is `React.PropTypes.shape`, `MyCustomValidator.shape`, or even the badly spelled `Raect.PropTypes.shape`?

This was not a problem up through v2.5.0. What changed between v2.5 and 2.6 that this became a problem?

yannickcr:
We support nested prop types validation since `2.6.0`, so we make more checks if you define a `shape`, `arrayOf`, `objectOf`or `oneOfType` prop type.

> I assume the problem is because it is named `shape`

Exactly.

[...]

PR #155, opened 2015-07-16
Fix/ignore child

CalebMorris:
This fixes a couple cases where I was having problems with shape and ignore.

Use case of failure:

eslintrc: 
```
"react/prop-types": [ 2, { ignore : [ 'get' ]} ],
```
test.js:
``` javascript
propTypes : {
  user : React.PropTypes.shape({
    id : React.PropTypes.node.isRequired,
  }).isRequired,
},

getInitialState() {
  console.log(this.props.user.get('foo'));
  return {};
}
```

despite `get` being an ignored attribute, it used to throw an error.

### DIFF
--- a/lib/rules/prop-types.js
+++ b/lib/rules/prop-types.js
@@ -93,14 +93,11 @@ module.exports = function(context) {
    * @returns {Boolean} True if the prop is declared, false if not.
    */
   function _isDeclaredInComponent(declaredPropTypes, keyList) {
-    var isDeclaredInComponentFlag = true;
-    var isFinalKey = false;
+    var isFinalKeyAndAcceptedProperty = true;
     for (var i = 0, j = keyList.length; i < j; i++) {
       var key = keyList[i];
-      isFinalKey = (i === j - 1);
       if (isIgnored(key)) {
-        isDeclaredInComponentFlag = true;
-        break;
+        return true;
       }
       var propType = (
         // Check if this key is declared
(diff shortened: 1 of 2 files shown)


CalebMorris:
Something specific to notice for this PR is that I change the expected output of 2 tests because of changes I made to cut off children of already thrown keys.

I/E
If `foo` is not in props before it would throw for `foo` and `foo.bar`, but now it will only throw for `foo`.

If you don't like this behavior change or thing it shouldn't be there, let me know and I will adjust this PR.

yannickcr:
Hi,

I'm not sure to want `ignore : [ 'foo' ]}` to ignore all `foo` properties on all objects. This could lead to a lot of ignored properties even if we only want to ignore a particular property on a particular object.

I think it should be more safe to specify exactly where we want to ignore the property, maybe with a configuration like this:

```json
"react/prop-types": [2, { 
  "ignore": {
    "user": ["get"]
  }
}]
```

(Also for your particular problem on `shape` and if you're using [react-immutable-proptypes](https://www.npmjs.com/package/react-immutable-proptypes), see #145)

CalebMorris:
That seems like a good alternative to blanket ignoring, although I would want to be able to mix the configuration with the current method. Mainly for the case like the `children` prop that I ignore.

Maybe use the `_` key or some other reservation to specify a global ignored key with support for the exclusive form if they don't need the specific. Do you have any opposition to this?
```
"react/prop-types": [2, { 
  "ignore": {
    "_": ["children"],
    "user": ["get"]
  }
}]
```

As to the discussion in #145 it doesn't look like there is a PR up for that, do you know if anyone has started one?

PR #156, opened 2015-07-16
Added `allowMultiline` property to the `jsx-curly-spacing` rule

mathieumg:
What do you think about this? I had hoped it would be like that by default when I first read about that rule being introduced! Which prompts my next question: should `allowMultiline` be on by default? I currently put it off by default and I'm ignoring it when the option is set to `"always"` as to not break current users (new warnings/errors) that do:

```js
<Hello name={
  firstname
} />;
```

using

```json
"jsx-curly-spacing": [2, "always"]
```

If it were on by default, I could leverage the option in all cases, meaning users would have a way to enforce:

```js
// Valid
<Hello name={ firstname } />;

// Invalid
<Hello name={
  firstname
} />;
```

with

```json
"jsx-curly-spacing": [2, "always", {"allowMultiline": false}]
```

However, current users relying on

```js
<Hello name={
  firstname
} />;
```

to warn with the current defaults, would not be warned anymore.

### DIFF
--- a/lib/rules/jsx-curly-spacing.js
+++ b/lib/rules/jsx-curly-spacing.js
@@ -10,19 +10,52 @@
 
 module.exports = function(context) {
   var spaced = context.options[0] === 'always';
+  var multiline = context.options[1] ? context.options[1].allowMultiline : true;
 
   // --------------------------------------------------------------------------
   // Helpers
   // --------------------------------------------------------------------------
 
   /**
-   * Determines whether two adjacent tokens are have whitespace between them.
+   * Determines whether two adjacent tokens have a newline between them.
+   * @param {Object} left - The left token object.
+   * @param {Object} right - The right token object.
+   * @returns {boolean} Whether or not there is a newline between the tokens.
+   */
+  function isMultiline(left, right) {
+    return left.loc.start.line !== right.loc.start.line;
+  }
+
+  /**
+   * Determines whether two adjacent tokens have whitespace between them.
    * @param {Object} left - The left token object.
    * @param {Object} right - The right token object.
    * @returns {boolean} Whether or not there is space between the tokens.
    */
   function isSpaced(left, right) {
     return left.range[1] < right.range[0];
+  }
+
+  /**
+  * Reports that there shouldn't be a newline after the first token
+  * @param {ASTNode} node - The node to report in the event of an error.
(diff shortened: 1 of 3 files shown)


yannickcr:
Thanks for this!

I think it's ok to make a small break on this rule to make it more flexible.

mathieumg:
So, which break is it? :stuck_out_tongue:  Default to `true`, but rule also considered with `"always"` if set to `false`? (Last example I showed)

If so, I'll update the PR!

moll:
:+1: This would then match http://eslint.org/docs/rules/object-curly-spacing.html which allows multilines in the `never` setting.

yannickcr:
> So, which break is it? :stuck_out_tongue: Default to true, but rule also considered with "always" if set to false? (Last example I showed)

Yep, that break ;)

lib/rules/jsx-curly-spacing.js
```diff
@@ -79,16 +112,28 @@ module.exports = function(context) {
    * @returns {void}
    */
   function validateBraceSpacing(node, first, second, penultimate, last) {
```
mathieumg: Refactored this part a bit because it was getting hard to follow! I don't know what your stance is on early returns to avoid deep nesting, but I can change this to an else statement if that is what you prefer.

lib/rules/jsx-curly-spacing.js
```diff
@@ -10,13 +10,24 @@
 
 module.exports = function(context) {
   var spaced = context.options[0] === 'always';
+  var multiline = context.options[1] ? context.options[1].allowMultiline : true;
```
mathieumg: This is a little fragile and relies on the fact that it's the only option and that no one would pass `{}` as the second parameter, I can change it to leverage `hasOwnProperty` if you want.

yannickcr: Should be ok since `allowMultiline` is the only option authorized by the rule schema.

lib/rules/jsx-curly-spacing.js
yannickcr: LGTM

lib/rules/jsx-curly-spacing.js
mathieumg: Yeah didn't think about the schema, good!

PR #158, opened 2015-07-17
Add rule to support ReactDOM (React 0.14 beta)

donabrams:
In React 0.14, JSX will use two packages, React and ReactDOM (see https://facebook.github.io/react/blog/2015/07/03/react-v0.14-beta-1.html). This adds a rule, `jsx-uses-reactdom` exactly similar to `jsx-uses-react`.

I took the approach of copying the `jsx-uses-react` rule rather than modifying it to take multiple vars to use. The relative merit of the approach I took is debatable.

I probably wouldn't accept this pull request until 0.14 goes out of beta but I found it useful for me today.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,6 +3,7 @@
 module.exports = {
   rules: {
     'jsx-uses-react': require('./lib/rules/jsx-uses-react'),
+    'jsx-uses-reactdom': require('./lib/rules/jsx-uses-reactdom'),
     'no-multi-comp': require('./lib/rules/no-multi-comp'),
     'prop-types': require('./lib/rules/prop-types'),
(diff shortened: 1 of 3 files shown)


yannickcr:
Mmm, if I understand correctly, the JSX elements still expands to a call to `React.createElement` so JSX will not use the ReactDOM package.

And for the calls to `ReactDOM` (like `ReactDOM.render`) the ESLint [no-unused-vars](http://eslint.org/docs/rules/no-unused-vars.html) rule will perfectly work.

Did I miss something?

donabrams:
Ooh, I thought this was extracting React.DOM instead of just the render methods (thus affecting jsx). The beta build in NPM supports your take. Sorry about that!

PR #160, opened 2015-07-19
Custom validator detection

CalebMorris:
A possible solution to the problem presented in #145.
Since we can't know what the `React.PropTypes` or `MyCustomValidator` is refering to when checking if it should be statically evaluated (or how to do so) I proposed adding an optional configuration to the prop-types rules: `customValidators`.

This allows us to ignore specific proptypes that the plugin doesn't have explicit knowledge of.

Example Usage:
``` javascript
"react/prop-types": [2, {customValidators: ['CustomValidator']}]

// Sample.js
  propTypes: {
    outer: React.PropTypes.shape({
      inner: CustomValidator.map
    })
  },
  render: function() {
    return <div>{this.props.outer.inner.get('test')}</div>;
  }
})
```

As a side note, it could possibly be added so the listed custom validators be actual modules that validate said props.

### DIFF
--- a/lib/rules/prop-types.js
+++ b/lib/rules/prop-types.js
@@ -610,6 +610,12 @@ module.exports.schema = [{
       items: {
         type: 'string'
       }
+    },
+    customValidators: {
+      type: 'array',
+      items: {
+        type: 'string'
+      }
     }
   },
   additionalProperties: false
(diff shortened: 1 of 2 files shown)


yannickcr:
Looks like to be a good solution to solve the custom validators problem. Thanks!

PR #161, opened 2015-07-19
Added rule for no duplicated props

### LINKED ISSUE #143: Rule proposal: Warn on duplicate props
When you're in strict-mode you get warnings when you have duplicate properties in an object, a similiar warning for props on React components would be nice.

```javascript
return (
   <Mycomponent className={classes}
                            onClick={this.handleClick}
                            someOtherProp={stuff}
                            onClick={this.onClick} /> // Duplicate prop, should produce an ESLint error
);
```

hummlas:
Added rule that checks for duplicate props sent to a React component written in JSX.
Has the options of specifying whether casing should be ignored when comparing, via `{ignoreCase: true}`

Closes #143


yannickcr:
Great! Thanks!

PR #164, opened 2015-07-21
Add event handlers to no-unknown-property

mkenyon:
Enforce casing for all event handlers from https://facebook.github.io/react/docs/events.html

I hate trying to debug my click handlers when I've typed `onclick` instead of `onClick` in my components.


yannickcr:
Thanks!

PR #176, opened 2015-08-01
New rule: 'no-literals'

CalebMorris:
This is new style rule that enforces wrapping literal strings in a `JSXExpressionContainer` element.

`<div>{'text'}</div>` and not allowing `<div>text</div>`.

Besides being sure of the actual rendered text, this also helps removed problems with some highlighters and JSX.

```javascript
<div>{"this can't possibly be a problem"}</div>
<div>this can't possibly be a problem</div>
```

As you see the `'` causes problems with the rest of the code, despite being valid.

I'm unsure if `no-literals` is an appropriate name, but since that seems to be the name of the node type I went with it.


yannickcr:
Seems good, but since it's a JSX related rule I think the rule name should be prefixed by `jsx-`: `jsx-no-literals`

CalebMorris:
@yannickcr Renamed. Any other suggestions?

yannickcr:
LGTM, Thanks!

Issue #178, opened 2015-08-03
Update peerDependencies for eslint@1.0.0


mxdubois:
eslint released 1.0.0 so it's time to bump the peerDependency.

I got this error when performing a fresh install of `eslint` and `eslint-plugin-react`:
```
/Users/XXXX/dev/ls-portal
└─┬ eslint-plugin-react@3.1.0
  └── UNMET PEER DEPENDENCY eslint@>=0.8.0 || ~1.0.0-rc-0

npm WARN ENOENT ENOENT, open '/Users/XXXXX/dev/ls-portal/package.json'
npm WARN EPEERINVALID eslint-plugin-react@3.1.0 requires a peer of eslint@>=0.8.0 || ~1.0.0-rc-0 but none was installed.
```

(using `npm@3.2.0` in case it matters, but I don't think it would)

yannickcr:
The specified range should also match `eslint@1.0.0`. But since `npm@3.0.0` [peerDepencencies are no longer installed automatically](https://github.com/npm/npm/releases/tag/v3.0.0), you have to do it manually. I think that's why you get this warning.

dozoisch:
@yannickcr With eslint 1.1.0, it doesn't match anymore though!

tleunen:
You should remove the peerDep now :)

yannickcr:
@dozoisch Actually, [according to semver](http://semver.npmjs.com) the `>=0.8.0 || ~1.0.0-rc-0` range should also match `eslint@1.1.0`. But with `npm@3.2.2` I still get an `EPEERINVALID` warning when installing eslint-plugin-react in global with `eslint@1.1.0`. No error with a local installation though. (npm bug ?)

@tleunen Yep, I'm agree. It does more harm than good right now.

dozoisch:
So weird... Seems like a bug!

madcapnmckay:
We're also blocked from updating to v1.1.0. Any chance we can get this dependency updated?

NPM v2.13.3

PR #185, opened 2015-08-09
Change peerDependencies

Keats:
It currently prevents installation of eslint 1.1.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "mocha": "2.2.5"
   },
   "peerDependencies": {
-    "eslint": ">=0.8.0 || ~1.0.0-rc-0"
+    "eslint": "^1.0.0"
   },
   "keywords": [
     "eslint",


nkbt:
:+1:

tleunen:
Instead of updating the peerDependencies, you should just remove it and put eslint as a simple dependency

nkbt:
@tleunen why?

AlexKVal:
`package.json` from [react-bootstrap](https://github.com/react-bootstrap/react-bootstrap) project
```js
"eslint": "1.0.0",
"eslint-plugin-react": "^3.1.0",
```
https://github.com/react-bootstrap/react-bootstrap/blob/master/package.json#L58
https://github.com/react-bootstrap/react-bootstrap/blob/master/package.json#L63

As you can see - all works fine without this PR :wink: 
I hope it helps :cherries:

tleunen:
@nkbt nvm. Actually, I was saying that because soon the dependencies set as peer would be installed, but in this case, the end user will have eslint installed already so that's fine

yannickcr:
The current `>=0.8.0 || ~1.0.0-rc-0` already match `eslint@1.1.0` but during my tests an error is still triggered when installing `eslint` + `eslint-plugin-react` in global (see #178), looks like an npm bug.

I think I will remove the peerDependency instead.

lamuertepeluda:
```bash
UNMET PEER DEPENDENCY eslint@^2.0.0 || ^3.0.0
```
with eslint v3.3.1

Shouldn't peerDependency have been removed or is this plugin not supposed to work with eslint 3.3.x?

yannickcr:
What version of npm do you use? How is installed eslint/eslint-plugin-react, globally or locally?

lamuertepeluda:
Npm v3.10.3
Node v6.4.0

(latest)

Tried both local (as devDependency) and global installation of eslint-plugin-react. Eslint installed globally.

ljharb:
You don't want anything installed globally, including eslint itself.

lamuertepeluda:
@ljharb do you mean is global installation that causes troubles? 
I use to install eslint and plugins globally because I have several projects managed in the same way, and I don't want to have node_modules replicated 100 times in my dev directory. Plus, I use nvm to manage node versions, I do `eslint --init`, ...

Probably this is more a npm issue, then, because semver also returns correctly.

ljharb:
Each project should have its own copy of eslint installed. Everything a given project needs should be installable with `npm install`.

yannickcr:
> Probably this is more a npm issue, then, because semver also returns correctly.

Yeah, seems the bug that made us remove peerDependencies one year ago (and that I had completely forgotten the existence) is still there :neutral_face:

ljharb:
I'm not sure what bug?

yannickcr:
@ljharb 

```sh
$ npm list -g
/home/yannick/.npm/lib
└── (empty)

$ npm install eslint -g
/home/yannick/.npm/bin/eslint -> /home/yannick/.npm/lib/node_modules/eslint/bin/eslint.js
/home/yannick/.npm/lib
└── eslint@3.3.1

$ npm install eslint-plugin-react -g
/home/yannick/.npm/lib
├── UNMET PEER DEPENDENCY eslint@^2.0.0 || ^3.0.0
└── eslint-plugin-react@6.1.2

npm WARN eslint-plugin-react@6.1.2 requires a peer of eslint@^2.0.0 || ^3.0.0 but none was installed.

$ npm list -g
/home/yannick/.npm/lib
├── eslint@3.3.1
└── eslint-plugin-react@6.1.2
```

Everything is working at the end, there is only a false `UNMET PEER DEPENDENCY` warning when installing `eslint-plugin-react`. No warning at all if eslint is installed locally.

ljharb:
That's an easily ignore able warning when following the bad practice of having eslint installed globally - I think the value of the peer dep far outweighs that.

PR #189, opened 2015-08-14
Add defensive coding to support experimental object spread properties

Morantron:
In a file like this with experimental object spread properties syntax:

```javascript
function lol(a, b, c) {
  return {...a, b, c};
}
```

The following rules were throwing an error because `node.property` is undefined in a property node of type `ExperimentalSpreadProperty`.

Not sure if this is the right approach, but at least it allows me to run eslint!


yannickcr:
Thanks for the fix!

PR #193, opened 2015-08-20
Check index is not negative.

mattyod:


yannickcr:
Hi, do you have a use case for this fix ?

mattyod:
I was getting an error thrown as i was -1. To be honest I didn't dig much deeper as this got my lint system working on .jsx files with an npm command:

```
"lint": "eslint . --ext .jsx,.js"
```

yannickcr:
Ok, I found that the following pattern trigger the bug:
```jsx
  static get propTypes() {} // <- empty body for the propTypes declaration
  render() {
    return <div>Hello World</div>;
  }
}
```
I'll add a this pattern to the test suite.

Thanks for the bugfix!

mattyod:
Thanks for the quick resolution.

PR #194, opened 2015-08-21
add `state` in lifecycle methods

mathieudutour:
In ESNext, the following is possible:
```
class extends React.Component {
  state = {
    key: value
  }
}
```


yannickcr:
Great! Thanks!

Issue #195, opened 2015-08-24
Is there a way to catch setting state?


moroshko:
I'd like to check whether I set state in my React app. More specifically, I'd like to catch things like:
```js
this.state = { ... };
maybeSomethingElse.state = { ... };
this.setState({ ... });
maybeSomethingElse.setState({ ... });
```
In other words, I'd like to restrict setting `state` and calling `setState()`.

Is that possible to catch these?

moroshko:
#197 Looks like a good start

markdalgleish:
Ha, didn't realise you opened this issue! Perfect timing :smile:

markdalgleish:
#197 got merged, this issue can be closed now.

remitbri:
Actually, @markdalgleish's `no-set-state` rule only prohibits the use of `setState`. Shouldn't `getInitialState()` be avoided as well in this context, i.e if you forbid one you forbid the other as well? 

From what I understand from @moroshko's initial question, there's still the problem of doing things like `this.state.foo = "bar"`… So should the issue be re-opened? Or maybe a new issue should be created for this particular thing?

yannickcr:
Yest, that the reason why I did not closed this issue at first, since all reported cases are not (yet) handled.

PR #197, opened 2015-08-25
Add no-set-state rule

markdalgleish:
As explained in the updated docs in this commit: "When using an architecture that separates your application state from your UI components (e.g. Flux), it may be desirable to forbid the use of local component state."

It's highly likely that this rule might need some changes, particularly since I haven't worked with ESLint internals before, but I wanted to get my first attempt in a pull request to at least get the conversation started.

Is this a rule that you'd consider including?


MoOx:
:+1:

yannickcr:
While not covering all cases listed in #195, the rule looks fine to me. I think it is a very good start.

Do you plan to handle more cases, or can I merge it as is and we'll improve it gradually ?

markdalgleish:
:+1: I think merging it in and improving it gradually is a good idea. I'd be keen to make this more comprehensive.

yannickcr:
Merged, thanks!

moroshko:
@yannickcr When can we expect to see this on npm?

yannickcr:
@moroshko I plan to release a new version by the end of the day tomorrow.

moroshko:
@yannickcr Thanks! Looking forward to it.

Daniel15:
A good variant of this rule would be to only allow `setState` from "container" components (ie. components with a "Container" suffix). The pattern we follow in my team is that UI components have no state and take everything via props, and container components wrap the UI components and connect them to stores. The container components only use state, no props. In this case we'd only want the "no setState" rule to apply to UI components, not container components.

PR #206, opened 2015-09-03
Treat component with render which returns createElement as Valid

epmatsw:
This component triggers a sort-comp warning:
```
return React.createClass({
  render: function() {
    return <div>Hello</div>;
  },
  displayName : 'Hello'
});
```

This one does not:
```
return React.createClass({
  render: function() {
    return React.createElement("div", null, "Hello");
  },
  displayName: 'Hello'
```

This change makes it so that classes of the latter type are correctly consider as react classes, and rules such as sort-comp will run against them.


yannickcr:
Great! Thanks!

PR #218, opened 2015-09-16
Configure jsx-closing-bracket-location for self-closing and non-empty tags

### LINKED ISSUE #208: Problem with jsx-closing-bracket-location rule
After enabling

```
    "react/jsx-closing-bracket-location": [2, {"location": "after-props"}],
```

rule in project's ESLint config, next element start producing errors,

```jsx
<Button
    bsStyle="success"
    data-id={showPromoteModal}
    onClick={this.handleUserPromote}>
    Promote
</Button>
```

But from rule definition I believe that it should only validate self-closing JSX tags.

I'm not sure how better handle this situation, maybe adding some configuration value to the rule, but in total, when rule enabled I only want to validate closing bracket location for self-closing JSX tags, like,

```jsx
<Route
    name="page"
    handler={Page}
    path="/page/:id" />
```

and do not validate normal tags, like `Button` above.

What do you think? Does it make sense for you?

evocateur:
Optionally, one can configure the behaviour of `jsx-closing-bracket-location`
when dealing with different types of tags. Additionally, it allows a simple
string config instead of the previous object with a single location property
(which is still supported).

This fixes #208

### DIFF
--- a/docs/rules/jsx-closing-bracket-location.md
+++ b/docs/rules/jsx-closing-bracket-location.md
@@ -32,10 +32,22 @@ The following patterns are not considered warnings:
 
 ## Rule Options
 
+There are two ways to configure this rule.
+
+The first form is a string shortcut corresponding to the `location` values specified below. If omitted, it defaults to `"tag-aligned"`.
+
 ```js
-...
-"jsx-closing-bracket-location": [<enabled>, { "location": <string> }]
-...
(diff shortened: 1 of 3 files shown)


yannickcr:
Great! Thank you very much for this!

evocateur:
No problem, glad I could help!

PR #219, opened 2015-09-16
Namespaced JSX fixes

zertosh:
cc: @yannickcr


yannickcr:
Oh, I was not aware of this syntax.
Thanks!

zertosh:
heh I just found out about it because no-undef wasn't working on a bunch of new files. Can you publish the fix please?

FYI, though it's valid JSX, and it's recognized by the parser, babel doesn't transpile it. I think you have to use jstransform or build a custom babel plugin (I think that's what we do).

yannickcr:
Thanks for the informations.

The fix was released in `v3.4.0`

Issue #201, opened 2015-08-30
rule proposal: no-direct-mutation-state


remitbri:
As suggested by @moroshko in #195 (but the issue was closed too quickly), there should be a rule that forbids direct state mutation. I.e
```js
this.state.foo = "newValue"
```
sends a warning,
whereas
```js
this.setState({foo: "newValue"})
```
or
```js
const bar = this.state.foo
```
sends none

mkenyon:
I noticed that this is a duplicate of #133.

(Though I think it's a good idea.)

PR #223, opened 2015-09-19
Add no-direct-mutation-state rule (fixes #133, #201)

### LINKED ISSUE #133: It does not detect when I'm mutating the state directly.
There's no error (or warning) for when I mutate the state instead of using setState.

so:
`this.state.foo = "bar"`
should yield an error, as opposed to
`this.setState({foo: "bar"});`

petersendidit:
Rule to call out direct state mutation.

First eslint rule I have written and am very open to feedback to make it better. Its very simple right now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ Finally, enable all of the rules that you would like to use.
     "react/no-danger": 1,
     "react/no-did-mount-set-state": 1,
     "react/no-did-update-set-state": 1,
+    "react/no-direct-mutation-state": 1,
     "react/no-multi-comp": 1,
     "react/no-set-state": 1,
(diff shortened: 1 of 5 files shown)


yannickcr:
Thanks.

Can you add some checks for properties existence (on `item.left`, `item.object` and `item.property`) to avoid any crash while checking a non existent property ?

For example currently the rule crash on: 

```js
module.exports = {};
```

Except this problem the rule looks good to me :)

petersendidit:
Ok should be fixed up to handle
```javascript
module.exports = {};
```

yannickcr:
LGTM, Thanks!

PR #224, opened 2015-09-22
add a rule that recommends alternatives to vague prop types

pwmckenna:
not super tied to the rule name, or the way options are passed in. here's the goal of the rule:

I don't ever want people in my repo using `React.PropTypes.any/array/object` because they could easily be replaced by other proptypes like `shape/arrayOf` that document the structure of the data better. Whether this rule checks for vagueness, or just allows for blacklisting of types, doesn't really matter to me. Thoughts? 

__Edit:__ Looks like this has already been suggested. Probably should have done a little browsing before hand: https://github.com/yannickcr/eslint-plugin-react/issues/215

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ Finally, enable all of the rules that you would like to use.
     "react/jsx-indent-props": 1,
     "react/jsx-max-props-per-line": 1,
     "react/jsx-no-duplicate-props": 1,
+    "react/jsx-no-literals": 1,
     "react/jsx-no-undef": 1,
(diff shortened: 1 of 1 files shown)


jimbolla:
I would do "no-forbidden-proptypes" and then make the rule customizable so that the user can specificy any proptypes they want to exclude, but make the defaults any/array/object.

pwmckenna:
good suggestion! https://github.com/yannickcr/eslint-plugin-react/pull/225

PR #225, opened 2015-09-22
add forbid-prop-types rule that by default forbids "any", "array", and "object"

pwmckenna:


pwmckenna:
@jimbolla made a new branch/pr (all the naming changed)
this pr addresses https://github.com/yannickcr/eslint-plugin-react/issues/215

yannickcr:
Looks great! Can you add some documentation ?

Also, does it handle nested propTypes ?
```jsx
    render: function () {
        return <div>{this.props.bar.baz.something}</div>
    }
}
Foo.propTypes = {
    bar: React.PropTypes.shape({               // ok
        baz: React.PropTypes.object.isRequired // not ok, PropTypes.object is forbidden
    })
};
```

pwmckenna:
@yannickcr just added a test that verifies support for nested props. I'll get started on some documentation now. I'll flatten these commits once the docs are looking good.

pwmckenna:
documentation added!

yannickcr:
:+1: 

If it is good for you, can you squash your commits ? Then I will merge the PR.

pwmckenna:
updated the docs to mention the suggested replacements, then squashed. should be good to merge once checks pass.

yannickcr:
Merged, thanks !

mydearxym:
this ` forbid-prop-types` seems not work in this situation:

```js
    const bar = this.props.bar;

    render: function () {
        return <div>{bar.baz.something}</div>
    }
}
Foo.propTypes = {
    bar: React.PropTypes.shape({               // ok
        baz: React.PropTypes.object.isRequired // not ok, PropTypes.object is forbidden
    })
};
```

ljharb:
That's correct behavior - you should use a shape, or objectOf, on `baz`.

mydearxym:
@ljharb   sorry for not make my self clear

```js
    const bar = this.props.bar;

    render: function () {
        return <div>{bar.baz}</div>
    }
}
Foo.propTypes = {
    bar: React.PropTypes.shape({  
        baz: React.PropTypes.string      // not ok, Prop defined but not used
    })
};
```

this is ok:
```js
    render: function () {
        return <div>{this.props.bar.baz}</div>
    }
}
Foo.propTypes = {
    bar: React.PropTypes.shape({  
        baz: React.PropTypes.string      // ok
    })
};
```

ljharb:
Gotcha, can you file a new issue for that?

mydearxym:
yes, i open #847

PR #234, opened 2015-10-02
Fix crash with destructured PropTypes

epmatsw:
Forbid-prop-types will crash if you have a bit of code that looks like:
```
var object = React.PropTypes.object;
var Comp = React.createClass({
  propTypes: {
    someProp: object
  },
  render: function() {.....}
});
```
This pull request fixes that issue and adds a test case to ensure that the above is correctly flagged as an error.


yannickcr:
Thanks!

PR #242, opened 2015-10-10
Update `jsx-sort-props` to allow sorting callbacks last

Daniel15:
In my team (ads at Facebook) we tend to sort callbacks to the end, in their own group:

```js
<Foo
  a="blah"
  b="blah"
  y="blah"
  z="blah"
  onAwesome={() => alert('Woohoo!')}
  onClick={blah}
  onFoo={blah}
  onZoom={blah}
```

This change adds a `callbacksLast` configuration option to allow this style of sorting.

Also fixed the `currenPropName` typo (missing `t`) and renamed `lastPropName` to `previousPropName` so it's clearer that it's the previously encountered prop and **not** the very last prop of the JSX tag.

### DIFF
--- a/docs/rules/jsx-sort-prop-types.md
+++ b/docs/rules/jsx-sort-prop-types.md
@@ -78,13 +78,32 @@ class Component extends React.Component {
 
 ```js
 ...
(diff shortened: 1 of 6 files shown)


yannickcr:
Nice addition, thanks!

Issue #252, opened 2015-10-16
Rule idea: disallow getInitialState and accessing this.state when setState is never used


mik01aj:
This would be useful for removing leftovers after refactoring.

I.e: a component should be either stateful, or stateless. A component should have either all or none of these things:

* calls to `setState`
* `getInitialState` implementation
* using `this.state`

PR #263, opened 2015-10-20
Introduce `state-usage` rule that detects unused/undeclared state

dtinth:
This rule detects unused state:

``` js
React.createClass({
  getInitialState: function () {
    return {unused: true};
  }
});
```

And undeclared state:

``` js
React.createClass({
  render: function () {
    return <span>{this.state.undeclared}</span>
  }
});
```

This PR partially complements with #252, which should ease in refactoring.

Note that this is just the first working implementation, which I hacked it up mainly to ease my workflow when refactoring React components at @taskworld. I may not have time to polish it and make it work with ES6-class style components. But I think this may be useful to someone looking for this linting capability.

### DIFF
--- a/lib/rules/state-usage.js
+++ b/lib/rules/state-usage.js
@@ -1,0 +1,149 @@
+/**
+ * @fileoverview Prevent unused state or using state that has never been set
+ * @author Thai Pangsakulyanont @ Taskworld.com
+ */
+'use strict';
+
+var componentUtil = require('../util/component');
+var ComponentList = componentUtil.List;
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = function(context) {
+
+  var componentList = new ComponentList();
+
+  /**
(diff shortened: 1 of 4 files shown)


yannickcr:
Thanks a lot! I'll have a look to add support for ES6 components (if you do not have the time to do it).

lib/rules/state-usage.js
```diff
@@ -0,0 +1,149 @@
...
+'use strict';
+
+var componentUtil = require('../util/component');
+var ComponentList = componentUtil.List;
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = function(context) {
```
petersendidit: This should use the [new eslint rule format](http://eslint.org/blog/2016/07/eslint-new-rule-format)

dtinth:
Sorry, but I don’t have plan to finish this for now. Closing this PR.

Please feel free to use this PR as a reference/base if anyone wants to implement this in the future.

PR #274, opened 2015-10-27
Add rule for disallowing `.bind` in JSX props

### LINKED ISSUE #184: Rule for `bind` inside JSX
Is that possible to create rule that would throw warning on this code:
```jsx
<a onClick={this.onClick.bind(this, true)}>click me</a>
```
And maybe suggest to do this:
```jsx
<a onClick={() => this.onClick(true)}>click me</a>
```

Daniel15:
A `bind` call or [arrow function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) in a JSX prop will create a brand new function on every single render. This is bad for performance, as it can result in the garbage collector being invoked way more than is necessary (especially on hot paths with components that have frequent updates). This rule disallows both `bind` calls and arrow functions, with options to re-enable them (`allowBind` and `allowArrowFunctions` respectively) in case you want to only disable one.

Fixes #184.


lib/rules/jsx-no-bind.js
```diff
@@ -0,0 +1,50 @@
...
+  return {
+    JSXAttribute: function(node) {
+      if (!node.value || !node.value.expression) {
+        return;
+      }
+      var valueNode = node.value.expression;
+      if (
+        !configuration.allowBind &&
+        valueNode.type === 'CallExpression' &&
+        valueNode.callee.property.name === 'bind'
```
Daniel15: This is a really naïve check for bind (it just checks for calls to any method called`.bind) but I figure it's Good Enough™. I hope nobody has objects with methods called `bind` and uses them in this context. I imagine that's an edge case. :stuck_out_tongue:

docs/rules/jsx-no-bind.md
```diff
@@ -0,0 +1,74 @@
...
+    );
+  }
+});
+
+var ListItem = React.createClass({
+  render() {
+    return (
+        <li onClick={this._onClick}>
+          ...
+        </li>
```
dtinth: You might want to remove 2 spaces from these 3 lines

Daniel15: Oops, thanks! I'll change that tomorrow. Not sure how I missed that :(

dtinth:
I like this rule, and the options are very helpful. I really like the ProTip section.

I personally would `allowArrowFunctions`, but not `allowBind`.

For non-hot code paths, using the arrow function leads to a more straightforward code, which is more comprehensible than the `.bind()` counterpart (even though it’s essentially the same). Only when performance impact becomes apparent, then I’ll bind the methods on the constructor.

yannickcr:
Thanks! I'll review this asap.

Can you also add the new rule to the README.md file?

Daniel15:
> Can you also add the new rule to the README.md file?

Will do :) Just let me know if any other changes are required, and I'll do all the updates at the same time.

yannickcr:
Currently I don't see any other changes to do, LGTM.

Daniel15:
Thanks! I just added the rule to the readme.

yannickcr:
Merged. Thank you very much!

jasonkylefrank:
Hello ES6 experts!

@Daniel15 mentioned above that:

> A bind call or **arrow function** in a JSX prop will create a brand new function on every single render. This is bad for performance...

Looking at docs such as those from Mozilla, I can see that _.bind()_ creates a _new_ function, corroborating what @Daniel15 mentioned.  **However, I am curious about the arrow function**.   I have not been able to find something that says that the arrow function would do something similar as far as making a copy (and thus bad for performance in the render method).  Can someone provide a link to a source explaining how the arrow function is similar to bind() in this regard?

I also thought that maybe it had to do with how arrow functions get **transpiled** - perhaps that they would end up using bind() once transpiled.  But I have not been able to confirm that either.  For example, [this test](https://babeljs.io/repl/#?evaluate=true&presets=es2015%2Creact%2Cstage-2&experimental=true&loose=true&spec=false&code=function%20foo()%20%7B%0A%20%20const%20bar%20%3D%20()%20%3D%3E%20%7B%0A%20%20%20%20console.log(this.id%2C%20arguments)%3B%0A%20%20%7D%0A%20%20bar()%3B%0A%7D%0Afoo.call(%7Bid%3A%2042%7D%2C%20'lol')%3B%0A) on Babel's online editor shows that this ES6:

```
function foo() {
  const bar = () => {
    console.log(this.id, arguments);
  }
  bar();
}
foo.call({id: 42}, 'lol');
```
transpiles to:
```
'use strict';

function foo() {
  var _this = this,
      _arguments = arguments;

  var bar = function bar() {
    console.log(_this.id, _arguments);
  }; // <----------- Notice .bind() was not added
  bar();
}
foo.call({ id: 42 }, 'lol');
```

In summary, **can someone please show how the arrow function has the same performance problems that bind() suffers from in React's render method?**

Daniel15:
In general, a new function is created every time the "function" keyword is
encountered, regardless of if you use bind or not. In your example, the
"bar" function only exists inside the closure crested by the "foo"
function. Every call to "foo" results in a new closure, and thus a new
instance of "bar" too. There's nothing special about bar, it behaves like
any other variable.

Some smarter optimisations could be done (such as having just one copy of
"bar" with copy-on-write semantics in case you set properties on the
function itself, like `bar.executedAt = Date.now()`) but I don't think any
JavaScript engine does that today.

jasonkylefrank:
@Daniel15 Ok, so the issue with arrow functions is not that they create a new function in the same way that bind() does, but rather that the runtime has overhead involved in creating the closure necessary to execute them?  That makes sense.

I have 2 follow-up questions.  

1. The suggested means to handle this problem (in the linter rule readme) is to create a separate React component to handle the callback situation.  There must additional overhead involved in initializing and rendering this extra React component (e.g., lifecycle methods, render(), etc.).  **So how does that added overhead compare to the overhead that would be incurred if you used arrow functions instead?**  I assume someone has tests comparing these 2 approaches right?  Does anyone have a link for one?

2.  The React docs, on their "[Communicate Between Components](https://facebook.github.io/react/tips/communicate-between-components.html)" page advocate using `bind()` to solve this passing-of-arguments problem.  Their example is a list example:
  ```
var handleClick = function(i, props) {
    console.log('You clicked: ' + props.items[i]);
}

function GroceryList(props) {  
  return (
    <div>
      {props.items.map(function(item, i) {
        return (
          <div onClick={handleClick.bind(this, i, props)} key={i}>{item}</div>
        );
      })}
    </div>
  );
}

ReactDOM.render(
  <GroceryList items={['Apple', 'Banana', 'Cranberry']} />, mountNode
);
```
**Is their example different somehow (i.e., is `bind()` _not_ a performance problem in their case)?**
@Daniel15 You work at Facebook, right?  If that example has the same performance problem, is there any chance you'd have some pull to get the docs updated to show the "proper" way?  It would be a lot less confusing for us React noobs then!

**Comment by dtinth** [comment]
@jasonkylefrank This is an interesting question, so I’ve done a quick comparison using [Benchmark.js](https://benchmarkjs.com/) against the __development__ build of React using example code from the rule documentation:

```js
import React from 'react'

const List = React.createClass({
  render () {
    return (
      <ul>
        {this.props.items.map(item =>
          <li key={item.id} onClick={this.props.onItemClick.bind(null, item.id)}>
            {item.text}
          </li>
        )}
      </ul>
    )
  }
})

export default List
```

```js
import React from 'react'

const List = React.createClass({
  render () {
    return (
      <ul>
        {this.props.items.map(item =>
          <ListItem key={item.id} item={item} onItemClick={this.props.onItemClick} />
        )}
      </ul>
    )
  }
})

var ListItem = React.createClass({
  render () {
    return (
      <li onClick={this._onClick}>
        {this.props.item.text}
      </li>
    )
  },
  _onClick () {
    this.props.onItemClick(this.props.item.id)
  }
})

export default List
```

This is the input data:

```js
function onItemClick (id) {
  alert(id + ' has been clicked!!')
}

const items = [ ]

for (let i = 0; i < 100; i ++) {
  items.push({ id: i, text: 'Item ' + i })
}
```

Here’s the benchmark suite:

```js
new Benchmark.Suite()
.add('Single component with bind()', function () {
  render(<Case1 items={items} onItemClick={onItemClick} />, div1)
})
.add('Two components without bind()', function () {
  render(<Case2 items={items} onItemClick={onItemClick} />, div2)
})
```

Note that `div1` and `div2` are divs with `display: none` CSS applied, so that it only tests raw DOM manipulation speed.

| Test | Speed |
| --- | --- |
| Single component with bind() | x 589 ops/sec ±1.75% (55 runs sampled) |
| Two components without bind() | x 516 ops/sec ±1.17% (54 runs sampled) |

In __development__, using single component takes __15% less time__.

## On Production

Now let’s try on __production__ build of React:

| Test | Speed |
| --- | --- |
| Single component with bind() | x 134,742 ops/sec ±2.91% (57 runs sampled) |
| Two components without bind() | x 134,721 ops/sec ±2.27% (58 runs sampled) |

Even though the numbers are almost close to each other, Benchmark.js reports that __Two components without bind() is faster__. Maybe %error is taken into account?

But is it _really_ because of whether I use bind or not? I added another test case: Two components but using `bind()`.

```js
import React from 'react'

const List = React.createClass({
  render () {
    return (
      <ul>
        {this.props.items.map(item =>
          <ListItem key={item.id} item={item} onItemClick={this.props.onItemClick} />
        )}
      </ul>
    )
  }
})

var ListItem = React.createClass({
  render () {
    return (
      <li onClick={this.props.onItemClick.bind(null, this.props.item.id)}>
        {this.props.item.text}
      </li>
    )
  }
})

export default List
```

| Test | Speed |
| --- | --- |
| Two components with bind() | 137,732 ops/sec ±1.94% (58 runs sampled) |
| Two components without bind() | 137,218 ops/sec ±1.92% (57 runs sampled) |

Benchmark.js reports that both ways are equally performant. In my opinion, the performance costs caused by `.bind()` is negligible. I’d still use `.bind()` (or arrow functions) in non-critical code paths.

__Disclaimer:__ I just did a preliminary testing in late night. Would be great if someone do a more formalized replication of this test.

dtinth:
Btw, when you call `React.createElement`, you are creating a new Virtual DOM node every time — just like creating a function. So I think it might make sense to compare what is faster — between creating a new function, and creating a React element.

The tests:

```js
const component = {
  props: {
    onClick (a, b) {
      return a + b
    }
  }
}

new Benchmark.Suite()
.add('Using Function.prototype.bind',
  new Function('return this.props.onClick.bind(null, 1, 2)').bind(component)
)
.add('Using closure',
  new Function('var that = this; return function () { return that.props.onClick(1, 2) }').bind(component)
)
.add('Creating an arrow function',
  new Function('return (() => this.props.onClick(1, 2))').bind(component)
)
.add('Creating a React element', function () {
  React.createElement('div')
})
```

Note: Function constructor is used to avoid code being transpiled by Babel.

The tests are performed on __Chrome 49__ on MacBook Pro (Retina, 13-inch, Late 2013)

| Test | Speed |
| --- | --- |
| Using Function.prototype.bind | x 1,924,886 ops/sec ±2.01% (56 runs sampled) |
| Using closure | x 26,304,860 ops/sec ±3.26% (52 runs sampled) |
| Creating an arrow function | x 29,261,815 ops/sec ±1.50% (56 runs sampled) |
| Creating a React element | x 19,198,690 ops/sec ±3.81% (54 runs sampled) |

You can see that __Function.prototype.bind is 10 times slower than a function literal__. On the other hand, creating a new function using the function literal or arrow function is faster than creating a single React element.

---

Note that Babel transpiles arrow functions without using `.bind()`, but by creating a `_this` variable:

```js
function b () {
  return () => this.props.onClick()
}
```

```js
function b() {
  var _this = this;

  return function () {
    return _this.props.onClick();
  };
}
```

Daniel15:
Benchmark.js tests might not show the full story, as the main perf cost comes from garbage collection rather than the actual runtime performance of the code itself. It's possible that you're not hitting garbage collection while testing and thus not experiencing a GC pause. It's probably worth watching memory consumption while running the test.

Does your test mount a brand new component every time, or does it update the props on the existing component? I don't see the source code of the `render` function used in the benchmark suite, is that just the standard `ReactDOM.render`? A more realistic benchmark might just mount the component once and then do lots of `setState` calls to re-render the component. I'm not sure if that would affect the results though.

> The React docs, on their "Communicate Between Components" page advocate using bind() to solve this passing-of-arguments problem.

Yeah, I'm not a fan of that part of the documentation. I may send a pull request to update it too.

jasonkylefrank:
@dtinth Those are great tests!  Thanks for running them!

Your second set of tests (where you compare using `bind()`, closures, arrow functions, and `React.createElement`) seem to indicate that arrow functions are significantly faster than using `bind()`.  So what happens if you run your "Single component with bind()" vs "Two components without bind()" tests using _arrow functions instead of `bind()`_?

dtinth:
@Daniel15 

The `render()` function comes directly from ReactDOM. What I’ve tried is using calling `render()` with the same DOM node, so it is the same as mounting once and call `setState()`. I’ve tried using Chrome’s timeline developer tools to capture the timeline. There are no difference in GC statistics.

screen shot 2016-03-20 at 7 35 24

From the picture above, you can see that the Minor GC runs on almost every render and Major GC is run several time throughout the tests.

Have you profiled the GC time when you reach the conclusion that the main perf comes from garbage collection? If so, can you share your research? By the way, I have no data to support this but I think Virtual DOM nodes as costly to garbage-collect as functions, so I really doubt that it’s GC…

@jasonkylefrank 

I replaced the `.bind()` call with arrow functions. There are almost no differences.

| Test | Speed |
| --- | ---
| One component with arrow functions | x 147,453 ops/sec ±1.92% (57 runs sampled) |
| Two components without bind() | x 148,354 ops/sec ±1.86% (57 runs sampled) |

-----

I’d appreciate it if someone can perform a larger scale test that represents the real world performance issue.

Based on my experience tuning React performance, __I’ve never found performance problems caused by creating new functions every `render()` call__. Most of the performance problems come from wasted render — render calls which result in the same Virtual DOM nodes as previous render.

Mostly it is that creating a new object or function to pass as props to child component every time sometimes breaks `PureRenderMixin` (or `shallowCompare`), as `this.props.onClick !== nextProps.onClick` is always the case. This is the only case I’ve found where performance issue is noticeable. Creating arrow functions in `render()` can lead to wasted renders, but that has nothing to do with creating or garbage-collecting functions/objects.

So here goes my conjecture: __If you create a new function every `render()` call, but it doesn’t lead to wasted renders, then performance difference will not be noticeable.__

Daniel15:
Thanks for your benchmarks @dtinth! Very useful :)

> Mostly it is that creating a new object or function to pass as props to child component every time sometimes breaks PureRenderMixin (or shallowCompare), as this.props.onClick !== nextProps.onClick is always the case. This is the only case I’ve found where performance issue is noticeable.

Yeah, this is what I found too. I should have highlighted it more thoroughly in the docs. You could write your own `shouldComponentUpdate` but it's usually good to follow the "happy path" of allowing `PureRenderMixin` to work as-is.

Feel free to send a pull request to make this clearer in the docs :)

jasonkylefrank:
@dtinth Thanks for the follow up tests - very helpful!  

Based on the tests that @dtinth has shown, I'm not seeing a benefit of using this linter rule, pending possible garbage-collection performance research from @Daniel15 or others.  I'd love to hear from anyone that has research on these issues.

Daniel15:
The fact that creating a new function on render breaks PureRenderMixin is
still worth keeping in mind, especially for components that are high up in
a component tree.

jasonkylefrank:
@dtinth Follow-up questions about your last test ("One component with arrow functions..."):

1. Was that with the production version of React?
2. Your "4 function tests" (where you compare using `bind()`, closures, arrow functions, and `React.createElement`) seemed to indicate that arrow functions are considerably faster than `bind()`.  Why then do you think that these last tests did not show improved performance over the `bind()` version?

dtinth:
@jasonkylefrank 

1. Yes, this is the production version of React.

2. The four tests only cover the time it takes to create functions or React elements, which take less than 1/2,000,000 seconds. Compared to the time it takes for React to render the Virtual DOM to DOM (which takes 1/150,000 seconds) this is extremely small. That’s why I think the performance issue here is negligible.

In my opinion, the real performance gains come from __avoiding wasted renders__, not by avoiding arrow functions or bound functions — but avoiding arrow functions or bound functions _is_ one way to avoid wasted renders.

__Note:__ I would not even avoid wasted renders until performance problems becomes noticeable; I don’t want to trade code readability and straightforwardness for it. But I _do_ take measures to ensure that _when I really need to optimize_, I can do it pretty easily — e.g. by always using immutable data.

Issue #243, opened 2015-10-11
jsx-closing-bracket-location when returning directly


FredyC:
With default settings, this is considered warning. I am wondering how do you write this kind of code.

        render() {
            return (<Application
                {...defaultProps}
                {...this.props}
                {...options}
            />);
        }

This looks totally weird in my opinion...

        render() {
            return (
            <Application
                {...defaultProps}
                {...this.props}
                {...options}
            />);
        }

yannickcr:
I personally write it this way:
```js
    render() {
        return (
            <Application
                {...defaultProps}
                {...this.props}
                {...options}
            />
        );
    }

**Comment by FredyC** [comment]
Ah I see, that doesn't look so bad except it's two extra lines. Hmm. Do you think it would be possible to extend the rule so it can recognize alignment to return ?

**Comment by yannickcr** [comment]
It is possible to add another location type, with the same behavior of `tag-aligned` except when the tag is preceded with return.

```jsx
render() {
  return (<App
    <Component
      {...defaultProps}
      {...this.props}
      {...options}
    />
  </App>);
}
```

csilvers:
I think it's not just `return`.  (We are seeing this same issue, and would prefer not to have the extra parens.)  For instance, we'd like to be able to do
```
   var x = <App
      ...
   />;

PR #279, opened 2015-10-29
Add 'line-aligned' option for jsx-closing-bracket-location

alopatin:
`'line-aligned'` acts like `'tag-aligned'` except when the opening JSX tag
is preceded by other code on the same line.

This commit adds the rules, the corresponding tests, and updates the
documentation.

Test Plan:
npm install
npm run test


alopatin:
Fixes #243

yannickcr:
Thanks! I think we'll keep `tag-aligned` has-is for now for backward compatibility and remove it in a future version.

PR #283, opened 2015-11-03
[jsx-no-bind] Ensure node is MemberExpression before checking property name

Daniel15:
When I actually started using this rule in real-life code, I noticed that it was crashing for normal function calls (eg. `foo={bar()}`) as `node.callee.property` is `null` for those. Instead, we want to explicitly check that the callee is a `MemberExpression`.


yannickcr:
Thanks for the quick fix :)

Issue #67, opened 2015-04-16
assert `key` properties on arrays


jonathanong:
might be a nice hint, though not necessary since react warns us anyways.

yannickcr:
I think this request is similar to #40

jonathanong:
yes it is! thanks

Issue #40, opened 2015-03-29
Rule proposal: require `key` prop for rendered array items


lencioni:
I'm not sure how feasible this would be to implement, but it would be really great if we could catch this before runtime.

Good:

```jsx
<div>
  {myArray.map(item => <div key={item.id}>{item.body}</div>)}
</div>
```

Bad:

```jsx
<div>
  {myArray.map(item => <div>{item.body}</div>)}
</div>
```

yannickcr:
Would be very useful but I am not sure how to do this. Need to investigate.

benmosher:
I'm willing to take a stab at this.

I feel like at a minimum, array literals can be reported:

```js
<dl>
  items.map(item => [<dt>{item.name}</dt>, <dd>{item.desc}</dd>])
</dl>
```

More generally, I'm trying to think of a case where something like

```js
object.method(x => <div>{x}</div>)
//or
fn(object, x => <div>{x}</div>)
```

_isn't_ an iterator callback. So, at least optionally, feel like reporting on callbacks without keys would report the right situations without a lot of false positives.

benmosher:
so my proposal would be `jsx-key` with `no-callback` option to disable the callback behavior if a given codebase has situations where it is over-reporting.

yannickcr:
Fixed by #293

PR #293, opened 2015-11-05
jsx-key

benmosher:
Implementation of proposal for reporting missing `key` props in array literals (#40, #67).

This draft has no options, and just reports on JSX elements in array literals and as the only expression in an arrow function body.

Let me know if you want to add an option to ignore arrow functions and just do array validation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ Finally, enable all of the rules that you would like to use.
     "react/jsx-closing-bracket-location": 1,
     "react/jsx-curly-spacing": 1,
     "react/jsx-indent-props": 1,
+    "react/jsx-key": 1,
     "react/jsx-max-props-per-line": 1,
     "react/jsx-no-bind": 1,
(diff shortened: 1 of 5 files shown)


benmosher:
A thought: this would be considered a warning:

```jsx
const props = { key: "x", foo: "bar" }
return [<Hello {...props} />]
```

I think that is an acceptable (and unlikely) edge case such that it would not be worth covering.

yannickcr:
> Let me know if you want to add an option to ignore arrow functions and just do array validation.

For what cases would you would disable the rule for arrow functions ? Is there a false positive risk ?

> I think that is an acceptable (and unlikely) edge case such that it would not be worth covering

Agreed to not handle this case for now.

Beside that, it's looks good to me. Can you squash your commits ?

benmosher:
Squashed to 1 commit.

yannickcr:
Merged, thanks!

PR #299, opened 2015-11-06
added new test case for another scenerio that was valid

hamiltondanielb:
Added a test case for this scenario that I was running into that I thought was valid, but was throwing and error without the ecmaFeatures. This test case it just for future reference if anything is ever changed.


yannickcr:
Hi, what is the problem exactly ?

hamiltondanielb:
The problem was without emcaFeature Module and Modules activated, this rule would fail for `export default class Hello extends React.Component {` I added the test case for this scenario, no code changes

Issue #245, opened 2015-10-12
Support fixable flag from esLint


carlosmmelo:
With the latest version of eslint, they added the support to automatically fix errors with the [fixable](https://github.com/eslint/eslint/blob/9a58b2494b7850a69c84c29540012f87c080a5a8/docs/rules/README.md#rules) flag on the rules and passing the `--fix` thru the cli.

Are you planning to add support to this on your rules by any chance? this is a nice feature to automatically fix the most obvious and easy ones. at least.

Thanks

yannickcr:
Yes, I'm planning to add support for the fixable flag, at least for formating rules.

But PR are welcome ;)

lencioni:
There are a number of rules that are autofixable now. This issue can probably be closed out.

PR #301, opened 2015-11-09
Improve jsx-closing-bracket-location error message

alopatin:
By adding expected column information for the closing bracket, as well
as reporting the error at the closing bracket instead of the opening
bracket. Parsing the error message should now be sufficient to allow
writing an external autofixer.

Example input `test.jsx` with default options:

```jsx
<Provider
    store> // <--
    <App
        foo
    />
</Provider>
```

Resulting error:

`test.jsx:2:10: Ereact/jsx-closing-bracket-location The closing bracket must be aligned with the opening tag (expected column 1 on the next line)`

Note that the reported error location now corresponds to the closing `>` bracket on line 2, and that the error message now includes the exact location of where that closing bracket should have been.

Test Plan:
`npm install`
`npm run test`

### DIFF
--- a/lib/rules/jsx-closing-bracket-location.js
+++ b/lib/rules/jsx-closing-bracket-location.js
@@ -9,7 +9,7 @@
 // ------------------------------------------------------------------------------
 module.exports = function(context) {
 
-  var MESSAGE = 'The closing bracket must be {{location}}';
+  var MESSAGE = 'The closing bracket must be {{location}}{{details}}';
   var MESSAGE_LOCATION = {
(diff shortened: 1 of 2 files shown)


yannickcr:
Great! Thanks for this!
> Parsing the error message should now be sufficient to allow writing an external autofixer.

Actually since ESLint now have an [API to apply fixes](http://eslint.org/docs/developer-guide/working-with-rules#applying-fixes) we could directly add the autofixer in the rule. There is already an issue for this (#245) but I did not yet had the time to do it.

PR #306, opened 2015-11-11
jsx-pascal-case

jakemmarsh:
Add a new rule to enforce PascalCase for user-defined JSX components


docs/rules/jsx-pascal-case.md
```diff
@@ -0,0 +1,37 @@
...
+
+```js
+<TestComponent>
+  <div />
+</TestComponent>
+```
+
+## When Not To Use It
+
+If you are not using JSX, or if you don't want your JSX components defined and referenced in PascalCase.
```
TheSavior: React will only work if user-land components are defined in PascalCase, otherwise it converts them to strings:

```
<foo />
<Foo />
```

becomes
```
  React.createElement("foo", null)
  React.createElement(Foo, null)
```

so perhaps this should just say

```
If you are not using JSX.
```

jakemmarsh: Good point, updated

README.md
```diff
@@ -98,6 +98,7 @@ Finally, enable all of the rules that you would like to use.
 * [jsx-sort-props](docs/rules/jsx-sort-props.md): Enforce props alphabetical sorting
 * [jsx-uses-react](docs/rules/jsx-uses-react.md): Prevent React to be incorrectly marked as unused
 * [jsx-uses-vars](docs/rules/jsx-uses-vars.md): Prevent variables used in JSX to be incorrectly marked as unused
+* [jsx-pascal-case](docs/rules/jsx-pascal-case.md): Enforce PascalCase for user-defined JSX components
```
Daniel15: Alphabetical order please :smile:

yannickcr:
Ok for me. Can you set the rules in alphabetical order in the list then squash your commits ? Thanks.

PR #308, opened 2015-11-11
no-unknown-property should not check custom elements

zertosh:
React 0.14 has support for custom elements and their attributes are passed through. It's discussed in https://github.com/facebook/react/issues/4933. The custom element detection is similar to the one done in https://github.com/facebook/react/blob/b4b1add/src/renderers/dom/shared/ReactDOMComponent.js#L493-L495


zertosh:
ping @yannickcr

yannickcr:
Sorry for the delay, I will look at it this week end.

yannickcr:
Merged, thanks!

PR #315, opened 2015-11-17
jsx-handler-names

jakemmarsh:
Add a new rule to enforce prefixing of event handler methods, both on components and when passed as props.


yannickcr:
Looks good to me. Can you squash your commits? Thanks.

jakemmarsh:
Done! @yannickcr

yannickcr:
Merged, thanks!

PR #321, opened 2015-11-18
Fix crash on incomplete class property declaration

### LINKED ISSUE #317: Crash while declaring a static class member
This is the file. It crashes on line 8.
http://pastebin.com/mW38qrr2

Edit: The bug can be reproduced with a much simpler file:
```jsx
import React from 'react';

export default class Test extends React.Component {
  displayName
};
```

This is my eslintrc:
http://pastebin.com/9HNcDhPL

And this is the exception:
screen shot 2015-11-18 at 11 07 00

dapetcu21:
Fixes #317


yannickcr:
Thanks!

dapetcu21:
You're welcome 😊

PR #324, opened 2015-11-20
Fix crash when using eslint 1.10.0

### LINKED ISSUE #323: breaks with ESlint 1.10.0
Breaks with eslint 1.10.0
Runs fine with eslint 1.9.0
1__tmux

eslint-plugin-react 3.9.0
node 4.2.1
npm 2.14.7

lukekarrys:
Instead of putting the `react` object on `context`, this PR creates it as a separate variable and passes it as a 3rd parameter to the fn passed to `Components.detect`.

fixes #323

### DIFF
--- a/lib/rules/display-name.js
+++ b/lib/rules/display-name.js
@@ -10,7 +10,7 @@ var Components = require('../util/Components');
 // Rule Definition
 // ------------------------------------------------------------------------------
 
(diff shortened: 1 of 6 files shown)


begizi:
Same here. Great job! :+1:

yannickcr:
Wow that was fast!

yannickcr:
Merged, thanks!

PR #325, opened 2015-11-21
Add `jsx-pascal-case` to complete list of rules

briandipalma:
The `jsx-pascal-case` isn't in the example complete config.


yannickcr:
Thanks!

PR #332, opened 2015-11-24
[jsx-key] Improve and fix false-positives

### LINKED ISSUE #320: False `Missing "key" prop for element in iterator` from jsx-key
First of all, great work on the jsx-key rule. I think this will save people a lot of time.

Now, I have the following code:

```jsx
<TransitionMotion
  styles={this._getStyles()}
  willEnter={this._willEnter}
  willLeave={this._willLeave}
  >
  {styles =>
    <div>
      {Object.keys(styles).map(key =>
        <Toast
          action={styles[key].action}
          key={key}
          message={styles[key].message}
          opacity={styles[key].opacity}
          translateY={`${styles[key].translateY}%`}
        />
      )}
    </div>
  }
</TransitionMotion>
```

using the jsx-key rule, I am getting the missing key prop on the line that has the opening `<div>`. I might be misunderstanding when `key` should be used, but I don't think I need one here. And, if I do, then the error message that this rule outputs is misleading since it claims that I am in an iterator when I am not.

I noticed in the documentation:

> Also, if you have some prevalent situation where you use arrow functions to return JSX that will not be held in an iterable, you may want to disable this rule.

which I believe is referring to this situation, but I figured I'd open an issue for this anyway in case there happens to be a solution lurking.

silvenon:
Currently only failing tests because I'm new to developing ESLint rules, so it'll take me a couple of days.

I was planning to do it by detecting if the calling method is `map`. This would apply to `Array.prototype.map` as well as `React.Children.map`. Are there any other cases where the rule should be looking for `key`?

Will fix #320.

### DIFF
--- a/lib/rules/jsx-key.js
+++ b/lib/rules/jsx-key.js
@@ -4,32 +4,69 @@
  */
 'use strict';
 
+// var Components = require('../util/Components');
+
 // ------------------------------------------------------------------------------
 // Rule Definition
 // ------------------------------------------------------------------------------
 
 module.exports = function(context) {
 
-  function isKeyProp(decl) {
-    if (decl.type === 'JSXSpreadAttribute') {
-      return false;
+  function hasKeyProp(node) {
+    return node.openingElement.attributes.some(function(decl) {
+      if (decl.type === 'JSXSpreadAttribute') {
+        return false;
+      }
+      return (decl.name.name === 'key');
+    });
+  }
+
+  function checkIteratorElement(node) {
+    if (node.type === 'JSXElement' && !hasKeyProp(node)) {
+      context.report(node, 'Missing "key" prop for element in iterator');
     }
-    return (decl.name.name === 'key');
+  }
+
+  function getReturnStatement(body) {
+    return body.filter(function(item) {
+      return item.type === 'ReturnStatement';
+    })[0];
   }
 
   return {
     JSXElement: function(node) {
-      if (node.openingElement.attributes.some(isKeyProp)) {
-        return; // has key prop
+      if (hasKeyProp(node)) {
+        return;
       }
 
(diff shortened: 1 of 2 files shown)


yannickcr:
> Currently only failing tests because I'm new to developing ESLint rules, so it'll take me a couple of days.

No problem :)

> I was planning to do it by detecting if the calling method is `map`.

Seems to be a good solution. Right now I don't see any other cases, but I prefer to miss some edge cases than having some false errors.

benmosher:
I wondered if this would come up. Couldn't imagine any cases for arrow functions in non-iterators when I wrote the rule, but the examples from #320 make sense.

silvenon:
Sorry for being insanely late :sweat_smile:

Valid cases:

* `[<App key={0} />, <App key={1} />];`
* `[1, 2, 3].map(function(x) { return <App key={x} /> });`
* `[1, 2, 3].map(x => <App key={x} />);`
* `[1, 2, 3].map(x => { return <App key={x} /> });`
* `var App = () => <div />;`

Invalid cases:

* `[<App />];`
* `[<App {...key} />];`
* `[<App key={0}/>, <App />];`
* `[1, 2 ,3].map(function(x) { return <App /> });`
* `[1, 2 ,3].map(x => <App />);`
* `[1, 2 ,3].map(x => { return <App /> });`

silvenon:
Note sure why coverage fell, it seems to have fallen in `/lib/rules/jsx-pascal-case.js` for some reason.

yannickcr:
Seems due to a missing test for a case in `jsx-key` https://coveralls.io/files/1701759218#L50

(for `jsx-pascal-case`, it already had a bad coverage, my fault!)

silvenon:
Done :smiley:

yannickcr:
Merged, thanks :)

PR #333, opened 2015-11-24
Allow consecutive uppercase letters in jsx-pascal-case

### LINKED ISSUE #328: Consecutive capital letters in jsx-pascal-case
I like the new pascal case rule, but I've run into a couple of cases where I have consecutive capital letters and am wondering if perhaps this should be allowed. For instance:

```jsx
<CSSTransitionGroup>
```

lencioni:
I've run into a couple of cases where I have consecutive capital letters
in React components (e.g. <CSSTransitionGroup>), and jsx-pascal-case is
complaining about them. This commit fixes this problem by allowing
consecutive uppercase letters.

Fixes #328

### DIFF
--- a/docs/rules/jsx-pascal-case.md
+++ b/docs/rules/jsx-pascal-case.md
@@ -20,6 +20,10 @@ The following patterns are considered warnings:
 <test_component />
 ```
 
+```js
+<YELLING />
+```
+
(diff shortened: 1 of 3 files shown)


yannickcr:
Merged, thanks!

PR #336, opened 2015-11-27
Add option to sort shorthand props on jsx-sort-props

lucasmotta:
You can now require that shorthand props should come before all other props, but still respecting the alphabetical order.

---

The following are valid situations:
```jsx
<Component a b="b" />
<Component x z a="a" b="b" />
```
And the following are invalid:
```jsx
<Component z x a="a" b="b" /> // should be in alphabetical order
<Component x z b="b" a="a" /> // should be in alphabetical order
<Component a="a" b="b" x z /> // shorthand props should be at the start
```
---
- [x] Tests
- [x] Backwards compatibility


yannickcr:
Great, thanks!

PR #344, opened 2015-12-01
fix: rules/prop-types crash

joeybaker:
Adds a defensive check to prevent a crash:

```
eslint-plugin-react/lib/rules/prop-types.js:403
        for (var i = 0, j = node.id.properties.length; i < j; i++) {
                                              ^

TypeError: Cannot read property 'length' of undefined
    at markPropTypesAsUsed (eslint-plugin-react/lib/rules/prop-types.js:403:47)
    at EventEmitter.VariableDeclarator (eslint-plugin-react/lib/rules/prop-types.js:577:7)
    at emitOne (events.js:82:20)
    at EventEmitter.emit (events.js:169:7)
    at NodeEventGenerator.enterNode (eslint/lib/util/node-event-generator.js:42:22)
    at CommentEventGenerator.enterNode (eslint/lib/util/comment-event-generator.js:98:23)
    at Controller.controller.traverse.enter (eslint/lib/eslint.js:767:36)
    at Controller.__execute (estraverse/estraverse.js:397:31)
    at Controller.traverse (estraverse/estraverse.js:495:28)
    at EventEmitter.module.exports.api.verify (eslint/lib/eslint.js:764:24)
```

### DIFF
--- a/lib/rules/prop-types.js
+++ b/lib/rules/prop-types.js
@@ -400,6 +400,10 @@ module.exports = Components.detect(function(context, components, utils) {
         }
         break;
       case 'VariableDeclarator':
+        if (!node.id || !node.id.properties) {
+          break;
+        }
+
         for (var i = 0, j = node.id.properties.length; i < j; i++) {
           // let {props: {firstname}} = this
           var thisDestructuring = (
--- a/lib/rules/prop-types.js
+++ b/lib/rules/prop-types.js
@@ -400,9 +400,9 @@ module.exports = Components.detect(function(context, components, utils) {
         }
         break;
       case 'VariableDeclarator':
-        if (!node.id || !node.id.properties) {
-          break;
-        }
+        // if (!node.id || !node.id.properties) {
+        //   break;
+        // }
 
         for (var i = 0, j = node.id.properties.length; i < j; i++) {
           // let {props: {firstname}} = this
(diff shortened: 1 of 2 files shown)


yannickcr:
Can you also add a test that reproduce the crash? Thanks.

joeybaker:
@yannickcr I could use some help with that. This error is coming from using `eslint-plugin-react` on decently sized code base. The stack trace gives very little information about what code is causing the crash. I can confirm this fixes the problem, but could use some advice: how would I got about tracking the input code that causes this crash?

craigbilner:
Joey,

I've used iron-node and ascertained so far that any function block which tries to assign a variable called props causes the issue:

```
var a = function() {
  var newProps1 = props1;
  const newProps2 = props2;
  let newProps3 = props3;
};

let b = function() {
  var newProps4 = props;
  const newProps5 = props;
  let newProps6 = props;
};

const c = function() {
  var newProps7 = props;
  const newProps8 = props;
  let newProps9 = props;
};

function d() {
  var newProps10 = props;
  const newProps11 = props;
  let newProps12 = props;
};

const e = () => {
  var newProps13 = props;
  const newProps14 = props;
  let newProps15 = props;
};
```

In the code above newProps1/2/3 don't fall into the "VariableDeclarator"

A cursory search of https://github.com/shakacode/react-webpack-rails-tutorial raised as an issue would support this.

joeybaker:
@craigbilner iron-node is very cool! I looks like the successor to node-inspector?

Thanks for tracking that down. I'll play around and see if I can add a test. That's quite an edge case!

joeybaker:
@yannickcr Thanks to @craigbilner's sleuthing, test added. It's a bit unlike the others, but it's apparently necessary :)

craigbilner:
@joeybaker,

does this fix the root cause?

To me this is the culprit

```
var statelessDestructuring = node.init && node.init.name === 'props' && utils.getParentStatelessComponent();
```

The example given is this:

```
let {firstname} = props
```

yet it doesn't test for the parent being an "ObjectPattern" and therefore having properties?

joeybaker:
@craigbilner I'm still getting used to the AST api. Let me try and report back.

joeybaker:
@craigbilner I gave this a shot, but I'm out of time to dig into this more. I pushed up what I believe to be your suggestion, but it breaks a different test. If you have a chance, can you take a look at this diff and tell me what I did wrong?

joeybaker:
@yannickcr If you'd rather merge it with the other fix, let me know. I can delete the most recent commit and the tests will pass.

yannickcr:
Thanks for all your help. I was able to find the problem and fix it in be3b7f30abc4ef2d61acef2692e27cfb9e1ef792.

joeybaker:
Great! Glad it got fixed correctly.

PR #347, opened 2015-12-02
Add missing JSDoc `@param` in variable.js

lencioni:
While looking into a bug, I noticed that there was a missing `@param`
here.


yannickcr:
Thanks!

PR #348, opened 2015-12-02
make jsx-handler-names error messages less specific

jakemmarsh:
I realized the error messages were previously specifying a more specific method name than the rule actually enforced. 

For example, `onChange={this.itemClicked}` would throw the warning `Handler function for onChange prop key must be named handleChange`, when really `this.handleItemClick` would (and should) pass.


jakemmarsh:
Not sure why the checks have failed?

yannickcr:
Thanks.

> Not sure why the checks have failed?

Just a very small coverage decrease because you removed some tested lines, not an error for me.

PR #349, opened 2015-12-03
Add `pragma` config option to `react-in-jsx-scope`

nealgranger:
Update the `react-in-jsx-scope` rule to match the behaviour of `jsx-uses-react`, allowing a custom `jsxPragma`.
```js
...
"react-in-jsx-scope": [2, { "pragma": "createElement" }]
...
```
Fixes: https://github.com/yannickcr/eslint-plugin-react/issues/228

### DIFF
--- a/docs/rules/react-in-jsx-scope.md
+++ b/docs/rules/react-in-jsx-scope.md
@@ -35,6 +35,26 @@ var Foo = require('foo');
 var Hello = <div>Hello {this.props.name}</div>;
 ```
 
+## Rule Options
+
+```js
+...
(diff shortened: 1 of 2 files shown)


yannickcr:
Sorry for the delay, I'm searching for a more generic solution to be able to set a custom pragma to all the rules at once.

yannickcr:
Fixed in 4cb2e9e

Now all rules relying on component detection (`display-name`, `prop-types`, etc.) or on the `React` variable (`jsx-uses-react`, `react-in-jsx-scope`, etc.) are using the shared settings.

PR #356, opened 2015-12-09
add rule to prevent deprecated use of React.render, etc.

graue:
This is a little rule that we've found useful to keep `React.render()`s and the like out of our code after [upgrading to React 0.14](https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#two-packages-react-and-react-dom).

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ Finally, enable all of the rules that you would like to use.
     "react/jsx-uses-react": 1,
     "react/jsx-uses-vars": 1,
     "react/no-danger": 1,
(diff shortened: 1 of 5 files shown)


yannickcr:
Thanks, I was making a [similar rule](https://github.com/yannickcr/eslint-plugin-react/tree/rule-no-deprecated) but I never finished it.

I'll try to merge the two.

yannickcr:
I merged the two rules, the `no-deprecated` rule should be available in next release. Thanks again!

PR #359, opened 2015-12-13
allow es6 class rule to be configured to discourage usage as well

pwmckenna:
Doesn't change the default behavior, but allows you to specify 'never' as an option that recommends createClass over using es6 classes.

[The argument for switching is still a little murky](http://reactjsnews.com/composing-components/). In the meantime its nice to be able to force it in either case.

### DIFF
--- a/lib/rules/prefer-es6-class.js
+++ b/lib/rules/prefer-es6-class.js
@@ -11,11 +11,14 @@ var Components = require('../util/Components');
 // ------------------------------------------------------------------------------
 
 module.exports = Components.detect(function(context, components, utils) {
+  var configuration = context.options[0] || 'always';
 
   return {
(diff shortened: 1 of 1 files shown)


yannickcr:
Why not.

Maybe I'll rename this rule in next major version to be more explicit about the different styles.

remitbri:
Documentation on the [rule page](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prefer-es6-class.md) needs an update

remitbri:
Suggestion: in the nitpicking matter of `[levelOfWarning, "value"]` vs. `[levelOfWarning, {key : "value"}]`, perhaps the rule could/should be `[levelOfWarning, {"frequency" : "always/never"]` ?

PR #361, opened 2015-12-15
add 'forbidExplicitDeclaration' option for display-name rule

jakemmarsh:
When using JSX and accepting the transpiler name for `displayName`, it may be preferable to then forbid explicit `displayName` definitions. This adds that option to the rule


docs/rules/display-name.md
```diff
@@ -85,6 +88,54 @@ function HelloComponent() {
 module.exports = HelloComponent();
 ```
 
+### `forbidExplicitDefinition`
+
+When `true` the rule will not accept any explicitly defined `displayName` property, allowing for the name to be set by the transpiler.
```
TheSavior: nit: 

```
to **only** be set by the transpiler.
```

jakemmarsh: added

lib/rules/display-name.js
```diff
@@ -191,10 +206,13 @@ module.exports = Components.detect(function(context, components, utils) {
       var list = components.list();
       // Report missing display name for all components
       for (var component in list) {
-        if (!list.hasOwnProperty(component) || list[component].hasDisplayName) {
+        if (forbidExplicitDefinition && list[component].hasDisplayName) {
```
yannickcr: The `hasDisplayName` property is `true` if we find an explicit displayName or if the transpiler can find one.

I think we should differentiate the 2 "types" of names and only warn if this is an explicit displayName.

yannickcr:
Hi, and sorry for the review delay.

Thanks for this useful option, I'm totally ok to add it to the `display-name` rule.

But I found a small issue, since the `hasDisplayName` property is also `true` if the transpiler can find a name, this option will warn in all cases, with or without an explicit displayName.

Failing example:
```js
  render() {
  }
}
```
Will warn:
```
Explicit display name definition is forbidden
```

jakemmarsh:
No worries about the delay @yannickcr. I'll work on differentiating between the two types of names ASAP

ljharb:
@jakemmarsh altho it’s been awhile, are you still interested in merging this?

Issue #341, opened 2015-12-01
New rule proposal: Check if `this.refs.someRef` properties are actually defined


Intellicode:
I had a few occasions myself that some of my components had typo's or missing reference attributes in components causing crashes. 

```
const Hello = React.createClass({
  onFocus() {
    const scrollResponder = this.refs.scrollView.getScrollResponder();
    ...
  }

  render() {
     return <ScrollList />
  }
});
```

jimbolla:
Just a heads up... I remember reading (but can't seem to find) a mention that `this.refs` is eventually going to be phased out by the React team. I believe the reasoning is that `ref=` is basically an escape hatch that breaks encapsulation and its use should be minimized. To make the code simpler, they'll phase out `ref="thing"` going into `this.refs.thing` and instead only support the function-based `ref={c => this.thing = c}` style. I can't seem to find the github issue where this was discussed, but I remember reading it in the past few months.

A better rule might be to disallow use of `this.refs`.

Intellicode:
Good point, I'll do some research and if I come across the discussion you mentioned, I'll reference it here.

Intellicode:
"React also supports using a string (instead of a callback) as a ref prop on any component, although this approach is mostly legacy at this point." https://facebook.github.io/react/docs/more-about-refs.html so indeed a rule to prevent using this.refs would make more sense.

PR #362, opened 2015-12-15
Prevent string literals being used in ref attributes

Intellicode:
Initially, the idea in #341 was to check if references in `this.refs` exist. Based on further discussion it was suggested to define a rule to disallow using string literals in reference properties.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ Finally, enable all of the rules that you would like to use.
     "react/no-direct-mutation-state": 1,
     "react/no-multi-comp": 1,
     "react/no-set-state": 1,
+    "react/no-string-refs": 1,
     "react/no-unknown-property": 1,
     "react/prefer-es6-class": 1,
     "react/prop-types": 1,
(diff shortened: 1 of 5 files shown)


yannickcr:
Hi, thanks for the rule.

2 minors things though:
* Maybe a less generic rule name (like `no-string-refs` ?), `no-refs` seems to forbid all types of refs.
* Can you add some titles in the documentations? See [other rules documentation](https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules) for example.

Beside that I'm totally ok to add this rule :)

Intellicode:
Sure! I'll add them tomorrow :)

Intellicode:
@yannickcr I've implemented your suggestions, let me know if there is anything else :)

yannickcr:
Just missing the title in docs/no-string-refs.md
Can you add it and squash your commits? Then I think we'll be good to go :)

Intellicode:
@yannickcr Sorry, my latest change to adding titles was not pushed correctly, should be good now and also I squashed all commits :)

yannickcr:
Perfect, thanks!

Issue #37, opened 2015-03-29
Rule proposal: no reference to `this.isMounted`in components


lencioni:
I've seen some people write code like `if (this.isMounted)` instead of `if (this.isMounted())`. It might be nice to have a rule that notices the former and encourages the latter.

Good:

```jsx
if (this.isMounted()) {
  ...
}
```

Bad:

```jsx
if (this.isMounted) {
  ...
}
```

glifchits:
It seems like this would be a feature in ESLint, not just this React plugin. Basically this is a warning/error when you reference a function instead of calling it. I tried searching the ESLint docs for it but I couldn't find anything. I don't know much about ESLint but is it even capable of knowing a variable's type? It seems like quite a challenge, esp. when you consider that we are concerned with object methods (eg. `this.isMounted`), and globals (which are often functions). Now that I think of it, the `global` config that is available in ESLint makes no distinction as to its type, so it seems quite unlikely that this feature can exist. Am I right?

lencioni:
> Basically this is a warning/error when you reference a function instead of calling it.

A general rule like that might be useful, but I think it would be difficult to implement. This request is specifically for `this.isMounted` in React components.

tleunen:
Not sure it's worth to work on that one...
> This method is not available on ES6 class components that extend React.Component. It may be removed entirely in a future version of React.
https://facebook.github.io/react/docs/component-api.html#ismounted

lencioni:
Indeed, it looks like `isMounted` is going away. https://github.com/facebook/react/issues/5465

Perhaps there should be a rule that prevents the use of it entirely?

PR #366, opened 2015-12-19
Add no-is-mounted rule

lencioni:
isMounted is an anti-pattern [0], is not available when using ES6
classes, and will eventually be officially deprecated. This commit adds
a rule that prevents the use of this method.

[0]: https://facebook.github.io/react/blog/2015/12/16/ismounted-antipattern.html

Finishes #37


yannickcr:
Merged, thanks!

PR #368, opened 2015-12-22
Fix link in no-is-mounted documentation

lencioni:
When I made this link, I used the wrong markdown syntax so it was
linking to "anti-pattern" instead of the intended URL.


yannickcr:
Thanks!

PR #372, opened 2015-12-24
sort-comp: say "should" instead of "must"

SystemParadox:
```
warning: componentWillUpdate must be placed after componentDidMount
```

Sorting component methods is only ever a style preference. "must" is a bit misleading - it implies that there will be errors in the program. Say "should" instead.

Thanks.

### DIFF
--- a/lib/rules/sort-comp.js
+++ b/lib/rules/sort-comp.js
@@ -42,7 +42,7 @@ module.exports = Components.detect(function(context, components) {
 
   var errors = {};
 
-  var MISPOSITION_MESSAGE = '{{propA}} must be placed {{position}} {{propB}}';
+  var MISPOSITION_MESSAGE = '{{propA}} should be placed {{position}} {{propB}}';
 
   var methodsOrder = getMethodsOrder({
     order: [
--- a/tests/lib/rules/sort-comp.js
+++ b/tests/lib/rules/sort-comp.js
@@ -205,7 +205,7 @@ ruleTester.run('sort-comp', rule, {
       '});'
     ].join('\n'),
     parserOptions: parserOptions,
-    errors: [{message: 'render must be placed after displayName'}]
+    errors: [{message: 'render should be placed after displayName'}]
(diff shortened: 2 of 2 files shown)


mjomble:
It seems you'll need to update tests as well to get the Travis CI build to pass.

SystemParadox:
Oops my bad. Should be sorted now. Thanks.

yannickcr:
Mmm. Not really sure about this change or, to be more precise, about the reasons for it :)

Imho it depends how the rule is configured, if configured as warning then `should` is more appropriate, but if configured as error then it is more likely `must` (since ESLint will exit with code 1 in this case).

After having a look at ESLint rules, it appears that there is no real rules for messages, `should` and `must` are used the same way, whether it is for styling rules or not.

(Same for us, due to the different rules authors and no rules about messages formatting, but that something I'd like to improve in future releases)

SystemParadox:
My reasoning is that "must" is absolutely not appropriate for warnings as it is misleading. The reverse is not true - "should" can also be used for errors without causing confusion. "must" is very definite, whereas "should" is a more lenient and flexible term.

Additionally:

- eslint is primarily and linting and style tool. It is far more likely that rules will be configured as warnings rather than errors, as they are not essential for program correctness
- even rules configured as errors are not necessarily fatal, depending how eslint is being used. This is the case in my setup, where eslint is run after the normal compilation step.
- the sort-comp rule in particular makes very little sense as an error and is likely to only ever be used as a warning

I haven't noticed any other instances yet, but at the moment this rule is causing confusion both for me and my coworkers.

Thanks.

yannickcr:
> My reasoning is that "must" is absolutely not appropriate for warnings as it is misleading. The reverse is not true - "should" can also be used for errors without causing confusion. "must" is very definite, whereas "should" is a more lenient and flexible term.

That's a sufficient reason for me :)

Thanks for the suggestion, do not hesitate to suggest other improvements to the error messages ;)

Issue #373, opened 2015-12-24
3.13.0 jsx-keys Errors


craigklem:
Using latest version 3.13.0, I see several issues when exceptions in jsx-keys when running eslint with our code. Errors 2 & 3 happen after adding if checks to avoid the exceptions. Could not identify code that made 3 happen - no WebStorm call stack :(

1.  TypeError: Cannot read property 'name' of undefined
line 49 - if (node.callee.property.name !== 'map') {
2. TypeError: Cannot read property 'type' of undefined
line 54 - var isFn = fn.type === 'FunctionExpression';
3. TypeError: Cannot read property 'argument' of undefined
line 64 - getReturnStatement(fn.body.body).argument

Code that caused issues:
<pre>
class MyComponent extends React.Component {
constructor(props) {
    super(props);//number 1 happens here
}
var IconRenderer= function() {
  this.fontsLoaded = false;
  this.checkFonts = function(loadedCallback) {
    this.fontsLoaded = lib.fontsLoaded();

    if (this.fontsLoaded) {
      if (loadedCallback) {
        loadedCallback();//number 2 happens here
      }
}
</pre>
WebStorm watches below:
![image](https://cloud.githubusercontent.com/assets/7660135/11996736/d6ff37f6-aa37-11e5-94a9-6a2d580a5fc8.png)
![image](https://cloud.githubusercontent.com/assets/7660135/11996927/9e4d7d14-aa3c-11e5-8fb6-2e1242d250a0.png)

PR #374, opened 2015-12-24
fix crashes in jsx-key

lukekarrys:
I was having the same crashes as #373 when using `jsx-key`.

It was triggering for non jsx stuff, so I added a few "valid" tests which don't actually test anything except that the rule doesn't crash. I'm not sure if its appropriate to have those tests in the specific rule test file, but I figured it would be good to prevent regressions.


yannickcr:
Thanks!

PR #380, opened 2016-01-01
fix another crash in jsx-key

nfcampos:
similar to PR #374

fix this crash:
```
TypeError: Cannot read property 'type' of null
    at checkIteratorElement (/Users/nuno/cp/james/node_modules/eslint-plugin-react/lib/rules/jsx-key.js:25:13)
    at EventEmitter.CallExpression (/Users/nuno/cp/james/node_modules/eslint-plugin-react/lib/rules/jsx-key.js:65:13)
```


yannickcr:
Thanks! I will add a test to confirm the fix.

PR #381, opened 2016-01-04
Docs: How to automatically sort-comp compliance

turadg:
There's now a jstransform codemod to do this automatically,
https://github.com/reactjs/react-codemod/blob/master/transforms/sort-comp.js

Added a note about it based on the pattern in http://eslint.org/docs/rules/no-trailing-spaces


yannickcr:
Wow! Great codemod!

Ideally it should be integrated to the rule, using the [fixable flag](http://eslint.org/docs/developer-guide/working-with-rules#applying-fixes) from ESLint (I should really get to work on #245).

But for now referencing it in the documentation could already be a great help for users.

Thanks!

PR #382, opened 2016-01-05
Support flow annotations for component props

phpnode:
This PR adds support for using [flow](http://flowtype.org/) type annotations for class props instead of `propTypes`. Flow added support for this quite recently in: https://github.com/facebook/flow/issues/850

The syntax looks like this:
```js
type Person = {
  name: {
    firstName: string;
    lastName: string;
  },
  age: number;
};
type Props = {
  greeting: string;
  people: Person[];
};
  props: Props;

  render () {
    const names = [];
    for (let i = 0; i < this.props.people.length; i++) {
      names.push(this.props.people[i].name.firstName);
    }
    return <div>{this.props.greeting}, {names.join(', ')}</div>;
  }
}
```

### DIFF
--- a/lib/rules/prop-types.js
+++ b/lib/rules/prop-types.js
@@ -19,8 +19,27 @@ module.exports = Components.detect(function(context, components, utils) {
   var configuration = context.options[0] || {};
   var ignored = configuration.ignore || [];
   var customValidators = configuration.customValidators || [];
+  // Used to track the type annotations in scope.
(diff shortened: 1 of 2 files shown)


yannickcr:
Wow! This is really great!

Can you squash your commits ? Then I think we'll be good to go :)

phpnode:
@yannickcr squashed

yannickcr:
Merged, thanks!

phpnode:
awesome, thanks!

Issue #357, opened 2015-12-09
[jsx-no-bind] add option to allow it in stateless components


silvenon:
For stateless components you don't really have a choice, do you? I see Dan Abramov use arrow functions all the time in stateless components. If Dan Abramov jumps of a cliff, it's ok for us to do it, right? :grin:

ALF-er:
Also, allow `ref={c => this.myComponent = c}` case please

eadz:
This would be great. Currently the following is allowed

```jsx
  <ButtonBlock onPress={function () { onContinue(8); }} text={'I Agree'} color={"black"}/>
```

but this is not:

```jsx
  <ButtonBlock onPress={() => onContinue(8) } text={"I Agree"} color={"black"}/>
```

silvenon:
@eadz really? I don't understand why the former is allowed :confused: 

Anyway, I since changed my mind about this issue. I don't see a reason for this level of control, we already have the `allowArrowFunctions` option.

silvenon:
@eadz oh, I see now. Well, [let's see](https://babeljs.io/repl/#?experimental=true&evaluate=true&loose=true&spec=true&code=React.createClass(%7B%0A%20%20render()%20%7B%0A%20%20%20%20return%20(%0A%20%20%20%20%20%20%3Cdiv%0A%20%20%20%20%20%20%20%20onClick%3D%7B()%20%3D%3E%20%7B%0A%20%20%20%20%20%20%20%20%20%20this.foo()%0A%20%20%20%20%20%20%20%20%7D%7D%0A%20%20%20%20%20%20%2F%3E%0A%20%20%20%20)%0A%20%20%7D%0A%7D)%3B%0A)… Is using an anonymous function directly bad? Does that create a new function on each render?

1. If yes, this should also be an error, bind or no bind.
2. If yes, this rule should only apply to bind, not arrow functions too.

Either way it seems like something needs to change. I don't know how to check whether a new function is created every time if you pass an anonymous function.

Daniel15:
@eadz - That's an oversight, the `function() { }` syntax should be caught by the lint rule too. I didn't implement that in the lint rule since I haven't actually seen that style in any live code.

In any case, you can just disable the lint rule if you don't agree with it :stuck_out_tongue:

ljharb:
@silvenon yes, creating any function inside `render`, including an arrow, is bad, because it creates a new one on each render.

[...]

PR #385, opened 2016-01-07
Add ignoreRefs to the jsx-no-bind rule

### LINKED ISSUE #330: [jsx-no-bind] Should allow arrow functions in ref property
React documentation suggests using callbacks instead of string value in ref:
https://facebook.github.io/react/docs/more-about-refs.html#the-ref-string-attribute

silvenon:
Not sure how you want to go about this. I first wanted to allow `ref` callbacks always, but it's still a bad practice for the same reasons, so I thought `allowRefs` option would be nice? However, it might be confusing that `allowArrowFunctions` and `allowBind` affects `ref` as well.

Also, I added other missing options to the docs. Is `|| false` a good way to document that it defaults to `false`?

Fixes #330.
Refs #357.

### DIFF
--- a/docs/rules/jsx-no-bind.md
+++ b/docs/rules/jsx-no-bind.md
@@ -18,6 +18,41 @@ The following patterns are not considered warnings:
 <div onClick={this._handleClick}></div>
 ```
 
+## Rule Options
+
+```js
+"jsx-no-bind": [<enabled>, {
+  "ignoreRefs": <boolean> || false,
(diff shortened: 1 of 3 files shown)


yannickcr:
Maybe `ignoreRefs` would be less confusing name for this option ?

Also, thanks for the doc update. I think the `|| false` is ok to document the default value.

silvenon:
Good idea, done!

yannickcr:
Merged, thanks!

PR #386, opened 2016-01-07
Add a new rule: filename matches component

tmcw:
For named components, this enforces that the camelcase component
name matches the underscore file name.


ljharb:
In Airbnb's case, we want the camelcase component name to match the *camelcase* file name - I'd want this to be enforceable too, regardless of the default.

However, this seems like a generic rule: that the name of the default export of a file matches the filename - either camelcase or snake_case - rather than a React-specific rule.

PR #411, opened 2016-01-26
Fix styleguide violations

luqmaan:
Fix two violations of the React styleguide: `_` underscore prefixing internal methods, and placement of the `render` function.

> Ordering for class extends React.Component:

https://github.com/airbnb/javascript/tree/master/react#ordering

> Do not use underscore prefix for internal methods of a React component.

https://github.com/airbnb/javascript/tree/master/react#methods


luqmaan:
Haha, I'm dumb. I though this was the Airbnb styleguide. Sorry for enforcing these rules on you.

Issue #244, opened 2015-10-12
Rule proposal: Space before closing bracket


kentor:
There should be some rule that warns when there isn't a space before a closing bracket: `<div/>` and does not warn when there is a space: `<div />`

PR #417, opened 2016-01-29
Add jsx-space-before-closing rule

ryym:
I implemented a new rule proposed at #244.
The rule simply enforces a space before `/>` on self closing elements like `<div />`.

* no option
* fixable


yannickcr:
Thanks. I think I'll add an option to inverse the rule too, with `always`/`never` like in the [jsx-boolean-value](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md) rule.

ryym:
Thanks! :smiley:

PR #428, opened 2016-02-05
Update babel-eslint 5.0.0-beta6 -> 5.0.0-beta10

lencioni:
The current version gives me the following error when I try to run tests
on master:

> TypeError: Cannot read property 'visitClass' of undefined

It looks like this has since been resolved by the related packages, so
we can avoid this problem by bumping versions.

More info: https://github.com/babel/babel-eslint/issues/243


lencioni:
Oops, looks like this was already fixed. 270959b3d26c90fe32f8c8a0e0239f9233bc1504

PR #429, opened 2016-02-05
Teach sort-comp rule about static class methods

lencioni:
A number of people have said that they think it makes sense for static
class methods to appear at the very top of classes. This PR allows
this to be configured by adding the `static-methods` keyword to the
sort-comp rule.

While I was at it, I fixed a couple of typos and made a minor perf tweak.

Fixes #128

PR #430, opened 2016-02-06
Teach prop-types about destructing in function signatures

### LINKED ISSUE #354: prop-types doesn't complain about destructured props in stateless functions
The following should cause an error for missing proptype validations, but it doesn't.
```
const Test = ({ name }) => {
  return (
    <div>{name}</div>
  );
};
```

If you replace the destructuring with using the props object, it works as expected.

lencioni:
The following should cause an error for missing propType validations,
but it doesn't.

```jsx
  const Test = ({ name }) => {
    return (
      <div>{name}</div>
    );
  };
```

If you replace the destructuring with using the props object, it works
as expected. This commit fixes this problem.

Fixes #354.

### DIFF
--- a/docs/rules/prop-types.md
+++ b/docs/rules/prop-types.md
@@ -23,6 +23,10 @@ var Hello = React.createClass({
     return <div>Hello {this.props.firstname} {this.props.lastname}</div>; // lastname type is not defined in propTypes
   }
 });
+
+function Hello({ name }) {
+  return <div>Hello {this.props.name}</div>;
+}
 ```
 
 The following patterns are not considered warnings:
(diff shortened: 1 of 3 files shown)


ljharb:
(ﾉ◕ヮ◕)ﾉ*:・ﾟ✧ awesome!

yannickcr:
Merged, thanks!

ljharb:
@yannickcr yay! version bump soon? :-D

lencioni:
Thanks! If you are thinking about releasing a new version, it would be really swell if #429 could sneak in it as well.

PR #433, opened 2016-02-08
Remove deprecated jsx-quotes

lencioni:
This was deprecated by f817e37b in September 2015 and was first released
in v3.4.0. Now that we are on 3.16.x, I think it is safe to finally
remove this deprecated rule.


yannickcr:
Thanks, but since it's a breaking change I will wait a little before removing this rule.

I should release a new major version soon but I would like to do a bit of cleanup before.

lencioni:
Sure thing, makes sense!

Issue #436, opened 2016-02-09
v4 Roadmap


yannickcr:
There is a few things that I'd like to changes for a while (default configuration changes, rules renaming/suppression and a few others) and with the upcoming ESLint `v2.0.0` release I think it is the right time to do this.

So here's what I have in mind for now:

* [x] [ESLint 2.0.0 compatibility](http://eslint.org/docs/2.0.0/user-guide/migrating-to-2.0.0)
  * [x] Remove `rulesConfig` and export a shareable config (#192, [documentation](http://eslint.org/docs/2.0.0/user-guide/migrating-to-2.0.0#plugins-no-longer-have-default-configurations))
  * [x] Stop using the `context` deprecated methods and use `sourceCode` ([documentation](http://eslint.org/docs/developer-guide/working-with-rules#the-context-object))
* [x] Definitely remove `jsx-quotes` (PR #433)
* [x] Add static method support to `sort-comp` (#128, PR #429)
* [x] Rename `jsx-sort-prop-types` to `sort-prop-types`
 (#87, PR #431)
* [x] In `display-name`, ~~set the acceptTranspilerName option to true by default~~ invert `acceptTranspilerName` option and rename it `ignoreTranspilerName`.

Do you see anything else that could be good to change/improve?

ljharb:
v2.0.0 is now released - time to get that out ASAP :-D

yannickcr:
It's should be done during this week-end :)

yannickcr:
Just published `4.0.0-rc.0` on npm, you can test it by running:

```bash
$ npm install eslint-plugin-react@next
```

Let me know if you find any issue or if you have suggestions about the new recommended configuration.

lukekarrys:
Thanks @yannickcr. v4 looks pretty good to me.

I did find one bug that I opened here: https://github.com/yannickcr/eslint-plugin-react/issues/443. I think it is due to https://github.com/yannickcr/eslint-plugin-react/commit/b2661aae3146f137032b4a5f8a7aef5e51bc7e67, but I'm not sure how to differentiate stateless component functions from other functions with destructured arguments.

yannickcr:
Thanks @lukekarrys, the issue should be fixed in 6643f8775851d765ce922b09f1826d0053d870a6.

I will publish another RC tomorrow.

yannickcr:
Just published `4.0.0-rc.1` on npm.

lukekarrys:
[...]

PR #439, opened 2016-02-13
Default display-name's acceptTranspilerName to true

lencioni:
This was a change that has been queued for v4.

Addresses #436

### DIFF
--- a/lib/rules/display-name.js
+++ b/lib/rules/display-name.js
@@ -14,7 +14,9 @@ module.exports = Components.detect(function(context, components, utils) {
 
   var sourceCode = context.getSourceCode();
   var config = context.options[0] || {};
-  var acceptTranspilerName = config.acceptTranspilerName || false;
+  var acceptTranspilerName = config.hasOwnProperty('acceptTranspilerName') ?
+    config.acceptTranspilerName :
+    true;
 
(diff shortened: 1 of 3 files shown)


lib/rules/display-name.js
```diff
@@ -14,7 +14,9 @@ module.exports = Components.detect(function(context, components, utils) {
 
   var sourceCode = context.getSourceCode();
   var config = context.options[0] || {};
-  var acceptTranspilerName = config.acceptTranspilerName || false;
+  var acceptTranspilerName = config.hasOwnProperty('acceptTranspilerName') ?
```
ljharb: instead of making "absence" and "falsy" be different, would it be better as a breaking change to rename the option to something like "ignoreTranspilerName"?

lencioni: Yeah, I definitely like that better--I was just doing what @yannickcr had written in the plan. I'm going to be afk for the next few days starting soon, so I'll push another version and he can choose his favorite.

lencioni: New version is at https://github.com/yannickcr/eslint-plugin-react/pull/440

ljharb: :+1:

yannickcr:
Merged #440 instead.

PR #440, opened 2016-02-13
Change display-name's acceptTranspilerName to ignoreTranspilerName

lencioni:
This was a change that has been queued for v4. I considered changing the
default from false to true, but I think it makes more sense to invert
the option to be an ignore option.

Addresses #436


yannickcr:
> I think it makes more sense to invert the option to be an ignore option.

I agree, merged. Thanks!

ljharb:
yay

PR #441, opened 2016-02-13
[readme] Move JSX-specific rules to their own section

lencioni:
Now that this list has grown considerably, I think it would be helpful
to people to break it up a bit. A natural break is JSX-specific rules
and non-JSX-specific rules.


lencioni:
This will likely conflict with some of my other open PRs, so it might be best to merge those first.

yannickcr:
Good idea. Merged.

PR #456, opened 2016-02-19
False positive jsx key rule

jkimbo:
Identified and added a fix for a false positive from the `jsx-key` rule. Happens when you pass a function that returns a `JSXElement` as the first parameter.


yannickcr:
Merges, thanks!

ljharb:
Perhaps a v4.0.1 soon? :-)

yannickcr:
Yes, I'll try to publish this quicky, probably today or tomorrow.

PR #461, opened 2016-02-22
Moved the recommended config up.

thaggie:
For the plugin to work "out of the box" you really need to use the recommended config so I've moved it up directly under the configuration section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,23 @@ With ESLint 2.x.x:
 }
 ```
 
+# Recommended configuration
+
+This plugin export a `recommended` configuration that enforce React good practices.
+
+To enable this configuration use the `extends` property in your `.eslintrc` config file:
+
+```js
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -117,22 +134,8 @@ Finally, enable all of the rules that you would like to use.
...
-  "plugins": [
-    "react"
-  ],
-  "extends": "plugin:react/recommended"
-}
-```
-
-See [ESLint documentation](http://eslint.org/docs/user-guide/configuring#extending-configuration-files) for more information about extending configuration files.
 
 The rules enabled in this configuration are:
```
yannickcr: The enabled rule list part should also be moved. 

But, while I think it's a good idea to make the recommended configuration more visible, it will be a bit weird to have this rule list before the actual full rule list.

PR #462, opened 2016-02-23
Added a note about the recommended configuration at the bottom of the configuration section.

thaggie:
Rather than moving the recommended configuration I've added a link to it and specifically called out the issue I had as without that rule it will always fail.


yannickcr:
Thanks. can you squash your commits ?

thaggie:
Done

yannickcr:
Merged, thanks!

PR #465, opened 2016-02-24
Fix jsx-curly-spacing rule name in documentation

alex-tan:


alex-tan:
Nevermind, looks like this is a convention in the documentation. Submitted this because I was getting this error when copying and pasting:

Definition for rule 'jsx-curly-spacing' was not found  jsx-curly-spacing

Issue #474, opened 2016-03-01
jsx-no-bind circumventable via Short-Circuit Evaluation 


iilei:
example code `myHandler={true && this.myHandler.bind(this, key)}` results in *no warnings*, while I would expect it to throw *error  JSX props should not use .bind()  react/jsx-no-bind*

Is this a bug or am I missing something?

ljharb:
Similarly, I've noticed that `const foo = this.whatever.bind(this);` and then `prop={foo}` doesn't seem to warn either.

iilei:
@ljharb I assume  `const foo = this.whatever.bind(this);` resides outside of the returned block jsx, right? I don't think that is related

ljharb:
@iilei yes it is - no functions should be created in the entire render path and passed as props, which is what this rule is attempting to prevent.

iilei:
I see ... @petersendidit :+1:

PR #475, opened 2016-03-01
Catch bind in render when assigned to a variable (#474)

petersendidit:
Takes care of part of the issue brought up in #474.


yannickcr:
Merged, thanks!

PR #509, opened 2016-03-21
Updated description to provide more context

coryhouse:
I've found this rule really helpful on Flux/Redux apps that don't use forms. I typically only use setState when working with forms.


yannickcr:
Thanks!

PR #511, opened 2016-03-21
Autofix for jsx-closing-bracket-location.

KevinGrandon:
@yannickcr - I'm fairly new to eslint and the inner workings of this so this could be wrong, or there could be better ways of doing things. Let me know what you think, thanks!

### DIFF
--- a/tests/lib/rules/jsx-closing-bracket-location.js
+++ b/tests/lib/rules/jsx-closing-bracket-location.js
@@ -358,12 +358,18 @@ ruleTester.run('jsx-closing-bracket-location', rule, {
       '<App ',
       '/>'
     ].join('\n'),
+    output: [
+      '<App />'
+    ].join('\n'),
     parserOptions: parserOptions,
     errors: MESSAGE_AFTER_TAG
   }, {
     code: [
       '<App foo ',
       '/>'
+    ].join('\n'),
+    output: [
+      '<App foo/>'
     ].join('\n'),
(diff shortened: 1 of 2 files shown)


tests/lib/rules/jsx-closing-bracket-location.js
```diff
@@ -596,6 +656,13 @@ ruleTester.run('jsx-closing-bracket-location', rule, {
       '         />',
       '}'
     ].join('\n'),
+    output: [
+      'var x = function() {',
+      '  return <App',
+      '    foo',
+      '/>',
```
yannickcr: I think it should be aligned with the `return` keyword in this case.

yannickcr:
Thanks for this! I think there is a small issue in a test case though.

KevinGrandon:
@yannickcr - Thank you for taking a look. I've fixed the test case and resolved this by walking up nodes until it finds one that it's looking for. Commit is here: https://github.com/yannickcr/eslint-plugin-react/pull/511/commits/0f9ff7f9a5e199618f8373c21a02a9153f3825a7

Please let me know if there's a better way of doing this, or if there's anything else you'd like to see here. Thanks!

yannickcr:
Great! Can you squash your commits? Then I think we'll be good to go :)

KevinGrandon:
This is now squashed, thanks!

yannickcr:
Merged. Thanks!

Issue #68, opened 2015-04-17
Parse extended class to read some informations


MoOx:
```js
import React, {Component} from "react"

class A extends Component {

  static propTypes = {
    whatever: PropTypes.isRequired,
  }

  render() {
    return {this.props.whatever}
  }
}

class B extends A {
  render() {
    return {this.props.whatever}
  }
}
```

B gets an error on propTypes undefined on this component.

yannickcr:
Mmm, I think we can only have a limited support for this case.

We can handle the above script but not this one for example:
```js
import A from "a"

class B extends A {
  render() {
    return {this.props.whatever}
  }
}
```
In this case the extended class is in another file and we cannot know if the propTypes property is defined in it or not.

MoOx:
I think the fs read might be a problem. But it's doable right ?
Maybe disabled by default but that can be handled when set to true ?

yannickcr:
Linting is only made on the current file content, starting to jump from file to file, resolving imports and finding how the class is constructed will be very complicated (but I accept PR :wink: ).

I can handle the original case and if you encounter the second one I think the best way to handle it should be to ignore the `prop-types` rule for this class:

```js
import A from "a"

class B extends A { // eslint-disable-line react/prop-types
  render() {
    return {this.props.whatever}
  }
}
```

gausie:
Is there a way we could do a decorator-style comment above a class manually to inform the parser that the class is a descendent of a Component and lint it thusly?

PR #513, opened 2016-03-24
Parse explicit declaration that class extends React.Component

gausie:
This small addition enables linting component-specific rules if the author uses a JSDoc explicitly indicating that the component is a descendent of React.Component

e.g.
```
/** @extends React.Component */
class MyClass extends ChildClass {
...
```

Provides a good workaround for issues such as #68

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,10 +22,12 @@
   },
   "homepage": "https://github.com/yannickcr/eslint-plugin-react",
   "bugs": "https://github.com/yannickcr/eslint-plugin-react/issues",
+  "dependencies": {
+    "doctrine": "^1.2.0"
+  },
   "devDependencies": {
     "babel-eslint": "6.0.0-beta.6",
     "coveralls": "2.11.8",
-    "doctrine": "^1.2.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -25,6 +25,7 @@
   "devDependencies": {
     "babel-eslint": "6.0.0-beta.6",
     "coveralls": "2.11.8",
+    "doctrine": "^1.2.0",
```
ljharb: this would need to be a full `dependency` since it's not related to actually developing the module.

gausie: Whoops! Quite right.

yannickcr:
Merged, thanks!

ljharb:
I wonder if we should make doctrine an optionalDependency, so people that don't use jsdoc don't need to install it?

yannickcr:
I don't like adding a dependency either.

But since this dependency is already used by ESLint I think it will already be installed in all cases (it will even be deduplicated with npm@3).

ljharb:
An, makes sense

PR #515, opened 2016-03-26
Add prefix-members rule

adrianton3:
I've mistyped `shouldComponentUpdate` and other React class members with long names too many times. This rule is intended to prevent that.

There's an unwritten convention that user methods should start with an `_` (flux tutorial uses this http://facebook.github.io/flux/docs/todo-list.html). This way it's easy to tell if a React method was misspelled by checking that member names that don't start with an `_` are in a short list of valid React methods.

This rule appeals to people who want to enforce the `_` prefix and also avoid bugs caused by typos.

Example that triggers a warning:

``` js
  componentDidUpdaet() { 
    // this never executes since there's a typo in the method name
  },
  render: function() {
  }
});
```

Example that doesn't trigger a warning:

``` js
  _handleClick() { 
    // ...
  },
  render: function() {
    return <div onClick={this._handleClick}>Hello {this.props.name}</div>;
  }
});
```

It has 100% coverage, but no doc yet - if this rule is good to be merged then I can write that too.

### DIFF
--- a/lib/rules/prefix-members.js
+++ b/lib/rules/prefix-members.js
@@ -1,0 +1,76 @@
+/**
+ * @fileoverview Enforce "_" prefix to user methods in a React component definition
+ * @author Adrian Toncean
+ */
+'use strict';
+
+var Components = require('../util/Components');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = Components.detect(function(context, components, utils) {
+  var reactMemberNames = [
+    'displayName',
+    'propTypes',
+    'contextTypes',
+    'childContextTypes',
+    'mixins',
+    'statics',
+    'defaultProps',
+    'constructor',
+    'getDefaultProps',
+    'state',
+    'getInitialState',
+    'getChildContext',
+    'componentWillMount',
+    'componentDidMount',
+    'componentWillReceiveProps',
+    'shouldComponentUpdate',
+    'componentWillUpdate',
+    'componentDidUpdate',
+    'render',
+    'componentWillUnmount'
+  ];
+
+  function isValid(name) {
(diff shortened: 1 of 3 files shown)


ljharb:
I strongly dislike this convention, as `_` is used usually to indicate "privacy" and not "user methods". Since JS does not have privacy, this convention often misleads people into thinking they don't have to test `_`-prefixed things, or that they can break `_`-prefixed things without a major version bump.

If this rule is to be included, could it please have a configurable prefix?? Ideally, even a *required* prefix so that it doesn't default to `_` and thus implicitly endorse its use.

yannickcr:
Also I am not totally convinced that it could be a useful rule since [`sort-comp`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md) could already warn you if a method is mistyped (since it will detect an user method in the lifecycle ones).

lib/rules/prefix-members.js
```diff
@@ -0,0 +1,76 @@
...
+ */
+'use strict';
+
+var Components = require('../util/Components');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = Components.detect(function(context, components, utils) {
```
petersendidit: This should use the [new eslint rule format](http://eslint.org/blog/2016/07/eslint-new-rule-format)

lencioni:
What if we had a rule that warned you about method names that had a small levenshtein distance (maybe <= 3) from lifecycle methods?

ljharb:
We have the `no-typos` rule: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-typos.md

I think this PR should be closed; any improvements should go to that rule.

Issue #516, opened 2016-03-28
`prefer-stateless-function` should detect `null` in if shorthand return


inyono:
With enabled rule `prefer-stateless-function`, the following does throw a warning although `null` may be returned:
```.js
class If extends Component {
  render() {
    const { condition, ...props } = this.props

    return condition ? <this.props.component {...props} /> : null
  }
}
```
If we write the result to a variable, though, there is no warning:
```.js
class If extends Component {
  render() {
    const { condition, ...props } = this.props

    const component = condition ? <this.props.component {...props} /> : null

    return component
  }
}
```

PR #520, opened 2016-03-29
Add failing test for #516

inyono:
Adds a failing test for my issue #516. I will try to solve this. If you have a fix for this, feel free to continue this PR.


yannickcr:
Thanks for this failing case. I think the problem comes from [isReturningJSX](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/util/Components.js#L171) where we only check if the ternary condition is returning JSX in one case (since this function is also used to detect stateless components).

PR #526, opened 2016-04-02
Fix grammar in prefer-stateless-function rule

ahoym:
While looking up this rule, this particular sentence read a little awkwardly.

This PR contains a minor grammar change in the markdown file for the `prefer-stateless-function` rule.


yannickcr:
Merged, thanks!

PR #535, opened 2016-04-07
Add a new rule to ensure position of first property in JSX

jseminck:
In our project we want to use the AirBnB React Style guidelines when it comes to React/JSX code style. One of the rules for JSX is to align components as such:

// bad
```js
<Foo superLongParam="bar"
     anotherSuperLongParam="baz" />
```

// good
```js
<Foo
  superLongParam="bar"
  anotherSuperLongParam="baz"
/>
```

However, I could not find a rule which enforces that the first property should be placed on a new line, in the case of a multiline JSX component.

As an experiment, I decided to try and implement this myself. The result is this PR.

I decided to make three options, although not sure if the other two are really relevant:
`always`: The first prop should always be on a new line, even when there is only 1 property.
`never`: The first prop should never be on a new line, so should always be on the same line as the opening component tag.
`multiline`: Follows AirBnB style. In the case of a multiline component the first prop should be on a new line, otherwise when all props are on a single line then this is allowed.

Please have a look at the code, tests, style, etc... If you have any comments or tips on how to improve this, please let me know. I'd love to see this make it to the project, as it already helped me find a few places which I missed out during my initial refactor to this AirBnB style.


jseminck:
Seems like there is already an open PR: https://github.com/yannickcr/eslint-plugin-react/pull/427

And there are other issues asking for the same functionality: https://github.com/yannickcr/eslint-plugin-react/issues/410

So would be nice to include this one way or another. I'm up for trying to improve this specific implementation or helping out with the other PR if needed.

ljharb:
It would be super ideal if one of these approaches made it in ASAP. I'd love to include it in the next release of the eslint plugin, which will be a semver-major anyways. (@yannickcr)

yannickcr:
Merged. Thanks!

PR #537, opened 2016-04-08
New rule: jsx-spread-operator-position

jseminck:
A rule we want to enforce in our project is consistent position when using spread operator as a prop in JSX components.

E.g. we either want it always at the beginning
```js
<MyComponent {...this.props.data} prop1 prop2 />
```

or at the end (we are still fighting about what approach to take)
```js
<MyComponent prop1 prop2 {this.props.data} />
```

I couldn't find any rule which enforces this... so decided to make another experiment and add it.  There are two options: "begin" and "end".

I think the rule naming could use some fine-tuning. Not sure what would be a good name.

Please review the idea, code, style and add comments where needed. If you want this rule to be added then I will do my best to get it included.


jseminck:
Closing this as it also contains other commits. Will create a different fork and make a new PR from that.

ljharb:
@jseminck in the future, please rebase and force push instead of closing and creating new PRs - it reduces clutter for maintainers.

PR #538, opened 2016-04-08
New rule: jsx-spread-operator-position

jseminck:
A rule we want to enforce in our project is consistent position when using spread operator as a prop in JSX components.

E.g. we either want it always at the beginning
```js
<MyComponent {...this.props.data} prop1 prop2 />
```

or at the end (we are still fighting about what approach to take)
```js
<MyComponent prop1 prop2 {this.props.data} />
```

I couldn't find any rule which enforces this... so decided to make another experiment and add it.  There are two options: "begin" and "end".

I think the rule naming could use some fine-tuning. Not sure what would be a good name.

Please review the idea, code, style and add comments where needed. If you want this rule to be added then I will do my best to get it included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,6 @@ The plugin has a [recommended configuration](#user-content-recommended-configura
 * [jsx-closing-bracket-location](docs/rules/jsx-closing-bracket-location.md): Validate closing bracket location in JSX (fixable)
 * [jsx-curly-spacing](docs/rules/jsx-curly-spacing.md): Enforce or disallow spaces inside of curly braces in JSX attributes (fixable)
 * [jsx-equals-spacing](docs/rules/jsx-equals-spacing.md): Enforce or disallow spaces around equal signs in JSX attributes (fixable)
-* [jsx-first-prop-new-line](docs/rules/jsx-first-prop-new-line.md): Enforce position of the first prop in JSX
 * [jsx-handler-names](docs/rules/jsx-handler-names.md): Enforce event handler naming conventions in JSX
 * [jsx-indent-props](docs/rules/jsx-indent-props.md): Validate props indentation in JSX (fixable)
 * [jsx-indent](docs/rules/jsx-indent.md): Validate JSX indentation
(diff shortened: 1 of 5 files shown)


ljharb:
What's the reasoning for requiring this? The ordering of props matters when using spread, because for example you might want an explicit prop overridden by an object only when the colliding prop exists in the object (when using the spread last).

README.md
```diff
@@ -102,6 +102,7 @@ The plugin has a [recommended configuration](#user-content-recommended-configura
 * [jsx-closing-bracket-location](docs/rules/jsx-closing-bracket-location.md): Validate closing bracket location in JSX (fixable)
 * [jsx-curly-spacing](docs/rules/jsx-curly-spacing.md): Enforce or disallow spaces inside of curly braces in JSX attributes (fixable)
 * [jsx-equals-spacing](docs/rules/jsx-equals-spacing.md): Enforce or disallow spaces around equal signs in JSX attributes (fixable)
+* [jsx-first-prop-new-line](docs/rules/jsx-first-prop-new-line.md): Enforce position of the first prop in JSX
```
ljharb: Is this line a mistake? If not it might be better in a separate PR.

jimbolla:
To build upon what @ljharb said, I question the applicability of this rule, as there are plenty of valid reasons to put the spread in seemingly arbitrary places. For example:

```
const DemoInput = (props) => (
  <Input
    overridableProp1="someDefault"
    overridableProp2="anotherDefault"
    {...props}
    requiredProp1="mustBeThis" />
);
```

This is a pattern I use quite a bit.

jseminck:
Fair point. 

We do not have these cases where we have overridable and required props mixed together. We just have a bunch of objects that we pass around and eventually spread into smaller components. We just use a mix of begin and end placement and otherwise have to correct it during code reviews.

But I understand that your use case can be quite common, perhaps way more than ours. If you think the rule is not applicable to most projects then feel free to dismiss it. It's just a suggestion.

yannickcr:
I find this rule pretty confusing. It seems to be pure stylistic rule but it can actually change how the properties will override / be overrided by the other props.

This could makes you introduce some subtle bugs while you try to follow what seems to just be a stylistic rule.

PR #558, opened 2016-04-21
Fix forbid-prop-types exploding on object spread

nevir:
If you pass an object spread to `propTypes`, `forbid-prop-types` currently freaks out.  This at least keeps it from throwing a `Cannot read property 'type' of undefined`


ljharb:
Thanks! Could you please add a test that would have caught this bug, prior to your fix?

nevir:
👍

yannickcr:
Sorry, missed your PR :( Seems it was fixed in https://github.com/yannickcr/eslint-plugin-react/commit/6950cbd73dd39385326cea233908df8232678450

But thank you for the fix :)

PR #559, opened 2016-04-21
Change "extentions" to "extensions" in require-extension

lencioni:
While working on a different rule, I noticed a small typo here and
thought it was worth fixing.


ljharb:
Does this constitute a breaking change?

ljharb:
Actually it seems like not, because it's all internal.

lencioni:
Indeed, it is all internal!

PR #561, opened 2016-04-22
Fix line-aligned jsx-closing-bracket-location fixer

### LINKED ISSUE #533: Autofix for react/jsx-closing-bracket-location line-aligned uses wrong indentation for nested elements
Basically all closing brackets are aligned to the OUTMOST element.

This code:
```
<div>
  <div
    className="foo">
  </div>
</div>
```

Is fixed like this:
```
<div>
  <div
    className="foo"
>
  </div>
</div>
```

dtinth:
Since we already know what the correct column would be, we can use that to derive the number of spaces to add.

Fixes #533.

Don’t have time to write tests right now, sorry.


lencioni:
I tried this out and it seemed to help on the cases I used it on.

yannickcr:
Seems to help, indeed. Thanks.

PR #565, opened 2016-04-24
Grammar fix

coryhouse:


ljharb:
Would you please merge #566, #568, #569, #570, all into the same PR? There's no need to even have these be separate commits.

coryhouse:
These were all performed via a mobile device on github.com so there was no easy way to merge.

coryhouse:
So to answer your question, no, I won't go through the re-work of merging these into one PR.

ljharb:
@coryhouse for future reference, even in that scenario you can continue editing files, in your fork, on the same branch, and it will add commits to the existing PR.

I'll manually squash these.

coryhouse:
Nice! News to me. Thanks for the tip.

PR #567, opened 2016-04-24
Add a failing test for sort-prop-types

oliviertassinari:
E.g, we are using this [pattern](https://github.com/callemall/material-ui/blob/master/src/TextField/TextFieldLabel.js) on the Material-UI repository.

### DIFF
--- a/tests/lib/rules/sort-prop-types.js
+++ b/tests/lib/rules/sort-prop-types.js
@@ -365,6 +365,28 @@ ruleTester.run('sort-prop-types', rule, {
     }]
   }, {
     code: [
+      'const propTypes = {',
+      '  muiTheme: PropTypes.object.isRequired,',
+      '  className: PropTypes.string,',
+      '};',
+      '',
+      'const TextFieldLabel = (props) => {',
+      '  return (',
+      '    <div />',
+      '  );',
+      '};',
+      '',
+      'TextFieldLabel.propTypes = propTypes;'
+    ].join('\n'),
+    parserOptions: parserOptions,
+    errors: [{
+      message: ERROR_MESSAGE,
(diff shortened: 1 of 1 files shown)


ljharb:
In your example, the propTypes aren't referenced anywhere in the function.

yannickcr:
We do not follow variables references, so with this pattern the propTypes order is not checked.

oliviertassinari:
> We do not follow variables references

@yannickcr Will this be fixed in the future? Or should we just not use this *pattern*.

oliviertassinari:
> In your example, the propTypes aren't referenced anywhere in the function. Is that reflective of your actual code?

@ljharb This is the [actual code](https://github.com/callemall/material-ui/blob/1f5762bd182fa16d3f68e734a52969f61c7d1558/src/TextField/TextFieldLabel.js#L28). I have removed the reference. I wanted to be consistent with the other tests.

yannickcr:
> Will this be fixed in the future? Or should we just not use this pattern?

We only have access to the AST and it is tricky to follow variables references, so it is not planned to support them. But you can still open an issue for it, if there is enough demand for this feature we could try to see if it is feasible.

If you want to be able to use the prop-types related rules (like `prop-types` or `sort-prop-types`) you should avoid this pattern for now, sorry :|

oliviertassinari:
I agree, it's not straightfoward.
Still, we have access to the scope. I think that it can be done in simples cases like this one.
E.g. This is how I'm doing it with babel: https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types/blob/master/src/index.js#L75.

Sure, I'm gonna open an issue. Thanks.

PR #582, opened 2016-05-02
New Rule: jsx-no-target-blank

Gasparila:
New rule to protect against a pretty serious security vulnerability: https://mathiasbynens.github.io/rel-noopener

Using `target='_blank'` to open a link in a new tab is inherently unsafe and should always be accompanied by `rel='noopener noreferrer'`


eddie:
Nice :) but why specific to JSX/react?

dmnd:
> why specific to JSX/react?

It's definitely specific to JSX. I suppose not necessarily React, though. @eddie can you suggest a more appropriate place to open this PR?

eddie:
@dmnd You are right (specific to jsx).. It was early when I wrote that comment. Here is probably the best place :+1:

yannickcr:
Merged. Thanks!

PR #586, opened 2016-05-03
no-render-return-value. closes #531

### LINKED ISSUE #531: RFC: Rule for legacy return value of ReactDOM.render
This is a RFC for creating an eslint rule to flag the return value of `ReactDOM.render` as a deprecated feature.

The documentation for this is in progress of being updated at: https://github.com/facebook/react/pull/6400

While there is no timeline on when this deprecation will be removed and made mainstream, I think it would be beneficial to the community. This deprecation in particular is something that cannot be done at runtime since it depends on knowledge outside of the callsite, which makes it a perfect candidate for an eslint rule.

There are two preferred recommendations to work with this:

```js
// Bad
const inst = ReactDOM.render(<App />, document.body);
doSomethingWithInst(inst);

// Good
ReactDOM.render(<App ref={doSomethingWithInst} />, document.body);
ReactDOM.render(<App />, document.body, doSomethingWithInst);
```

cc @jimfb @gaearon

iamdustan:
As documented initially in https://github.com/facebook/react/pull/6400 and on the documentation website at http://facebook.github.io/react/docs/top-level-api.html#reactdom.render

The return value of `ReactDOM.render` is considered legacy. It's rather hard to deprecate the return value of an element with runtime code, but fairly straight-forward with static analysis so this is intended to support the communities education and eventual migration from using that pattern.

### DIFF
--- a/lib/rules/no-render-return-value.js
+++ b/lib/rules/no-render-return-value.js
@@ -1,0 +1,62 @@
+/**
+ * @fileoverview Prevent usage of the return value of React.render
+ * @author Dustan Kasten
+ */
+'use strict';
+
+var versionUtil = require('../util/version');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = function(context) {
+
+  // --------------------------------------------------------------------------
+  // Public
+  // --------------------------------------------------------------------------
+
+  return {
+
+    CallExpression: function(node) {
+      var callee = node.callee;
+      var parent = node.parent;
+      if (callee.type !== 'MemberExpression') {
+        return;
+      }
+
+      var calleeObjectName = /^ReactDOM$/;
+      if (versionUtil.test(context, '15.0.0')) {
+        calleeObjectName = /^ReactDOM$/;
+      } else if (versionUtil.test(context, '0.14.0')) {
+        calleeObjectName = /^React(DOM)?$/;
+      } else if (versionUtil.test(context, '0.13.0')) {
+        calleeObjectName = /^React$/;
+      }
+
+      if (
+        callee.object.type !== 'Identifier' ||
+        !calleeObjectName.test(callee.object.name) ||
+        callee.property.name !== 'render'
+      ) {
+        return;
+      }
+
+      if (
+        parent.type === 'VariableDeclarator' ||
+        parent.type === 'Property' ||
+        parent.type === 'ReturnStatement' ||
+        parent.type === 'ArrowFunctionExpression'
+      ) {
+        context.report({
+          node: callee,
+          message: 'Do not depend on the return value from ' + callee.object.name + '.render'
+        });
+      }
+    }
+  };
+
+};
+
+module.exports.schema = [];
+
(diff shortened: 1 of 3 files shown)


ljharb:
This should probably only apply in React 0.14 and higher - and should apply to `React.render` in 0.13

iamdustan:
good call @ljharb. Updated accordingly.

lib/rules/no-render-return-value.js
```diff
@@ -0,0 +1,62 @@
...
+
+    CallExpression: function(node) {
+      var callee = node.callee;
+      var parent = node.parent;
+      if (callee.type !== 'MemberExpression') {
+        return;
+      }
+
+      var calleeObjectName;
+      if (versionUtil.test(context, '15.0.0')) {
```
ljharb: this should probably be the default case, to automatically get support for newer react versions.

lib/rules/no-render-return-value.js
```diff
@@ -0,0 +1,62 @@
...
+
+      if (
+        parent.type === 'VariableDeclarator' ||
+        parent.type === 'Property' ||
+        parent.type === 'ReturnStatement' ||
+        parent.type === 'ArrowFunctionExpression'
+      ) {
+        context.report({
+          node: callee,
+          message: 'Do not depend on the return value from render'
```
ljharb: should the message also vary based on React version?

lib/rules/no-render-return-value.js
iamdustan: how would you recommend doing that? `versionUtil` seems to be `true` for any greater version. I initially had:

```
var calleeObjectName = /^ReactDOM$/;
if (versionUtil.test(context, '0.13.0')) {
  // ...
} else if (versionUtil.test(context, '0.14.0')) {
  // ...
}
```

lib/rules/no-render-return-value.js
iamdustan: we could have the message be `'Do not depend on the return value from ' + callee.object.name + '.render'`.

lib/rules/no-render-return-value.js
ljharb: ah, if `versionUtil` already has "this or greater" semantics, then this is totally fine. however, i feel like your initial approach might be better because it explicitly has a default case in the code.

lib/rules/no-render-return-value.js
ljharb: seems legit

lib/rules/no-render-return-value.js
iamdustan: I like the initial approach more, but it would also end up being whatever the final `else if` branch would be. I could just duplicate it in the variable declaration so that is more explicit.

lib/rules/no-render-return-value.js
```diff
@@ -0,0 +1,62 @@
...
+        calleeObjectName = /^ReactDOM$/;
+      } else if (versionUtil.test(context, '0.14.0')) {
+        calleeObjectName = /^React(DOM)?$/;
+      } else if (versionUtil.test(context, '0.13.0')) {
+        calleeObjectName = /^React$/;
+      }
+
+      if (
+        callee.object.type !== 'Identifier' ||
+        !calleeObjectName.test(callee.object.name) ||
```
iamdustan: do any tests currently verify that the `React/ReactDOM` identifier is from a `require`/`import` call?

lib/rules/no-render-return-value.js
ljharb: i like the `else` because then i don't have to know how `versionUtil.test` works.

iamdustan:
Lemme know whatever else is needed for this (if anything).

yannickcr:
Merged, thanks!

Issue #572, opened 2016-04-25
Why `self-closing-comp` allows empty divs?


alexfedoseev:
I noticed that [self-closing-comp](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/self-closing-comp.md) doesn't report empty divs (eg `<div></div>`). Is there any particular reason that it allows empty html tags? Or it's valid b/c it's valid html?

ljharb:
Seems like a bug to me. That should be `<div />`.

alexfedoseev:
@ljharb I'd be sure it is, if there wasn't this in the doc:

<img width="463" alt="screen shot 2016-04-25 at 21 56 39" src="https://cloud.githubusercontent.com/assets/4244251/14795349/cd1a0252-0b30-11e6-8731-8c453a7eee92.png">

ljharb:
That doesn't make much sense to me - but if it's documented, then it's not a bug.

However, I'd love to see an option added to this rule that requires self-closing tags for *every* kind of tag, including HTML tags. @yannickcr?

yannickcr:
This is the intended behavior. This rule only enforce empty React Components to be self-closed, and HTML tags are not concerned.

I think we could modify this rule to check all type of tags by default, then add some options to filter by type (components and/or HTML tags) if needed.

It seems to be the more logical behavior to me, but it will be a breaking change.

ljharb:
Airbnb's React styleguide requires every empty element to be self-closed.

Alternatively, we could add an option in a minor version, and then flip the default behavior later, in a major?

gitim:
Have the same issue. Is anybody already working on it? If not I'll prepare a pull request, which will add the option suggested by @ljharb to the rule.

GeoffGodwin:
Just wanted to leave my comment in here for future developers reaching this page. `<div/>` is not a valid HTML 4.x tag. It's the only base HTML tag (that I know of) that cannot be legally self-closing. If you use a `<div/>` it can cause very strange issues that can bubble up through a variety of frameworks (looking at you jQuery) making it difficult to trace as it does not throw distinct errors. One particular example being in IE8 where it breaks the DOM tree fairly silently.

It is always better practice to use `<div></div>` even if no content exists between the tags.

As a note, `<div/>` is legitimate XHTML 1.x. However for the sake of compatibility across browsers and versions there is no reason to use it.

I realize this is ooooold world HTML but it's still very valid today. Cheers!

[...]

PR #600, opened 2016-05-14
Add option to the self-closing-comp

gitim:
This pr adds an option to the self-closing-comp suggested #572, which is corresponding what type of tags (`{component: <boolean>, html: <boolean>}`) should be self-closed when this is possible. **Note**: this not changes current behavior, default option is `{component: true, html: false}`, but in the next major
release we can change it to `{component: true, html: true}` to enforce all tags to be self-closing when that is possible.

### DIFF
--- a/docs/rules/self-closing-comp.md
+++ b/docs/rules/self-closing-comp.md
@@ -26,9 +26,9 @@ The rule can take one argument to select types of tags, which should be self-clo
 
 ```js
 ...
-"self-closing-comp": [<enabled>, {
-  "component" <boolean> || true,
-  "html" <boolean> || false
+"self-closing-comp": ["error", {
+  "component": true,
+  "html": false
 }]
 ...
 ```
--- a/docs/rules/self-closing-comp.md
+++ b/docs/rules/self-closing-comp.md
@@ -17,5 +17,79 @@ var contentContainer = <div className="content"></div>;
 
 var HelloJohn = <Hello name="John" />;
 
-var Profile = <Hello name="John"><img src="picture.png" /></Hello>; 
+var Profile = <Hello name="John"><img src="picture.png" /></Hello>;
+```
+
+## Rule Options
+
+It takes an option as the second parameter, which corresponding what components tags should be self-closed when this is possible.
+
+```js
+...
+"self-closing-comp": [<enabled>, 'all'|'component'|'html'>]
+...
+```
+
+### `all`
+
+All tags, including custom components and html components should be self-closed.
+
+The following patterns are considered warnings:
+
+```js
+var HelloJohn = <Hello name="John"></Hello>;
+
+var contentContainer = <div className="content"></div>;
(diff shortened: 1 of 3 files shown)


docs/rules/self-closing-comp.md
```diff
@@ -17,5 +17,79 @@ var contentContainer = <div className="content"></div>;
 
 var HelloJohn = <Hello name="John" />;
 
-var Profile = <Hello name="John"><img src="picture.png" /></Hello>; 
+var Profile = <Hello name="John"><img src="picture.png" /></Hello>;
+```
+
+## Rule Options
+
+It takes an option as the second parameter, which corresponding what components tags should be self-closed when this is possible.
```
ljharb: We should indicate which is the default here - "component" i think.

docs/rules/self-closing-comp.md
```diff
@@ -17,5 +17,79 @@ var contentContainer = <div className="content"></div>;
...
+
+```js
+...
+"self-closing-comp": [<enabled>, 'all'|'component'|'html'>]
```
ljharb: i wonder if rather than "all", this should have the format `{ component: true, html: true }` - that allows for other categorizations, and works just as well for the default of `{ component: true, html: false }`

gitim: yea, your solution is better, thanks, I'll fix this

docs/rules/self-closing-comp.md
```diff
@@ -17,5 +17,56 @@ var contentContainer = <div className="content"></div>;
...
+```
+
+## Rule Options
+
+The rule can take one argument to select types of tags, which should be self-closed when this is possible. By default only custom components tags should be self-closed.
+
+```js
+...
+"self-closing-comp": [<enabled>, {
+  "component" <boolean> || true,
```
ljharb: `|| true` means it will never be false, so i don't think this is good notation. I'd just provide the default object:
```js
"self-closing-comp": ["error", {
  "component": true,
  "html": false
}]
```

I think it's pretty self-apparent that you can override them with your own booleans

gitim: ok, fixed

ljharb:
Just to be safe, could we also add tests with the options object omitted?

gitim:
> Just to be safe, could we also add tests with the options object omitted?

Yes, sure. Just copied all tests without options and pasted with adding `options: []`.

ljharb:
LGTM but i'll let @yannickcr do the merging since i'm pretty weak on parser code

gitim:
Is there any plans to release this fix?

yannickcr:
@gitim Yes, but I'm currently in holidays and this until the 1st June, then I will be at React Europe on 2-3 June. So do not expect a lot of activity from me until the 4th June.

yannickcr:
Thanks!

PR #601, opened 2016-05-15
Adds new require-optimization rule.

### LINKED ISSUE #240: Rule Proposal: Require "shouldComponentUpdate" or PureRenderMixin to be applied
On our team, we want to ensure all components are optimized. This means either PureRenderMixin must be applied on the component, or the component must have a `shouldComponentUpdate` method defined. It would be great if ESLint could warn developers when they have forgotten to optimize.

EvNaverniouk:
Fixes #240. Adds a new rule called `require-optimization` which ensures all components either have a `shouldComponentUpdate` method or a `PureRenderMixin`.

Tested on several large React project repos without any issues using a beta release of this rule here (https://travis-ci.org/Eight-Solutions/eslint-plugin-cumul8). General feedback from users of this rule so far have been very positive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,7 @@ The plugin has a [recommended configuration](#user-content-recommended-configura
 * [prop-types](docs/rules/prop-types.md): Prevent missing props validation in a React component definition
 * [react-in-jsx-scope](docs/rules/react-in-jsx-scope.md): Prevent missing `React` when using JSX
 * [require-extension](docs/rules/require-extension.md): Restrict file extensions that may be required
+* [require-optimization](docs/rules/require-optimization.md): Enforce React components to have a shouldComponentUpdate method
 * [require-render-return](docs/rules/require-render-return.md): Enforce ES5 or ES6 class for returning value in render function
 * [self-closing-comp](docs/rules/self-closing-comp.md): Prevent extra closing tags for components without children
 * [sort-comp](docs/rules/sort-comp.md): Enforce component methods order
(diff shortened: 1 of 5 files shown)


ljharb:
Thanks! Can you add some tests that ensure that SFCs (both normal function and arrow function ones) do *not* warn with this rule?

EvNaverniouk:
@ljharb Ah, of course. Forgot about SFCs. Updated rule, added tests here: https://github.com/yannickcr/eslint-plugin-react/pull/601/commits/94fa138d138a528dc43d46b9a791a72ce0544db6

EvNaverniouk:
Actually, now that I think about it, maybe ignoring SFCs should be done behind an option flag. Since, technically, SFCs are **not** optimized components so they **should** throw an error. But at the moment, there is no way to optimize them. So I can imagine some authors might want to ban the user of SFCs for this reason.

What do you think?

ljharb:
I think throwing an error on all SFCs, especially when other rules exist to require SFCs whenever possible, is heavy handed enough that it would need to be defaulted off, or omitted entirely.

SFCs don't _need_ optimization, that's the point.

EvNaverniouk:
Alright, I'll leave the rule as is then. Ready to merge?

ljharb:
LGTM, but @yannickcr should take a look just to be sure.

What about a shouldComponentUpdate that's the default, i.e., `return true`?

yannickcr:
LGTM too, great work!

PR #602, opened 2016-05-15
Fix display-name when JSON style is used for defining components

### LINKED ISSUE #590: display-name rule quotes handling
If JSON style is used for defining components, rule is not handle "displayName" option.
`var Hello = React.createClass({
  "displayName": "Hello",
  "render": function() {
  }
});`
but its valid code i think.

gitim:
Fixes #590. 

Imho, it is not very common style to define components, but I think it should be fixed.

### DIFF
--- a/lib/rules/display-name.js
+++ b/lib/rules/display-name.js
@@ -24,23 +24,25 @@ module.exports = Components.detect(function(context, components, utils) {
    * @returns {Boolean} True if we are declaring a display name, false if not.
    */
   function isDisplayNameDeclaration(node) {
-    // Special case for class properties
-    // (babel-eslint does not expose property name so we have to rely on tokens)
-    if (node.type === 'ClassProperty') {
-      var tokens = sourceCode.getFirstTokens(node, 2);
-      if (
-        tokens[0].value === 'displayName' ||
-        (tokens[1] && tokens[1].value === 'displayName')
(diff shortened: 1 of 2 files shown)


lib/rules/display-name.js
```diff
@@ -24,23 +24,25 @@ module.exports = Components.detect(function(context, components, utils) {
...
+      case 'Identifier':
+        return node.name === 'displayName';
+      case 'Literal':
+        return node.value === 'displayName';
+      default:
+        return false;
     }
-
-    return Boolean(
-      node &&
```
gitim: Is this check is necessary? All tests passes without it.

lib/rules/display-name.js
```diff
@@ -24,23 +24,25 @@ module.exports = Components.detect(function(context, components, utils) {
...
-    if (node.type === 'ClassProperty') {
-      var tokens = sourceCode.getFirstTokens(node, 2);
-      if (
-        tokens[0].value === 'displayName' ||
-        (tokens[1] && tokens[1].value === 'displayName')
-      ) {
-        return true;
-      }
-      return false;
+    switch (node.type) {
```
gitim: I rewrote this with switch statement. I had an option to add another if statement or ternary operator to the last return, but switch seems more explicitly defining possible cases

PR #608, opened 2016-05-20
Autofix: `jsx-indent`: Add support for tabs

jayphelps:
Everything seems correct except for the very last test. 

The real output is:

``` js
function App() {
  return (
    <App>
  <Foo />
</App>
  );
}
```

But I would _expect_ the output to be:

``` js
function App() {
  return (
    <App>
        <Foo />
    </App>
  );
}
```

Anyone have any ideas what I'm doing wrong or if somehow this is indeed expected behavior?

### DIFF
--- a/lib/rules/jsx-indent.js
+++ b/lib/rules/jsx-indent.js
@@ -39,8 +39,6 @@ module.exports = {
       category: 'Stylistic Issues',
       recommended: false
     },
-    // fixer will fix whitespace and tabs indentation.
-    // It won't replace tabs with whitespaces and vice-versa.
     fixable: 'whitespace',
     schema: [{
       oneOf: [{
@@ -77,20 +75,16 @@ module.exports = {
      * Responsible for fixing the indentation issue fix
      * @param {ASTNode} node Node violating the indent rule
      * @param {Number} needed Expected indentation character count
-     * @param {Number} gotten Indentation character count in the actual node/code
      * @returns {Function} function to be executed by the fixer
      * @private
      */
-    function getFixerFunction(node, needed, gotten) {
-      if (needed > gotten) {
-        var spaces = indentChar.repeat(needed - gotten);
-        return function(fixer) {
-          return fixer.insertTextBeforeRange([node.range[0], node.range[0]], spaces);
-        };
-      }
-
+    function getFixerFunction(node, needed) {
       return function(fixer) {
-        return fixer.removeRange([node.range[0] - (gotten - needed), node.range[0]]);
+        var indent = Array(needed + 1).join(indentChar);
+        return fixer.replaceTextRange(
+          [node.start - node.loc.start.column, node.start],
+          indent
+        );
       };
     }
 
@@ -99,7 +93,7 @@ module.exports = {
      * @param {ASTNode} node Node violating the indent rule
      * @param {Number} needed Expected indentation character count
      * @param {Number} gotten Indentation character count in the actual node/code
-     * @param {Object=} loc Error line and column location
+     * @param {Object} loc Error line and column location
      */
     function report(node, needed, gotten, loc) {
       var msgContext = {
@@ -115,14 +109,14 @@ module.exports = {
           loc: loc,
           message: MESSAGE,
           data: msgContext,
-          fix: getFixerFunction(node, needed, gotten)
+          fix: getFixerFunction(node, needed)
         });
       } else {
         context.report({
           node: node,
           message: MESSAGE,
           data: msgContext,
-          fix: getFixerFunction(node, needed, gotten)
+          fix: getFixerFunction(node, needed)
         });
       }
     }
--- a/tests/lib/rules/jsx-indent.js
+++ b/tests/lib/rules/jsx-indent.js
@@ -233,6 +233,11 @@ ruleTester.run('jsx-indent', rule, {
       '    <Foo />',
       '</App>'
     ].join('\n'),
+    output: [
+      '<App>',
+      '\t<Foo />',
+      '</App>'
+    ].join('\n'),
     options: ['tab'],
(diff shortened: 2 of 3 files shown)


tests/lib/rules/jsx-indent.js
```diff
@@ -159,6 +188,15 @@ ruleTester.run('jsx-indent', rule, {
...
     ].join('\n'),
+    output: [
+      'function App() {',
+      '  return (',
+      '    <App>',
+      '  <Foo />',
+      '</App>',
+      '  );',
+      '}'
+    ].join('\n'),
```
jayphelps: I would expect the output to be

```js
function App() {
  return (
    <App>
      <Foo />
    </App>
  );
}
```

Why is it not?

CiGit:
In this case, only the `<App>` tag is an error so the `fix` function only applies to it.

[Test case](https://github.com/yannickcr/eslint-plugin-react/blob/be33738b098ea145a6598d9ab4b009952d06c913/tests/lib/rules/jsx-indent.js#L164)

To have the output you (and I) expect, the way errors are computed should be changed.
ie JSXElement's indentation compared to something like `ROOT - 1line`, not to its direct parent / opening tag.

IMO that's an other issue

zouxuoz:
@jayphelps @CiGit How can I help to merge this PR?

tests/lib/rules/jsx-indent.js
mmiller42: This is an error test case, right? If I'm reading it right, it's saying that `</App>` should be indented by 4 spaces but is only indented by 0.

jayphelps:
Abandoning due to lack of response from maintainers.

tests/lib/rules/jsx-indent.js
ljharb: I agree - let's not merge this without the fix for this.

jayphelps: @ljharb a fix included in this PR or a fix elsewhere? 

@CiGit had said this:

> To have the output you (and I) expect, the way errors are computed should be changed. ie JSXElement's indentation compared to something like ROOT - 1line, not to its direct parent / opening tag.
> 
> IMO that's an other issue

If I understand correctly, this will require detection logic changes. I unfortunately don't have cycles to dig that deep into the internals at the moment. If anyone else does, feel free to continue--I don't need commit "credit" hehe

jayphelps:
@ljharb oooo when I tried to rebase, I discovered it appears this has already been added in #830. Our two implementation seem mutually exclusive. Do you know if there is something this fixes that #830 did not?

mmiller42:
@jayphelps I know of one -- I don't think #830 doesn't support tabs. That could be filed as a separate bug. (Currently it counts the number of characters in the indent and inserts spaces, I believe.)

jayphelps:
@mmiller42 oh! my PR takes a different approach that _does_ support that. Instead of trying to insert or remove indention, it just always just replaces the indention entirely using `fixer.replaceTextRange`.

I've rebased and updated my PR, uncommented the tests which weren't working in #830 but now are. Seems OK, but it's been a while since I've looked at this code so definitely have a critical eye.

jayphelps:
Note that I have no fixed the pre-existing issue of the previously mentioned odd output in the test:

```js
function App() {
  return (
    <App>
  <Foo />
</App>
  );
}
```

Presumably, #830 just didn't catch that.

lib/rules/jsx-indent.js
```diff
@@ -77,20 +75,16 @@ module.exports = {
      * Responsible for fixing the indentation issue fix
      * @param {ASTNode} node Node violating the indent rule
      * @param {Number} needed Expected indentation character count
-     * @param {Number} gotten Indentation character count in the actual node/code
      * @returns {Function} function to be executed by the fixer
      * @private
      */
-    function getFixerFunction(node, needed, gotten) {
-      if (needed > gotten) {
-        var spaces = indentChar.repeat(needed - gotten);
```
jayphelps: The old code used `String.prototype.repeat()` which is ES2015. I can switch back to using that instead of the `join()` trick I used, if it's a safe to assume repeat is supported.

ljharb: Alternatively you can just bring in a dep like https://www.npmjs.com/package/string-repeat

tests/lib/rules/jsx-indent.js
ljharb: I'd rather leave the broken but correct test, and comment it out, than leave this incorrect test. Could we do that?

lib/rules/jsx-indent.js
jayphelps: just let me know which way you'd like. _to me_ it seems silly to bring in a lib since the join trick is short, simple, and I think fairly obvious what it does. 😏  aka left-pad

ljharb: In general, `left-pad` in no way changes that more deps is better. `left-pad` was only a problem for people that were already ignoring a decade of best practices and depending on third-party registries for their dependencies.

In this case, keeping your join trick is fine too.

jayphelps:
> I'd rather leave the broken but correct test, and comment it out, than leave this incorrect test. Could we do that?

Done.

jayphelps:
lmk if there's anything else I missed 🎉

yannickcr:
Merged, thanks! And sorry for the lack of responses.

sassanh:
Is it released to npm? How can we use it?

yannickcr:
@sassanh yes, it's available in v6.5.0. You can use it with `eslint --fix file.js` (see [ESLint documentation](http://eslint.org/docs/user-guide/command-line-interface#fix)).

sassanh:
That was fast. Thanks.

PR #610, opened 2016-05-22
Fix jsx-curly-spacing: spacing around spread operator

gitim:
Fixes https://github.com/yannickcr/eslint-plugin-react/issues/606
Added two type of tests:
* creating jsx elements using only JSXSpreadAttribute syntax
* creating jsx elements using both JSXSAttribute  and JSXSpreadAttribute syntax

### DIFF
--- a/lib/rules/jsx-curly-spacing.js
+++ b/lib/rules/jsx-curly-spacing.js
@@ -143,13 +143,14 @@ module.exports = function(context) {
   /**
    * Determines if spacing in curly braces is valid.
    * @param {ASTNode} node The AST node to check.
-   * @param {Token} first The first token to check (should be the opening brace)
-   * @param {Token} second The second token to check (should be first after the opening brace)
-   * @param {Token} penultimate The penultimate token to check (should be last before closing brace)
-   * @param {Token} last The last token to check (should be closing brace)
    * @returns {void}
    */
-  function validateBraceSpacing(node, first, second, penultimate, last) {
+  function validateBraceSpacing(node) {
+    var first = context.getFirstToken(node);
+    var last = sourceCode.getLastToken(node);
+    var second = context.getTokenAfter(first);
+    var penultimate = sourceCode.getTokenBefore(last);
+
     if (spaced) {
       if (!sourceCode.isSpaceBetweenTokens(first, second)) {
         reportRequiredBeginningSpace(node, first);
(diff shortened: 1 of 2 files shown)


lib/rules/jsx-curly-spacing.js
```diff
@@ -143,13 +143,14 @@ module.exports = function(context) {
...
-   * @param {Token} penultimate The penultimate token to check (should be last before closing brace)
-   * @param {Token} last The last token to check (should be closing brace)
    * @returns {void}
    */
-  function validateBraceSpacing(node, first, second, penultimate, last) {
+  function validateBraceSpacing(node) {
+    var first = context.getFirstToken(node);
+    var last = sourceCode.getLastToken(node);
+    var second = context.getTokenAfter(first);
+    var penultimate = sourceCode.getTokenBefore(last);
```
gitim: just put these declarations inside the function to reduce copy pasting

tests/lib/rules/jsx-curly-spacing.js
```diff
@@ -83,6 +83,88 @@ ruleTester.run('jsx-curly-spacing', rule, {
...
+      '...bar',
+      '}/>;'
+    ].join('\n'),
+    options: ['always'],
+    parserOptions: parserOptions
+  }, {
+    code: '<App foo={bar/* comment */} {...baz/* comment */} />;',
+    options: ['never'],
+    parserOptions: parserOptions,
+    parser: 'babel-eslint'
```
ljharb: why do you need the babel-eslint parser for jsx object spread? that should only be needed for JS object spread.

gitim: ah, sorry, just copy pasted and modified this case https://github.com/yannickcr/eslint-plugin-react/blob/master/tests/lib/rules/jsx-curly-spacing.js#L81-L86

ljharb: hm, i wouldn't expect that one to need it either

gitim: this test case was added in https://github.com/yannickcr/eslint-plugin-react/pull/584, and it seems that the issue wasn't related with babel-eslint parse, so I think it is safe to remove them both

yannickcr:
Merged, thanks!

PR #616, opened 2016-05-31
Add new rule: no-comment-textnodes

benvinegar:
Our team has run into issues where people mistakenly commit comments as text nodes, and they have managed to get through code review and appeared (embarrassingly) in our UI. I threw together this rule to try and catch such occasions, and thought it might be useful to other users of this ESLint plugin.

Feedback, comments greatly appreciated.

### DIFF
--- a/docs/rules/no-comment-textnodes.md
+++ b/docs/rules/no-comment-textnodes.md
@@ -1,0 +1,68 @@
+# Prevent comments from being inserted as text nodes (no-comment-textnodes)
+
+This rule prevents comment strings (e.g. beginning with `//` or `/*`) from being accidentally
+injected as a text node in JSX statements.
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```js
+var Hello = React.createClass({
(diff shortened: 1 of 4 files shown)


ljharb:
What about false positives where the literal JS comment text is what's desired?

benvinegar:
@ljharb – They have two choices. Either disable the rule, or do:

```javascript
<div>{'// Not a comment'}</div>
```

I personally think that's acceptable. This strikes me as a particularly rare edge case.

ljharb:
@benvinegar want to add documentation for the rule, and include that info in it explicitly? :-D

benvinegar:
@ljharb – Updated with documentation.

yannickcr:
Seems to be a useful rule. Merged, thanks!

remitbri:
a bite late, but: shouldn't the name of the rule be prefixed with `jsx-`?

MoOx:
I agree that this rule should have been prefixed with jsx-.

PR #623, opened 2016-06-05
`jsx-curly-spacing`: fix “alternative” option

ljharb:
Rule now accepts a `spacing` object, which takes its defaults from the first always/never option. It accepts:
 - “default”: which controls all values unless otherwise specified.
 - “objectLiterals”: controls object literal values.
(in other words, omitted spacing option values default to the first option, which is "always" or "never")

Warnings around object literals themselves are removed, because the built-in `object-curly-spacing` rule handles that.

(Re https://github.com/yannickcr/eslint-plugin-react/pull/604#issuecomment-223824638)

### DIFF
--- a/docs/rules/jsx-curly-spacing.md
+++ b/docs/rules/jsx-curly-spacing.md
@@ -113,12 +113,10 @@ You can specify an additional `spacing` property that is an object with the foll
 
 ```json
 "jsx-curly-spacing": [2, "always", {"spacing": {
-  "default": "always",
   "objectLiterals": "never"
 }}]
 ```
 
-* `default`: This controls the default behavior for all scenarios, unless a more specific override applies.
 * `objectLiterals`: This controls different spacing requirements when the value inside the jsx curly braces is an object literal.
 
 All spacing options accept either the string `"always"` or the string `"never"`. Note that the default value for all "spacing" options matches the first "always"/"never" option provided.
--- a/docs/rules/jsx-curly-spacing.md
+++ b/docs/rules/jsx-curly-spacing.md
@@ -107,25 +107,35 @@ The following patterns are not warnings:
 <Hello name={ {firstname: 'John', lastname: 'Doe'} } />;
 ```
 
-#### Alternative
+#### Granular spacing controls
 
-When setting the `alternative` option to `true` you must collapse the curly braces:
+You can specify an additional `spacing` property that is an object with the following possible values:
 
 ```json
-"jsx-curly-spacing": [2, "always", {"alternative": true}]
+"jsx-curly-spacing": [2, "always", {"spacing": {
+  "default": "always",
+  "objectLiterals": "never"
+}}]
 ```
 
-When `"always"` is used and `alternative` is `true`, the following pattern is not warnings:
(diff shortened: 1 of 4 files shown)


ljharb:
(hold up on this one, i want to add some more test cases)

ljharb:
@yannickcr k, this should be gtg :-) thoughts?

ljharb:
ping!

docs/rules/jsx-curly-spacing.md
```diff
@@ -107,26 +107,36 @@ The following patterns are not warnings:
...
 ```json
-"jsx-curly-spacing": [2, "always", {"alternative": true}]
+"jsx-curly-spacing": [2, "always", {"spacing": {
+  "default": "always",
+  "objectLiterals": "never"
+}}]
 ```
 
-When `"always"` is used and `alternative` is `true`, the following pattern is not warnings:
+* `default`: This controls the default behavior for all scenarios, unless a more specific override applies.
```
yannickcr: I don't see the need for the `default` control, is it not the same as the main option ?

docs/rules/jsx-curly-spacing.md
```diff
@@ -107,26 +107,36 @@ The following patterns are not warnings:
...
+* `objectLiterals`: This controls different spacing requirements when the value inside the jsx curly braces is an object literal.
```
yannickcr: What about using `true`/`false` instead of `always`/`never`? (to get the same behavior as `object-curly-spacing`)

docs/rules/jsx-curly-spacing.md
ljharb: Yeah, that's a good point. My hope was to actually remove the need for the initial option - ie, to allow "always/never", OR the options object, but I don't know enough about eslint rules to configure that. Is that possible?

docs/rules/jsx-curly-spacing.md
ljharb: I suppose that's reasonable, but what if someone later wants a third option? Using strings is extensible, using booleans forever limits us to two options.

docs/rules/jsx-curly-spacing.md
yannickcr: I'd prefer to keep the initial option to get a signature similar to the `object-curly-spacing` rule.

docs/rules/jsx-curly-spacing.md
yannickcr: Right, I'm ok to keep `always`/`never` in this case.

docs/rules/jsx-curly-spacing.md
ljharb: ok - so to sum up, i'll remove "default" since the initial option covers that.

yannickcr:
Merged, thanks!

ljharb:
Yay, thanks!

Any idea when the next release will be? I'm waiting on it to release the next `eslint-config-airbnb` :-)

yannickcr:
Yep, I plan to do a new release this week (likely before the week-end).

PR #631, opened 2016-06-10
Add ignore config to no-unknown-property rule

insin:
Context: insin/babel-plugin-react-html-attrs#4

### DIFF
--- a/docs/rules/no-unknown-property.md
+++ b/docs/rules/no-unknown-property.md
@@ -22,6 +22,17 @@ var React = require('react');
 var Hello = <div className="hello">Hello World</div>;
 ```
 
+## Rule Options
+
+```js
+...
+"no-unknown-property": [<enabled>, { ignore: <ignore> }]
+...
+```
+
+* `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
+* `ignore`: optional array of property and attribute names to ignore during validation.
+
 ## When Not To Use It
 
 If you are not using JSX you can disable this rule.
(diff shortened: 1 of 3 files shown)


docs/rules/no-unknown-property.md
```diff
@@ -22,6 +22,16 @@ var React = require('react');
 var Hello = <div className="hello">Hello World</div>;
 ```
 
+## Rule Options
+
+HTML properties and attributes which would normally be considered warnings can be whitelisted. If you wanted to allow use of 'class' and 'for' attributes because your Babel configuration handles transforming them to DOM attributes, the configuration would be:
+
+```js
+"rules": {
+  "react/no-unknown-property": [1, { "allow": ["class", "for"] }],
```
ljharb: As far as I know, React itself doesn't support "class" or "for", and requires "className" and "htmlFor". Can you point to some documentation that allows for this usage?

insin: React doesn't support it, but you can use Babel plugins to handle usage of 'class' and 'for' at transpile time for teams who prefer to keep using them, e.g: https://github.com/insin/babel-plugin-react-html-attrs/issues/4#issuecomment-225307332

yannickcr:
That's a very specific use case, but I am not against adding such option.

Just a couple of things:

* I'd prefer to use an `ignore` option instead of `allow` (like [prop-types](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md) or [jsx-pascal-case](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-pascal-case.md))
* You are referring to your specific use case in the documentation. I think it can be confusing. Maybe just saying that you can ignore some properties is enough ?

insin:
Changed the option to `ignore` and made the documentation more generic, as per other rules.

yannickcr:
Merged, thanks!

PR #633, opened 2016-06-14
[added] react/ prefix to rule names

appsforartists:
The README didn't make it clear that rule names should be prefixed with `react/`, so I fixed it.  Reasonable solutions:

1. Show an example `rules` block.
2. Add `react/` to each rule name in the index.

I've done both here.  Feel free to choose one or accept both.  :smiley:

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,8 +72,8 @@ Finally, enable all of the rules that you would like to use.  Use [our preset](#
 
 ```json
   "rules": {
-    "react/jsx-uses-react": "warn",
-    "react/jsx-uses-vars": "warn",
+    "react/jsx-uses-react": "error",
+    "react/jsx-uses-vars": "error",
   }
 ```  
 
--- a/README.md
+++ b/README.md
@@ -68,60 +68,65 @@ With ESLint 2.x.x:
 }
 ```
 
-Finally, enable all of the rules that you would like to use.
+Finally, enable all of the rules that you would like to use.  Use [our preset](#user-content-recommended-configuration) to get reasonable defaults quickly, and/or choose your own:
 
-The plugin has a [recommended configuration](#user-content-recommended-configuration) that enforces React good practices.
+```json
+  "rules": {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,60 +68,65 @@ With ESLint 2.x.x:
 }
 ```
 
-Finally, enable all of the rules that you would like to use.
+Finally, enable all of the rules that you would like to use.  Use [our preset](#user-content-recommended-configuration) to get reasonable defaults quickly, and/or choose your own:
 
-The plugin has a [recommended configuration](#user-content-recommended-configuration) that enforces React good practices.
+```json
+  "rules": {
+    "react/jsx-uses-react": "warn",
```
ljharb: if we're going to use an example, let's use "error" instead of "warn" - warnings are often ignored and thus mostly useless :-)

appsforartists: Fixed.

Honestly, I wasn't sure what the right word was to use there.  `jsx-uses-react` fixes `eslint` errors, rather than introducing them.

ljharb:
I think I like this change - it's a bit noisier but also more explicit.

I'll let @yannickcr merge it, but it'd be good if @appsforartists wouldn't mind rebasing down to a single commit first? :-)

appsforartists:
@ljharb Isn't that what the squash and merge button is for? 😉 

I've been lazy and just made these changes from the GitHub UI.

ljharb:
It's what the "rebase and merge" button should be for :-p blindly squashing commits on merge messes with the changelog :-( but no worries if you're editing in the UI.

appsforartists:
Oh.  I [thought they fixed that](https://github.com/blog/2141-squash-your-commits).  😢

yannickcr:
Thanks. It will really help those who are new to ESLint plugins.

appsforartists:
Glad I could be helpful!

Thanks for putting this module together, and for landing my patch. :smiley:

PR #634, opened 2016-06-15
Integrate jsx-ast-utils

evcohen:
as discussed in evcohen/eslint-plugin-jsx-a11y#54 we can integrate the [jsx-ast-utils](https://github.com/evcohen/jsx-ast-utils) package here for better code reuse across these packages. 

We can start adding API functions to fit use cases for this project as well to `jsx-ast-utils` in the next few iterations (i.e. styling and spacing functions)

### DIFF
--- a/lib/rules/jsx-no-bind.js
+++ b/lib/rules/jsx-no-bind.js
@@ -5,6 +5,8 @@
  */
 'use strict';
 
+var propName = require('jsx-ast-utils').propName;
+
 // -----------------------------------------------------------------------------
 // Rule Definition
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


lib/rules/jsx-no-bind.js
```diff
@@ -5,6 +5,8 @@
  */
 'use strict';
 
+var propName = require('jsx-ast-utils').propName;
```
ljharb: it'd be nice if we could require via path, ie, `var propName = require('jsx-ast-utils/propName')` rather than having to pull everything in and grab one property - but obv this is not a blocker :-)

ljharb:
Awesome, love the code sharing

lib/rules/jsx-no-bind.js
evcohen: ah yes ill export from root directory too 😄

yannickcr:
This is really great!

LGTM. if it is good to merge for you too I think we can merge it.

evcohen:
Good for me - will try and continually send PR's as we flesh out `jsx-ast-utils`

yannickcr:
Merged. Thank you very much :)

PR #611, opened 2016-05-23
Adds new `no-unused-prop-types` rule

EvNaverniouk:
Fixes #226 by adding an new `unused-prop-types` rule.

The rule is based very heavily on the `prop-types` rule and shares a lot of the same code. For the time being, I have not made any attempts at breaking out the shared code into a reusable utility file. Please advise on the best way to do this. I was simply thinking of dropping a `proptypes.js` file into the `utils` directory with all the shared functions. If that's acceptable, let me know and I will do that.

Have tested against several React projects with great success so far. No false positive problems yet, but it needs a lot more testing against other repos. I'm hoping other users will report issues as they try it. For my needs, this rule is working great.

Can be tested by following these steps:

```
npm install EvNaverniouk/eslint-plugin-react#6d309e13a65a1032577ae0c98081d778f85d515a
```

Then enable the rule in your eslintrc:

```
"rules": {
    "react/unused-prop-types": 2
}
```

PR #635, opened 2016-06-16
Rule proposal: jsx-no-undef-props

beefancohen:
This rule enforces that no props have the value `undefined`. In React, these props will not be passed down to the instance of the component, so just suggest to the user to leave off the prop instead of passing undefined as its value.

Not sure if this is wanted or needed, but I thought it could be useful! Don't be afraid to say you think it's stupid though 😝


docs/rules/jsx-no-undef-props.md
```diff
@@ -0,0 +1,23 @@
...
+
+This rule prevents setting a prop value as `undefined` and suggests you just don't pass the prop down. In React, that prop will be dropped by the parent, anyway.
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```js
+<Hello name={undefined} />;
+<Hello name={foo ? undefined : bar} />;
```
ljharb: How else would you make a prop conditionally present?

ljharb:
This seems like it adds a little value when the prop value is solely the literal `undefined` - but it wouldn't be able to catch any variables whose contents were undefined, and the conditional case seems like a useful way to conditionally specify a prop.

docs/rules/jsx-no-undef-props.md
evcohen: True - you can move the logic outside of the prop value and just pass a var, but should probably just special case this to ignore. In this case, the rule is kinda pointless :)

evcohen:
Yeah that was kind of the point to check for literally `undefined` - this was more for newcomers to React to say you don't need to pass down an undefined prop that you've maybe defined in `propTypes`. I can see, also, that this conflicts with #611 which I think is more valuable, anyway.

ljharb:
I don't see how it conflicts with #611 - that's checking for unused propTypes within the same component file, at the definition site, while this one checks for undefined props at the rendering site.

We already have a rule that checks for literal `true`, so I think it's helpful to have one for literal `undefined` as well.

evcohen:
Ah you're right I misunderstood - it would only conflict if someone defined a prop in `propTypes` and then weirdly just passed it down with undefined as a value, but I don't think that's much to worry about. I'll update to just check for the literal `undefined`.

ljharb:
LGTM

lib/rules/jsx-no-undef-props.js
```diff
@@ -0,0 +1,36 @@
...
+
+'use strict';
+
+var propName = require('jsx-ast-utils/propName');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = function(context) {
```
petersendidit: This should use the [new eslint rule format](http://eslint.org/blog/2016/07/eslint-new-rule-format)

ljharb:
@evcohen this seems like a bad merge - can you force push to the branch and update the PR in-place?

evcohen:
@ljharb what the heck just happened, i rebased against master 😢

ljharb:
@evcohen I think it needs a rebase, NOT a merge, so as to reduce the 56 commits included in this PR to a more manageable number

evcohen:
@ljharb as soon as i get a few mins i will rebase and squash

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
+<<<<<<< HEAD
```
ljharb: i think you need to fix some merge conflicts :-)

evcohen: that's not supposed to be there?!? 😛

lib/rules/jsx-no-undef-props.js
```diff
@@ -0,0 +1,44 @@
...
+          return;
+        } else if (typeof getPropValue(node) !== 'undefined') {
+          return;
+        }
+
+        var name = propName(node);
+
+        context.report({
+          node: node,
+          message: 'Leave off the prop `' + name + '` instead of passing undefined as its value.'
```
ljharb: should this check for `null` also, since React ignores them both?

evcohen: also re: following spread operator- if the above is true and you want to omit a value, we should probably ignore in the case that you want the defaultProp to be applied. or maybe this renders this rule futile because of that distinction.

ljharb: ah, good to know, thanks

ljharb: right, i'm suggesting always ignoring cases or either that appear after a spread prop.

Issue #651, opened 2016-06-28
Request: Disallow iteration in render


TheSavior:
What are your thoughts on a rule like this:

Invalid

```
React.createClass({
  propTypes: {
    array: React.PropTypes.array
  },

  render: function() {
    return (
      <div>
      {
        this.props.array.map((item) => <MyComponent {...item})
      }
      </div>
    );
  }
});
```

Valid

```
React.createClass({
  propTypes: {
    array: React.PropTypes.array
  },

  render: function() {
    return (
      <div>
      {this.renderItems()}
      </div>
    );
  },

  renderItems: function() {
    return this.props.array.map((item) => <MyComponent {...item});
  }
});
```

ljharb:
Why is this preferable? Personally I much prefer the former, because then I can make the component an SFC - since it has no state.

TheSavior:
Our team likes and enforces this rule because we find it easier and quicker to read the render function to see what a component is doing by seeing the well named function calls instead of inline maps.

Real code sample from our repo:

```
React.createClass({
  // ...
  render() {
    return (
      <div>
        <table>
          <tbody>
            {this.renderCashRows()}
            {this.renderTransferRows()}
            {this.renderUnsureRows()}
            {this.renderUnsupportedRows()}
          </tbody>
        </table>
        {this.renderTaxConsequences()}
      </div>
    );
  }
  // ...
});
```

We find it way more readable to see that than to see the inline maps those functions are doing. Because of this, we socially enforce it everywhere, but want to be able to lint for it as well.

TheSavior:
@ljharb, since you are a 👎  on the PR, can you provide some more insight into your thoughts on the code sample in my above comment? I'd like to better understand how you would approach that type of render with multiple iterations. I'm sure I'll learn something.

What if the calls to map weren't just passing an item to a prop, what if the map statements were many lines doing some work on each item and then passing the results as props to a child item? What if those functions had `if` statements in them to conditionally return out instead of doing the map? Do you still believe that work should be done directly in the render function?

ljharb:
Here's how I'd write it:
```js
function Component({ array }) {
   return (
     <div>
       {array.map(item => (
         <MyComponent {...item} />
       ))}
     </div>
  );
}
Component.propTypes = {
  array: React.PropTypes.array,
};
```

[...]

PR #652, opened 2016-06-29
New Rule: no-render-iteration

TheSavior:
Implementing #651


ljharb:
I'm -1 on this rule, because doing an inline map is idiomatic React, because `.map` doesn't mean it's `Array#map`, and because doing an autofix for this rule wouldn't be feasible since there's so many ways to work around it.

(If we do end up adding this rule, then I'd want an alternative rule for eslint-config-airbnb that *forbids* instance methods that contain iterations, and requires that they be inline + SFCs - but I'd prefer having neither, tbh)

TheSavior:
Alright, I'll close this out and our team will discuss a different approach. I'd like to continue the conversation over in #651 to keep it all together.

ljharb:
Let's leave it open for now, until @yannickcr has weighed in.

PR #657, opened 2016-07-02
bumps deps; peerDeps allows ESLint 2 or 3

jokeyrhyme:
- specify ESLint 2 **or 3** as a peerDependency to help `npm` users identify compatibility issues

- specify Node.js 0.10 as the minimum (same as ESLint 2)

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,11 +27,17 @@
     "jsx-ast-utils": "^1.2.1"
   },
   "devDependencies": {
-    "babel-eslint": "6.0.5",
+    "babel-eslint": "^6.0.5",
     "coveralls": "2.11.9",
-    "eslint": "2.13.1",
+    "eslint": "^3.0.0",
     "istanbul": "0.4.4",
     "mocha": "2.5.3"
+  },
+  "peerDependencies": {
+    "eslint": "^2.13.1 || ^3.0.0"
+  },
+  "engines": {
+    "node": ">=0.10"
   },
   "keywords": [
     "eslint",


ljharb:
Wouldn't it be wiser to wait on updating the rule format, so there can be a version of the plugin that supports both eslint 2 and eslint 3 for awhile?

jokeyrhyme:
@ljharb my changes here are just a proposal of course, with the maintainers having final say. I made these changes assuming that the MAJOR SemVer number would be bumped, so ESLint 2 users could still use the previous version without any problems.

Do we want to continue making changes to the ESLint 2 version of this plugin? Or can we safely leave it as-is?

ljharb:
Yes, ideally we would keep making changes to the eslint "2 + 3" version for awhile, so as to give eslint 2 users (and node 0.10 users) as smooth an upgrade path as possible, for as long as possible. The only time this PR will become necessary is after eslint v4 comes out, so I'm inclined not to rush it.

jokeyrhyme:
Okay. If I get time later, I'll change this PR to keep the old ESLint 2 rule format. Or, someone else can beat me to it by creating a less turbulent PR. :)

jokeyrhyme:
@ljharb okies, I've removed all rule format changes, this is purely about enabling usage with ESLint 3 without touching existing support for ESLint 2

package.json
```diff
@@ -27,12 +27,18 @@
     "jsx-ast-utils": "^1.2.1"
   },
   "devDependencies": {
-    "babel-eslint": "6.0.5",
+    "babel-eslint": "^6.0.5",
```
ljharb: can this not be bumped to 6.1.0?

ljharb:
LGTM

elliottsj:
What's blocking this from being merged?

ljharb:
@yannickcr it'd be great to get a release of this :-)

yannickcr:
@ljharb Thanks for your help, I'm a bit late on this :( I'll try to do a release tomorrow.

ljharb:
@yannickcr thanks! looking forward to releasing a new eslint-config-airbnb ;-)

ljharb:
@jokeyrhyme turns out eslint v2 supports the new rule format already (https://github.com/benmosher/eslint-plugin-import/issues/406#issuecomment-231132889) so if you'd like to make a new PR that upgrades the rules, id be happy to merge it :-)

jokeyrhyme:
@ljharb woah, that's really good planning on ESLint's part. I'll see how extra-curricular I'm feeling this weekend. :)

yannickcr:
@ljharb Since adding a peerDepencency is a big change I prefer to consider it as a breaking change. 

I am not ready yet to publish the `v6.0.0` (I'd like to add a few more waiting changes in it) so I published it as `v6.0.0-alpha.1`. You can install it with `npm install eslint-plugin-react@next`.

ljharb:
@yannickcr understood - in this case I don't think anybody is still using eslint v1 with eslint-plugin-react, but it's good to be cautious. I won't be able to release an updated eslint-config-airbnb without a non-alpha eslint-plugin-react anyways, but [as you can see](https://github.com/airbnb/javascript/issues/936#issuecomment-230108045) I've got a few more blockers, so within the next week or so would be ideal if that's ok :-)

jokeyrhyme:
Do you have a GitHub Milestone for 6.0, or a list somewhere? Want some help?

yannickcr:
@ljharb @jokeyrhyme I do not have a lot of time this week but I'll try to do an roadmap before the week-end (like I made for 4.0 #436)

PR #658, opened 2016-07-03
`react/display-name`: `ignoreTranspilerName` should only work with `React.createClass`

sheepsteak:
Babel's plugin to add `displayName` ([`babel-plugin-transform-react-display-name`](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-react-display-name)) to components only works with `React.createClass` components (some more info [here](https://phabricator.babeljs.io/T1331)).

The [`react/display-name`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/display-name.md) rule in this repo, however, gives the impression that Babel will add a `displayName` to any `class` or `function` component too as the rule reports them as valid when they are missing `displayName` and the `ignoreTranspilerName` option is `false`.

This change brings the rule into line with what actually happens by not assuming that a `displayName` will be added by Babel for `class` and `function` components.

I realise this is quite a big change for this rule but I think it's causing a lot of people to think a `displayName` is being added automatically when it isn't at the moment. Especially since most people don't really use `React.createClass` anymore. Also, I think because React will use the `function.name` or `class.name` in error messages and the developer tools when `displayName` is missing this problem is being hidden. 

I've updated the tests and tried to keep good coverage but please let me know if anything is missing.

### DIFF
--- a/lib/rules/display-name.js
+++ b/lib/rules/display-name.js
@@ -91,32 +91,8 @@ module.exports = Components.detect(function(context, components, utils) {
         node.parent.parent &&
         node.parent.parent.type === 'VariableDeclarator'
     );
-    var namedClass = (
-      (node.type === 'ClassDeclaration' || node.type === 'ClassExpression') &&
-      node.id &&
-      node.id.name
-    );
 
-    var namedFunctionDeclaration = (
-      (node.type === 'FunctionDeclaration' || node.type === 'FunctionExpression') &&
-      node.id &&
-      node.id.name
-    );
-
-    var namedFunctionExpression = (
-      (node.type === 'FunctionExpression' || node.type === 'ArrowFunctionExpression') &&
-      node.parent &&
-      (node.parent.type === 'VariableDeclarator' || node.parent.method === true)
-    );
-
-    if (
-      namedObjectAssignment || namedObjectDeclaration ||
-      namedClass ||
-      namedFunctionDeclaration || namedFunctionExpression
-    ) {
-      return true;
-    }
-    return false;
+    return namedObjectAssignment || namedObjectDeclaration;
   }
 
   // --------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


ljharb:
I'm pretty sure that Babel will in fact add the inferred class/function name, per ES6, which will then be used as the display name by React.

Do you have an example case where this isn't happening?

sheepsteak:
How I discovered this was trying to run unit tests on a Node 6 preset - `babel-preset-es2015-node6` so that fewer plugins were being used and hopefully the Mocha tests would speed up for our project. I noticed I was getting failing tests with `enzyme` though - especially with Higher Order Components (`connect` in `redux`) where wrapping a component and producing a `displayName` like `connect(MyComponent)` was coming out as `connect(Component)`.

When I looked into this more I realised that it was because there were places where the fallback to `function.name` was broken for various reasons. I'd always assumed (along with most people I think) that Babel was able to add `displayName` for all types of components so it was pretty confusing at first.

I've made a repo here you can run to get a glimpse of the problem - https://github.com/sheepsteak/eslint-node-6

It will generate output like this:
```
ES5
TestComponent1 (createClass) displayName: TestComponent1
TestComponent1 (createClass) name:
TestComponent2 (func) displayName: undefined
TestComponent2 (func) name: TestComponent2
TestComponent3 (class) displayName: undefined
TestComponent3 (class) name: TestComponent3
```
```
ES2015
TestComponent1 (createClass) displayName: TestComponent1
TestComponent1 (createClass) name: Constructor
TestComponent2 (func) displayName: undefined
TestComponent2 (func) name:
TestComponent3 (class) displayName: undefined
TestComponent3 (class) name: TestComponent3
```

You can see that for `TestComponent2` it compiles down to this for ES5 where a normal function with a name is generated:
```javascript

var TestComponent2 = exports.TestComponent2 = function TestComponent2() {
  return _react2.default.createElement(
    'div',
    null,
    'TestComponent2'
  );
};
```

However, for ES2015 it compiles to this where function name is lost and V8 is unable to guess what it is as the `exports.TestComponent2` confuses it I think:
```javascript
const TestComponent2 = exports.TestComponent2 = () => _react2.default.createElement(
  'div',
  null,
  'TestComponent2'
);
```

Anyway, this isn't a major bug or anything but I just thought it was interesting how things can break when dropping Babel plugins and moving to native equivalents. You just assume it will kind of work 😄 

I thought the safest option was to force developers to add `displayName` where Babel couldn't provide one for them and prevent _possible_ problems later on for them. The best outcome would be to update the Babel plugin that handles `displayName` so that it provides them for `class` and `function` components so everyone can just carry on as normal.

Do you think it's maybe better to just update the docs on the rule with a warning instead? Or just drop it entirely?

ljharb:
That seems like a Babel bug that I'd file with them. Losing the explicit function name is not OK.

As for https://github.com/sheepsteak/eslint-node-6/blob/master/src/component.js#L11 - the Airbnb style guide recommends avoiding arrow functions for SFCs, due to function name inference. However, the JS spec does actually require that `TestComponent2.name === 'TestComponent2'` in that example - so again I think it's just a Babel bug.

ljharb:
This seems answered.

PR #665, opened 2016-07-06
Fix no-comment-textnodes incorrectly catching urls (fixes #664)

### LINKED ISSUE #664: `no-comment-textnodes` false positive on URL
I have JSX like the following:
```js
<strong>
  &nbsp;https://www.example.com/attachment/download/1
</strong>
```

Since the line doesn't start with `//` or `/*`, there shouldn't be any way this is identified as an accidental leaked comment.

petersendidit:
Fixes #664.

### DIFF
--- a/lib/rules/no-comment-textnodes.js
+++ b/lib/rules/no-comment-textnodes.js
@@ -19,7 +19,7 @@ module.exports = function(context) {
 
   return {
     Literal: function(node) {
-      if (/\s*\/(\/|\*)/.test(node.value)) {
+      if (/^\s*\/(\/|\*)/m.test(node.value)) {
         // inside component, e.g. <div>literal</div>
         if (node.parent.type !== 'JSXAttribute' &&
             node.parent.type !== 'JSXExpressionContainer' &&
--- a/tests/lib/rules/no-comment-textnodes.js
+++ b/tests/lib/rules/no-comment-textnodes.js
@@ -109,6 +109,15 @@ ruleTester.run('jsx-needs-i18n', rule, {
       args: [1],
       parser: 'babel-eslint'
     },
+    {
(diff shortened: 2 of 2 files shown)


ljharb:
@petersendidit if you rebase on latest master, your tests should pass.

ljharb:
I'd forgotten to update the appveyor config - it's done now.

lib/rules/no-comment-textnodes.js
```diff
@@ -19,7 +19,7 @@ module.exports = function(context) {
 
   return {
     Literal: function(node) {
-      if (/\s*\/(\/|\*)/.test(node.value)) {
+      if (/^\s*\/(\/|\*)/m.test(node.value)) {
```
ljharb: since this is just testing the start of a quasi-comment, does it need the "m"?

petersendidit: Without the `m` modifier things like this would fail:
```javascript
<div>
  asdjfl
  /* invalid */
  foo
</div>
```

ljharb: ah right, i'm assuming the text node will be split up by lines, and that's not likely to be the case.

Could we add your example as a test, to prevent a regression?

petersendidit: It's already a test https://github.com/yannickcr/eslint-plugin-react/blob/master/tests/lib/rules/no-comment-textnodes.js#L170

ljharb: ah, understood, thanks for clarifying :-)

ljharb:
@yannickcr ping :-)

yannickcr:
Merged, thanks!

PR #667, opened 2016-07-06
[Fix] Don’t check filenames on text input.

### LINKED ISSUE #662: jsx-filename-extension errors on "executeOnText"
When using the JS API for eslint, I'm getting jsx-filename-extension errors.

```js
import { CLIEngine } from 'eslint';
const cli = new CLIEngine({
  useEslintrc: false,
  plugins: ['react'],
  parserOptions: {
    ecmaFeatures: {
      jsx: true,
    },
  },
  ecmaFeatures: {
    jsx: true
  },
  rules: {
    'react/jsx-filename-extension': 2,
  },
});

const { results } = cli.executeOnText(`
var React = require('react');
module.exports = function SFC() { return <div />; };
`);
```
I would expect `results[0].messages` to be an empty array, since `results[0].filePath` is `<text>`.

However, I get:
```js
{ filePath: '<text>',
  messages: 
   [ { ruleId: 'react/jsx-filename-extension',
       severity: 2,
       message: 'JSX not allowed in files with extension \'\'',
       line: 3,
       column: 42,
       nodeType: 'JSXElement',
       source: 'module.exports = function SFC() { return <div />; };' } ],
  errorCount: 1,
  warningCount: 0 }
```

ljharb:
Fixes #662.

### DIFF
--- a/lib/rules/jsx-filename-extension.js
+++ b/lib/rules/jsx-filename-extension.js
@@ -32,9 +32,12 @@ module.exports = function(context) {
   return {
 
     JSXElement: function(node) {
-      var allowedExtensions = getExtensionsConfig();
       var filename = context.getFilename();
+      if (filename === '<text>') {
+        return;
+      }
 
+      var allowedExtensions = getExtensionsConfig();
       var isAllowedExtension = allowedExtensions.some(function (extension) {
         return filename.slice(-extension.length) === extension;
       });
--- a/tests/lib/rules/jsx-filename-extension.js
+++ b/tests/lib/rules/jsx-filename-extension.js
@@ -34,6 +34,11 @@ ruleTester.run('jsx-filename-extension', rule, {
 
   valid: [
     {
+      filename: '<text>',
+      code: withJSX,
+      parserOptions: parserOptions
+    },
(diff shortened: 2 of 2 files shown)


lib/rules/jsx-filename-extension.js
```diff
@@ -32,9 +32,12 @@ module.exports = function(context) {
   return {
 
     JSXElement: function(node) {
-      var allowedExtensions = getExtensionsConfig();
       var filename = context.getFilename();
+      if (filename === '<text>') {
```
lencioni: I wonder if this comparison could reference a variable or something instead of hard-coding `<text>`

lencioni:
LGTM! I didn't even realize this was a thing.

lib/rules/jsx-filename-extension.js
ljharb: Sadly the only place it exists is https://github.com/eslint/eslint/blob/master/lib/cli-engine.js#L220

ljharb:
@yannickcr ping :-)

yannickcr:
Merged, thanks!

PR #669, opened 2016-07-07
New rule options allowDecorators in require-optimization

Tom910:
I like this require-optimization rule, but it is not suitable for our implementation. I think a lot of people use https://www.npmjs.com/package/pure-render-decorator or similar solutions. It is the reason I made an additional option. Now it will be possible to register a list of names of decorators, which allow to pass validation.

I hope my change will be useful and will get in release =)

### DIFF
--- a/docs/rules/require-optimization.md
+++ b/docs/rules/require-optimization.md
@@ -52,14 +52,30 @@ React.createClass({
 
 ```js
 ...
-"require-optimization": [<enabled>]
+"require-optimization": [<enabled>, { allowDecorators: [<allowDecorator>] }]
 ...
+```
+
+* `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
+* `allowDecorators`: optional array of decorators names to allow validation.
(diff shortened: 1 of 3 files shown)


ljharb:
How does this handle that decorators are actually in-scope bindings, and not string names? In other words, the name "foo" might mean different things in different files.

Tom910:
> In other words, the name "foo" might mean different things in different files.

Yes, the name of the decorator may vary. It's up to the developer to make up some convention and to name imports conveniently and in the same way.

There is no need to perform any deep checks. This addition allows the developer to use any optimization decorator, not only pure mixin, but any other, including his own utilities. The only purpose of allowDecorators option is to provide names of this tools to the rule.

How can this be made in a different way? Analyze the decorator? Or check imports? There can be too many variations.

ljharb:
That's totally a fair point - just wanted to make sure it was considered.

yannickcr:
Merged, thanks!

Tom910:
Thanks!

PR #670, opened 2016-07-08
Added `shorthandFirst` option to `jsx-sort-props` rule. (Fixes #391)

### LINKED ISSUE #391: shorthandLast to match shorthandFirst in `jsx-sort-props`
Would this be welcome? Or would it be better to keep it a simple property with different values?

Thanks!

mathieumg:
I think `callbackLast` having precedence is a better fit than the other way around.


yannickcr:
Merged, thanks!

PR #676, opened 2016-07-12
Fix jsx-no-bind reporting errors on render functions

### LINKED ISSUE #663: `jsx-no-bind` errors on non-react components.
I have Backbone views with a "render" method that are being warned when they have `.bind` used within them. Did something change recently that loosened the criteria for identifying a React component?

If the `render` method doesn't return JSX, and the component doesn't `extend React.Component`, and most importantly if the file itself doesn't even import/require `React`, why are these rules applying to it?

petersendidit:
Fix jsx-no-bind reporting errors on render functions that don't return JSX (Fixes #663)

### DIFF
--- a/lib/rules/jsx-no-bind.js
+++ b/lib/rules/jsx-no-bind.js
@@ -5,13 +5,14 @@
  */
 'use strict';
 
+var Components = require('../util/Components');
 var propName = require('jsx-ast-utils/propName');
 
 // -----------------------------------------------------------------------------
 // Rule Definition
 // -----------------------------------------------------------------------------
 
-module.exports = function(context) {
+module.exports = Components.detect(function(context, components, utils) {
   var configuration = context.options[0] || {};
 
   return {
@@ -30,10 +31,12 @@ module.exports = function(context) {
           (ancestors[i].type === 'MethodDefinition' && ancestors[i].key.name === 'render') ||
           (ancestors[i].type === 'Property' && ancestors[i].key.name === 'render')
         ) {
-          context.report({
-            node: callee,
-            message: 'JSX props should not use .bind()'
-          });
+          if (utils.isReturningJSX(ancestors[i])) {
+            context.report({
+              node: callee,
+              message: 'JSX props should not use .bind()'
+            });
+          }
           break;
         }
       }
@@ -74,7 +77,7 @@ module.exports = function(context) {
       }
     }
   };
-};
+});
 
 module.exports.schema = [{
   type: 'object',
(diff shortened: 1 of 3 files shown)


lib/rules/jsx-no-bind.js
```diff
@@ -5,13 +5,25 @@
...
 
+var Components = require('../util/Components');
 var propName = require('jsx-ast-utils/propName');
 
 // -----------------------------------------------------------------------------
 // Rule Definition
 // -----------------------------------------------------------------------------
 
-module.exports = function(context) {
+function renderReturnsJSX(node, utils) {
```
ljharb: this seems like it might be common behavior that all the rules could use - and that some may already have implemented?

petersendidit: I agree. Sounds like we should move it to the `utils` provided by the `Components` utility?  What should it be named? There is already a `isReturningJSX` that does the bulk of this work but this deals with finding the `ReturnStatement`

ljharb: Is there a reason that `isReturningJSX` isn't already checking for a return statement?

petersendidit: It currently expects a `ReturnStatement` or a `ArrowFunctionExpression` to be passed in.

petersendidit: We could make the default for the case statement to find a return the `ReturnStatement`

ljharb: I think it makes sense for "is returning JSX" to accept any kind of function or `class`, and it should be able to accurately return a boolean.

petersendidit: Updated it to make `isReturningJSX` to handle more nodes.

tests/lib/rules/jsx-no-bind.js
```diff
@@ -121,6 +144,18 @@ ruleTester.run('jsx-no-bind', rule, {
       errors: [{message: 'JSX props should not use .bind()'}],
       parser: 'babel-eslint'
     },
+    {
+      code: [
+        'const foo = {',
+        '  render: function() {',
```
ljharb: can we also cover the `render() {` case?

tests/lib/rules/jsx-no-bind.js
```diff
@@ -72,6 +72,29 @@ ruleTester.run('jsx-no-bind', rule, {
...
+        '    this.onTap.bind(this);',
+        '  }',
+        '});'
+      ].join('\n'),
+      parser: 'babel-eslint'
+    },
+    {
+      code: [
+        'const foo = {',
+        '  render: function() {',
```
ljharb: similarly here with `render() {`

petersendidit: Added

tests/lib/rules/jsx-no-bind.js
petersendidit: Added

yannickcr:
Merged, thanks!

PR #677, opened 2016-07-12
Add support for typeAnnotions in sort-comp

dozoisch:
Hey! I did a basic pass to add support for typeAnnotions in sort-component. Let me know what you think about that. I tested on files from react-native lib & couple of my own files and it everything seemed to be working properly so far.

https://github.com/yannickcr/eslint-plugin-react/issues/235

For anyone wanting to use that right now put this in your package.json:
```
"eslint-plugin-react": "https://github.com/dozoisch/eslint-plugin-react#typeannotations",
```

### DIFF
--- a/docs/rules/sort-comp.md
+++ b/docs/rules/sort-comp.md
@@ -89,6 +89,7 @@ The default configuration is:
 * `lifecycle` is referring to the `lifecycle` group defined in `groups`.
 * `everything-else` is a special group that match all the methods that do not match any of the other groups.
 * `render` is referring to the `render` method.
+* `type-annotations`. This group is not speficied by default, but can be used enforce flow annotations to be at the top.
 
 You can override this configuration to match your needs.
 
@@ -170,6 +171,48 @@ var Hello = React.createClass({
   }
 });
 ```
+
+If you want to flow annotations to be at the top:
+
+```js
+"react/sort-comp": [1, {
+  order: [
+    'type-annotations',
+    'static-methods',
+    'lifecycle',
+    'everything-else',
+    'render',
+  ],
+}]
+```
+
+With the above configuration, the following patterns are considered warnings:
+
+```js
+class Hello extends React.Component<any, Props, void> {
+  onClick() { this._someElem = true; }
+  props: Props;
+  _someElem: bool;
+  render() {
+    return <div>Hello</div>;
+  }
+}
+```
+
+With the above configuration, the following patterns are not considered warnings:
+
+```js
+type Props = {};
+class Hello extends React.Component<any, Props, void> {
+  props: Props;
+  _someElem: bool;
+  onClick() { this._someElem = true; }
+  render() {
+    return <div>Hello</div>;
+  }
+}
+```
+
 
 ## When Not To Use It
 
--- a/lib/rules/sort-comp.js
+++ b/lib/rules/sort-comp.js
@@ -136,6 +136,15 @@ module.exports = {
         }
       }
 
+      if (method.typeAnnotation) {
+        for (i = 0, j = methodsOrder.length; i < j; i++) {
+          if (methodsOrder[i] === 'type-annotations') {
+            indexes.push(i);
+            break;
+          }
+        }
+      }
+
       // Either this is not a static method or static methods are not specified
       // in the methodsOrder.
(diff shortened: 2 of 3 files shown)


ljharb:
Since this is generically named "type annotations", does this work with Typescript and/or JSDoc as well, or only with Flow?

dozoisch:
@ljharb I suspect it could. Im not sure how the ast looks for these. I'm using the typeAnnotation field from the ast which seems generic.

dozoisch:
Any update on this! I can change the name if that is what is needed.

dozoisch:
@yannickcr Anything missing/to change? I'd rebase the branch at the same time.

acusti:
I would love to see this merged! Tried applying @dozoisch’s changes [in my own fork](https://github.com/Brandcast/eslint-plugin-react/tree/typeannotations) based on newer eslint (first tried to rebase, but 10c6f23866b58dff5a7f8d16eb8b8a62cc997a66 made it a bit too hairy for me), then applied it to my .eslintrc, which extends the airbnb config, with this rule:

```
        'react/sort-comp': [2, {
            order: [
                'type-annotations',
                'static-methods',
                'lifecycle',
                'everything-else',
                '/^render.+$/',
                'render'
            ],
        }],
```

and it worked a charm! I also use eslint-plugin-flowtype along with `flowtype/define-flow-type` and `'flowtype/use-flow-type` and no conflicts there.

dozoisch:
@yannickcr Let me know if you want me to do anything. I'll rebase if everything is ok

yannickcr:
@dozoisch Is this introduces a breaking change in the `sort-comp` rule?

dozoisch:
@yannickcr the change it introduces is that it will now check for type annotation fields to be present a the top by default. 

I can create the type category (like it is right now) but not include it in the default configuration, if that's not a wanted behavior.

yannickcr:
@dozoisch I prefer to not include it in the default configuration for now. We'll wait for the next major release to update it.

Also I'm not very experienced in Flow so I trust @acusti and you on this 😉 

If it's good for you, you can rebase your branch, remove `type-annotations` from the default configuration and update the documentation. Then I think we'll be good to go 😃

dozoisch:
Perfect will do that today :)!

dozoisch:
@yannickcr rebased & updated! Let me know what you think!

docs/rules/sort-comp.md
```diff
@@ -171,6 +172,48 @@ var Hello = React.createClass({
...
+```
+
+With the above configuration, the following patterns are considered warnings:
+
+```js
+class Hello extends React.Component<any, Props, void> {
+  onClick: function() { this._someElem = true; }
+  props: Props;
+  _someElem: bool;
+  render: function() {
```
ljharb: not a big deal, but `class` examples should be using concise method syntax - i'm actually not even sure it will work with the `:` syntax you've got here. (ie `onClick() {` not `onClick: function () {`

dozoisch: whoop true that! You're  actually right.

dozoisch:
Ok, fixed the class issue!

docs/rules/sort-comp.md
```diff
@@ -171,6 +172,48 @@ var Hello = React.createClass({
...
+  ],
+}]
+```
+
+With the above configuration, the following patterns are considered warnings:
+
+```js
+class Hello extends React.Component<any, Props, void> {
+  onClick() { this._someElem = true; }
+  props: Props;
```
ljharb: what does `props: Props` desugar to in Flow? Does this whole line just get removed, because it's just a type declaration, or does it have another effect?

dozoisch: It gets completely removed altogether!

http://babeljs.io/repl/#?evaluate=true&lineWrap=false&presets=es2015%2Ces2015-loose%2Creact%2Cstage-0&experimental=true&loose=true&spec=false&code=type%20Props%20%3D%20%7B%20a%3A%20number%20%7D%3B%0Aclass%20B%20%7B%0A%20%20props%3A%20Props%3B%0A%20%20%0A%7D

dozoisch:
@yannickcr :) should be good now!

yannickcr:
Merged, thanks!

PR #679, opened 2016-07-13
Handle spread props in jsx-no-target-blank rule

randycoulman:
When running the `jsx-no-target-blank` rule against code like the following:

```
<a {...this.props} target='_blank' />
```

the following error would be raised:

```
TypeError: Cannot read property 'name' of undefined
```

The root cause is that the node representing the spread props does not have a `name` property.

Note that if the desired `rel` attribute appears before the spread props, the error does not occur.

### DIFF
--- a/lib/rules/jsx-no-target-blank.js
+++ b/lib/rules/jsx-no-target-blank.js
@@ -15,7 +15,7 @@ module.exports = function(context) {
         var relFound = false;
         var attrs = node.parent.attributes;
         for (var idx in attrs) {
-          if (attrs[idx].name.name === 'rel') {
+          if (attrs[idx].name && attrs[idx].name.name === 'rel') {
             var tags = attrs[idx].value.value.split(' ');
             if (tags.indexOf('noopener') >= 0 && tags.indexOf('noreferrer') >= 0) {
               relFound = true;
--- a/tests/lib/rules/jsx-no-target-blank.js
+++ b/tests/lib/rules/jsx-no-target-blank.js
@@ -27,7 +27,8 @@ ruleTester.run('jsx-no-target-blank', rule, {
   valid: [
     {code: '<a href="foobar"></a>', parserOptions: parserOptions},
     {code: '<a randomTag></a>', parserOptions: parserOptions},
-    {code: '<a href="foobar" target="_blank" rel="noopener noreferrer"></a>', parserOptions: parserOptions}
+    {code: '<a href="foobar" target="_blank" rel="noopener noreferrer"></a>', parserOptions: parserOptions},
+    {code: '<a target="_blank" {...spreadProps} rel="noopener noreferrer"></a>', parserOptions: parserOptions}
   ],
   invalid: [
     {code: '<a target="_blank"></a>', parserOptions: parserOptions,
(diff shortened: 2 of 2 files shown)


tests/lib/rules/jsx-no-target-blank.js
```diff
@@ -27,7 +27,8 @@ ruleTester.run('jsx-no-target-blank', rule, {
   valid: [
     {code: '<a href="foobar"></a>', parserOptions: parserOptions},
     {code: '<a randomTag></a>', parserOptions: parserOptions},
-    {code: '<a href="foobar" target="_blank" rel="noopener noreferrer"></a>', parserOptions: parserOptions}
+    {code: '<a href="foobar" target="_blank" rel="noopener noreferrer"></a>', parserOptions: parserOptions},
+    {code: '<a target="_blank" {...spreadProps} rel="noopener noreferrer"></a>', parserOptions: parserOptions}
```
ljharb: since the ordering of "rel" and the spread props seems to matter, should we also have a test for one where "rel" appears first?

lib/rules/jsx-no-target-blank.js
```diff
@@ -15,7 +15,7 @@ module.exports = function(context) {
         var relFound = false;
         var attrs = node.parent.attributes;
         for (var idx in attrs) {
```
ljharb: @yannickcr is there any reason this uses `for..in` instead of `Object.keys(…).forEach(…)`?

tests/lib/rules/jsx-no-target-blank.js
randycoulman: It only matters because we break out of the `for` loop as soon as we find the `rel` attribute we're looking for.  I could add another test case, but it would pass on the old code, so doesn't illustrate the problem.  I didn't think it worth it at the time, but if you think it is, I'm happy to add it.

ljharb: I think it's important to add, to prevent a regression - if there's already order-dependent bugs, there could easily be more added later.

randycoulman: Good thought.  Added in 3db97e49912ce1cccffbbebb595d9bfb14f65512.

lib/rules/jsx-no-target-blank.js
yannickcr: @ljharb If I remember correctly `Object.keys` is only supported since Node.js 4.0

ljharb: @yannickcr `Object.keys` is ES5, and is supported since node 0.6.

yannickcr: @ljharb  Ho ok, my bad. No reason to not using it then.

randycoulman: Do you want this as part of this PR, or should it be a separate one?  Feels unrelated to the bug IMO, but I'm willing to do it if you want.

ljharb: I'm fine with separate, but either way :-)

randycoulman: I looked at this today.  There is a `break` inside the `for` loop which doesn't work in the `forEach` version.  I'm sure there's a way around that, but I think this should be in a separate PR.  I'd like to have this one considered done and merged (if acceptable).

ljharb: Yeah that'd require `.some` or similar. Separate PR is fine.

yannickcr:
Merged, thanks!

PR #681, opened 2016-07-15
Rule for requiring HTML entities to be escaped

pfhayes:
I will also add in some documentation around this, but wanted to get this out there to make sure I was on the right track

I ran this on our codebase and it caught a bunch of errors so it appears to work

### DIFF
--- a/lib/rules/no-unescaped-entities.js
+++ b/lib/rules/no-unescaped-entities.js
@@ -1,0 +1,80 @@
+/**
+ * @fileoverview HTML special characters should be escaped.
+ * @author Patrick Hayes
+ */
+'use strict';
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+// NOTE: '<' and '{' are also problematic characters, but they do not need
+// to be included here because it is a syntax error when these characters are
+// included accidentally.
+var DEFAULTS = ['>', '"', '\'', '}'];
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Detect unescaped HTML entities, which might represent malformed tags',
+      category: 'Possible Errors',
+      recommended: false
+    },
+    schema: [{
+      type: 'object',
+      properties: {
+        forbid: {
+          type: 'array',
+          items: {
+            type: 'string'
+          }
+        }
+      },
+      additionalProperties: false
+    }]
+  },
+
+  create: function(context) {
+    function isInvalidEntity(node) {
+      var configuration = context.options[0] || {};
+      var entities = configuration.forbid || DEFAULTS;
+
+      // HTML entites are already escaped in node.value (as well as node.raw),
+      // so pull the raw text from context.getSourceCode()
+      for (var i = node.loc.start.line; i <= node.loc.end.line; i++) {
+        var rawLine = context.getSourceCode().lines[i - 1];
+        var start = 0;
+        var end = rawLine.length;
+        if (i === node.loc.start.line) {
+          start = node.loc.start.column;
+        }
+        if (i === node.loc.end.line) {
+          end = node.loc.end.column;
+        }
+        rawLine = rawLine.substring(start, end);
+        for (var j = 0; j < entities.length; j++) {
+          for (var index = 0; index < rawLine.length; index++) {
+            var c = rawLine[index];
+            if (c === entities[j]) {
+              context.report({
+                loc: {line: i, column: start + index},
+                message: 'HTML entities must be escaped.',
+                node: node
+              });
+            }
+          }
+        }
+      }
+    }
+
+    return {
+      Literal: function(node) {
+        if (node.type === 'Literal' && node.parent.type === 'JSXElement') {
+          if (isInvalidEntity(node)) {
+            context.report(node, 'HTML entities must be escaped.');
+          }
+        }
+      }
+    };
+  }
+};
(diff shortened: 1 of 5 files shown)


tests/lib/rules/no-unescaped-entities.js
```diff
@@ -0,0 +1,143 @@
...
+        '}'
+      ].join('\n'),
+      args: [1],
+      parser: 'babel-eslint',
+      errors: [{message: 'HTML entities must be escaped.'}]
+    }, {
+      code: [
+        'class Comp1 extends Component {',
+        '  render() {',
+        '    return <div>Multiple errors: \'>></div>;',
```
ljharb: can we also ensure that `{'>'}` is allowed, by adding a test for it?

pfhayes: the test on line 70 should cover that case

ljharb: whoops, missed that. thanks, that works.

tests/lib/rules/no-unescaped-entities.js
```diff
@@ -0,0 +1,143 @@
...
+        '  }',
+        '}'
+      ].join('\n'),
+      args: [1],
+      parser: 'babel-eslint',
```
ljharb: wait, why is babel-eslint required? the default parser should be able to handle all of these examples.

pfhayes: Okay - I can remove this. I copied this from another test and didn't know it could be removed

ljharb: if it works when removed, then i'd say please do :-) if not, then obv you should keep it

pfhayes: Hm, I tried removing it and now all the tests fail with "Parsing error: Unexpected token <"

Is that correct? how is the JSX compiled if babel-eslint is not used?

ljharb: ah - we may need to enable the `jsx` parser option.

pfhayes: Is that something I can do? I tried setting `parser: 'jsx'` but that didn't work

ljharb: See https://github.com/yannickcr/eslint-plugin-react/blob/master/tests/lib/rules/jsx-sort-props.js#L15-L20

pfhayes: great thanks! the latest version should use the correct parser

pfhayes:
Any thoughts on this rule? Happy to make any recommended updates.

We've been running this in development and have already gotten a lot of value out of it. It has caught several errors around mistyped closing tags

cc @yannickcr

lib/rules/no-unescaped-entities.js
```diff
@@ -0,0 +1,68 @@
...
+ * @fileoverview HTML special characters should be escaped.
+ * @author Patrick Hayes
+ */
+'use strict';
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+var DEFAULTS = ['>', '<', '"', '\'', '}'];
```
lencioni: I see the value in linting for `>`, since you have provided such a good example of what that can be problematic. However, I'm not so sure about the other characters listed in the defaults here. Can you provide similar reasoning for each of these characters?

ljharb: I think `<<div>` is just as problematic as `</div>>`, and `{<Foo />}}` and `{{<Foo />}` so those seem obvious to me (let's add `{`?)

I'm not sure I see the purpose in having double or single quotes in here, though.

lencioni: Yeah, I also found the omission of `{` curious. In any case, it would be really great if the documentation included an example of why each of the defaults can be problematic.

pfhayes: I omitted { because it's not actually possible to include this character inside a tag - this character begins a curly-brace section and is a syntax error. I could include it here for completeness but AFAICT would have no effect.

This does raise a good point that < is also extraneous, I should remove that.

Here's an example where quotes can bite you:

```jsx
<Foo
  a="b"
  c="d">
  e="f"
```

Perhaps you get into this state by sorting props alphabetically (this is where it has happened to me). I'm happy to update the documentation to include this example as well

pfhayes: I've added examples for all the cases to the documentation.

ljharb:
@pfhayes please rebase so as to remove merge commits from the PR :-)

pfhayes:
I've squashed this into a single commit

lib/rules/no-unescaped-entities.js
```diff
@@ -0,0 +1,71 @@
...
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+// NOTE: '<' and '{' are also problematic characters, but they do not need
+// to be included here because it is a syntax error when these characters are
+// included accidentally.
+var DEFAULTS = ['>', '"', '\'', '}'];
+
+module.exports = function(context) {
```
petersendidit: This should use the [new eslint rule format](http://eslint.org/blog/2016/07/eslint-new-rule-format)

pfhayes: done

lencioni:
This looks good. Thanks for your contribution!

PR #683, opened 2016-07-15
Add in an "all" config

pfhayes:
Much like eslint:all

http://eslint.org/docs/user-guide/configuring#using-eslintall

### DIFF
--- a/tests/index.js
+++ b/tests/index.js
@@ -38,5 +38,8 @@ describe('configurations', function() {
       assert.equal(configName.indexOf('react/'), 0);
       assert.equal(plugin.configs.all.rules[configName], 2);
     });
+    rules.forEach(function(ruleName) {
+      assert(plugin.configs.all.rules['react/' + ruleName]);
+    });
   });
 });
--- a/tests/index.js
+++ b/tests/index.js
@@ -30,7 +30,7 @@ describe('configurations', function() {
   it('should export a \'all\' configuration', function() {
     assert(plugin.configs.all);
     rules.forEach(function(ruleName) {
-      assert(plugin.configs.all.rules[ruleName]);
+      assert.equal(plugin.configs.all.rules[ruleName], 2);
     });
   });
 });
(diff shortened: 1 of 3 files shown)


ljharb:
These should probably be a superset of the "recommended" config - what about these options https://github.com/yannickcr/eslint-plugin-react/blob/master/index.js#L68-L69 ?

pfhayes:
That's a good point. I would say that "all" should be every rule in it's default configuration, and I would also say it's okay for "recommended" to deviate from default configurations (to ones that might be preferred). But I'm happy to update this to be a superset of recommended if you disagree

pfhayes:
Also, what do you think is the best for `react/jsx-sort-prop-types` ? It's deprecated and replaced by `react/sort-prop-types` so maybe it should be excluded from `all`?

tests/index.js
```diff
@@ -26,5 +26,17 @@ describe('all rule files should be exported by the plugin', function() {
...
     assert(plugin.configs.recommended);
+    Object.keys(plugin.configs.recommended.rules).forEach(function (configName) {
+      assert.equal(configName.indexOf('react/'), 0);
+      var ruleName = configName.substring('react/'.length);
+      assert(plugin.rules[ruleName]);
+    });
+  });
+  it('should export a \'all\' configuration', function() {
+    assert(plugin.configs.all);
+    Object.keys(plugin.configs.all.rules).forEach(function(configName) {
```
ljharb: this seems like it now no longer verifies that every rule is included?

It is important to have a test that fails when someone adds a new rule but forgets to add it to "all"

pfhayes: Yes, you are right, I will fix tha

yannickcr:
Nice. Can you also update the Readme?

pfhayes:
Sure thing, I've updated the README now

yannickcr:
Thanks, can you rebase your branch? Then I think we'll be good to go :)

pfhayes:
Done, should be good to go now

yannickcr:
Merged, thanks!

PR #684, opened 2016-07-15
Repair `--fix` for props-aligned in jsx-closing-bracket-location

pfhayes:
Use the correctColumn attribute that is populated above.

I don't quite understand the error, but running with `--fix` I would still get errors that look like

"  247:37  error  The closing bracket must be aligned with the last prop (expected column 9)   react/jsx-closing-bracket-location"

But with this change, `--fix` does the right thing.

### DIFF
--- a/tests/lib/rules/jsx-closing-bracket-location.js
+++ b/tests/lib/rules/jsx-closing-bracket-location.js
@@ -685,6 +685,70 @@ ruleTester.run('jsx-closing-bracket-location', rule, {
     }]
   }, {
     code: [
+      'const Button = function(props) {',
+      '  return (',
+      '    <Button',
+      '      size={size}',
+      '      onClick={onClick}',
+      '                                    >',
+      '      Button Text',
+      '    </Button>',
+      '  );',
+      '};'
+    ].join('\n'),
+    output: [
+      'const Button = function(props) {',
+      '  return (',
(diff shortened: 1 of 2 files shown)


yannickcr:
Mmm, I don't quite see what is the current problem.

Can you add a test that reproduce it and that is fixed with your patch?

pfhayes:
@yannickcr sure - the test I've added fails on master. This is a modifed example taken from our own codebase that was the impetus for this change.

tests/lib/rules/jsx-closing-bracket-location.js
```diff
@@ -653,6 +653,38 @@ ruleTester.run('jsx-closing-bracket-location', rule, {
...
+      '    <Button',
+      '      size={size}',
+      '      onClick={onClick}',
+      '      >',
+      '      Button Text',
+      '    </Button>',
+      '  );',
+      '};'
+    ].join('\n'),
+    options: ['props-aligned'],
```
ljharb: can we also add a test for the "line-aligned" option? We might want to add explicit tests for all of them.

pfhayes: I added a similar test for line-aligned and tag-aligned

yannickcr:
Merged, thanks!

Issue #599, opened 2016-05-13
sort-comp: Consider separating static fields & instance fields into separate groups


rapilabs:
Currently sort-comp asks me to place instance fields after methods in my class.  It's conventional to place instance fields above the constructor, underneath static fields.

Possibly also consider allowing static fields be grouped separately from static methods.

ljharb:
I'd expect static fields and static methods to both be the last items in the class body, after everything instance-related is done - although I agree that instance fields could belong above the constructor.

leonid-bauxy:
I've decided to take a stab at this and wonder how exactly do you guys see this work?

So far i've come up with an `instance-properties` keyword, which would **not** include arrow functions, only properties. So the default order would be:

```js
[
    'static-methods',
    'instance-properties',
    'lifecycle',
    'everything-else',
    'render'
]
```

Would this be alright with you?

le0nik:
I tried to add this feature with this PR: https://github.com/yannickcr/eslint-plugin-react/pull/685.

Will gladly take suggestions and feedback on it.

yannickcr:
Also, to continue in this direction, it could be good to have separate groups for `static-method`, `static-properties`, `instance-methods` and `instance-properties`.

le0nik:
@yannickcr I like it, but it raises a couple of questions:

1. What would be considered `instance-methods` and `static-methods`? Would they be just arrow function properties or also regular functions?

Statics fields:

```js
class Hello extends React.Component {
  static displayName = 'Hello'; // static-properties
  static fetch() {} // static-methods
  static process = () => {} // static-methods ?
}
```

Instance fields:

```js
class Hello extends React.Component {
  count = 0; // instance-properties
  handleClick = () => {} //  instance-methods
  renderSeciton() {} // instance-methods ?
}
```

I think it should be consistent for both `static-methods` and `instance-methods` `methods`, but then `renderSection` in the example above would be semantically incorrect, because it wouldn't be an **instance** method.

2. Is it possible to combine methods and properties in configuration to allow users to mix them up by default?

Right now `static-methods` allows this:

```js
class Hello extends React.Component {
  static prop1 = 'Hello';
  static fetch() {}
  static prop2 = 'Bye';
}
```

but configuration like this wouldn't allow this and users wouldn't be able to go back to the way it was before without applying codemod:

```js
[
  'static-properties',
  'static-methods'
]
```

Thoughts?

ljharb:
[...]

PR #685, opened 2016-07-17
Add instance-properties group to sort-comp. Issue #599.

le0nik:
Relevant issue: https://github.com/yannickcr/eslint-plugin-react/issues/599

It's a common convention to have instance properties on top(right below the static properties/methods), like this:

``` js
  static displayName = 'Hello';

  count = 0;

  componentDidMount() {
    // ...
  }

  handleClick = () => {
    this.count += 1;
    // ...
  }

  render() {
    // ...
  }
}
```

Right now `sort-comp` rule doesn't allow that and wants `count` to be below `componentDidMount`, unless you specify it by name in the configuration.

This PR's aim is to allow placing instance properties separately.

The default configuration is not changed.

### DIFF
--- a/docs/rules/sort-comp.md
+++ b/docs/rules/sort-comp.md
@@ -9,9 +9,10 @@ When creating React components it is more convenient to always follow the same o
 With default configuration the following organisation must be followed:
 
   1. static methods and properties
-  2. lifecycle methods: `displayName`, `propTypes`, `contextTypes`, `childContextTypes`, `mixins`, `statics`,`defaultProps`, `constructor`, `getDefaultProps`, `getInitialState`, `state`, `getChildContext`, `componentWillMount`, `componentDidMount`, `componentWillReceiveProps`, `shouldComponentUpdate`, `componentWillUpdate`, `componentDidUpdate`, `componentWillUnmount` (in this order).
-  3. custom methods
-  4. `render` method
+  2. instance properties
+  3. lifecycle methods: `displayName`, `propTypes`, `contextTypes`, `childContextTypes`, `mixins`, `statics`,`defaultProps`, `constructor`, `getDefaultProps`, `getInitialState`, `state`, `getChildContext`, `componentWillMount`, `componentDidMount`, `componentWillReceiveProps`, `shouldComponentUpdate`, `componentWillUpdate`, `componentDidUpdate`, `componentWillUnmount` (in this order).
+  4. custom methods
+  5. `render` method
 
 The following patterns are considered warnings:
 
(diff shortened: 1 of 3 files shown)


ljharb:
Allowing it to be configurable is excellent. However, changing the default order makes it a breaking change. Should this preserve the default order, and be semver-minor, and change the default later?

le0nik:
@ljharb I've reverted the default order. My initial way of thinking was that this change could be part of `6.0.0`, considering that it's in `alpha` now.

ljharb:
That's a fair point. @yannickcr should weigh in on that.

yannickcr:
Since `6.0.0` is imminent we could do this, but I don't know if it is a good thing to force instance properties on top in the default configuration.

What is the most common order used by React users here?

ljharb:
Instance properties are still stage 1, so we don't use them yet at Airbnb. However, since instance properties run almost as if they were in the constructor, i think they absolutely should all be placed just like the constructor - at the top.

kylpo:
This has become a common pattern for the react community, and one that I would certainly like to enforce in my projects. Mind taking another look at this and weighing in on why it is not accepted? (other than the conflicts to be resolved now)

docs/rules/sort-comp.md
```diff
@@ -9,9 +9,10 @@ When creating React components it is more convenient to always follow the same o
 With default configuration the following organisation must be followed:
 
   1. static methods and properties
-  2. lifecycle methods: `displayName`, `propTypes`, `contextTypes`, `childContextTypes`, `mixins`, `statics`,`defaultProps`, `constructor`, `getDefaultProps`, `getInitialState`, `state`, `getChildContext`, `componentWillMount`, `componentDidMount`, `componentWillReceiveProps`, `shouldComponentUpdate`, `componentWillUpdate`, `componentDidUpdate`, `componentWillUnmount` (in this order).
-  3. custom methods
-  4. `render` method
+  2. instance properties
```
lencioni: It looks like this documentation needs to be updated, since the default order has not been modified yet.

It would be nice to open a PR that updates the default and documentation that we could bring in as part of v7.0.0.

klimashkin:
Public Class Fields are on stage 2 already, why it still hasn't been merged into `7.x`?...

ljharb:
@klimashkin this needs reviews from more collabs first.

Separately, stage 3 is the time when it's safe to start using a feature, so that's when it will become critical to be supported here.

noahbenham:
Class fields is now a [stage 3 proposal](https://github.com/tc39/proposals). Happy to contribute if this needs someone else to take it on.

ljharb:
Unfortunately @le0nik deleted their fork, so this PR can never land.

PR #695, opened 2016-07-20
Minor text change in docs.

mik01aj:


docs/rules/jsx-max-props-per-line.md
```diff
@@ -58,4 +58,4 @@ The following patterns are not considered warnings:
 
 ## When not to use
 
-If you are not using JSX then you can disable this rule.
+If you are not using JSX or you already use the [`max-len` rule](http://eslint.org/docs/rules/max-len) then you can disable this rule.
```
ljharb: I can see how the max-len rule might overlap sometimes with this rule, but all the places I want to use this one wouldn't hit max-len.

mik01aj: That's why I didn't write "you MUST disable this rule". I see it like a suggestion.

ljharb: Generally this section of rules tells you when it's useless to use the rule. Using `max-len` doesn't make it useless.

mik01aj: Okay. Perhaps you can just propose some other wording to mention the possible rule overlap?

ljharb: I'm not sure it needs mentioning. Many rules happen to overlap in some cases, but aren't conceptually exclusive rules - this case applies.

I guess "if you are not using jsx, or if you are primarily concerned with line length and the max-len rule is sufficient for you" but I'm not sure what value it adds.

lencioni:
Thanks for your contribution! I agree with @ljharb so I am closing this out.

I actually think the text that is currently there has little value as well.

Thanks again for helping to make this plugin better. 💃

Issue #686, opened 2016-07-17
Roadmap v6.0.0


yannickcr:
So here's what I have in mind for now:

* [x] Add ESLint as a peerDependency (PR #657)
* [x] Add `all` configuration (#674, PR #683)
* [x] Remove deprecated `jsx-sort-prop-types` rule (#549)
* [x] Remove rules deprecated options (like https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-deprecated.md) (PR #700)
* [x] Rename some JSX-specific rules to include `jsx-` prefix (keep old rules with a warning like `jsx-sort-prop-types`) (#668, PR #689)
* [x] Add `findDOMNode()` warning rule (#678)
* [x] Enable `allow-in-func` mode by default in `no-did-mount-set-state` and `no-did-update-set-state` rules (PR #702)
* [x] Solve `recommended` configuration discussions and update it accordingly (PR #636, #596)
* [x] Support ESLint 3 and new rule format (#661, PR #699)

Do you see anything else that could be good to change/improve?

ljharb:
That seems good to me.

ljharb:
5 down, 4 to go!

lencioni:
On this roadmap, you have

> Enable allow-in-func mode by default in no-did-mount-set-state and no-did-update-set-state rules

I'm not entirely sure what you want done here? Does this amount to removing the option and just having it always be the case, or should this be the case by default and there would be a new option like `disallow-in-func`?

yannickcr:
> should this be the case by default and there would be a new option like `disallow-in-func`

I was more thinking about something like this.

ljharb:
It looks like the last 3 issues all have PRs up - how quickly do you think we could get to a v6?

yannickcr:
@ljharb I hope to publish an RC before friday and the final this week-end if there is no major issue.

ljharb:
@yannickcr there's still one unchecked item on the list - is that punted to v7?

petersendidit:
I just rebased #699 so it should be able to be merged cleanly.

yannickcr:
@ljharb no, still planned for v6 :)

ljharb:
Should this be reopened until that's merged?

yannickcr:
Yep

ljharb:
@yannickcr any more things to add to the above checklist, or are we good for a final release? :-)

yannickcr:
[...]

PR #700, opened 2016-07-23
Remove rules deprecated options

lencioni:
These options were deprecated in a previous release. Now that we are preparing for a major version bump, we want to clear them out. 

Addresses #686.

### DIFF
--- a/lib/rules/jsx-uses-react.js
+++ b/lib/rules/jsx-uses-react.js
@@ -33,12 +33,4 @@ module.exports = function(context) {
 
 };
 
-module.exports.schema = [{
-  type: 'object',
-  properties: {
-    pragma: {
-      type: 'string'
-    }
-  },
-  additionalProperties: false
-}];
+module.exports.schema = [];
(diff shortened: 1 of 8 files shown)


lib/rules/jsx-uses-react.js
```diff
@@ -33,12 +33,4 @@ module.exports = function(context) {
...
-module.exports.schema = [{
-  type: 'object',
-  properties: {
-    pragma: {
-      type: 'string'
-    }
-  },
-  additionalProperties: false
-}];
+module.exports.schema = [];
```
lencioni: I just tried this out with a different rule that has a schema of `[]` and it produced an error like:

```
Configuration for rule "plugin/rule" is invalid:
Value "config value" has more items than allowed.
```

ljharb: Awesome, thanks for confirming :-)

yannickcr:
Merged, thanks!

PR #701, opened 2016-07-23
Reference right prop for ListItem

mfn:
It seems to me that in the `ListItem` component, the referred prop should be `key` as the example above is using `<ListItem key...`.

### DIFF
--- a/docs/rules/jsx-no-bind.md
+++ b/docs/rules/jsx-no-bind.md
@@ -99,7 +99,7 @@ var ListItem = React.createClass({
     );
   },
   _onClick() {
-    this.props.onItemClick(this.props.item.id);
+    this.props.onItemClick(this.props.key);
   }
 });
 ```


lencioni:
It's an interesting thought, but I believe that `key` is a special "prop" that doesn't actually get passed down to the element it is added to. 

Here's a quick example I whipped up to demonstrate: https://jsfiddle.net/e1nmfme9/1/

tleunen:
> Warning: Goodbye: `key` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop. (https://fb.me/react-special-props)

mfn:
Ops. N00b alert here; please disregard the PR and sorry for the noise `¯\_(ツ)_/¯`

lencioni:
No worries! Thanks for contributing! 💃

PR #702, opened 2016-07-24
Enable allow-in-func mode by default in no-did-mount-set-state and no-did-update-set-state rules

lencioni:
There are many people who regularly call `this.setState` in a callback
in `componentDidMount` and `componentDidUpdate`. Although there
was an option for this rule that would allow this to happen, we have
decided to make it the default.

Addresses #686.


yannickcr:
Merged, thanks!

PR #703, opened 2016-07-24
Add rule to forbid className and style being passed to Components

### LINKED ISSUE #314: Rule Proposal: Disallow className prop on components
(Or perhaps make it customisable for any particular prop.)

I've seen it passing `className` to components which are not DOM elements as a common anti-pattern. Perhaps we can disallow declaring them in `propTypes`? Wrote a blog post about it recently, and it would be nice to capture this in a rule (https://medium.com/brigade-engineering/don-t-pass-css-classes-between-components-e9f7ab192785).

lencioni:
Passing `className` or `style` to your Components is a source of much
hidden complexity that can be solved either by using a wrapper to add
styling or by adding props for specific styling. This rule aims at
preventing these props from being passed to your Components, to help
minimize hidden complexity.

The original issue proposed having this rule check propTypes. I
considered that route, but I worry that it won't be as effective for
people who do not have every prop used by the component listed in the
propTypes. Additionally, propTypes seem to have waning interest from the
React team in favor of Flow annotations, so I think in the long run
checking for this at the callsite will be more stable.

More information can be found by reading:

  https://medium.com/brigade-engineering/don-t-pass-css-classes-between-components-e9f7ab192785

Also:

  https://twitter.com/sebmarkbage/status/598971268403073024

Fixes #314

cc: @janpaul123 @sebmarkbage

### DIFF
--- a/docs/rules/forbid-component-props.md
+++ b/docs/rules/forbid-component-props.md
@@ -1,0 +1,44 @@
+# Forbid certain props on Components (forbid-component-props)
+
+By default this rule prevents passing of [props that add lots of complexity](https://medium.com/brigade-engineering/don-t-pass-css-classes-between-components-e9f7ab192785) (`className`, `style`) to Components. This rule only applies to Components (e.g. `<Foo />`) and not DOM nodes (e.g. `<div />`). The list of forbidden props can be customized with the `forbid` option.
+
+## Rule Details
+
+This rule checks all JSX elements and verifies that no forbidden props are used
+on Components. This rule is off by default.
+
+The following patterns are considered warnings:
+
+```jsx
+<Hello className='foo' />
+```
+
+```jsx
+<Hello style={{color: 'red'}} />
+```
+
+The following patterns are not considered warnings:
+
+```jsx
+<Hello name='Joe' />
+```
+
+```jsx
+<div className='foo' />
+```
+
+```jsx
+<div style={{color: 'red'}} />
+```
+
+## Rule Options
+
+```js
+...
+"forbid-component-props": [<enabled>, { "forbid": [<string>] }]
+...
+```
+
+### `forbid`
+
+An array of strings, with the names of props that are forbidden.
--- a/docs/rules/forbid-component-props.md
+++ b/docs/rules/forbid-component-props.md
@@ -41,4 +41,4 @@ The following patterns are not considered warnings:
 
 ### `forbid`
 
-An array of strings, with the names of props that are forbidden.
+An array of strings, with the names of props that are forbidden. The default value of this option is `['className', 'style']`.
--- a/lib/rules/forbid-component-props.js
+++ b/lib/rules/forbid-component-props.js
@@ -14,46 +14,52 @@ var DEFAULTS = ['className', 'style'];
 // Rule Definition
 // ------------------------------------------------------------------------------
 
-module.exports = function(context) {
+module.exports = {
+  meta: {
+    docs: {},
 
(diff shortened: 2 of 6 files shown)


docs/rules/forbid-component-props.md
```diff
@@ -0,0 +1,44 @@
...
+
+```jsx
+<div style={{color: 'red'}} />
+```
+
+## Rule Options
+
+```js
+...
+"forbid-component-props": [<enabled>, { "forbid": [<string>] }]
```
ljharb: would this not need a "react/" prefix?

lencioni: Yes, I think so. I was copying how it is in other rules documentation though so we should probably fix them all at the same time.

ljharb: Sounds good.

lib/rules/forbid-component-props.js
```diff
@@ -0,0 +1,59 @@
...
+// Constants
+// ------------------------------------------------------------------------------
+
+var DEFAULTS = ['className', 'style'];
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = function(context) {
```
petersendidit: This should use the [new eslint rule format](http://eslint.org/blog/2016/07/eslint-new-rule-format)

lencioni: Thanks for the heads up! This PR has been updated.

docs/rules/forbid-component-props.md
```diff
@@ -0,0 +1,44 @@
...
+...
+```
+
+### `forbid`
+
+An array of strings, with the names of props that are forbidden.
```
ljharb: If i want to forbid certain props on Components, but *not* forbid className/style, how do I do that?

lencioni: You would set the `forbid` option to include the props that you want to forbid and not include className/style. This is identical to forbid-prop-types.

https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/forbid-prop-types.md

ljharb: It might be good to explicitly note that the default value of this option is `['className', 'style']`

lencioni: Sure, done!

lencioni:
@yannickcr @ljharb anything else you'd like to see on this before we merge it in?

lencioni:
I plan to merge this soon unless I hear otherwise.

cristian-sima:
This rule is nice, but what if we are required by default to pass the `className` as a `prop`. Let's take for instance, the `react-modal` case:

```jsx
      <Modal
        className="Modal__Bootstrap modal-dialog"
        onRequestClose={this.handleCloseModal}
        isOpen>
```

ljharb:
Per the info in the OP, you'd simply not use the rule in that case - the idea is that passing a `className` is an antipattern, and you'd want to avoid it.

cristian-sima:
Thanks for your reply.

In order to not use it, I have to disable the rule, which in turn may be contra-productive, as I can have in the same function or file other components which should indeed not receive the `className` or `style` props.

I propose a different approach. What about introducing an `option` to tell the rule which components should be `ignored`.

As in my example, `react-modal` has always the `Modal` component. I can specify `eslint` to ignore any component with the name `Modal`.

Well, yes this approach has a down-side as I may write in the future components with that name, but I think is less likely.

PR #704, opened 2016-07-24
Teach jsx-uses-vars about components as object properties

### LINKED ISSUE #694: eslint doesn't catch some unused imports
When I'm importing a component (say `Input`) which has other subcomponents as it's properties (like `Input.Placeholder`, `Input.Button`, etc), and there is another different component imported by a same name (say `Button`), and then if I use `<Input.Button>` in my code then eslint doesn't tell that `<Button>` is unused.

Reproducible code:

```jsx
import React from 'react';
import Button from './Button';
import Input from './Input';

export default class X extends React.Component {
  render() {
    return <Input.Button />;
  }
}
```

What should be expected?

The `import Button from './Button';` line should be highlighted with a `no-unused-vars` error. You can see this if you just use `<Input>` in the render function instead of `<Input.Button>`.

lencioni:
When I'm importing a component (say Input) which has other subcomponents
as it's properties (like Input.Placeholder, Input.Button, etc), and
there is another different component imported by a same name (say
Button), and then if I use `<Input.Button>` in my code then eslint doesn't
tell that `<Button>` is unused.

Reproducible code:

```jsx
import React from 'react';
import Button from './Button';
import Input from './Input';

export default class X extends React.Component {
  render() {
    return <Input.Button />;
  }
}
```

What should be expected?

The import Button from './Button'; line should be highlighted with a
no-unused-vars error. You can see this if you just use `<Input>` in the
render function instead of `<Input.Button>`.

This was happening because the rule was checking for all JSXIdentifiers,
which included the object properties. I fixed this by having it check
for JSXOpeningElements and then drilling down to get the variable name
used from there.

Fixes #694

### DIFF
--- a/tests/lib/rules/jsx-uses-vars.js
+++ b/tests/lib/rules/jsx-uses-vars.js
@@ -115,6 +115,22 @@ ruleTester.run('no-unused-vars', rule, {
     }, {
       code: '\
         /*eslint jsx-uses-vars:1*/\
+        var App;\
+        var Hello;\
+        React.render(<App:Hello/>);',
+      errors: [{message: '\'Hello\' is defined but never used'}],
+      parserOptions: parserOptions
+    }, {
+      code: '\
+        /*eslint jsx-uses-vars:1*/\
+        var Button;\
+        var Input;\
+        React.render(<Button.Input unused=""/>);',
+      errors: [{message: '\'Input\' is defined but never used'}],
+      parserOptions: parserOptions
+    }, {
+      code: '\
+        /*eslint jsx-uses-vars:1*/\
         class unused {}',
       errors: [{message: '\'unused\' is defined but never used'}],
       parserOptions: parserOptions
--- a/lib/rules/jsx-uses-vars.js
+++ b/lib/rules/jsx-uses-vars.js
@@ -20,11 +20,22 @@ module.exports = function(context) {
       variableUtil.markVariableAsUsed(context, node.expression.name);
     },
 
-    JSXIdentifier: function(node) {
-      if (node.parent.type === 'JSXAttribute') {
+    JSXOpeningElement: function(node) {
+      var name;
+      if (node.name.namespace && node.name.namespace.name) {
+        // <Foo:Bar>
+        name = node.name.namespace.name;
+      } else if (node.name.name) {
+        // <Foo>
+        name = node.name.name;
+      } else if (node.name.object && node.name.object.name) {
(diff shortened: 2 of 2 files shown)


tests/lib/rules/jsx-uses-vars.js
```diff
@@ -115,6 +115,22 @@ ruleTester.run('no-unused-vars', rule, {
     }, {
       code: '\
         /*eslint jsx-uses-vars:1*/\
+        var App;\
+        var Hello;\
+        React.render(<App:Hello/>);',
```
lencioni: I'm not at all familiar with this namespace syntax, so I'm not really sure if this is correctly marked as an error or not. @ljharb can you verify?

ljharb: I had no idea this syntax was a thing (test added in e5e29cec / #219) - the [official docs](https://facebook.github.io/react/docs/jsx-in-depth.html#namespaced-components) don't mention it either.

@zertosh, could you verify this is correct? Also, can you link to where in the JSX spec this syntax is supported? If it's solely a custom addition then I don't think it's something we should try hard to support.

zertosh: It's in [the spec](https://facebook.github.io/jsx/). It's a `JSXNamespacedName`.

ljharb: Thanks! What would it transform to that's different than a dot?

zertosh: Stock babel doesn't transform them. You have to write your own transform to do anything with them.

ljharb: Is that a bug filed on babel, or an intentional decision?

ljharb: what i meant was, conceptually what is the purpose of that production in the JSX spec, if the recommended transpiler doesn't support it?

zertosh: The transpiler supports them insofar as it parses the AST, and supports custom transforms on that node type. I wouldn't call the lack of an out-of-box transform for namespaced JSX a bug. What would you transform them to anyway? The value I see in them is in the ability to write custom transforms.

ljharb: I guess that was my question, what would you transform them to anyways. If it's just meant as an otherwise useless injection point, then that makes sense.

yannickcr:
Merged, thanks!

PR #705, opened 2016-07-24
Prevent jsx-handler-names from incorrectly flagging `only`

### LINKED ISSUE #571: react/jsx-handler-names should not trigger on "only"
The rule was triggered for a property `only`. I assume it checks for any prop key starting with `on`, but it would maybe make sense changing this to match `on[A-Z]` (i.e. camelcased keys) instead?

lencioni:
The rule was triggered for a property `only`. This is because the regex
was checking for the prefix (by default "on") followed by any
characters. We can fix this by ensuring that the prefix we are checking
for is always followed by an uppercase character.

It was suggested in the issue that we might want to allow this rule to
accept a regex, which would allow for more flexibility in how this rule
is configured. I think this might be a nice enhancement, but since this
fix was so simple I decided to go with the quick and easy win at this
time.

Fixes #571


ljharb:
LGTM

lencioni:
Rebased

PR #706, opened 2016-07-25
Add jsx-closing-bracket-atomic rule (Fixes #693)

### LINKED ISSUE #693: Rule proposal: jsx-tag-spacing
For some reason, `/ >` is valid JSX, as is `/` followed by any kind of whitespace until the `>`. This rule would require that the self-closing terminator is always written as an atomic `/>`.

Not having a check for this also allows for funny workarounds for [jsx-closing-bracket-location](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-closing-bracket-location.md), such as this (when using `after-props`):

``` js
<SomeComponent
  with='many'
  props='specified' /
>
```

Update: https://github.com/yannickcr/eslint-plugin-react/issues/693#issuecomment-234842584

Kovensky:
Minimal working implementation. This is probably missing things :)

I also don't really like the error message, but I couldn't think of a better one.

### DIFF
--- a/lib/rules/jsx-closing-bracket-atomic.js
+++ b/lib/rules/jsx-closing-bracket-atomic.js
@@ -1,0 +1,39 @@
+/**
+ * @fileoverview Ensures the "/>" and "</" closing indicators are written as a single token
+ * @author Diogo Franco (Kovensky)
+ */
+'use strict';
+
+// ------------------------------------------------------------------------------
+// Constants
+// ------------------------------------------------------------------------------
+
+var MESSAGE = 'JSX tag closing has embedded whitespace';
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = function(context) {
+  return {
+    JSXOpeningElement: function(node) {
+      if (node.selfClosing) {
+        var sourceCode = context.getSourceCode();
+        var sourceText = sourceCode.getText(node);
+        if (sourceText.lastIndexOf('/>') !== sourceText.length - 2) {
+          context.report(node, MESSAGE);
+        }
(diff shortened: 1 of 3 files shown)


lib/rules/jsx-closing-bracket-atomic.js
```diff
@@ -0,0 +1,39 @@
...
+ * @fileoverview Ensures the "/>" and "</" closing indicators are written as a single token
+ * @author Diogo Franco (Kovensky)
+ */
+'use strict';
+
+// ------------------------------------------------------------------------------
+// Constants
+// ------------------------------------------------------------------------------
+
+var MESSAGE = 'JSX tag closing has embedded whitespace';
```
ljharb: How about
```
Self-closing JSX elements must not have whitespace between the `/` and the `>`
```
?

Kovensky: I don't check the position of the tag name here, only for the presence of `/>` and `</`.

Thinking about it, checking for the tag name sounds like a job for a possible "jsx-space-after-opening" rule...

ljharb: In this case it'd be `jsx-no-space-in-closing` tho.

Kovensky: Hm... conceptually, at least in my head, a rule that checks for `</ foo>` should also check for `< foo>` and `< foo/>`.

ljharb: yeah, that's true. perhaps making this rule handle all of those would be helpful?

Kovensky:
Closed in favour of #707.

PR #719, opened 2016-07-26
Add new dom-elements-no-danger-with-children rule (Fixes #710)

### LINKED ISSUE #710: Rule proposal: prevent both children and dangerouslySetInnerHTML at the same time
React will issue a warning if you try to use dangerouslySetInnerHTML and children at the same time.

https://github.com/facebook/react/blob/85dcbf83/src/renderers/dom/shared/ReactDOMComponent.js#L170

This should be easy enough to enforce for most cases with a linter rule.

Bad:

```jsx
<div dangerouslySetInnerHTML={{ __html: 'HTML' }}>
  Children
</div>
```

Good:

```jsx
<div dangerouslySetInnerHTML={{ __html: 'HTML' }} />
```

```jsx
<div>
  Children
</div>
```

petersendidit:
Prevents dangerouslySetInnerHTML and children from being used at the same time (Fixes #710)

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@ Finally, enable all of the rules that you would like to use.  Use [our preset](#
 * [react/forbid-component-props](docs/rules/forbid-component-props.md): Forbid certain props on Components
 * [react/forbid-prop-types](docs/rules/forbid-prop-types.md): Forbid certain propTypes
 * [react/no-danger](docs/rules/no-danger.md): Prevent usage of dangerous JSX properties
+* [react/no-danger-with-children](docs/rules/no-danger-with-children.md): Prevent problem with children and props.dangerouslySetInnerHTML
 * [react/no-deprecated](docs/rules/no-deprecated.md): Prevent usage of deprecated methods
 * [react/no-did-mount-set-state](docs/rules/no-did-mount-set-state.md): Prevent usage of `setState` in `componentDidMount`
 * [react/no-did-update-set-state](docs/rules/no-did-update-set-state.md): Prevent usage of `setState` in `componentDidUpdate`
--- a/docs/rules/no-danger-with-children.md
+++ b/docs/rules/no-danger-with-children.md
@@ -1,0 +1,53 @@
+# Prevent problem with children and props.dangerouslySetInnerHTML (no-danger-with-children)
+
+This rule helps prevent problems caused by using children and the dangerouslySetInnerHTML prop at the same time.
+React will throw a warning if this rule is ignored.
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```jsx
+<div dangerouslySetInnerHTML={{ __html: "HTML" }}>
+  Children
+</div>
+
+<Hello dangerouslySetInnerHTML={{ __html: "HTML" }}>
+  Children
+</Hello>
+
+```
+
+```js
+React.createElement("div", { dangerouslySetInnerHTML: { __html: "HTML" } }, "Children");
+
+React.createElement("Hello", { dangerouslySetInnerHTML: { __html: "HTML" } }, "Children");
+```
+
+The following patterns are not considered warnings:
+
+```jsx
+<div dangerouslySetInnerHTML={{ __html: "HTML" }} />
+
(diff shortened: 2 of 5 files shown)


tests/lib/rules/dom-elements-no-danger-with-children.js
```diff
@@ -0,0 +1,67 @@
...
+    },
+    {
+      code: 'React.createElement("div", {}, "Children");',
+      parserOptions: parserOptions
+    }
+  ],
+  invalid: [
+    {
+      code: [
+        '<div dangerouslySetInnerHTML={{ __html: "HTML" }}>',
```
ljharb: Can we also add a test that sets the "children" prop explicitly, in both cases?

petersendidit: Good idea. Added.

ljharb: I see you've added it for `React.createElement`, but not for the jsx example.

petersendidit: Didn't even think about passing a children prop in JSX. Seems just wrong to even do such thing. Will update the rule.

ljharb: It's very wrong but we should still cover it :-)

petersendidit: Added

lencioni: Sounds like another good rule would be to prevent a `children` prop on DOM elements.

ljharb: I'd say on any elements - children should be actual children (in `React.createElement` as well)

lencioni: Ah, right. Good point!

lencioni: #720

docs/rules/dom-elements-no-danger-with-children.md
```diff
@@ -0,0 +1,39 @@
+# Prevent problem with children and props.dangerouslySetInnerHTML (dom-elements-no-danger-with-children)
+
+This rule helps prevent problems caused by using children and the dangerouslySetInnerHTML prop at the same time
```
lencioni: This sentence needs ending punctuation.

Also, it might be worth mentioning that React will issue a warning for this case.

petersendidit:
New commit should take care of the comments. Let me know if there is anything else.

yannickcr:
Thanks, just a few small things:

* Can you rename the rule to `no-danger-with-children`
* Pass the `recommended` to `false` in the metadata
* Rebase your branch

And then I think we'll be good to go :)

lib/rules/dom-elements-no-danger-with-children.js
```diff
@@ -0,0 +1,80 @@
...
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Report when a DOM element is using both children and dangerouslySetInnerHTML',
+      category: '',
+      recommended: true
```
ljharb: It should be, though, since it's already a React error.

petersendidit:
Renamed and removed it from the recommended rules.

I do think these type of rules that are catching errors that React will throw itself should be recommend ASAP.

lencioni:
Agreed. I think we should write them all, put out minor releases, and then
add them all to recommended config in a major release ASAP. That way we can
maximize benefit and minimize pain.

ljharb:
Although, I kind of feel like it's not really a semver-major if React would be throwing the error at runtime anyways.

lencioni:
I think it is a warning

yannickcr:
> I do think these type of rules that are catching errors that React will throw itself should be recommend ASAP.

Yes, it will be added to the recommended config on next major release.

yannickcr:
Merges, thanks!

PR #722, opened 2016-07-27
Add new no-children-prop rule (Fixes #720)

### LINKED ISSUE #720: Rule proposal: disallow children prop
Children should always be actual children, not passed in as a prop.

Bad:

```jsx
<div children='Children' />
```

Good:

```jsx
<div>
  Children
</div>
```

This should also apply to `React.createElement`.

Name: `no-children-prop`?

benstepp:
Prevents children being passed as props. Children should be actual children.

I've never written an eslint rule, so any suggestions are definitely appreciated.

### DIFF
--- a/docs/rules/no-children-prop.md
+++ b/docs/rules/no-children-prop.md
@@ -1,0 +1,36 @@
+# Prevent passing of children as props (no-children-prop)
+
+Children should always be actual children, not passed in as a prop.
+
+When using JSX, the children should be nested between the opening and closing
+tags. When not using JSX, the children should be passed as additional
+arguments to `React.createElement`.
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```jsx
+<div children='Children' />
+
+<MyComponent children={<AnotherComponent />} />
+<MyComponent children={['Child 1', 'Child 2']} />
+
+React.createElement("div", { children: 'Children' })
+```
+
+The following patterns are not considered warnings:
+
+```jsx
+<div>Children</div>
+
+<MyComponent>Children</MyComponent>
+
+<MyComponent>
+  <span>Child 1</span>
+  <span>Child 2</span>
+</MyComponent>
+
+React.createElement("div", {}, 'Children')
+React.createElement("div", 'Child 1', 'Child 2')
+```
--- a/index.js
+++ b/index.js
@@ -57,7 +57,8 @@ var rules = {
   'no-find-dom-node': require('./lib/rules/no-find-dom-node'),
   'no-danger-with-children': require('./lib/rules/no-danger-with-children'),
   'style-prop-object': require('./lib/rules/style-prop-object'),
-  'no-unused-prop-types': require('./lib/rules/no-unused-prop-types')
+  'no-unused-prop-types': require('./lib/rules/no-unused-prop-types'),
+  'no-children-prop': require('./lib/rules/no-children-prop')
 };
 
 var ruleNames = Object.keys(rules);
--- a/lib/rules/no-children-prop.js
+++ b/lib/rules/no-children-prop.js
@@ -1,0 +1,69 @@
+/**
+ * @fileoverview Prevent passing of children as props
+ * @author Benjamin Stepp
+ */
+'use strict';
+
+// ------------------------------------------------------------------------------
+// Helpers
+// ------------------------------------------------------------------------------
+
+/**
+ * Checks if the node is a createElement call with a props literal.
+ * @param {ASTNode} node - The AST node being checked.
+ * @returns {Boolean} - True if node is a createElement call with a props
+ * object literal, False if not.
+*/
+function isCreateElementWithProps(node) {
+  return node.callee
+    && node.callee.type === 'MemberExpression'
+    && node.callee.property.name === 'createElement'
+    && node.arguments.length > 1
+    && node.arguments[1].type === 'ObjectExpression';
+}
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Prevent passing of children as props.',
+      category: 'Best Practices',
+      recommended: false
+    },
+    schema: []
+  },
+  create: function(context) {
+    return {
+      JSXAttribute: function(node) {
+        if (node.name.name !== 'children') {
+          return;
+        }
+
+        context.report({
+          node: node,
+          message: 'Do not pass children as props. Instead, nest children between the opening and closing tags.'
+        });
+      },
+      CallExpression: function(node) {
+        if (!isCreateElementWithProps(node)) {
+          return;
+        }
+
+        var props = node.arguments[1].properties;
+        var childrenProp = props.find(function(prop) {
+          return prop.key.name === 'children';
+        });
+
+        if (childrenProp) {
+          context.report({
+            node: node,
+            message: 'Do not pass children as props. Instead, pass them as additional arguments to React.createElement.'
+          });
+        }
+      }
+    };
+  }
+};
--- a/tests/lib/rules/no-children-prop.js
+++ b/tests/lib/rules/no-children-prop.js
@@ -1,0 +1,237 @@
+/**
+ * @fileoverview Tests for no-children-prop
+ * @author Benjamin Stepp
+ */
+
+'use strict';
+
+// -----------------------------------------------------------------------------
+// Requirements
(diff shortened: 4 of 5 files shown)


ljharb:
LGTM

docs/rules/no-children-prop.md
```diff
@@ -0,0 +1,21 @@
+# Prevent passing of children as props (no-children-prop)
+
+Children should always be actual children, not passed in as a prop.
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```js
```
lencioni: These should use jsx, not js, for better syntax highlighting.

index.js
```diff
@@ -50,7 +50,8 @@ var rules = {
   'jsx-no-target-blank': require('./lib/rules/jsx-no-target-blank'),
   'jsx-filename-extension': require('./lib/rules/jsx-filename-extension'),
   'require-optimization': require('./lib/rules/require-optimization'),
-  'no-find-dom-node': require('./lib/rules/no-find-dom-node')
+  'no-find-dom-node': require('./lib/rules/no-find-dom-node'),
+  'no-children-prop': require('./lib/rules/no-children-prop')
```
lencioni: I think this should go in the recommended config as well. Thoughts @yannickcr?

tests/lib/rules/no-children-prop.js
```diff
@@ -0,0 +1,124 @@
...
+    {
+      code: 'React.createElement("div", {children: React.createElement("div")});',
+      errors: [{message: 'Do not pass children as props.'}],
+      parserOptions: parserOptions
+    },
+    {
+      code: 'React.createElement("div", {children: [React.createElement("div"), React.createElement("div")]});',
+      errors: [{message: 'Do not pass children as props.'}],
+      parserOptions: parserOptions
+    }
```
lencioni: This rule should also warn when Components receive a `children` prop. For example, this is bad too:

```jsx
<MyComponent children='Children' />
```

lib/rules/no-children-prop.js
```diff
@@ -0,0 +1,71 @@
...
+      JSXAttribute: function(node) {
+        if (isTagName(node.parent.name.name) && node.name.name === 'children') {
+          context.report({
+            node: node,
+            message: 'Do not pass children as props.'
+          });
+        }
+      },
+      CallExpression: function(node) {
+        if (isCreateElementWithProps(node)) {
```
lencioni: nit: I think it would be slightly nicer to use an early return here to keep your code flatter.

lencioni:
@benstepp You've done a really great job with your first rule! 🎊

lib/rules/no-children-prop.js
```diff
@@ -0,0 +1,71 @@
...
+    docs: {},
+    schema: []
+  },
+  create: function(context) {
+    return {
+      JSXAttribute: function(node) {
+        if (isTagName(node.parent.name.name) && node.name.name === 'children') {
+          context.report({
+            node: node,
+            message: 'Do not pass children as props.'
```
lencioni: It might be nice to additionally tell people what they should do. How about this error message?

> Do not pass children as a prop. Instead, nest children between the opening and closing tags.

lib/rules/no-children-prop.js
```diff
@@ -0,0 +1,71 @@
...
+        if (isCreateElementWithProps(node)) {
+          var props = node.arguments[1].properties;
+          var childrenProp = props.find(function(prop) {
+            return prop.key.name === 'children';
+          });
+
+          if (childrenProp) {
+            context.report({
+              node: node,
+              message: 'Do not pass children as props.'
```
lencioni: Same here. What do you think about this?

> Do not pass children as a prop. Instead, pass them as additional arguments to `React.createElement`.

benstepp:
I went ahead and made changes to the error messages, and added the additional clarity regarding the desired behavior to the introduction of the docs.

* Added a few tests for a `<MyComponent />`.
* Switched to an early return to reduce the nesting.

lib/rules/no-children-prop.js
```diff
@@ -0,0 +1,65 @@
...
+
+        var props = node.arguments[1].properties;
+        var childrenProp = props.find(function(prop) {
+          return prop.key.name === 'children';
+        });
+
+        if (childrenProp) {
+          context.report({
+            node: node,
+            message: 'Do not pass children as props. Instead, pass them as the third argument to React.createElement.'
```
lencioni: I believe that `React.createElement` can take an arbitrary number of arguments. Starting with the third argument they are considered children. This error message should probably be updated to reflect this.

tests/lib/rules/no-children-prop.js
```diff
@@ -0,0 +1,142 @@
...
+    },
+    {
+      code: '<MyComponent>Children</MyComponent>',
+      parserOptions: parserOptions
+    }
+  ],
+  invalid: [
+    {
+      code: '<div children="Children" />;',
+      errors: [{message: 'Do not pass children as props.'}],
```
lencioni: I think you need to update your tests to match the new error messages.

tests/lib/rules/no-children-prop.js
```diff
@@ -0,0 +1,142 @@
...
+    {
+      code: 'React.createElement("div", undefined, [React.createElement("div"), React.createElement("div")]);',
+      parserOptions: parserOptions
+    },
+    {
+      code: '<MyComponent />',
+      parserOptions: parserOptions
+    },
+    {
+      code: '<MyComponent>Children</MyComponent>',
```
lencioni: Will you please include a valid case using React.createElement with MyComponent and children?

tests/lib/rules/no-children-prop.js
```diff
@@ -0,0 +1,142 @@
...
+    {
+      code: 'React.createElement("div", undefined);',
+      parserOptions: parserOptions
+    },
+    {
+      code: '<div>Children</div>;',
+      parserOptions: parserOptions
+    },
+    {
+      code: 'React.createElement("div", {}, "Children");',
```
lencioni: It might be nice to include an example with multiple children. e.g.

```js
React.createElement("div", {}, "Child 1", "Child 2");
```

lencioni: Also, a valid case where there is at least one non-children prop with React.createElement and JSX.

benstepp:
Went ahead and added test cases with non-children props and multiple children, as well as clarified the error regarding `React.createElement`.

docs/rules/no-children-prop.md
ljharb: I've been told `js` gives better highlighting than `jsx`, even for JSX content.

index.js
ljharb: I agree, but that change could be considered breaking whereas this change is minor. If this is going to make it into v6 then that's fine, though.

docs/rules/no-children-prop.md
lencioni: Oh? I usually see lots of blocks with red backgrounds when using `js` on JSX code. Here's an example.

Before: https://github.com/Khan/aphrodite/tree/436325f8848db974e6f9ffba18afa6af1c426095#api
After: https://github.com/Khan/aphrodite/tree/a1fdb520fd9fcc50148a30de2c49c7c1253a604e#api

lencioni:
LGTM

index.js
lencioni: I think this is ready for v6 so might as well include it in recommended.

docs/rules/no-children-prop.md
ljharb: That does look better :-) either way's fine, but i'm pretty sure it's come up before. Maybe Github improved the highlighting since.

lencioni:
@benstepp It looks like this needs a rebase.

lencioni:
@benstepp any chance you can rebase this?

benstepp:
Forgot I still had this over here.

Rebased over current master. Same 13 failing tests as master that would probably be better fixed as a separate commit.

lencioni:
Agreed on all accounts! Thanks for rebasing. 💃

index.js
```diff
@@ -96,6 +97,7 @@ module.exports = {
         'react/jsx-no-undef': 2,
         'react/jsx-uses-react': 2,
         'react/jsx-uses-vars': 2,
+        'react/no-danger': 2,
```
lencioni: I think you unintentionally added this line, which is causing specs to fail.

benstepp: Looks like you're right. Went ahead and removed that line.

lencioni:
Thanks!

PR #723, opened 2016-07-27
Don't include deprecated rules in `all` config

pfhayes:
Since they are replaced by better versions

### DIFF
--- a/tests/index.js
+++ b/tests/index.js
@@ -39,7 +39,9 @@ describe('configurations', function() {
       assert.equal(plugin.configs.all.rules[configName], 2);
     });
     rules.forEach(function(ruleName) {
-      assert(plugin.configs.all.rules['react/' + ruleName]);
+      if (!plugin.deprecatedRules[ruleName]) {
+        assert(plugin.configs.all.rules['react/' + ruleName]);
+      }
     });
   });
 });
--- a/tests/index.js
+++ b/tests/index.js
@@ -7,13 +7,13 @@ var assert = require('assert');
 var fs = require('fs');
 var path = require('path');
 
-var rules = fs.readdirSync(path.resolve(__dirname, '../lib/rules/'))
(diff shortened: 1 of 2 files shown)


tests/index.js
```diff
@@ -39,7 +39,9 @@ describe('configurations', function() {
       assert.equal(plugin.configs.all.rules[configName], 2);
     });
     rules.forEach(function(ruleName) {
-      assert(plugin.configs.all.rules['react/' + ruleName]);
+      if (!plugin.deprecatedRules[ruleName]) {
```
ljharb: we should assert instead that every rule is either in `deprecatedRules` or `all`, rather than just skipping the deprecated ones.

ljharb:
Good catch!

tests/index.js
pfhayes: good call, I'll make that more explicit

tests/index.js
```diff
@@ -38,8 +38,10 @@ describe('configurations', function() {
       assert.equal(configName.indexOf('react/'), 0);
       assert.equal(plugin.configs.all.rules[configName], 2);
     });
-    rules.forEach(function(ruleName) {
-      assert(plugin.configs.all.rules['react/' + ruleName]);
+    ruleFiles.forEach(function(ruleName) {
+      var inDeprecatedRules = Boolean(plugin.deprecatedRules[ruleName]);
```
ljharb: fwiw `!!foo` and `Boolean(foo)` are equivalent, except that the former is faster, and can't be forged

pfhayes: I'm happy to change, but it looked like in this project `Boolean` (12 occurrences) was preferred over `!!` (0 occurrences)

ljharb: Sounds good, keep it as `Boolean` for now then. We can change it to `!!` separately, if desired.

yannickcr:
Merged, thanks!

PR #725, opened 2016-07-28
Add Pragma for createClass factory method

zurawiki:
Allows the the plugin to lint components created in a non-standard way with
configuration such as

```js
var MyComponent = SpecialReact.createSpecialClass({
  ...
});
```

with config looking like:
```json
{
  "settings": {
    "react": {
      "createClass": "createSpecialClass",
      "pragma": "SpecialReact", 
    }
  }
}
```

### DIFF
--- a/lib/util/pragma.js
+++ b/lib/util/pragma.js
@@ -5,6 +5,21 @@
 'use strict';
 
 var JSX_ANNOTATION_REGEX = /^\*\s*@jsx\s+([^\s]+)/;
+// Does not check for reserved keywords or unicode characters
+var JS_IDENTIFIER_REGEX = /^[_$a-zA-Z][_$a-zA-Z0-9]*$/;
+
+
+function getCreateClassFromContext(context) {
+  var pragma = 'createClass';
+  // .eslintrc shared settings (http://eslint.org/docs/user-guide/configuring#adding-shared-settings)
+  if (context.settings.react && context.settings.react.createClass) {
+    pragma = context.settings.react.createClass;
+  }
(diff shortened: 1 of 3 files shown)


zurawiki:
@yannickcr I would love to help you merge this in. I didn't see any notes on contributing to this project in the README or on the wiki. 

Please let me know what next steps I need to take.

yannickcr:
Thanks for your contribution.

Sorry, I'm a bit busy working on the v6.0.0 right now, but I'll try to have a look on this right after 😃

lib/util/pragma.js
```diff
@@ -6,6 +6,15 @@
 
 var JSX_ANNOTATION_REGEX = /^\*\s*@jsx\s+([^\s]+)/;
 
+function getCreateClassFromContext(context) {
+  var pragma = 'createClass';
+  // .eslintrc shared settings (http://eslint.org/docs/user-guide/configuring#adding-shared-settings)
+  if (context.settings.react && context.settings.react.createClass) {
+    pragma = context.settings.react.createClass;
+  }
+  return pragma.split('(')[0];
```
yannickcr: Why do you make a `.split('(')` here?

zurawiki: I matched the `.split('.')` on line 24.
It's there because `(` cannot be part of a function's name

ljharb: If someone sets an invalid pragma value, shouldn't the plugin error out, instead of trying to fix it for them?

zurawiki:
Ok, I added a regex and I set config to throw an error when an invalid config is given for either pragma field

yannickcr:
Merged, thanks!

PR #728, opened 2016-07-30
Fix wrap-multilines rule

akozhemiakin:
Wrap multilines was broken after switching to new eslint rule format.

### DIFF
--- a/lib/rules/wrap-multilines.js
+++ b/lib/rules/wrap-multilines.js
@@ -20,6 +20,7 @@ module.exports = {
       category: 'Stylistic Issues',
       recommended: false
     },
+    fixable: 'code',
 
     schema: [{
       type: 'object',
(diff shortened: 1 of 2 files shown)


ljharb:
Thanks! Can you add a test that would have caught this bug?

akozhemiakin:
@ljharb I thought about adding a test but it is a deprecated rule that uses jsx-wrap-multilines rule under the hood.

So I'm not sure how to test it properly. Should we just copy this test https://github.com/yannickcr/eslint-plugin-react/blob/master/tests/lib/rules/jsx-wrap-multilines.js? Or should we create separate test with a single random valid code snippet to just test that it works?

ljharb:
I think that as long as we're keeping the deprecated test we should keep the deprecated test file - so yeah, just duplicating https://github.com/yannickcr/eslint-plugin-react/blob/master/tests/lib/rules/jsx-wrap-multilines.js is probably good.

akozhemiakin:
Yeah, it is definitely a good approach to duplicate that test file :) It revealed another bug with missing meta property (fixable: 'code'). I fixed it and added the test.

ljharb:
Thanks!

PR #736, opened 2016-08-03
Add no-context rule (fixes #455)

### LINKED ISSUE #455: Rule Proposal: no context
I'd like a rule that warns if it finds either static `contextTypes` being defined, or `this.context` being used.

zachguo:
Fixes #455

### DIFF
--- a/docs/rules/no-context.md
+++ b/docs/rules/no-context.md
@@ -1,0 +1,59 @@
+# Prevent usage of context (no-context)
+
+[Context](https://facebook.github.io/react/docs/context.html) is an advanced and experimental feature. The API is likely to change in future releases. Most applications will never need to use context. Especially if you are just getting started with React, you likely do not want to use context. Using context will make your code harder to understand because it makes the data flow less clear. It is similar to using global variables to pass state through your application. Consider higher order components or Flux architecture instead.
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```jsx
+class Button extends React.Component {
+  render() {
+    return (
+      <button style={{background: this.context.color}}>
+        {this.props.children}
+      </button>
+    );
+  }
+}
+Button.contextTypes = {
+  color: React.PropTypes.string
+};
+```
+
+```jsx
+const Button = (props, context) =>
+  <button style={{background: context.color}}>
+    {props.children}
+  </button>;
+Button.contextTypes = {
+  color: React.PropTypes.string
+};
+```
+
+The following patterns are not considered warnings:
+
+```jsx
+class Button extends React.Component {
+  render() {
+    return (
+      <button style={{background: this.props.color}}>
+        {this.props.children}
+      </button>
+    );
+  }
+}
+Button.propTypes = {
+  color: React.PropTypes.string
+};
+```
+
+```jsx
+const Button = (props) =>
+  <button style={{background: props.color}}>
+    {props.children}
+  </button>;
+Button.propTypes = {
+  color: React.PropTypes.string
+};
+```
--- a/lib/rules/no-context.js
+++ b/lib/rules/no-context.js
@@ -1,0 +1,134 @@
+/**
+ * @fileoverview Prevent usage of context
+ * @author Zach Guo
+ */
+'use strict';
+
+var Components = require('../util/Components');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Prevent usage of context',
+      category: 'Best Practices',
+      recommended: true
+    },
+    schema: []
+  },
+
+  create: Components.detect(function(context) {
+
+    /**
+     * Checks if we are using context
+     * @param {ASTNode} node The AST node being checked.
+     * @returns {Boolean} True if we are using context, false if not.
+     */
+    function isContextUsage(node) {
+      return Boolean(
+        node.object.type === 'ThisExpression' &&
+        node.property.name === 'context'
+      );
+    }
+
+    /**
+   * Checks if we are declaring a context
+   * @param {ASTNode} node The AST node being checked.
+   * @returns {Boolean} True if we are declaring a context, false if not.
+   */
+    function isContextTypesDeclaration(node) {
+      // Special case for class properties
+      // (babel-eslint does not expose property name so we have to rely on tokens)
+      if (node && node.type === 'ClassProperty') {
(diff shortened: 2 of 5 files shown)


docs/rules/no-context.md
```diff
@@ -0,0 +1,51 @@
+# Prevent usage of context (no-context)
+
+[Context](https://facebook.github.io/react/docs/context.html) is an advanced and experimental feature. The API is likely to change in future releases. Most applications will never need to use context. Especially if you are just getting started with React, you likely do not want to use context. Using context will make your code harder to understand because it makes the data flow less clear. It is similar to using global variables to pass state through your application.
```
lencioni: It would be nice to point people toward things like higher order components if they want to use context.

docs/rules/no-context.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```js
```
lencioni: Please use `jsx` here for better syntax highlighting

docs/rules/no-context.md
```diff
@@ -0,0 +1,51 @@
...
+        {this.props.children}
+      </button>
+    );
+  }
+}
+Button.contextTypes = {
+  color: React.PropTypes.string
+};
+
+const Button = ({children}, context) =>
```
lencioni: It might be nice to put this in a separate code block to make it more obvious that there are two examples here.

lencioni: Also, it might be nice to change `{children}` to `props` to simplify the example, which might make it clearer to folks who haven't encountered destructuring like this before.

docs/rules/no-context.md
```diff
@@ -0,0 +1,51 @@
...
+        {this.props.children}
+      </button>
+    );
+  }
+}
+Button.propTypes = {
+  color: React.PropTypes.string
+};
+
+const Button = ({children, color}, context) =>
```
lencioni: Did you mean to remove `, context` from the signature?

zachguo:
@lencioni thanks for reviewing, all fixed. I gonna test this rule against my project a bit and let you know how it goes.

tests/lib/rules/no-context.js
```diff
@@ -0,0 +1,155 @@
...
+      '};'
+    ].join('\n'),
+    parserOptions: parserOptions,
+    errors: [{
+      message: 'Using context is not allowed.'
+    }, {
+      message: 'Using context is not allowed.'
+    }]
+  }]
+});
```
lencioni: It would be nice to include examples where:

- component is an SFC that uses `function` syntax (not arrow function)
- `context` is destructured in the signature
- the second argument has a name that is not `context`
- access of context via shenanigans
  - `(...args) => { args[1] }`
  - `function Hello() { arguments[1] }`
  - `this['context']`
  - `var foo = 'context'; this[foo]`
  - `const { context } = this;`

zachguo:
@lencioni how can I capture static properties like
```
class Foo extends Component {
  static contextTypes = {
    bar: PropTypes.object
  };
...
```

lib/rules/no-context.js
```diff
@@ -75,6 +87,15 @@ module.exports = {
 
     return {
 
+      ClassProperty: function(node) {
```
zachguo: @lencioni how can I capture static properties like
```
class Foo extends Component {
  static contextTypes = {
    bar: PropTypes.object
  };
...
```

lencioni: Not entirely sure. Since that's not part of standard JS yet (still stage 2), you might have to jump through some extra hoops to make sure there is a parser that can understand that experimental syntax.

In astexplorer.net it looks like espree fails on this syntax, but babel-eslint does not. So you probably need to specify `parser: 'babel-eslint'` in your test for these.

tests/lib/rules/no-context.js
```diff
@@ -36,12 +36,59 @@ ruleTester.run('no-context', rule, {
...
+      '  render() {',
+      '    return <div>Hello {this.props.name}</div>;',
+      '  }',
+      '};'
+    ].join('\n'),
+    parser: 'babel-eslint',
+    parserOptions: parserOptions
+  }, {
+    code: [
+      'class Hello extends React.Component {',
```
ljharb: let's also ensure it works with `extends React.PureComponent`

lencioni: Might as well. Note: that will likely require a rebase onto origin/master (which you probably want to do anyway).

tests/lib/rules/no-context.js
zachguo: I don't how to capture these shenanigans.... 😞

zachguo:
@lencioni @ljharb Everything done except [tests of context via shenanigans](https://github.com/yannickcr/eslint-plugin-react/pull/736#discussion_r73412640). Advice needed on how to capture those. I also tested it against my project, it's working well.

zachguo:
I found a false positive in my project.
```jsx
<div>
  {menuItems.map((e, i) => <MenuItem value={e.path} key={i} primaryText={e.text}/>)}
</div>
```

[My current rule for stateless functional component](https://github.com/yannickcr/eslint-plugin-react/pull/736/files#diff-2d7e1cec73d88c8503d3b7467d2f1d02R66) is that, for any React component detected by `/util/Components`, if it has a second argument then it uses `context`. 
Is something with `/util/Components`'s detection, or with my rule?

lib/rules/no-context.js
```diff
@@ -61,6 +62,19 @@ module.exports = {
     /**
      * @param {ASTNode} node We expect either an ArrowFunctionExpression,
      *   FunctionDeclaration, or FunctionExpression
+     * @returns {Boolean} True if node is a stateless functional component,
+     *   false if not.
+     */
+    function isStatelessComponent(node) {
```
zachguo: `Component.detect` seems not working, so I tried detecting stateless functional component manually. However, `utils.isReturningJSX` didn't work as expected, it's returning `false` no matter what.

zachguo:
@ljharb @lencioni I'm now stuck at the fact that I cannot reliably detect stateless function(see #741). What should I do now?

lencioni:
@zachguo It sounds like `utils.isReturningJSX` might need to be fixed? That seems like a good area to focus your efforts and would likely help a lot of other people.

ljharb:
@zachguo are you no longer interested in completing this PR?

zachguo:
Probably yes for this PR.
When I have time I may try to fix https://github.com/yannickcr/eslint-plugin-react/issues/741 first.

PR #742, opened 2016-08-05
no-did-(update|mount)-set-state: `disallow-via-methods` option

avaly:
[RFC] Adding a new rule to complement the existing `no-did-update-set-state`.

This rule will look for any occurrences of `this.setState` and check whether the method it is being called from is also called inside `componentDidUpdate`.

If this gets a positive feedback I'll also add a doc page for it.

I could see this as an option on the `no-did-update-set-state` rule, but the logic is more complex than the original rule, so I decided to keep it separate for now.

### DIFF
--- a/docs/rules/no-did-mount-set-state.md
+++ b/docs/rules/no-did-mount-set-state.md
@@ -53,7 +53,7 @@ var Hello = React.createClass({
 
 ```js
 ...
-"no-did-mount-set-state": [<enabled>, <mode>]
+"no-did-mount-set-state": [<enabled>, <modeInFunctions>, <modeViaMethods>]
 ...
 ```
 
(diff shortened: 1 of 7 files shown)


lencioni:
I'm not sure this needs to be a separate rule. It would also be nice if it allowed these function calls as callbacks with an option to disallow, like we do with the `disallow-in-func` option.

What do other folks think?

avaly:
I've updated my PR to not create a separate rule for this use case. Instead it uses the logic in both `no-did-update-set-state` and `no-did-mount-set-state` rules using a new mode `disallow-via-methods`.

avaly:
:+1: / :-1: ? Friendly bump!

docs/rules/no-did-mount-set-state.md
```diff
@@ -90,3 +90,38 @@ var Hello = React.createClass({
...
 });
 ```
+
+### `disallow-via-methods` mode
+
+By default this rule ignores any call to `this.setState` via methods called in `componentDidMount`. The `disallow-via-methods` mode makes this rule more strict by disallowing calls to `this.setState` even within methods.
+
+
+```js
```
lencioni: nit: these should be `jsx`

docs/rules/no-did-update-set-state.md
```diff
@@ -88,3 +88,38 @@ var Hello = React.createClass({
...
 });
 ```
+
+### `disallow-via-methods` mode
+
+By default this rule ignores any call to `this.setState` via methods called in `componentDidUpdate`. The `disallow-via-methods` mode makes this rule more strict by disallowing calls to `this.setState` even within methods.
+
+
+```js
```
lencioni: `jsx`

docs/rules/no-did-update-set-state.md
```diff
@@ -51,7 +51,7 @@ var Hello = React.createClass({
 
 ```js
 ...
-"no-did-update-set-state": [<enabled>, <mode>]
+"no-did-update-set-state": [<enabled>, <modeInFunctions>, <modeViaMethods>]
```
lencioni: It might be a good idea to use an options object for these modes instead of positional arguments.

That would be a breaking change, so perhaps it should be done in a followup PR.

lencioni (approved):

docs/rules/no-did-update-set-state.md
ljharb: Both could be supported with eslint schemas as semver-minor - ie, a "mode" string, and then otherwise, an options object.

lencioni:
Ah yes, that would be preferable

ljharb (changes requested):
This needs a rebase, and it'd be great to use an options object per https://github.com/yannickcr/eslint-plugin-react/pull/742/files#r101368104

ljharb:
I've rebased this, but a few tests are failing.

PR #744, opened 2016-08-06
[Fix] In require-optimization corrected test for function declaration

Tom910:
There was a problem with the ad functions and methods in classes, sample code -
```jsx
import React from "react";
class YourComponent extends React.Component {
  handleClick() {}
  render() {
     return <div onClick={this.handleClick}>123</div>;
  }
}
```
```json
{
  "rules": {
    "react/require-optimization": 2  
  }
}
```
This code has been checked and is not output error, it is wrong

### DIFF
--- a/tests/lib/rules/require-optimization.js
+++ b/tests/lib/rules/require-optimization.js
@@ -39,6 +39,17 @@ ruleTester.run('react-require-optimization', rule, {
     parserOptions: parserOptions
   }, {
     code: [
+      'import React, {Component} from "react";',
+      '@reactMixin.decorate(PureRenderMixin)',
+      'class YourComponent extends Component {',
+      '  componetnDidMount () {}',
+      '  render() {}',
+      '}'
+    ].join('\n'),
+    parser: 'babel-eslint',
+    parserOptions: parserOptions
+  }, {
+    code: [
       'import React from "react";' +
       'React.createClass({' +
       'shouldComponentUpdate: function () {}' +
@@ -119,6 +130,21 @@ ruleTester.run('react-require-optimization', rule, {
       'import React from "react";' +
       'class YourComponent extends React.Component {}'
     ].join('\n'),
+    errors: [{
+      message: MESSAGE
+    }],
+    parserOptions: parserOptions
+  }, {
+    code: [
+      'import React from "react";',
+      'class YourComponent extends React.Component {',
+      '  componentDidMount() {}',
+      '  render() {',
+      '    return <div>123</div>',
+      '  }',
+      '}'
+    ].join('\n'),
+    parser: 'babel-eslint',
     errors: [{
       message: MESSAGE
     }],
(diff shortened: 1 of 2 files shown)


tests/lib/rules/require-optimization.js
```diff
@@ -39,6 +39,17 @@ ruleTester.run('react-require-optimization', rule, {
     parserOptions: parserOptions
   }, {
     code: [
+      'import React, {Component} from "react";',
+      '@reactMixin.decorate(PureRenderMixin)',
+      'class YourComponent extends Component {',
+      '  componetnDidMount () {}',
```
ljharb: should this be `componentDidMount`?

ljharb:
The example in your OP should indeed pass the "require-optimization" check - because it has a lifecycle method. That the method is empty is irrelevant; adding the empty method clearly indicates developer intent to flesh it out later, and as such, they should not be required to make it an SFC.

tests/lib/rules/require-optimization.js
Tom910: Replace with a new example

Tom910:
@ljharb In the code, there is no such logic. A similar operation is due to not a correct execution "FunctionExpression"

ljharb:
In your OP you say "And this code passed checks, it is wrong" - if the code passes checks, what's the bug?

Tom910:
@ljharb Thanks, corrected the wording

ljharb:
OK so the code in your example:
```js
import React from "react";
class YourComponent extends React.Component {
  handleClick() {}
  render() {
     return <div onClick={this.handleClick}>123</div>;
  }
}
```

has a class method, so it should indeed NOT output a `require-optimization` error (unless the rule is smart enough to recognize that the method doesn't use `this` or `super`, and isn't a lifecycle method - but I don't think that's the case). Are you saying you do get one?

Tom910:
@jkimbo For whatever reason, in your example should not be a mistake? require-optimization should be checked for statelesFunction, some mixins and shouldComponentUpdate. For all other cases, the error should be.

I turned on this rule on my big project, 5000+ react components and has received 0 such errors, although there are a lot of components, without optimization)

ljharb:
Sorry for my earlier confusion - I was confusing this with a "prefer SFCs" rule :-)

Afaik `require-optimization` should skip any `createClass` component with `PureRenderMixin`, and should skip any `class` component with a `shouldComponentUpdate` method, and should skip any SFC entirely.

If you have components that don't match these criteria, and `require-optimization` is not warning on them, then that seems like it might indeed be a bug.

lencioni:
It should also skip classes that extend `React.PureComponent`.

Tom910:
@lencioni Yes, I agree, but it has already implemented @yannickcr 

@ljharb Yes, it is =)

yannickcr:
Merged, thanks!

Issue #730, opened 2016-08-02
`self-closing-comp` warns on elements with whitespace children.


ljharb:
`<div>    </div>` is NOT supposed to be self-closing - the whitespace is reduced to a single space, but it's definitely not zero spaces. Currently the rule recommends that this JSX be self-closed, but that would result in the removal of a space.

petersendidit:
Seems difficult for this rule to know when the space is intentional vs a result of some refactoring. The rule currently supports `<div>{' '}</div>` as valid. Which helps the rule know that the space is intentional. Seems like if we make the rule support `<div> </div>` it could miss the more likely case where it was a left over from refactoring.

Seems like it might be better to update the documentation for the rule to show that `<div>{' '}</div>` is valid.

ljharb:
The point of the rule is that I should be able to autofix the violations and not change the meaning of the JSX.

If I have `<div> </div>` the rule should be telling me to change it to `<div>{' '}</div>`, but NOT to change it to `<div />` - and since it's "self-closing-comp", it should only be warning on things that changing to self-closing is *100% identical*.

ljharb:
This, and #1717, were fixed by #2210.

PR #745, opened 2016-08-06
Update self-closing-comp doc to show example of a intentional space

petersendidit:
This came up when reviewing #730

### DIFF
--- a/docs/rules/self-closing-comp.md
+++ b/docs/rules/self-closing-comp.md
@@ -15,9 +15,13 @@ The following patterns are not considered warnings:
 ```js
 var contentContainer = <div className="content"></div>;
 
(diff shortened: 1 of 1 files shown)


ljharb:
I think the added example is good 👍 but I don't think it resolves #730 whatsoever.

docs/rules/self-closing-comp.md
```diff
@@ -15,9 +15,13 @@ The following patterns are not considered warnings:
 ```js
 var contentContainer = <div className="content"></div>;
 
+var intentionalSpace = <div>{' '}</div>;
```
lencioni: It would be nice to include this example in the html section down below.

lencioni: Also the component section, I guess.

PR #755, opened 2016-08-08
Add new dom-elements-style-is-object rule (Fixes #715)

### LINKED ISSUE #715: Rule proposal: ensure style prop on DOM element tags is an object
React will issue a warning in dev if you pass a non-object to the `style` prop on DOM element tags

https://github.com/facebook/react/blob/85dcbf83/src/renderers/dom/shared/ReactDOMComponent.js#L204-L207

This should be easy enough to enforce for most cases with a linter rule.

Bad:

```jsx
<div style="foo" />
```

```jsx
<div style={true} />
```

```jsx
const styles = true;
<div style={styles} />
```

Good:

```jsx
<div style={{ color: 'red' }} />
```

```jsx
const styles = { color: 'red' };
<div style={styles} />
```

It would be even nicer if it validated that the values of the style object are all strings.

petersendidit:
Fixes #715

### DIFF
--- a/index.js
+++ b/index.js
@@ -54,7 +54,8 @@ var rules = {
   'jsx-filename-extension': require('./lib/rules/jsx-filename-extension'),
   'require-optimization': require('./lib/rules/require-optimization'),
   'no-find-dom-node': require('./lib/rules/no-find-dom-node'),
-  'no-danger-with-children': require('./lib/rules/no-danger-with-children')
+  'no-danger-with-children': require('./lib/rules/no-danger-with-children'),
+  'style-prop-object': require('./lib/rules/style-prop-object')
 };
 
 var ruleNames = Object.keys(rules);
--- a/README.md
+++ b/README.md
@@ -105,6 +105,7 @@ Finally, enable all of the rules that you would like to use.  Use [our preset](#
 * [react/self-closing-comp](docs/rules/self-closing-comp.md): Prevent extra closing tags for components without children (fixable)
 * [react/sort-comp](docs/rules/sort-comp.md): Enforce component methods order
 * [react/sort-prop-types](docs/rules/sort-prop-types.md): Enforce propTypes declarations alphabetical sorting
+* [react/style-prop-object](docs/rules/style-prop-object.md): Enforce style prop value being an object
 
 ## JSX-specific rules
 
--- a/docs/rules/style-prop-object.md
+++ b/docs/rules/style-prop-object.md
@@ -1,0 +1,51 @@
+# Enforce style prop value being an object (style-prop-object)
+
+Require that the value of the prop `style` be an object or a variable that is
+an object.
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```jsx
+<div style="color: 'red'" />
+
+<div style={true} />
+
+<Hello style={true} />
+
+const styles = true;
+<div style={styles} />
+```
+
+```js
+React.createElement("div", { style: "color: 'red'" });
+
+React.createElement("div", { style: true });
+
+React.createElement("Hello", { style: true });
+
+const styles = true;
+React.createElement("div", { style: styles });
+```
+
+
+The following patterns are not considered warnings:
+
+```jsx
+<div style={{ color: "red" }} />
+
+<Hello style={{ color: "red" }} />
+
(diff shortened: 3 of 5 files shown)


index.js
```diff
@@ -72,6 +73,7 @@ module.exports = {
         }
       },
       rules: {
+        'react/dom-elements-style-is-object': 2,
```
lencioni: I think we want to wait to enable this in the recommended configuration until the next major version bump, since doing so may end up breaking people's builds.

lib/rules/dom-elements-style-is-object.js
```diff
@@ -0,0 +1,100 @@
...
+var variableUtil = require('../util/variable');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Enforce style prop value being an object and havng sting values',
```
lencioni: typo: sting -> string

lencioni: Also, I don't think that the string values part of this rule is true.

lencioni: It would be nice for this rule to stick to the type of check that is made by React, which simply verifies that it is null or an object.

https://github.com/facebook/react/blob/85dcbf83/src/renderers/dom/shared/ReactDOMComponent.js#L204

For example, these should all be valid:

- Integer value, e.g. `style={{ width: 100 }}`
- Some string values and some integer values, e.g. `style={{ color: 'red', width: 100 }}`
- Interpolation in a value, e.g. `style={{ border: `1px solid ${primaryColor}` }}`
- Function call in the value, e.g. `style={{ border: borderFactory(1) }}`
- Variable as the value, e.g. `style={{ color: red }}` and `style={{ color }}`
- String concatenation in the value, e.g. `style={{ border: '1px solid ' + primaryColor }}`

lencioni: I think it might be nice to have a rule that validates the shape of the style object when that can be discerned, but I think that could be a different rule.

yannickcr:
Great! Can you fix the issues raised by @lencioni?

Also I'm not really a fan of the rule name: too long, should start by `jsx-` like the other rules.

`jsx-style-object`? Any other idea?

ljharb:
It shouldn't have jsx since it should apply to createElement DOM elements too.

petersendidit:
Yea been hoping to get back to fixing up this rule. Hoping I can do it sometime this weekend.

Yes the name is long but I think it accurately describes the rule.

lib/rules/dom-elements-style-is-object.js
petersendidit: What should the rule for validating the style object be named?  `style-prop-is-valid`?

petersendidit:
Checking that the style object values are stings has been pulled out and addressed the other comments.

yannickcr:
@ljharb A lot of current JSX rules can also apply to `createElement`. So even if it also apply to `createElement` I still think it should be prefixed by `jsx-`, just by consistency with the other rules.

ljharb:
@yannickcr hm, I feel the opposite - that those rules should be renamed to not have jsx unless that's the only thing they apply to.

petersendidit:
So what is the decision on the name? Are we going with `jsx-style-object` or are we sticking with the longer name?

ljharb:
I don't care about the length; `style-prop-object` would be fine with me too. I just think that `jsx-` should be reserved only for things that can only apply to jsx.

yannickcr:
Sorry for the delay.

> those rules should be renamed to not have jsx unless that's the only thing they apply to

Yeah, in fact the `jsx-` rules should only be styling rules for JSX since the other ones can apply to `creactElement` too. A rename would be the solution but I'm not really a fan of renaming rules since it introduces a breaking change for almost no added value 😐. But that's another discussion 😉 

I'm ok with `style-prop-object` 👍

petersendidit:
Renamed the rule. Let me know if there is anything else left to do.

yannickcr:
Merged. Thank you!

antoinerousseau:
does not work if i'm using `componentWillReceiveProps`'s first argument (next props)

PR #761, opened 2016-08-11
Fix jsx-uses-var to handle nested object properties

yayalice:
When importing an object and using a deeply nested property as a component, mark that variable as used.


yannickcr:
Great, thanks!

Issue #524, opened 2016-04-01
Does jsx-no-bind contradicts with no-string-refs?


guocheng:
The `no-string-refs` document provides the following example:

```js
render() {
    return <div ref={c => this.hello = c}>Hello, world.</div>;
  }
```

I am curious if the arrow function `ref={c => this.hello = c}` gets recreated every time render is called and thus contradicts with the `jsx-no-bind` rule.

Thanks.

anthonator:
I'm curious about this too!

ljharb:
Yes, it would, unless you've enabled the override options to it for arrow functions and/or refs.

Kovensky:
How does React actually behave if the identity of the `ref` pseudo-prop changes?

If you put it on a DOM element, it's not a prop that is reflected on the DOM, so even if the diff algorithm finds that it needs to do "something" because it changed, it would not ultimately write to the DOM.

If you put it on a custom component, it's not a prop that can be read by the component (AFAIK), so it shouldn't do the full `shouldComponentUpdate` / `render` cycle (if it does, even though it's a prop that the component cannot do anything with, it sounds like a React bug).

cristian-sima:
+1

ljharb:
@cristian-sima please don't reply "+1" to posts on github - that's what "emoji reactions on the original post" are for. I'll continue to clean up any such posts and I encourage you to add a +1 to the original post.

ixisio:
@Kovensky, @ljharb, @anthonator, @guocheng 

[from eslint docs](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md): 
> A bind call or arrow function in a JSX prop will create a brand new function on every single render. This is bad for performance, as it will result in the garbage collector being invoked way more than is necessary.

So, try it like that:

````js
export class Test extends React.Component {
  constructor(props) {
    super(props);

    this.elementRef = this.elementRef.bind(this);
  }

  elementRef (element) {
    this.element = element;
  }
  
  render () {
    return (
        <div ref={this.elementRef}>Hello</div>
    )
  }
}
```

Cheers Andy

anthonator:
We've been doing something similar but slightly different:

```js
export class Test extends React.Component {
  constructor(props) {
    super(props);
  }

  elementRef (element) {
    this.element = element;
  }

  elementRef = this.elementRef.bind(this)

  render () {
    return (
        <div ref={this.elementRef}>Hello</div>
[...]

PR #764, opened 2016-08-13
Do not recommend arrow functions in no-string-refs documentation

toxicFork:
See https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md

Apart from jsx-no-bind reasons ( #524 ) , this function would actually also **get called every render**, so it should be explicitly recommended. 

See https://github.com/facebook/react/blob/master/src/renderers/shared/stack/reconciler/ReactReconciler.js#L150-L165

If necessary, additional comments can be added regarding this explanation.

### DIFF
--- a/docs/rules/no-string-refs.md
+++ b/docs/rules/no-string-refs.md
@@ -6,7 +6,7 @@ Currently, two ways are supported by React to refer to components. The first one
 
 Invalid:
 
-```js
+```jsx
 var Hello = React.createClass({
  render: function() {
   return <div ref="hello">Hello, world.</div>;
@@ -14,7 +14,7 @@ var Hello = React.createClass({
 });
 ```
 
-```js
+```jsx
 var Hello = React.createClass({
   componentDidMount: function() {
     var component = this.refs.hello;
@@ -28,14 +28,17 @@ var Hello = React.createClass({
 
 Valid:
 
-```js
+```jsx
 var Hello = React.createClass({
   componentDidMount: function() {
     var component = this.hello;
     // ...do something with component
   },
+  _helloRef(c){
+    this.hello = c;
+  },
   render() {
-    return <div ref={(c) => { this.hello = c; }}>Hello, world.</div>;
+    return <div ref={this._helloRef}>Hello, world.</div>;
   }
 });
 ```


ljharb:
The intention is that it's called every render, otherwise it wouldn't get an up-to-date DOM node. Your example gets called every render too.

toxicFork:
@ljharb you are incorrect, because if the dom node changes then the ref will be called as well. 

The problem is when the dom node does not change but the function gets called again and again every time you render.

E.g. equivalent of:

before:

```js
this._domNode = undefined;

domNode = document.createElement('div');

render(); // initial render
this._domNode = domNode;

render(); // domNode unchanged
this._domNode = domNode;

render(); // domNode unchanged
this._domNode = domNode;

render(); 
domNode = document.createElement('div'); // something changes dom node because of render
this._domNode = domNode;

```

after:
```js
this._domNode = undefined;

domNode = document.createElement('div');

render(); // initial render
this._domNode = domNode;

render(); // domNode unchanged

render(); // domNode unchanged

render(); 
domNode = document.createElement('div'); // something changes dom node because of render
this._domNode = domNode;

```

You can also test this by doing something like:
```jsx
render() {
    return (<div
        ref={(c) => {
            console.log('this._domNode is being set but should it have?', this._domNode === c);
            this._domNode = c;}
        }
    />);
}
```

toxicFork:
However I do need to clarify the PR comment to explain what "this function" is

ljharb:
Ah, you're implying that because the function is `===` that it won't think the prop has changed? I was under the impression that React didn't actually do that optimization.

However, there's also no appreciable performance impact to having it called on every render unconditionally, nor to using an arrow function - and it can be much clearer than an instance method.

toxicFork:
Regarding the prop change, here's how react will consider this scenario:

1. first render
    - Just created this element, call its ref function :+1: 
1. second render and for each render after that
    - This ref is different than last time, even though this component may not have changed... [call its previous ref function with null](https://github.com/facebook/react/blob/master/src/renderers/shared/stack/reconciler/ReactRef.js#L72-L80) and [call the new ref function with the new value](https://github.com/facebook/react/blob/master/src/renderers/shared/stack/reconciler/ReactRef.js#L36-L44) :-1: 

I agree with you regarding the performance impact, it may not affect things that much since it will technically be "just a couple more function calls". Usually it won't affect the performance significantly at all, unless you're doing a lot of updates.

Up to the authors to consider the significance of the impacts, I am guessing in 99% of the scenarios it should be OK.

docs/rules/no-string-refs.md
```diff
@@ -28,14 +28,17 @@ var Hello = React.createClass({
...
 Valid:
 
-```js
+```jsx
 var Hello = React.createClass({
   componentDidMount: function() {
     var component = this.hello;
     // ...do something with component
   },
+  _helloRef(c){
```
ljharb: at the very least, let's make sure this function does not use the harmful convention of pretending a leading underscore is somehow "private" :-) also, there always needs to be a space between `)` and `{`.

toxicFork: @ljharb updated :)

pixelass:
How should a `ref` be assigned?
Neither a string nor a function are accepted while it is in the documentation of React.js itself.

https://facebook.github.io/react/docs/refs-and-the-dom.html

Help apprechiated

ljharb:
@pixelass a constructor-bound instance method - ie, `this.foo = this.foo.bind(this)`, `foo(node) { this.something = node; }`, and `ref={this.foo}`.

pixelass:
@ljharbthx for the fast reply.
Is this documented anywhere on the react documentation pages? I can't find official docs to this concept.

ljharb:
React doesn't always document best practices.

pixelass:
React is open source so someone who know how "it should work" should open a PR for that. 

Sadly I'm not sure what the current best practice is. ;)

PR #770, opened 2016-08-15
Add auto fix for self-closing-comp

pl12133:
If you would like comments removed just let me know. Thanks for the awesome project.

### DIFF
--- a/tests/lib/rules/self-closing-comp.js
+++ b/tests/lib/rules/self-closing-comp.js
@@ -107,12 +107,14 @@ ruleTester.run('self-closing-comp', rule, {
   invalid: [
     {
       code: 'var contentContainer = <div className="content"></div>;',
+      output: 'var contentContainer = <div className="content"/>;',
       parserOptions: parserOptions,
       errors: [{
         message: 'Empty components are self-closing'
       }]
     }, {
       code: 'var contentContainer = <div className="content"></div>;',
+      output: 'var contentContainer = <div className="content"/>;',
       options: [],
       parserOptions: parserOptions,
       errors: [{
@@ -120,18 +122,21 @@ ruleTester.run('self-closing-comp', rule, {
       }]
     }, {
       code: 'var HelloJohn = <Hello name="John"></Hello>;',
+      output: 'var HelloJohn = <Hello name="John"/>;',
       parserOptions: parserOptions,
       errors: [{
         message: 'Empty components are self-closing'
       }]
     }, {
       code: 'var HelloJohn = <Hello name="John">\n</Hello>;',
+      output: 'var HelloJohn = <Hello name="John"/>;',
       parserOptions: parserOptions,
       errors: [{
         message: 'Empty components are self-closing'
       }]
     }, {
       code: 'var HelloJohn = <Hello name="John"> </Hello>;',
+      output: 'var HelloJohn = <Hello name="John"/>;',
       parserOptions: parserOptions,
       errors: [{
         message: 'Empty components are self-closing'
@@ -139,6 +144,7 @@ ruleTester.run('self-closing-comp', rule, {
     },
     {
       code: 'var HelloJohn = <Hello name="John"></Hello>;',
+      output: 'var HelloJohn = <Hello name="John"/>;',
       options: [],
       parserOptions: parserOptions,
       errors: [{
@@ -146,6 +152,7 @@ ruleTester.run('self-closing-comp', rule, {
       }]
     }, {
       code: 'var HelloJohn = <Hello name="John">\n</Hello>;',
+      output: 'var HelloJohn = <Hello name="John"/>;',
       options: [],
       parserOptions: parserOptions,
       errors: [{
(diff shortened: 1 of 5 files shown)


tests/lib/rules/self-closing-comp.js
```diff
@@ -107,73 +107,84 @@ ruleTester.run('self-closing-comp', rule, {
   invalid: [
     {
       code: 'var contentContainer = <div className="content"></div>;',
+      output: 'var contentContainer = <div className="content"/>;',
```
ljharb: these should all autofix with a space before the `/>`.

pl12133: I'm not sure I understand, as that change makes the tests fail. It seems that if there is no space before the `>` then after fixing there should continue to be no space before the `>`.

Would you prefer that the replacing value located in [lib/rules/self-closing-comp.js - Line 88](https://github.com/yannickcr/eslint-plugin-react/pull/770/files#diff-0f0a3a3442c0b696947282a7eac6cedfR88) be changed from `"/>"` to `" />"`?

ljharb: Yes, that's exactly what i'm suggesting. `<div />` is correct, not `<div/>`

pl12133: Thank you for clarifying, I didn't realize the white space was part of the rule. I have added that behavior now. Do you specifically want to enforce a single white space character, or would multiple be valid? For example, would the following be invalid:

```js
<Foo  /> // double white space
```

ljharb: I think multiple would be ok, but we'd want to encourage single.

tleunen: @ljharb, @pl12133 - This should also take the rule `jsx-space-before-closing` into consideration. No space could be valid if `jsx-space-before-closing` is set as `never`.

ljharb: That's true - but I don't think eslint exposes that information from one rule to another.

pl12133: Good catch @tleunen, I have now tested this rule with `jsx-space-before-closing` which is also fixable, and both of these fixes are working together to fix all warnings when using my initial commit. Adding a white space may be beyond the scope of this rule, as that is better left to the fix implemented in `jsx-space-before-closing`.

If you and @ljharb agree after your own testing that the fix for `jsx-space-before-closing` works well with this fix, then I think f6cc36a should be reverted.

Also from our discussion I noticed that the [jsx-space-before-closing documentation](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-space-before-closing.md) is missing the **Fixable** notice, I can add that to this PR if you like.

ljharb: Please do update the documentation to be correct :-)

As long as the combination of those rules (jsx-space-before-closing set to never, and this one) yields code that passes the linter, I think this works great :-)

pl12133: Great, and to be clear here is how I'm seeing the two rules interact:

With  `'react/jsx-space-before-closing': [2, 'never']`:
Initial: `<Foo></Foo>`
Fixed: `<Foo/>`


With  `'react/jsx-space-before-closing': 2`:
Initial: `<Foo></Foo>`
Fixed: `<Foo />`

I have now reverted f6cc36a and updated documentation for `jsx-space-before-closing` page.

ljharb: Why revert it? The default fix for this rule should be the same as the default spacing on jsx-space-before-closing - which is to include the space.

pl12133: I reverted because white space considerations seem outside the scope of this rule. When running `--fix` the linter will first find the warning for `self-closing-comp` and fix it. The linter will then be run again against the code with fixed `self-closing-comp` rule, and apply the fix for `jsx-space-before-closing`. So it seems that white space and closing tags are unrelated, and this way the fixes for the two rules work together to give the result of producing a self-closing tag with the appropriate white space.

pl12133: I believe that is the only ordering possible, because the `jsx-space-before-closing` rule does not produce a warning unless there is a self-closing tag. This obviously cannot happen if `self-closing-comp` is in violation, because that rule implies there is no self-closing tag. Therefore I think it is impossible for `jsx-space-before-closing` to produce an error on the same node as `self-closing-comp`, and the warning conditions for the two rules are mutually exclusive.

ljharb: Yes but someone may not have enabled `jsx-space-before-closing` - I want the burden of running that extra autofixer to be on someone who has chosen "never" - which is a very unconventional choice - not on the vast majority of JSX authors who always put a space there.

pl12133: I see, in that case I can re-commit the white space character before a self-closing tag. Should there be any other behavior for white space than what is in f6cc36a?

ljharb: nope, that looks fine to me!

pl12133: Sounds good, let me know if you want commits squashed before merging.

yannickcr:
Merged, thanks!

Issue #777, opened 2016-08-18
React.PureComponent and react/prefer-stateless-function


kassiansun:
PureComponent will be warned with react/prefer-stateless-function like:
```javascript
class App extends React.PureComponent {
    render() {
        return (<h2>Hello</h2>);
    }
}
```

ljharb:
You should be preferring a stateless function there, since there's no state. "PureComponent" usage doesn't affect this rule by design.

kassiansun:
@ljharb So how to make this function stateless?

ljharb:
```js
function App() {
  return (<h2>Hello</h2>);
}
```

kassiansun:
@ljharb Thanks for your tips! I misunderstood 'pure' and 'stateless' for a long time.

PR #781, opened 2016-08-21
Make prefer-stateless-function PureComponent aware

tiemevanveen:
prefer-stateless-function should not return a warning when the user uses `PureComponent` to opt in for free optimisations with `shouldComponentUpdate` and is using props or context. 

Current false positive: 

```{js}
class Foo extends React.PureComponent {
  render() {
  }
}
```

I first only added the same `isPureComponent` check as [require-optimization](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/require-optimization.js#L92) does. But realized that components that are extending form PureComponent but don't use `this.props` or `this.context` should be written as a pure function as well. So I had to change the existing checks a bit to check that too. 

Yeah I know, I could have first reacted to the discussion at #777, but I thought it would be cool to find out how all the 'rules' in eslint work by trying to fix this issue. Hope you agree about the false positive. I think this case is different from #777 because that case it not using props or context at all.

Please correct any formatting fails, first contribution here.

### DIFF
--- a/docs/rules/prefer-stateless-function.md
+++ b/docs/rules/prefer-stateless-function.md
@@ -16,7 +16,7 @@ If none of these elements are found, the rule will warn you to write this compon
 
 The following pattern is considered a warning:
 
-```js
+```jsx
 var Hello = React.createClass({
   render: function() {
     return <div>Hello {this.props.name}</div>;
@@ -26,7 +26,7 @@ var Hello = React.createClass({
 
 The following pattern is not considered a warning:
 
-```js
+```jsx
 const Foo = function(props) {
   return <div>{props.foo}</div>;
 };
@@ -34,7 +34,7 @@ const Foo = function(props) {
 
 The following pattern is not considered a warning in React <15.0.0:
 
-```js
+```jsx
 class Foo extends React.Component {
   render() {
     if (!this.props.foo) {
@@ -45,9 +45,25 @@ class Foo extends React.Component {
 }
 ```
 
-The following pattern is not considered a warning:
+
+## Rule Options
 
 ```js
+...
+"prefer-stateless-function": [<enabled>, { "ignorePureComponent": <ignorePureComponent> }]
+...
+```
+
+* `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
+* `ignorePureComponent`: optional boolean set to `true` to ignore components extending from `React.PureComponent` (default to `false`).
+
+### `ignorePureComponent`
+
+When `true` the rule will ignore Components extending from `React.PureComponent` that use `this.props` or `this.context`.
+
+The following patterns is considered okay and does not cause warnings:
+
+```jsx
 class Foo extends React.PureComponent {
   render() {
     return <div>{this.props.foo}</div>;
@@ -55,3 +71,12 @@ class Foo extends React.PureComponent {
 }
 ```
 
+The following pattern is considered a warning because it's not using props or context:
+
+```jsx
+class Foo extends React.PureComponent {
+  render() {
+    return <div>Bar</div>;
+  }
+}
+```
--- a/docs/rules/prefer-stateless-function.md
+++ b/docs/rules/prefer-stateless-function.md
@@ -8,12 +8,13 @@ This rule will check your class based React components for
 
 * methods/properties other than `displayName`, `propTypes`, `render` and useless constructor (same detection as ESLint [no-useless-constructor rule](http://eslint.org/docs/rules/no-useless-constructor))
 * instance property other than `this.props` and `this.context`
+* extension of `React.PureComponent` ()
 * presence of `ref` attribute in JSX
 * `render` method that return anything but JSX: `undefined`, `null`, etc. (only in React <15.0.0, see [shared settings](https://github.com/yannickcr/eslint-plugin-react/blob/master/README.md#configuration) for React version configuration)
 
-If none of these 4 elements are found, the rule will warn you to write this component as a pure function.
+If none of these elements are found, the rule will warn you to write this component as a pure function.
 
-The following pattern is considered warnings:
+The following pattern is considered a warning:
 
 ```js
 var Hello = React.createClass({
@@ -23,7 +24,7 @@ var Hello = React.createClass({
 });
 ```
 
-The following pattern is not considered warnings:
+The following pattern is not considered a warning:
 
 ```js
 const Foo = function(props) {
@@ -31,7 +32,7 @@ const Foo = function(props) {
 };
 ```
 
-The following pattern is not considered warning in React <15.0.0:
+The following pattern is not considered a warning in React <15.0.0:
 
 ```js
 class Foo extends React.Component {
@@ -43,3 +44,14 @@ class Foo extends React.Component {
   }
 }
 ```
+
+The following pattern is not considered a warning:
+
+```js
+class Foo extends React.PureComponent {
+  render() {
+    return <div>{this.props.foo}</div>;
+  }
+}
+```
+
--- a/lib/rules/prefer-stateless-function.js
+++ b/lib/rules/prefer-stateless-function.js
@@ -7,7 +7,6 @@
 'use strict';
 
 var Components = require('../util/Components');
-var pragmaUtil = require('../util/pragma');
 var versionUtil = require('../util/version');
 
 // ------------------------------------------------------------------------------
@@ -21,13 +20,22 @@ module.exports = {
       category: 'Stylistic Issues',
       recommended: false
     },
-    schema: []
+    schema: [{
+      type: 'object',
+      properties: {
+        ignorePureComponents: {
+          default: false,
+          type: 'boolean'
+        }
+      },
+      additionalProperties: false
+    }]
   },
 
   create: Components.detect(function(context, components, utils) {
 
-    var pragma = pragmaUtil.getFromContext(context);
-    var pureComponentRegExp = new RegExp('^(' + pragma + '\\.)?PureComponent$');
+    var configuration = context.options[0] || {};
+    var ignorePureComponents = configuration.ignorePureComponents || false;
 
     var sourceCode = context.getSourceCode();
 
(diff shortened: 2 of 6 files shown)


lib/rules/prefer-stateless-function.js
```diff
@@ -65,6 +69,19 @@ module.exports = {
...
+     * @param {ASTNode} node The AST node being checked.
+     * @returns {Boolean} True if node extends React.PureComponent, false if not.
+     */
+    var isPureComponent = function (node) {
+      if (node.superClass) {
+        return pureComponentRegExp.test(sourceCode.getText(node.superClass));
+      }
+
+      return false;
+    };
```
lencioni: I wonder if this function should go in lib/util/Components.js

lencioni:
This seems like a good change to me. I think it could be DRYed up a little with some of the code borrowed from other rules.

The only question I have is--should this be the behavior for everyone, or should it be an option? I'm inclined to think it can be the behavior for everyone for the sake of simplicity, but I'd like to hear opinions on that.

tiemevanveen:
Thanks for the quick response. 

Yeah I changed as little as possible. Just let me know if you want me to move some of it to the utils. That makes sense.

docs/rules/prefer-stateless-function.md
```diff
@@ -8,10 +8,11 @@ This rule will check your class based React components for
 
 * methods/properties other than `displayName`, `propTypes`, `render` and useless constructor (same detection as ESLint [no-useless-constructor rule](http://eslint.org/docs/rules/no-useless-constructor))
 * instance property other than `this.props` and `this.context`
+* extension of `React.PureComponent` ()
 * presence of `ref` attribute in JSX
 * `render` method that return anything but JSX: `undefined`, `null`, etc. (only in React <15.0.0, see [shared settings](https://github.com/yannickcr/eslint-plugin-react/blob/master/README.md#configuration) for React version configuration)
 
-If none of these 4 elements are found, the rule will warn you to write this component as a pure function.
+If none of these 5 elements are found, the rule will warn you to write this component as a pure function.
```
ljharb: Let's just make this say "these elements" to avoid having to update the count in the future.

docs/rules/prefer-stateless-function.md
```diff
@@ -43,3 +44,14 @@ class Foo extends React.Component {
   }
 }
 ```
+
+The following pattern is not considered warnings:
```
ljharb: a warning

docs/rules/prefer-stateless-function.md
```diff
@@ -43,3 +44,14 @@ class Foo extends React.Component {
...
+
+```js
+class Foo extends React.PureComponent {
```
ljharb: I still think this pattern should be considered a warning - being a PureComponent does not mean this is better as a stateful component.

lib/rules/prefer-stateless-function.js
ljharb: Agreed

ljharb:
I'm -1 on this, per #777.

docs/rules/prefer-stateless-function.md
tiemevanveen: haha yeah that's what I wrote, but the rest of docs we're like this so I didn't change it. Will fix this.

docs/rules/prefer-stateless-function.md
tiemevanveen: Being a PureComponent does not make it a stateful component right? It just makes it benefit from `shouldComponentUpdate`

tiemevanveen:
on #777 
> You should be preferring a stateless function there, since there's no state. "PureComponent" usage doesn't affect this rule by design.

But if the user chose to use PureComponent instead of Component and uses props or context, doesn't that indicate that the user want's the component to only update if the props or context changes? And that's not possible with a stateless component.

And if this is not desired then there should also be a warning when the user implemented `shouldComponentUpdate` or used the `PureRenderMixin` right? And both are not triggering a warning now.

docs/rules/prefer-stateless-function.md
ljharb: Being a PureComponent does not make it stateless either - the two things are orthogonal. I'm saying `prefer-stateless-function` is far more important than `prefer-optimization`.

ljharb:
If the user wants to convey their intentions to ignore this rule that prefers components without state be SFCs, they can do that with an eslint override.

While I agree that implementing `shouldComponentUpdate` means that the component can't be blindly made into an SFC, I think that if a `React.createClass` implements `PureRenderMixin`, but in all other respects should be an SFC - that it should be a warning.

lencioni:
Yeah I was originally thinking that using PureComponent was a strong
indication that the developer prefers this over an SFC, but taking a step
back I don't think that's accurate. I agree that this should be handled on
a case by case basis with an eslint override and not by weakening this rule.

tiemevanveen:
Okay, fair enough.

But currently it's not possible to globally override this rule in `.eslintrc` so that it doesn't trigger a warning when PureComponent is used right?

ljharb:
@tiemevanveen that's true - making an option to `prefer-stateless-function` like "ignorePureComponents" might be worth doing, however?

tiemevanveen:
Added the option. Also moved `isPureComponent(node)` to the utils in `Components.js`.

lib/util/Components.js
```diff
@@ -200,9 +200,22 @@ function componentRule(rule, context) {
...
+        return new RegExp('^(' + pragma + '\\.)?PureComponent$').test(sourceCode.getText(node.superClass));
+      }
+      return false;
+    },
+
+    /**
      * Check if the node is returning JSX
      *
-     * @param {ASTNode} ASTnode The AST node being checked
+     * @param {ASTNode} node The AST node being checked
```
tiemevanveen: all the other methods say 'node'..

docs/rules/prefer-stateless-function.md
```diff
@@ -43,3 +44,14 @@ class Foo extends React.Component {
   }
 }
 ```
+
+The following pattern is not considered a warning:
```
ljharb: I think this documentation needs to be updated to describe the option?

lib/rules/prefer-stateless-function.js
```diff
@@ -244,6 +276,12 @@ module.exports = {
     }
 
     return {
+      ClassDeclaration: function (node) {
+        if (utils.isPureComponent(node) && ignorePureComponents === true) {
```
ljharb: I don't think we need the `=== true` :-) also we should probably check the boolean first to avoid the possibly unnecessary isPureComponent call.

tiemevanveen:
You're right, thanks. 

Fixed and added to last commit. 

Also removed the `=== true` form 'no-multi-comp' where I stole the example from in the first place ;)

docs/rules/prefer-stateless-function.md
```diff
@@ -43,3 +44,36 @@ class Foo extends React.Component {
...
+
+```js
+...
+"prefer-stateless-function": [<enabled>, { "ignorePureComponent": <boolean> }]
+...
+```
+
+### `ignorePureComponent`
+
+When `true` the rule will ignore Components extending from `React.PureComponent` that use `this.props` or `this.context`.
```
yannickcr: Can you indicate the default value here please?

tiemevanveen: Done (added to commit).

docs/rules/prefer-stateless-function.md
```diff
@@ -43,3 +44,36 @@ class Foo extends React.Component {
...
+
+The following patterns is considered okay and does not cause warnings:
+
+```js
```
lencioni: nit: please make these `jsx` instead of `js`, for best syntax highlighting

lib/util/Components.js
lencioni: This token should match the argument name, so you should update line 222 at the same time.

docs/rules/prefer-stateless-function.md
tiemevanveen: forgot to push.. done now

docs/rules/prefer-stateless-function.md
tiemevanveen: done

lib/util/Components.js
tiemevanveen: Check. And that's not possible because node is reassigned afterwards. Changed back tot original value. Thanks.

yannickcr:
Merged, thanks!

tiemevanveen:
Thanks!

docs/rules/prefer-stateless-function.md
```diff
@@ -45,13 +45,38 @@ class Foo extends React.Component {
 }
 ```
 
-The following pattern is not considered a warning:
+
+## Rule Options
 
 ```js
+...
+"prefer-stateless-function": [<enabled>, { "ignorePureComponent": <ignorePureComponent> }]
```
swarthy: ignorePureComponents not ignorePureComponent

docs/rules/prefer-stateless-function.md
```diff
@@ -45,13 +45,38 @@ class Foo extends React.Component {
...
+...
+```
+
+* `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
+* `ignorePureComponent`: optional boolean set to `true` to ignore components extending from `React.PureComponent` (default to `false`).
+
+### `ignorePureComponent`
```
swarthy: ignorePureComponents

docs/rules/prefer-stateless-function.md
```diff
@@ -45,13 +45,38 @@ class Foo extends React.Component {
...
+## Rule Options
 
 ```js
+...
+"prefer-stateless-function": [<enabled>, { "ignorePureComponent": <ignorePureComponent> }]
+...
+```
+
+* `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
+* `ignorePureComponent`: optional boolean set to `true` to ignore components extending from `React.PureComponent` (default to `false`).
```
swarthy: ignorePureComponents

ljharb: Thanks, fixed in 9f764599a537728e7ac9fa332d0ac6e3a6deb788

docs/rules/prefer-stateless-function.md
ljharb: Thanks, fixed in 9f764599a537728e7ac9fa332d0ac6e3a6deb788

docs/rules/prefer-stateless-function.md
ljharb: Thanks, fixed in 9f764599a537728e7ac9fa332d0ac6e3a6deb788

PR #788, opened 2016-08-27
improved docs for jsx-space-before-closing rule

mvolkmann:


docs/rules/jsx-space-before-closing.md
```diff
@@ -8,7 +8,7 @@ Enforce or forbid spaces before the closing bracket of self-closing JSX elements
 
 This rule checks if there is one or more spaces before the closing bracket of self-closing JSX elements.
 
-This rule takes one argument. If it is `"always"` then it warns whenever a space is missing before the closing bracket. If `"never"` then it warns if a space is present before the closing bracket. The default value of this option is `"always"`.
+This rule takes two arguments. If `[1, "always"]` then it warns whenever a space is missing before the closing bracket. If `[1, "never"]` then it warns if a space is present before the closing bracket. The default value of this option is `[1, "always"]`.
```
lencioni: I think we can avoid repeating ESLint documentation and clarify here by calling these "options" instead of "arguments", and including an example of how to configure it after the text. What do you think about that?

lencioni:
Thanks for the contribution!

docs/rules/jsx-space-before-closing.md
ljharb: I agree - let's not attempt to explain severity levels here. That knowledge is a prerequisite for using an eslint plugin, since it's required to use eslint.

PR #791, opened 2016-08-29
jsx-wrap-multilines check arrow functions without block body

### LINKED ISSUE #790: jsx-wrap-multilines to check stateless components
Given this code, check should not pass

```
const Grade = ({ grade }) =>
  <View style={styles.gradeContainer}>
    <Text style={styles.grade}>{grade}</Text>
  </View>;
```

For this code, check should pass:

```
const Grade = ({ grade }) => (
  <View style={styles.gradeContainer}>
    <Text style={styles.grade}>{grade}</Text>
  </View>
);
```

ColCh:
Aka "lambdas"

Close #790

I will duplicate issue desc here:

Given this code, check should not pass

```jsx
const Grade = ({ grade }) =>
  <View style={styles.gradeContainer}>
    <Text style={styles.grade}>{grade}</Text>
  </View>;
```

For this code, check should pass:

```jsx
const Grade = ({ grade }) => (
  <View style={styles.gradeContainer}>
    <Text style={styles.grade}>{grade}</Text>
  </View>
);
```

I leaved this behaviour by-default, but it can break backward compatibility?

May be It should be turned off by default?

### DIFF
--- a/tests/lib/rules/jsx-wrap-multilines.js
+++ b/tests/lib/rules/jsx-wrap-multilines.js
@@ -73,6 +73,18 @@ var ASSIGNMENT_NO_PAREN = '\
     <p>Hello</p>\n\
   </div>;';
 
+var ARROW_SINGLE_LINE = 'var hello = () => <p>Hello</p>;';
+
+var ARROW_PAREN = '\
+  var hello = () => (<div>\n\
+    <p>Hello</p>\n\
+  </div>);';
+
+var ARROW_NO_PAREN = '\
+  var hello = () => <div>\n\
+    <p>Hello</p>\n\
+  </div>;';
+
 // ------------------------------------------------------------------------------
 // Tests
 // ------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


yannickcr:
Thanks for this.

> I leaved this behaviour by-default, but it can break backward compatibility?
> May be It should be turned off by default?

If I'm not mistaken, according to [ESLint Semantic Versioning Policy](https://github.com/eslint/eslint#semantic-versioning-policy) this is a breaking change, so I'll wait for the next major release to merge it.

ColCh:
okay, thanks!

tests/lib/rules/jsx-wrap-multilines.js
```diff
@@ -73,6 +73,18 @@ var ASSIGNMENT_NO_PAREN = '\
...
+
+var ARROW_PAREN = '\
+  var hello = () => (<div>\n\
+    <p>Hello</p>\n\
+  </div>);';
+
+var ARROW_NO_PAREN = '\
+  var hello = () => <div>\n\
+    <p>Hello</p>\n\
+  </div>;';
```
lencioni: @ljharb After your Node 4+ PR this could get cleaned up a little.

ljharb: Good call; I'll include this in my PR.

PR #792, opened 2016-08-29
Fixes a few false positive failures in the `no-unused-prop-types` rule

EvNaverniouk:
Fixes several false positives failures in the `no-unused-prop-types` rules:

- Ensures destructured props in component lifecycle methods are properly counted towards used props 
- Ensure common `nextProps` and `prevProps` property names in lifecycle methods are correctly counted towards used props
- Fixes https://github.com/yannickcr/eslint-plugin-react/issues/801

### DIFF
--- a/tests/lib/rules/no-unused-prop-types.js
+++ b/tests/lib/rules/no-unused-prop-types.js
@@ -1208,6 +1208,180 @@ ruleTester.run('no-unused-prop-types', rule, {
       ].join('\n'),
       options: [{skipShapeProps: true}],
       parser: 'babel-eslint'
+    }, {
+      // Destructured props in componentWillReceiveProps shouldn't throw errors
+      code: [
+        'class Hello extends Component {',
+        '  static propTypes = {',
+        '    something: PropTypes.bool',
+        '  }',
+        '  componentWillReceiveProps (nextProps) {',
+        '    const {something} = nextProps;',
+        '    doSomething(something);',
+        '  }',
+        '}'
+      ].join('\n'),
+      parser: 'babel-eslint'
+    }, {
+      // Destructured function props in componentWillReceiveProps shouldn't throw errors
+      code: [
+        'class Hello extends Component {',
+        '  static propTypes = {',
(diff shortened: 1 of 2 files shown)


tests/lib/rules/no-unused-prop-types.js
```diff
@@ -1208,6 +1208,96 @@ ruleTester.run('no-unused-prop-types', rule, {
...
+    }, {
+      // Destructured props in the constructor shouldn't throw errors
+      code: [
+        'class Hello extends Component {',
+        '  static propTypes = {',
+        '    something: PropTypes.bool',
+        '  }',
+        '  constructor (props) {',
+        '    super(props);',
+        '    const {something} = props;',
```
ljharb: can we also test all these cases with the destructuring being done inline in the signature? ie, `constructor ({ something }) {`?

EvNaverniouk: Done!

ljharb:
Test cases LGTM.

cristian-sima:
Is it merged ?

yannickcr:
Merged, thanks!

quanru:
@EvNaverniouk hi, i wonder know why must i use destructured props: 
'Ensures destructured props in component lifecycle methods are properly counted towards used props'

ljharb:
You don't have to; this just ensures it works if you do.

quanru:
@ljharb if i don't do it in this way, it goes wrong. is this config correct ?
'no-unused-prop-types': [2, { customValidators: ['nextProps'], skipShapeProps: true }],

ismay:
Using nextProps with an object spread operator doesn't prevent the no-unused-proptypes error (even though all props are used), I assume that's by design? (Since it's still only a proposal)

ljharb:
The rule has no way of knowing what props you are using there because it can't know what properties are on it at runtime. Using object spread is dynamic; prefer explicit prop referencing.

ismay:
Ah yeah of course. That makes sense.

PR #796, opened 2016-08-31
Case insensitive comparisons for no-target-blank

dmnd:
`<a href="foo" target="_BLANK">` should also trigger no-target-blank.
Now it does.

### DIFF
--- a/lib/rules/jsx-no-target-blank.js
+++ b/lib/rules/jsx-no-target-blank.js
@@ -25,12 +25,12 @@ module.exports = {
           return;
         }
 
-        if (node.name.name === 'target' && node.value.value === '_blank') {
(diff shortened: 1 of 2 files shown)


ljharb:
Do browsers natively treat them as case-insensitive (and does react normalize the case)? If not, they should *not* be targeted by this rule.

dmnd:
Looks like I was overzealous with the HTML attribute name. Those seem to be case sensitive. But target="_BLANK" definitely works: http://codepen.io/dmnd/pen/LRPmZE?editors=1010

dmnd:
Amended to not include all-caps html attributes.

yannickcr:
Thanks. Can you rebase your branch ? Then I think we'll be good to go.

dmnd:
Rebased.

yannickcr:
Perfect, thanks!

PR #800, opened 2016-08-31
Add no-duplicate-props to the recommended rule set

Wilfred:
I'd like to propose adding no-duplicate-props to the recommended rules. I don't see any use case for allow duplicate props, so it seems worthwhile to me.

### DIFF
--- a/index.js
+++ b/index.js
@@ -97,6 +97,7 @@ module.exports = {
         'react/jsx-uses-vars': 2,
         'react/no-deprecated': 2,
         'react/no-direct-mutation-state': 2,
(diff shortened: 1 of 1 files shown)


Wilfred:
Oops, misread the code. Please ignore.

ljharb:
I actually agree that this rule should be part of the recommended set. This rule catches bugs often.

lencioni:
It is: https://github.com/yannickcr/eslint-plugin-react/blob/fd06c063b32f433f93a4d1fad11575deb6a69d1d/index.js#L94

ljharb:
ah, understood :-)

PR #802, opened 2016-09-01
Add default for jsx-first-prop-new-line

kokarn:
Right now just enabling the rule does nothing, you actually have to pass in a value. 
This seems very counterintuitive to me.

If you feel the default should be something else I'm fine with whatever.


lencioni:
I think this might technically be a breaking change, so I'll mark this for the v7.0.0 release.

ljharb:
@kokarn would you please check the "allow edits" checkbox, and/or rebase and update the default from "multiline" to "multiline-multiprop"?

kokarn:
Shit, sorry about that. Allowed now!

kokarn:
Thanks!

PR #813, opened 2016-09-09
Add examples of correctly specified propTypes

AndiDog:
Because examples for ES6 are not immediately clear - I tried `propTypes: { ... }` in ES6 and it didn't work unless it's `static`.

### DIFF
--- a/docs/rules/prop-types.md
+++ b/docs/rules/prop-types.md
@@ -29,6 +29,39 @@ function Hello({ name }) {
 }
 ```
 
+Examples of correct usage without warnings:
+
+```jsx
+var Hello = React.createClass({
+  propTypes: {
+    name: React.PropTypes.string.isRequired,
+  },
+  render: function() {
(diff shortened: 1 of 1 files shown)


docs/rules/prop-types.md
```diff
@@ -29,6 +29,29 @@ function Hello({ name }) {
...
+    name: React.PropTypes.string.isRequired,
+  },
+  render: function() {
+    return <div>Hello {this.props.name}</div>;
+  }
+});
+
+// Or in ES6:
+class Hello extends React.Component {
+  static propTypes = {
```
ljharb: this isn't actually ES6 - it's using a stage 2 proposal for public class properties.

The current standard JS way of doing this is `Hello.propTypes = { … }`, after the class definition.

AndiDog:
Now all three variants are documented.

ljharb:
LGTM

Issue #825, opened 2016-09-13
`no-unused-prop-types` crash when destructuring `prevState`


daltones:
If I use destructuring on the second argument of `componentDidUpdate`, like this:
```js
import React from 'react';

export default class Test extends React.Component {
  componentDidUpdate(prevProps, {state1, state2}) {
  }
}
```

It gets this error:
```
Error: Cannot read property 'length' of undefined
TypeError: Cannot read property 'length' of undefined
    at markPropTypesAsUsed ([...]/eslint-plugin-react/lib/rules/no-unused-prop-types.js:622:41)
    at [...]/eslint-plugin-react/lib/rules/no-unused-prop-types.js:893:15
    at Array.forEach (native)
    at EventEmitter.ObjectPattern ([...]/eslint-plugin-react/lib/rules/no-unused-prop-types.js:891:27)
    at emitOne (events.js:95:20)
    at EventEmitter.emit (events.js:182:7)
    at NodeEventGenerator.enterNode ([...]/eslint/lib/util/node-event-generator.js:40:22)
    at CodePathAnalyzer.enterNode ([...]/eslint/lib/code-path-analysis/code-path-analyzer.js:608:23)
    at CommentEventGenerator.enterNode ([...]/eslint/lib/util/comment-event-generator.js:97:23)
    at Controller.enter ([...]/eslint/lib/eslint.js:918:36)
    at Controller.__execute ([...]/estraverse/estraverse.js:397:31)
    at Controller.traverse ([...]/estraverse/estraverse.js:501:28)
    at Controller.Traverser.controller.traverse ([...]/eslint/lib/util/traverser.js:36:33)
    at EventEmitter.module.exports.api.verify ([...]/eslint/lib/eslint.js:915:23)
    at processText ([...]/eslint/lib/cli-engine.js:262:31)
    at CLIEngine.executeOnText ([...]/eslint/lib/cli-engine.js:749:26)
```

eslint: v3.5.0
eslint-plugin-react: v6.2.1

milesj:
Also having this problem.

```
Cannot read property 'length' of undefined
TypeError: Cannot read property 'length' of undefined
    at markPropTypesAsUsed (/Users/miles_johnson/Sites/airlab/repos/lux/node_modules/eslint-plugin-react/lib/rules/no-unused-prop-types.js:622:41)
[...]

PR #829, opened 2016-09-13
Fixes Issue #825

EvNaverniouk:
This PR fixes issue #825 by ensuring that properties are valid arrays before trying to loop through their destructured values.

### DIFF
--- a/tests/lib/rules/no-unused-prop-types.js
+++ b/tests/lib/rules/no-unused-prop-types.js
@@ -1382,6 +1382,33 @@ ruleTester.run('no-unused-prop-types', rule, {
       ].join('\n'),
       parser: 'babel-eslint',
       parserOptions: parserOptions
+    }, {
+      // Destructured state props in `componentDidUpdate` [Issue #825]
+      code: [
+        'class Hello extends Component {',
+        '  static propTypes = {',
+        '    something: PropTypes.bool',
+        '  }',
+        '  componentDidUpdate ({something}, {state1, state2}) {',
+        '    return something;',
+        '  }',
+        '}'
+      ].join('\n'),
+      parser: 'babel-eslint',
+      parserOptions: parserOptions
+    }, {
+      // Destructured state props in `componentDidUpdate` without custom parser [Issue #825]
+      code: [
+        'var Hello = React.Component({',
+        '  propTypes: {',
+        '    something: PropTypes.bool',
+        '  },',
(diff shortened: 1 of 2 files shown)


tests/lib/rules/no-unused-prop-types.js
```diff
@@ -1382,6 +1382,20 @@ ruleTester.run('no-unused-prop-types', rule, {
...
+        'class Hello extends Component {',
+        '  static propTypes = {',
+        '    something: PropTypes.bool',
+        '  }',
+        '  componentDidUpdate ({something}, {state1, state2}) {',
+        '    return something;',
+        '  }',
+        '}'
+      ].join('\n'),
+      parser: 'babel-eslint',
```
ljharb: can we also have a test case that uses the default parser (ie, not using class properties)

tests/lib/rules/no-unused-prop-types.js
```diff
@@ -2337,6 +2351,24 @@ ruleTester.run('no-unused-prop-types', rule, {
...
+        'class Hello extends Component {',
+        '  static propTypes = {',
+        '    something: PropTypes.bool',
+        '  }',
+        '  componentDidUpdate (prevProps, {state1, state2}) {',
+        '    return something;',
+        '  }',
+        '}'
+      ].join('\n'),
+      parser: 'babel-eslint',
```
ljharb: here too

tests/lib/rules/no-unused-prop-types.js
EvNaverniouk: Does the default parser support object destructuring? I can't seem to get it to work. I keep getting: `Parsing error: Unexpected token }`

ljharb: it does, but you have to specify `ecmaVersion` as `6` or higher in `parserOptions` (which should be done anyways)

EvNaverniouk: Pushed an update to include an additional test with a default parser for this.

tests/lib/rules/no-unused-prop-types.js
EvNaverniouk: Pushed an update to include an additional test with a default parser for this.

ljharb:
LGTM

yannickcr:
Merged, thanks!

PR #830, opened 2016-09-14
jsx-indent rule autofixer

eelyafi:
Due to limitations of the RuleFixer class there is no way to safely fix these two cases:
1. When a file uses spaces but options requires tab and vice-versa.
2. If a parent node didn't pass the rule, autofixer won't fix folded child nodes. Second run will actually fix the next child node.

Even with these limitations the autofixer found to be very useful and was already used on Pinterest codebase.

### DIFF
--- a/lib/rules/jsx-indent.js
+++ b/lib/rules/jsx-indent.js
@@ -39,6 +39,8 @@ module.exports = {
       category: 'Stylistic Issues',
       recommended: false
     },
+    // fixer will fix whitespace and tabs indentation.
+    // It won't replace tabs with whitespaces and vice-versa.
     fixable: 'whitespace',
     schema: [{
       oneOf: [{
@@ -69,7 +71,7 @@ module.exports = {
       }
     }
 
-    var indentChar = indentType === 'space' ? ' ' : '\u0009';
+    var indentChar = indentType === 'space' ? ' ' : '\t';
 
     /**
      * Responsible for fixing the indentation issue fix
--- a/lib/rules/jsx-indent.js
+++ b/lib/rules/jsx-indent.js
@@ -39,7 +39,7 @@ module.exports = {
       category: 'Stylistic Issues',
       recommended: false
     },
-
+    fixable: 'whitespace',
     schema: [{
       oneOf: [{
         enum: ['tab']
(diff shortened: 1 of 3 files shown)


ljharb:
Can you explain more about why you can't autofix *to* tabs?

eelyafi:
Let's say we have "\t<App />" and we want to replace "\t" with 4 spaces. In order to do so we have to run fixer.replace(). Unfortunately fixer.replace() won't let you increase number of characters. 
Basically, if there are two spaces it won't replace them with more than two characters. On the other hand fixer won't let you run more than one command. Because of that there is no way to replace spaces with empty string first and then add needed amount of spacing.

ljharb:
@eelyafi does that mean it's impossible to have a fixer for `indent` in eslint core?

eelyafi:
@ljharb, after additional debugging I found that replaceTextRange will let you replace text. The actual reason why it won't replace one type of spacing with another is in getNodeIndent function. 
The regExp in this function will look for spaces or tabs depending on the options passed in. Thus gotten variable will be always zero if file uses spaces and tabs are passed as an option.

eelyafi:
Ping :)

ljharb:
I'm very hesitant to merge an autofixer that doesn't fix all forms of the rule. If we can't support tabs, we shouldn't support spaces, and vice versa.

eelyafi:
In reality many existing autofixers are far away from perfection. This autofixer will cover most of the cases and won't break code.

lib/rules/jsx-indent.js
```diff
@@ -69,6 +69,30 @@ module.exports = {
       }
     }
 
+    var indentChar = indentType === 'space' ? ' ' : '\u0009';
```
ljharb: Can this just be `'\t'` instead of the character code?

lib/rules/jsx-indent.js
```diff
@@ -39,7 +39,7 @@ module.exports = {
       category: 'Stylistic Issues',
       recommended: false
     },
-
+    fixable: 'whitespace',
```
ljharb: can we add a comment here detailing what cases are not covered? Ideally we'd also add skipped/commented-out test cases for that future work.

lib/rules/jsx-indent.js
eelyafi: sure

eelyafi:
Addressed the comments

eelyafi:
Rebased

eelyafi:
@ljharb is it good to merge ?

ljharb:
@eelyafi I'm waiting for one of the other contributors to review it.

lencioni:
It would be good to note the limitations of the autofixer in the documentation, and also open an issue to improve it. Otherwise LGTM.

eelyafi:
@lencioni I updated docs and created an issue https://github.com/yannickcr/eslint-plugin-react/issues/865

eelyafi:
@ljharb @lencioni ping

PR #835, opened 2016-09-15
feat(no-unused-prop-types): support nextProps

yutin1987:
Sample

```
shouldComponentUpdate(nextProps: Object) {
  return !!nextProps.shouldUpdate;
}
```

### DIFF
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -14,7 +14,7 @@ var variable = require('../util/variable');
 // Constants
 // ------------------------------------------------------------------------------
 
-var DIRECT_PROPS_REGEX = /^props\s*(\.|\[)/;
+var DIRECT_PROPS_REGEX = /^(props|nextProps)\s*(\.|\[)/;
 
 // ------------------------------------------------------------------------------
 // Rule Definition
@@ -936,4 +936,3 @@ module.exports = {
     };
   })
 };
-


lib/rules/no-unused-prop-types.js
```diff
@@ -14,7 +14,7 @@ var variable = require('../util/variable');
 // Constants
 // ------------------------------------------------------------------------------
 
-var DIRECT_PROPS_REGEX = /^props\s*(\.|\[)/;
+var DIRECT_PROPS_REGEX = /^(props|nextProps)\s*(\.|\[)/;
```
lencioni: I think it might be good to only allow `nextProps` to work on the specific lifecycle methods it is provided by React. `shouldComponentUpdate`, `componentWillReceiveProps`, `componentWillUpdate`. And, it would probably be better to use positional argument detection for these, since people can name the arguments anything they like.

Also, it would be good to support `prevProps` on `componentDidUpdate` in the same way.

ljharb: They can also destructure it inline, so we should support that too.

EvNaverniouk: This is actually already done. See `isPropAttributeName()` function.

lencioni:
I'm closing this out due to inactivity. @yutin1987 please feel free to re-open.

PR #838, opened 2016-09-15
Add JSXExpressionContainer support for jsx-indent rule

eelyafi:
Right now jsx-indent linter won't catch cases like:
```
<App>
   {test}
</App>
```
This diff will add support for JSXExpressionContainer


eelyafi:
@ljharb I've added tests for tabs

lencioni:
LGTM, thanks!

PR #839, opened 2016-09-15
Handle `undefined` tag

pfhayes:
I have observed this error:

```
Cannot read property '0' of undefined
TypeError: Cannot read property '0' of undefined
    at EventEmitter.JSXAttribute (/Users/patrick/sigopt-api/node_modules/eslint-plugin-react/lib/rules/forbid-component-props.js:51:16)
    at emitOne (events.js:101:20)
    at EventEmitter.emit (events.js:188:7)
    at NodeEventGenerator.enterNode (/Users/patrick/sigopt-api/node_modules/eslint/lib/util/node-event-generator.js:40:22)
    at CodePathAnalyzer.enterNode (/Users/patrick/sigopt-api/node_modules/eslint/lib/code-path-analysis/code-path-analyzer.js:608:23)
    at CommentEventGenerator.enterNode (/Users/patrick/sigopt-api/node_modules/eslint/lib/util/comment-event-generator.js:97:23)
    at Controller.enter (/Users/patrick/sigopt-api/node_modules/eslint/lib/eslint.js:918:36)
    at Controller.__execute (/Users/patrick/sigopt-api/node_modules/eslint/node_modules/estraverse/estraverse.js:397:31)
    at Controller.traverse (/Users/patrick/sigopt-api/node_modules/eslint/node_modules/estraverse/estraverse.js:501:28)
    at Controller.Traverser.contMakefile:37: recipe for target 'eslint' failed
```

### DIFF
--- a/lib/rules/forbid-component-props.js
+++ b/lib/rules/forbid-component-props.js
@@ -48,7 +48,7 @@ module.exports = {
     return {
       JSXAttribute: function(node) {
         var tag = node.parent.name.name;
-        if (tag[0] !== tag[0].toUpperCase()) {
+        if (tag && tag[0] !== tag[0].toUpperCase()) {
           // This is a DOM node, not a Component, so exit.
           return;
         }


pfhayes:
Here is a minimal example that causes the error for me:

```jsx
const Parent = React.createClass({
  render: function() {
    return <Child a={this.props.a}/>;
  }
});
```

However, I tried adding this as a test and the test appears to pass with the existing code. I'm not sure what would cause this or how to induce this error in test-mode, but I can repeatedly generate this error in my environment.

ljharb:
Is your environment in strict or sloppy mode?

pfhayes:
How do I check that? I assume it is whatever the default is because I haven't configured anything like that to my knowledge

ljharb:
@pfhayes if you have `'use strict'`, or if you've configured eslint to use sourceType modules, then it's in strict mode. Otherwise it'd be in sloppy mode.

If you add `'use strict';` to the top of the test code, does it fail?

pfhayes:
I don't believe I have configured eslint to use sourceType modules because I do not know how to do that. I am also not using 'use strict'

ljharb:
@pfhayes are you using babel? any code inside `class`, or any file using `import` or `export` is automatically in strict mode.

pfhayes:
Yes, the code is being compiled with babel. The example provided has no imports/exports or class declarations (the example provided was the entire contents of the file that introduced the error)

I tried to reproduce this error recently and was unable to. I'm not sure what the cause was - perhaps an error in a dependent module that was fixed by an `npm install`. i'm fine closing this

EdwardDrapkin:
Here's one way to replicate it:


    foo(bar) {
        return <bar.baz className="error goes here" />;
    }

lencioni:
Thanks for the contribution. Feel free to re-open with a test!

xeodou:
Got the exactly the same issue here , seems @pfhayes 's PR been closed, but still have the issue. So how do I solve this @lencioni  ?

ljharb:
@xeodou if you can provide a test case in a PR, or better, a test case with a fix, then we'd love to merge it!

xeodou:
Sure @ljharb  i'll do.

xeodou:
Hi @ljharb  I just found the use case and create pull request, please feel free to make some comments.

PR #846, opened 2016-09-16
Allow skipping prop-types validation when not declared

pfhayes:
Adding the `prop-types` rule can be a daunting task for
projects that have not been consistent about using prop
types to begin with.

This allows users to slowly opt-in to prop-types validation
by only erroring on "incomplete" propTypes declarations.

### DIFF
--- a/lib/rules/prop-types.js
+++ b/lib/rules/prop-types.js
@@ -42,6 +42,9 @@ module.exports = {
           items: {
             type: 'string'
           }
+        },
+        skipUndeclared: {
+          type: 'boolean'
         }
       },
       additionalProperties: false
@@ -54,6 +57,7 @@ module.exports = {
     var configuration = context.options[0] || {};
     var ignored = configuration.ignore || [];
     var customValidators = configuration.customValidators || [];
+    var skipUndeclared = configuration.skipUndeclared || false;
     // Used to track the type annotations in scope.
     // Necessary because babel's scopes do not track type annotations.
     var stack = null;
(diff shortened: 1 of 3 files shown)


pfhayes:
@ljharb sure thing - let me know if this is what you had in mind

tests/lib/rules/prop-types.js
```diff
@@ -2366,6 +2386,38 @@ ruleTester.run('prop-types', rule, {
...
+      parserOptions: parserOptions,
+      errors: [{
+        message: '\'firstname\' is missing in props validation',
+        line: 2,
+        column: 22
+      }]
+    }, {
+      code: [
+        'class Hello extends React.Component {',
+        '  static get propTypes() {',
```
ljharb: actually this would be `static propTypes = () => {` - i don't think `propTypes` is expected to be anything but a data property.

pfhayes: I copied how the other static propTypes were declared in this file. upon closer inspection, this file is inconsistent in how it declares static proptypes. I see three different types:

`static get propTypes() {`
`static propTypes = {`
`static propTypes: {`

do you think I should introduce a fourth type? i'm also happy to consolidate them but I think that is outside the scope of this PR

ljharb: Yes, I'm not sure why we have the getter version, and the "fourth" type is in fact the most standard. The third type there isn't actually valid because of the `static` keyword, but I think you mean its use in `React.createClass`, which is fine.

ljharb: Let's not worry about consolidating them in this PR, but please do add a `Hello.propTypes = { … }` variant.

lib/rules/prop-types.js
```diff
@@ -179,10 +182,12 @@ module.exports = {
      * @returns {Boolean} True if the component must be validated, false if not.
      */
     function mustBeValidated(component) {
+      var isSkippedByConfig = (typeof component.declaredPropTypes === 'undefined' && skipUndeclared);
```
pfhayes: fixed!

yannickcr:
Merged, thanks!

PR #850, opened 2016-09-18
Add includeDOM option to rule "forbid-component-props".

EdwardDrapkin:
I wanted to forbid some prop types on DOM nodes as well as Components, so I added the option here.  Let me know if there's anything that needs to be done here to get this approved.  Thanks in advance!


docs/rules/forbid-component-props.md
```diff
@@ -35,10 +35,14 @@ The following patterns are not considered warnings:
...
-"forbid-component-props": [<enabled>, { "forbid": [<string>] }]
+"forbid-component-props": [<enabled>, { "forbid": [<string>], "includeDOM": [<boolean>]}]
 ...
 ```
 
 ### `forbid`
 
 An array of strings, with the names of props that are forbidden. The default value of this option is `['className', 'style']`.
+
+### `includeDOM`
```
ljharb: This seems like something that will vary per prop name. Perhaps the enhancement should be that `forbid` can have strings or objects, and if an object, it can be something like `{ "property": "foo", "includeDOM": true }`?

ljharb (changes requested):
Either way, this will need tests.

EdwardDrapkin:
How about `forbid` and `forbidOnDOM`?  `forbidOnDOM` would be an array of strings just like `forbid` except that it's empty by default.

ljharb:
I think that allowing string-or-object in the array gives far greater flexibility - that would also let me potentially forbid a prop on one list of components, but not on others.

EdwardDrapkin:
I think the flexibility is the same: 

     {
      "forbid":  ["className", "style"], 
      "forbidOnDOM": ["style"] 
     }  

would achieve the same thing, no?

I think that's easier for users to manage than:  `

       {
         "forbid": [ "style", { "className": { "forbidOnDOM": true } } ]
       }

Or
       {
         "forbid": [ "style", { "property": "className", "forbidOnDOM": true } ]
       }

It's a completely subjective decision and ultimately it's your code, so I'll do what you think is best, but IMO the first example is the easiest to read if I'm a hypothetical engineer and I'm reading someone else's .eslintrc and I'm not necessarily familiar with this particular plugin.

ljharb:
@EdwardDrapkin imagine a hypothetical future option where you could forbid "foo" on a Foo component, but allow "foo" on anything else - or where you want to forbid "style" on both React and DOM components, but simultaneously forbid "className" only on React components.

The latter certainly can be achieved with your suggestion, but that won't allow for the former, and that also requires that property names be duplicated in both lists, which can be hard to maintain.

EdwardDrapkin:
Ahhhhh, I wasn't even considering that!  You make excellent points.

Perhaps we could also take this opportunity to allow people the flexibility you imagine?  I have two ideas.

First, we could allow passing custom validator functions like:

      {
        "property": "style",
        "validator": "no-dom-components"|"only-dom-components"|"all-components"|function 
      }

where the function is just `(componentName:string) => boolean` that the plugin would use to choose whether or not to apply the filters to the component?  Alternatively, I could do `(componentName:string, property:string) => boolean` and expose the validation step itself to the user, but I'm having trouble imagining a case where that more useful than the former. 

This would be pretty simple to implement, and is probably the most flexible approach, but not exactly user friendly because it's too complex for what I believe is likely to be the common use case...

So maybe, instead we just allow a regular expression to be passed?  

      {
        "property": "style",
        "isValid": "^[A-Z]" 
      }

This is probably the easiest for users to use and covers like 99% of the cases people would want, but would require documentation to tell people to use something like `isValid: "."` to match all components' names, or a special case.

What do you think?

ljharb:
eslint configs are defined in JSON or yaml, so it'd be impossible to provide a function.

I think adding regex support is unneeded complexity - being able to configure a boolean for DOM and another for React, along with simple component names per-property that can override the booleans, seems like it would cover all the current and potential use cases pretty well.

EdwardDrapkin:
Can't you use a JS file as an .eslintrc.js?  It's irrelevant anyway...

How does this look?

     {
       "property": <string>,
       "include": "React"|"DOM"|"Both"|<array> 
     }

Is `include` a good param name?

ljharb:
I think I'd prefer:
```js
{
  "property": <string>,
  "allowOnComponents": <boolean>, // default false
  "allowOnDOM": <boolean>, // default false
  "forbid": <array of component name strings>, // overrides previous two booleans
}
```
where passing a simple string "foo" as the property name is equivalent to:
```js
{
    "property": "foo",
    "allowOnComponents": false,
    "allowOnDOM": true, // simply so it's not a breaking change
    "forbid": [],
}
```

EdwardDrapkin:
Perfect.  I'll get this PR updated sometime in the next 24-48 hours with these changes!

docs/rules/forbid-component-props.md
```diff
@@ -35,10 +37,94 @@ The following patterns are not considered warnings:
 
 ```js
 ...
-"forbid-component-props": [<enabled>, { "forbid": [<string>] }]
+"forbid-component-props": [<enabled>, { "forbid": [<string>|<object>], "defaultAllowOnComponents": <bool>, "defaultAllowOnDOM": <bool>}]
```
ljharb: the two "default" booleans need to be on the object, not on the outer config object. not sure how to denote this tho.

ljharb: unless you're suggesting that they'd be top-level settings as well? if so, i do think i like that.

EdwardDrapkin:
Okay, the football sucked, so I went ahead and got to work on this.

I did some extra work that we hadn't discussed, but I was adding examples to the documentation and surfaced some cases I wanted to address.

I added `defaultAllowOnComponents` and `defaultAllowOnDOM` as top level config options to do exactly what they say.  I also allowed a white `allow` alongside the blacklist `forbid` you asked for, because I imagined a case where someone was okay with font icon i tags having classNames but everything else going through something like css modules.

I updated the documentation and I think it's okay, but please do read and update it if you feel it's lacking.  Or tell me and I gladly will!

I also added more test cases so Istanbul reports 100% coverage for me.

Let me know if there's any issues this time.

Thanks so much for your time helping me get this working (and for the rule in the first place)!

ljharb:
I think this is looking much better, and I like the additions you've made! I think we'll want to add a very large quantity of test cases, that cover as many interesting combinations of rules as possible.

docs/rules/forbid-component-props.md
EdwardDrapkin: yeah, there's an outer config option for specifying the default behavior and an option for overloading the default, but i wasn't quite sure how to make that clear in the documentation other than in the examples...

EdwardDrapkin:
I added some more test cases, up to 21 now.  I'm having trouble coming up with cases that aren't already tested.

lib/rules/forbid-component-props.js
```diff
@@ -22,48 +31,117 @@ module.exports = {
...
-      properties: {
-        forbid: {
-          type: 'array',
-          items: {
-            type: 'string'
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          defaultAllowOnDOM: {
```
ljharb: i think the schema is missing `defaultAllowOnComponents`?

lib/rules/forbid-component-props.js
```diff
@@ -22,48 +31,117 @@ module.exports = {
...
+                }
+              ]
+            }
           }
-        }
-      },
-      additionalProperties: true
-    }]
+        },
+        additionalProperties: true
```
ljharb: i think we can maybe set this to `false`, so nobody accidentally passes an invalid config option?

lib/rules/forbid-component-props.js
```diff
@@ -22,48 +31,117 @@ module.exports = {
...
-          items: {
-            type: 'string'
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          defaultAllowOnDOM: {
+            type: 'boolean'
+          },
+          forbid: {
```
ljharb: the schema also seems to be missing `allow`, which i think should probably share a schema with `forbid`? (ie, reuse the array schema in both)

lib/rules/forbid-component-props.js
```diff
@@ -22,48 +31,117 @@ module.exports = {
...
+    }
 
-    function isForbidden(prop) {
-      var configuration = context.options[0] || {};
+    function getForbidden(input) {
+      var forbidden = {};
 
-      var forbid = configuration.forbid || DEFAULTS;
-      return forbid.indexOf(prop) >= 0;
+      for (var i = 0; i < input.length; i++) {
```
ljharb: can we use `input.forEach` here, instead of a for loop?

lib/rules/forbid-component-props.js
```diff
@@ -22,48 +31,117 @@ module.exports = {
...
+          rules[prop].allow.indexOf(tag) === -1 && (
+              rules[prop].forbid.indexOf(tag) > -1
+              || (isDOMNode(tag) && !rules[prop].allowOnDOM)
+              || (!isDOMNode(tag) && !rules[prop].allowOnComponents)
+          );
+    }
+
+    function getError(prop, tag) {
+      if (rules[prop].forbid.indexOf(tag) > -1) {
+        return 'Prop `' + prop + '` is specifically forbidden on tags named ' + tag;
```
ljharb: `'… on ' + tag + ' components'`?

tests/lib/rules/forbid-component-props.js
```diff
@@ -143,11 +196,181 @@ ruleTester.run('forbid-component-props', rule, {
...
+      {
+        message: TAG_CLASSNAME_ERROR_MESSAGE,
+        line: 4,
+        column: 36,
+        type: 'JSXAttribute'
+      }
+    ]
+  }, {
+    code: [
+      'var First = React.createClass({',
```
ljharb: for each of the tests you're using for `React.createClass`, let's please also add a test case for an SFC, and a class-based component.

ljharb (changes requested):

lib/rules/forbid-component-props.js
EdwardDrapkin: okay, I can change it... I just didn't wanna mess with what you already had

ljharb: yeah that's a good point, it might constitute a breaking change. probably best to leave it `true` for now.

lib/rules/forbid-component-props.js
EdwardDrapkin: oops is all i got

lib/rules/forbid-component-props.js
EdwardDrapkin: okay i was a little excited

lib/rules/forbid-component-props.js
EdwardDrapkin: yep, done

lib/rules/forbid-component-props.js
EdwardDrapkin: It swaps between "$tag Components" and "$tag DOM nodes" now.

lib/rules/forbid-component-props.js
```diff
@@ -38,6 +45,9 @@ module.exports = {
           defaultAllowOnDOM: {
             type: 'boolean'
           },
+          defaultAllowOnComponents: {
+            type: 'boolean'
+          },
           forbid: {
```
ljharb: is there not a top-level `allow` as well?

EdwardDrapkin: Wouldn't that mean the entire plugin would have to switch into whitelist mode?  Would that be useful to anyone?

ljharb: Hm, yeah i guess if i wanted to override a `forbid` i'd now just modify the object inside the array. This is fine as-is.

tests/lib/rules/forbid-component-props.js
EdwardDrapkin: I only added one total test case for each. I figure if one test case works for all three, all test cases should, because we're only getting at the JSX tags inside the (render) function body, right?  I can add more if you think they're necessary.

ljharb: Ah, that's another thing - we should check for `React.createElement` calls too (ie, non-JSX) :-)

EdwardDrapkin: If you want me to, that's cool and I'm down to learn how, but I don't know... so can you point me in the right direction if there's another plugin I might be able to read?  If not, I'll just RTFM I guess.

And do you have any problem if I do it in a separate pull request?  It seems outside the scope of this one and I might not have time to get to it until this upcoming weekend.

EdwardDrapkin: Hmm, you may have opened a can of worms, at first glance there are several other plugins that do things with props that rely on JSXExpression too... no-duplicate-props, no-unknown-property, sort-props, maybe more....

What is the best way to go forward if this is an issue that affects multiple rules?  Open a new issue and then read each rule one by one and make sure they're not affected?

ljharb: There are a number of rules that address both - however, I'm fine if this rule covers jsx right now, but is expanded to non-JSX in a future PR.

EdwardDrapkin:
I think I addressed everything from your notes - and I'll make a separate pull request to add `React.createElement` support  - but is there anything else that needs to be done now?

burabure:
this might clash with https://github.com/yannickcr/eslint-plugin-react/pull/936

ljharb:
@burabure i think perhaps we might want to follow this PR's lead for "forbidPatterns", and let the "dom or component" options be pattern-specific instead of rule-specific.

burabure:
@EdwardDrapkin @ljharb maybe I should take this PR as an starting point for #936 does that sound like a good idea?

ljharb:
@EdwardDrapkin would you be alright with @burabure building off your commits to achieve both this PR and #936?

burabure:
@EdwardDrapkin ping

burabure:
@ljharb I don't think @EdwardDrapkin is going to reply. Should I go ahead?

ljharb:
Sure, go for it.

EdwardDrapkin:
I'm sorry, I've been really sick lately fighting off MRSA (yikes!) and I won't have time to get to this.  You should definitely go ahead.

ljharb:
Thanks for confirming - @burabure, do your thing :-)

burabure:
@EdwardDrapkin sorry to hear that, has a friend that had MRSA and I know it really sucks. Hope you feel better soon!

burabure:
I don't have time to see this through, if anyone is up to it please do =)

PR #860, opened 2016-09-21
Update the rules of recommended config in readme

fson:
Just bringing the list up to date with what's currently in the recommended config.


yannickcr:
Thanks!

PR #862, opened 2016-09-23
Handle spread props in no-children-prop rule

randycoulman:
Nodes of type `ExperimentalSpreadProperty` don’t have a `key` property, causing the `no-children-prop` to raise a “Cannot read property ‘name’ of undefined” error.

Code such as:

```
return React.createElement(
  elementType,
  { className: classNames, ...otherProps },
  children
)
```

will trigger the error.

### DIFF
--- a/tests/lib/rules/no-children-prop.js
+++ b/tests/lib/rules/no-children-prop.js
@@ -15,6 +15,7 @@ var RuleTester = require('eslint').RuleTester;
 var parserOptions = {
   ecmaVersion: 6,
   ecmaFeatures: {
+    experimentalObjectRestSpread: true,
     jsx: true
   }
 };
@@ -170,6 +171,14 @@ ruleTester.run('no-children-prop', rule, {
     {
       code: 'React.createElement(MyComponent, {className: "class-name"}, "Children");',
       parserOptions: parserOptions
+    },
+    {
+      code: '<MyComponent className="class-name" {...props} />;',
+      parserOptions: parserOptions
+    },
+    {
+      code: 'React.createElement(MyComponent, {className: "class-name", ...props});',
+      parserOptions: parserOptions
     }
   ],
   invalid: [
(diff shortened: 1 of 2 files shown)


tests/lib/rules/no-children-prop.js
```diff
@@ -15,6 +15,7 @@ var RuleTester = require('eslint').RuleTester;
 var parserOptions = {
   ecmaVersion: 6,
   ecmaFeatures: {
+    experimentalObjectRestSpread: true,
```
ljharb: let's not necessarily enable this for every test - can we enable it just for the new tests?

randycoulman: I thought about doing that, but in looking through all of the tests in the project, there isn't a single other case where different `parserOptions` are used for different test cases.  There are eight other test cases that add `experimentalObjectRestSpread: true` to their top-level `parserOptions`, though.

That said, if you think it's important to establish a new pattern here, I'm happy to make the change.

ljharb: I'm not 100% on it either way - it just seems like it'd be useful to test the parser with common (non-experimental) options.

randycoulman: I think you've got a valid point.

But I'm pretty big on code consistency, so I wonder if it would be better to do a separate PR with that change made in all of the cases like this.

ljharb: I'm not sure why - this is the PR that adds `experimentalObjectRestSpread`. but i don't think this needs to block if other maintainers think it's fine.

randycoulman: Yes, this PR adds it here.  But there are eight other tests that include `experimentalObjectRestSpread` in their default `parserOptions` just like this one does.

If it's important to only set that option for certain test cases, then I think we should make that consistent across all of the tests that use that option.  That's why I suggested a separate PR to make that change.

ljharb: Gotcha, that's a fair point.

yannickcr:
Merged, thanks!

ljharb:
@yannickcr any update on a release?

yannickcr:
@ljharb sorry, I was pretty busy last days but I'll try to craft a release this week-end.

PR #876, opened 2016-10-02
Adds missing rule name for `jsx-no-bind` documentation.

manovotny:
The rule name is missing from the `jsx-no-bind` title / header.

This PR adds it, like all the other rules.


ljharb:
Thanks!

PR #877, opened 2016-10-02
Add ES6 import example to "jsx in scope" docs

gaearon:
This way beginners using ES6 today won’t be confused about using something called "require" which they might be seeing for the first time.


yannickcr:
Thanks

PR #882, opened 2016-10-04
Jsx max props per line updates

kentor:
addresses parts of https://github.com/yannickcr/eslint-plugin-react/issues/878

will happily change option names

### DIFF
--- a/docs/rules/jsx-max-props-per-line.md
+++ b/docs/rules/jsx-max-props-per-line.md
@@ -4,12 +4,16 @@ Limiting the maximum of props on a single line can improve readability.
 
 ## Rule Details
 
-This rule checks all JSX elements and verifies that the number of props per line do not exceed the maximum allowed. A spread attribute counts as one prop. This rule is off by default and when on the default maximum of props on one line is `1`.
+This rule checks all JSX elements and verifies that the number of props per line do not exceed the maximum allowed. Props are considered to be in a new line if there is a line break between the start of the prop and the end of the previous prop. A spread attribute counts as one prop. This rule is off by default and when on the default maximum of props on one line is `1`.
 
 The following patterns are considered warnings:
 
 ```jsx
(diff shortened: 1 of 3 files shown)


tests/lib/rules/jsx-max-props-per-line.js
```diff
@@ -89,5 +100,24 @@ ruleTester.run('jsx-max-props-per-line', rule, {
     ].join('\n'),
     errors: [{message: 'Prop `this.props` must be placed on a new line'}],
     parserOptions: parserOptions
+  }, {
+    code: [
+      '<App',
+      '  foo={{',
```
ljharb: Let's also include some tests where the tagname and first prop are on the same line, but there's another prop. Also, please add variants of all tests that include spread props

kentor: sure thing

revisionfour:
I think this is ready to be merged and should be merged.

docs/rules/jsx-max-props-per-line.md
```diff
@@ -56,6 +60,28 @@ The following patterns are not considered warnings:
...
 
+### `when`
+
+Possible values:
+- `always` (default) - Always check for max props per line.
+- `multiline` - Only check for max props per line when jsx tag spans multiple lines.
+
+The following patterns are considered warnings:
+```jsx
+// [1, {when: always}]
```
lencioni: Should this be

```js
// [1, {when: 'always'}]
```

kentor: Not sure. It's valid if we're talking yaml, but not if it's json/js format.

ljharb: I think json format is the most common by a massive amount, so that's the format we should stick to in docs.

princed:
Any chance to get this merged?

PR #883, opened 2016-10-04
jsx-first-prop-new-line: add multiline-multiprop option

kentor:
addresses another part of https://github.com/yannickcr/eslint-plugin-react/issues/878

### DIFF
--- a/docs/rules/jsx-first-prop-new-line.md
+++ b/docs/rules/jsx-first-prop-new-line.md
@@ -7,7 +7,8 @@ Ensure correct position of the first property.
 This rule checks whether the first property of all JSX elements is correctly placed. There are three possible configurations:
 * `always`: The first property should always be placed on a new line.
 * `never` : The first property should never be placed on a new line, e.g. should always be on the same line as the Component opening tag.
-* `multiline`: The first property should always be placed on a new line when the properties are spread over multiple lines.
+* `multiline`: The first property should always be placed on a new line when the JSX tag takes up multiple lines.
+* `multiline-multiprop`: The first property should always be placed on a new line if the JSX tag takes up multiple lines and there are multiple properties.
 
 The following patterns are considered warnings when configured `"always"`:
 
@@ -58,6 +59,11 @@ The following patterns are considered warnings when configured `"multiline"`:
     prop />
 ```
 
+```jsx
+<Hello foo={{
+}} />
+```
+
 The following patterns are not considered warnings when configured `"multiline"`:
 
 ```js
@@ -66,6 +72,27 @@ The following patterns are not considered warnings when configured `"multiline"`
 <Hello
     personal={true}
     foo="bar"
+/>
+```
+
+The following patterns are considered warnings when configured `"multiline-multiprop"`:
+
+```jsx
+<Hello foo={{
+    }}
+    bar />
+```
+
+The following patterns are not considered warnings when configured `"multiline-multiprop"`:
+
+```jsx
+<Hello foo={{
+}} />
+
+<Hello
+    foo={{
+    }}
+    bar
 />
 ```
 
--- a/docs/rules/jsx-first-prop-new-line.md
+++ b/docs/rules/jsx-first-prop-new-line.md
@@ -12,7 +12,7 @@ This rule checks whether the first property of all JSX elements is correctly pla
 
 The following patterns are considered warnings when configured `"always"`:
 
-```js
+```jsx
 <Hello personal={true} />
 
 <Hello personal={true}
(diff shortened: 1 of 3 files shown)


docs/rules/jsx-first-prop-new-line.md
```diff
@@ -69,6 +75,27 @@ The following patterns are not considered warnings when configured `"multiline"`
...
+The following patterns are not considered warnings when configured `"multiline-multiprop"`:
+
+```js
+<Hello foo={{
+}} />
+
+<Hello
+    foo={{
+    }}
+    bar />
```
lencioni: I think you have a typo here

docs/rules/jsx-first-prop-new-line.md
```diff
@@ -7,7 +7,8 @@ Ensure correct position of the first property.
 This rule checks whether the first property of all JSX elements is correctly placed. There are three possible configurations:
 * `always`: The first property should always be placed on a new line.
 * `never` : The first property should never be placed on a new line, e.g. should always be on the same line as the Component opening tag.
-* `multiline`: The first property should always be placed on a new line when the properties are spread over multiple lines.
+* `multiline`: The first property should always be placed on a new line when the jsx tag takes up multiple lines.
```
lencioni: JSX should be capitalized here and the next line

docs/rules/jsx-first-prop-new-line.md
```diff
@@ -58,6 +59,11 @@ The following patterns are considered warnings when configured `"multiline"`:
     prop />
 ```
 
+```js
```
lencioni: It would be ideal if these used jsx for syntax highlighting instead of just js

tests/lib/rules/jsx-first-prop-new-line.js
```diff
@@ -144,6 +177,24 @@ ruleTester.run('jsx-first-prop-new-line', rule, {
...
+      parser: parserOptions
+    },
+    {
+      code: [
+        '<Foo bar={{',
+        '}} baz />'
+      ].join('\n'),
+      options: ['multiline-multiprop'],
+      errors: [{message: 'Property should be placed on a new line'}],
+      parser: parserOptions
```
lencioni: Thanks for adding tests!

docs/rules/jsx-first-prop-new-line.md
ljharb: Is this not just a `bar` prop?

lencioni: It is, but on the next line we see `/>` again.

docs/rules/jsx-first-prop-new-line.md
```diff
@@ -69,6 +75,27 @@ The following patterns are not considered warnings when configured `"multiline"`
...
+```jsx
+<Hello foo={{
+    }}
+    bar />
+```
+
+The following patterns are not considered warnings when configured `"multiline-multiprop"`:
+
+```jsx
+<Hello foo={{
```
ljharb: wait, why is this not a warning? The JSX tag is spanning multiple lines, so it should look like:
```jsx
<Hello
  foo={{
    …
  }}
/>
```

tests/lib/rules/jsx-first-prop-new-line.js
```diff
@@ -91,6 +91,39 @@ ruleTester.run('jsx-first-prop-new-line', rule, {
...
+    {
+      code: [
+        '<Foo bar baz />'
+      ].join('\n'),
+      options: ['multiline-multiprop'],
+      parser: parserOptions
+    },
+    {
+      code: [
+        '<Foo prop={{',
```
ljharb: similarly, this should be invalid.

just like an object literal, or an array literal, if 100% of the expression can't fit on one line, then every item needs to appear on a line by itself.

docs/rules/jsx-first-prop-new-line.md
kentor: That would be handled by the existing `multiline` option. This is the `multiline-multiprop` option which does not warn because there is only 1 prop. 

I wanted this option to allow something like this:
```jsx
<Hello style={{
  'background': 'black',
}} />
```

but not
```jsx
<Hello style={{
  'background': 'black',
}} secondProp />
```

docs/rules/jsx-first-prop-new-line.md
kentor: ok sure

docs/rules/jsx-first-prop-new-line.md
ljharb: hmm, i misunderstood the purpose of this option - it seems like Airbnb's config already matches `multiline`, and we wouldn't use this particular option.

This thus seems like it does match what you're looking for.

MartinDoyleUK:
Would love this to happen. The current multiline looks very messy with multiline attribute values. 👍

PR #890, opened 2016-10-06
Add forbid-elements rule

kentor:
May specify a list of forbidden elements and their desired replacement
elements.

addresses https://github.com/yannickcr/eslint-plugin-react/issues/887

### DIFF
--- a/docs/rules/forbid-elements.md
+++ b/docs/rules/forbid-elements.md
@@ -1,0 +1,58 @@
+# Forbid certain elements (forbid-elements)
+
+You may want to forbid usage of certain elements in favor of others, (e.g. forbid all `<div />` and use `<Box />` instead). This rule allows you to configure a list of forbidden elements and to specify their desired replacements.
+
+## Rule Details
+
+This rule checks all JSX elements and `React.createElement` calls and verifies that no forbidden elements are used. This rule is off by default. If on, no elements are forbidden by default.
+
+## Rule Options
+
+```js
+...
+"forbid-elements": [<enabled>, { "forbid": [<string|object>] }]
+...
+```
+
+### `forbid`
+
+An array of strings and/or objects. An object in this array may have the following properties: 
+
+* `element` (required): the name of the forbidden element (e.g. `'button'`, `'Modal'`)
+* `message`: additional message that gets reported
+
+A string item in the array is a shorthand for `{ element: string }`. 
+
+The following patterns are not considered warnings:
+
+```jsx
+// [1, { "forbid": ["button"] }]
+<Button />
+
+// [1, { "forbid": [{ "element": "button" }] }]
+<Button />
+```
+
+The following patterns are considered warnings:
+
+```jsx
+// [1, { "forbid": ["button"] }]
+<button />
+React.createElement('button');
+
+// [1, { "forbid": ["Modal"] }]
+<Modal />
+React.createElement(Modal);
+
+// [1, { "forbid": ["Namespaced.Element"] }]
+<Namespaced.Element />
+React.createElement(Namespaced.Element);
+
+// [1, { "forbid": [{ "element": "button", "message": "use <Button> instead" }, "input"] }]
+<div><button /><input /></div>
+React.createElement('div', {}, React.createElemet('button', {}, React.createElement('input')));
+```
+
+## When not to use
+
+If you don't want to forbid any elements.
--- a/lib/rules/forbid-elements.js
+++ b/lib/rules/forbid-elements.js
@@ -1,0 +1,112 @@
+/**
+ * @fileoverview Forbid certain elements
+ * @author Kenneth Chung
+ */
+'use strict';
+
+var has = require('has');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Forbid certain elements',
+      category: 'Best Practices',
+      recommended: false
+    },
+
+    schema: [{
+      type: 'object',
+      properties: {
+        forbid: {
+          type: 'array',
+          items: {
+            anyOf: [
+              {type: 'string'},
+              {
+                type: 'object',
+                properties: {
+                  element: {type: 'string'},
+                  message: {type: 'string'}
+                },
+                required: ['element'],
+                additionalProperties: false
+              }
+            ]
+          }
+        }
+      },
+      additionalProperties: false
+    }]
+  },
+
+  create: function(context) {
+    var sourceCode = context.getSourceCode();
+    var configuration = context.options[0] || {};
+    var forbidConfiguration = configuration.forbid || [];
+
+    var indexedForbidConfigs = {};
+
+    forbidConfiguration.forEach(function(item) {
+      if (typeof item === 'string') {
+        indexedForbidConfigs[item] = {element: item};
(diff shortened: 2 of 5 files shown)


docs/rules/forbid-elements.md
```diff
@@ -0,0 +1,39 @@
+# Forbid certain elements (forbid-elements)
+
+You may want to forbid usage of certain elements in favor of others, (e.g. forbid all `<div />` and use `<Box />` instead). This rule allows you to configure a list of forbidden elements and to specify their desired replacements.
+
+## Rule Details
+
+This rule checks all JSX components and verifies that no forbidden elements are used. When it detects a forbidden element, the error message will contain the specified elements that you should use instead. This rule is off by default. If on, no elements are forbidden by default.
```
ljharb: this rule should also check `React.createElement` usages as well.

docs/rules/forbid-elements.md
```diff
@@ -0,0 +1,39 @@
...
+The following patterns are considered warnings:
+
+```jsx
+// [1, {forbid: ['button']}]
+<button />
+
+// [1, {forbid: ['button', 'input']}]
+<div><button /><input /></div>
+
+// [1, {forbid: [{button: 'Button']}]
```
ljharb: i think maybe a better object schema would be:
`{ component: 'button', replacement: 'Button', message: 'hey yo buttons suck, use Button' }` - where "replacement" and "message" would both be optional.

That way you can use the object form for everything, and the string form is just a convenience.

docs/rules/forbid-elements.md
```diff
@@ -0,0 +1,39 @@
...
+<button />
+
+// [1, {forbid: [{div: ['Box', 'View']}, 'button']]
```
ljharb: I'm actually kind of confused what the point is of replacement elements is, unless it's to autofix or populate the message - but "message" would handle the latter, and an autofixer would need only a single replacement to be able to work. So, why multiple?

kentor: I don't personally have a use case of multiple replacements, but yeah your custom message suggestion would make specifying multiple replacements obsolete.

docs/rules/forbid-elements.md
kentor: I like this suggestion. The only thing is I don't want to use the word `component` because to me that means a  react component which excludes simple DOM nodes. I like to use `element` because the it aligns with `React.createElement()`

docs/rules/forbid-elements.md
kentor: Sounds reasonable.

docs/rules/forbid-elements.md
ljharb: oh sure, "element" makes more sense

docs/rules/forbid-elements.md
kentor: If we specify `forbid: ['Modal']`, then for which of these do we warn?

```js
React.createElement('Modal');
React.createElement(Modal);
```

or something crazy like with `forbid: ['button']`:

```js
var button = 'div';
React.createElement(button);
React.createElement('button');
```

i'm guessing we should follow the same rules as JSX when translating names into either strings or identifiers

ljharb: `<Modal />` is the same as `React.createElement(Modal)`, and I'd expect the same setting to forbid both.

I'm not sure about the latter example.

docs/rules/forbid-elements.md
```diff
@@ -0,0 +1,49 @@
...
+
+An array of strings, objects, or a mixture of string and objects. If an item is a string, then an element matching the string will be forbidden and no replacement will be suggested. If an item is an object, then the keys will be the forbidden element and their values can be either a string or an array of strings of desired replacement elements.
+
+The following patterns are not considered warnings:
+
+```jsx
+// [1, {forbid: ['button']}]
+<Button />
+
+// [1, {forbid: [{button: 'Button']}]
```
lencioni: Codebases may have multiple components with the same name. Would it be more useful to be able to specify a path to a component instead of or in addition to just a component name?

lib/rules/forbid-elements.js
```diff
@@ -0,0 +1,109 @@
...
+      if (replacements.length === 1) {
+        return replacements[0];
+      }
+
+      if (replacements.length === 2) {
+        return replacements.join(' or ');
+      }
+
+      var last = replacements.pop();
+      return replacements.join(', ') + ', or ' + last;
```
lencioni: Thanks for including the Oxford comma! A+++

docs/rules/forbid-elements.md
kentor: If we consider `React.createElement` calls that can be mapped back to JSX as "valid", then  `React.createElement('Modal')`, and `React.createElement(button)` would not be valid.

`<Modal />` translates to `React.createClass(Modal)`, and there's no way I know that we can get it to translate to `React.createClass('Modal')`.

That's why I think this rule should not bother to make checks that cannot be mapped back to JSX, and can be left to a different rule.

kentor:
@ljharb I've made the suggested schema changes. I've left out the auto fixing capability for now and I could add it in later.

docs/rules/forbid-elements.md
ljharb: I'm not sure we're on the same page here - `<Modal />` is identical to `React.createElement(Modal)`. `React.createElement('Modal')` is the same as `<{'Modal'} />` which is simply invalid react/jsx.

In other words, if `Button` is forbidden, i could still do `const NotButton = Button; <NotButton />` and work around this rule.

kentor: Sorry, made some typos earlier. I guess I am arguing for two things:

1) If 'button' is forbidden, I don't think this rule should stop someone from doing `const button = 'div'; React.createElement(button)` because that looks like that's done for a reason other than to circumvent the rule (otherwise they would just use `eslint-ignore`).

2) In general:
* if `'button'` is forbidden, then the rule should check for `React.createElement('button')` and ignore  `React.createElement(button)`
* if `'Modal'` is forbidden, then the rule should check for `React.createElement(Modal)` and ignore `React.createElement('Modal')`.
* if `'a.b'` is forbidden, then the rule should check for `React.createElement(a.b)` and ignore `React.createElement('a.b')`.

because this aligns with how React translates JSX element names into createElement calls

ljharb: I agree with your summary (although I'm pretty sure `React.createElement` won't work at all with strings that aren't DOM components)

kentor:
I'm no longer interested in having this merged. If no one else wants this rule it's fine to close it imo.

ljharb:
@kentor sorry this fell through the cracks - I still think it's a great rule to add.

TheXardas:
Will it work for named imports like:
```js
import MyBlock from './Block.jsx';

    render() {
        return <MyBlock />;
    }
}
```
with forbidden `Block` element?

ljharb:
@TheXardas no, there is no block element in that file.

TheXardas:
But there is!
It is aliased as MyBlock, but it's still a Block component.

ljharb:
@TheXardas right but there's no way static analysis can determine that within that file. The only way a linter rule can determine what kind of component a thing is is its name in the file. If someone intentionally names it MyBlock to get around the prohibition, there's nothing a linter rule could do about it (just like if they used an eslint override comment)

Issue #540, opened 2016-04-11
jsx-indent behavior with { bool && jsx } syntax


MoOx:
I just upgraded to v4.3 and got issue on this code

```js
        {
          head.title &&
          <h1>
            { head.title }
          </h1>
        }
```

```
  63:11  error  Expected indentation of 12 space characters but found 10  react/jsx-indent
```

But the change is weird:

```js
        {
          head.title &&
            <h1>
              { head.title }
            </h1>
        }
```

Is this on purpose? I don't see any change in the changelog about that. I also added some other rules it should not interfere right?

yannickcr:
`jsx-indent` was not updated since `v4.0.0` so I don't know what happened here.

Can you give me your eslint, babel-eslint and eslint-plugin-react versions before and after the upgrade?

MoOx:
Ok nevermind, I wasn't using jsx-indent before... (Sorry about the false alarm).

So my issue would be "Is there a way to disable jsx-indent for this use case?" (Renaming accordingly).

ljharb:
@MoOx whether you surround the `<h1 />` in parens or not (as other rules/guides require), semantically it's a nested expression, and as such, it should be indented.

MoOx:
Sorry but I don't really get what you are explaining.

Why _jsx_-indent is warning me about non jsx indentation? I would expect that jsx-indent rule only warn for indentation _inside_ jsx nodes, like [it's explained in the doc](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-indent.md).


Btw, this seems consistent

```js
{
  condition1 &&
  condition2
}
```

But this doesn't
```js
{
  condition1 &&
    condition2
}
```

I don't understand why the first JSX level need to be indented, seems like a bug to me.

ljharb:
OK so, `jsx-indent` handles indentation for everything inside JSX, which includes this JS - namely because the normal `indent` rule won't handle it.

I agree with your "condition1" and "condition2" analogy - but that's not the same thing. Imagine that instead of "condition2", you had a multiline array:
```js
// bad
{
  condition1 && [
  1,
  2,
  ]
}

// good
{
  condition1 && [
    1,
    2,
  ]
}
```

The reason this is the right analogy is because every JSX expression has implicit parens around them - the fact that they're optional to type, like semicolons, doesn't mean they aren't always semantically present (like semicolons). In other words, your original example, with parens added:
```js
// bad
        {
          head.title && (
[...]

PR #896, opened 2016-10-09
failing test for jsx-indent without parens

lukekarrys:
Starting in 6.4.0 I got some new errors from `jsx-indent` in some of my projects.

Here's a simple example that I added as a failing test:

```js
{head.title &&
  <h1>
    {head.title}
  </h1>
}
```

I could wrap the `<h1>` with parens and then it works (I added a test for that also), but I'm wondering if it could be made to pass without parens. Or there's probably a good reason why it shouldn't pass that I'm missing 😄 . If this should be ok, I'd be more than willing to work on a fix for this.

### DIFF
--- a/tests/lib/rules/jsx-indent.js
+++ b/tests/lib/rules/jsx-indent.js
@@ -167,6 +167,26 @@ ruleTester.run('jsx-indent', rule, {
     ].join('\n'),
     parserOptions: parserOptions,
     options: [2]
+  }, {
+    code: [
+      '{head.title && (',
+      '  <h1>',
+      '    {head.title}',
+      '  </h1>',
+      ')}'
+    ].join('\n'),
+    parserOptions: parserOptions,
+    options: [2]
+  }, {
+    code: [
+      '{head.title &&',
+      '  <h1>',
+      '    {head.title}',
+      '  </h1>',
+      '}'
+    ].join('\n'),
+    parserOptions: parserOptions,
+    options: [2]
   }],
 
   invalid: [{


yannickcr:
Hi, thanks for the failing test, it was a big help 😃 

The `jsx-indent` rule was previously asking you to indent like this but everyone was not really happy with this behavior. If you follow the discussion in #540 you'll see that there is not really a "right" way to indent this case. 

I fixed the problem in https://github.com/yannickcr/eslint-plugin-react/commit/963267ad9108c46c76818f28003315a7bf24d589 but accidentally disallowed this style at the same time 😞 

I made a fix https://github.com/yannickcr/eslint-plugin-react/commit/cc84a4629a19d49700dfae74757a4ac5263580a9 and it should now allow the following cases:

```jsx
// No parens, same indentation
{
  head.title &&
  <h1>
    {head.title}
  </h1>
}

// No parens, one indentation deeper
{
  head.title && 
    <h1>
      {head.title}
    </h1>
}

// Parens, same indentation
{
  head.title && (
  <h1>
    {head.title}
  </h1>)
}

// Parens, one indentation deeper
{
  head.title && (
    <h1>
      {head.title}
    </h1>
  )
}
```
I'll publish `v6.4.1` with the fix shortly.

ljharb:
@yannickcr does it allow all those at once, or can i force one? For my codebases I'd want to force only the last option

yannickcr:
@ljharb it allow all those at once. I tried to be pretty flexible on this since ESLint is pretty loose on `LogicalExpression` too. But I am not against adding an option to enforce one style in particular.

ljharb:
Awesome, I'd love to add that option :-)

PR #905, opened 2016-10-12
Valid self reference jsx element.

xeodou:
Please check the information here https://github.com/yannickcr/eslint-plugin-react/pull/839 

I'm trying to include [xo](https://github.com/sindresorhus/xo) in one of our projects, i hit the issue exactly the same as @pfhayes . And after dig into the codebase, i found we have a self reference linked jsx element and that cause the issue, Like this 
```
class Foo extends React.createClass {
   render() {
       return <this.Bar  className='demo' />
   }
}
```

That's cause the issue. I don't think it's a best practice, but at least it shouldn't crash the workflow.

### DIFF
--- a/tests/lib/rules/forbid-component-props.js
+++ b/tests/lib/rules/forbid-component-props.js
@@ -81,6 +81,17 @@ ruleTester.run('forbid-component-props', rule, {
     ].join('\n'),
     options: [{forbid: ['style', 'foo']}],
     parserOptions: parserOptions
+  }, {
+    code: [
+      'var First = React.createClass({',
+      '  propTypes: externalPropTypes,',
+      '  render: function() {',
+      '    return <this.Foo className="bar" />;',
+      '  }',
+      '});'
+    ].join('\n'),
+    options: [{forbid: ['style', 'foo']}],
+    parserOptions: parserOptions
   }],
 
   invalid: [{
--- a/tests/lib/rules/forbid-component-props.js
+++ b/tests/lib/rules/forbid-component-props.js
@@ -101,6 +101,13 @@ ruleTester.run('forbid-component-props', rule, {
     ].join('\n'),
     options: [{forbid: ['style']}],
     parserOptions: parserOptions
+  }, {
+    code: [
+      'const First = (props) => (',
+      '  <this.Foo {...props} />',
+      ');'
+    ].join('\n'),
+    parserOptions: parserOptions
   }],
 
   invalid: [{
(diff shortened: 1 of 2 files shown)


tests/lib/rules/forbid-component-props.js
```diff
@@ -81,6 +81,17 @@ ruleTester.run('forbid-component-props', rule, {
     ].join('\n'),
     options: [{forbid: ['style', 'foo']}],
     parserOptions: parserOptions
+  }, {
+    code: [
+      'var First = React.createClass({',
```
ljharb: could you also add examples with an SFC, and with a `class`-based component?

tests/lib/rules/forbid-component-props.js
```diff
@@ -81,6 +81,17 @@ ruleTester.run('forbid-component-props', rule, {
...
+      '  propTypes: externalPropTypes,',
+      '  render: function() {',
+      '    return <this.Foo className="bar" />;',
+      '  }',
+      '});'
+    ].join('\n'),
+    options: [{forbid: ['style', 'foo']}],
```
ljharb: hmm - why would "foo" lowercase catch "Foo", which is a distinct property in JS?

xeodou: Let me check the test.

tests/lib/rules/forbid-component-props.js
xeodou: Sure.

xeodou: By the way, SFC is short for self-closing component right ?

ljharb: no - it's for "stateless functional component" - ie, a function that takes props and returns JSX

tests/lib/rules/forbid-component-props.js
xeodou: I think i just forget remove the line `options: [{forbid: ['style', 'foo']}]`, it should related to `forbid` option.

xeodou:
hi @ljharb  I just made some changes, Are there what you expect ?

tests/lib/rules/forbid-component-props.js
xeodou: ah, Thanks. I know 'stateless functional component' , just have no idea what does SFC meaning. Thanks for point out

EvNaverniouk:
LGTM

xeodou:
Cool. Thanks @ljharb and @EvNaverniouk

xeodou:
Hi @ljharb   When can you merge the pull request ?

xeodou:
Hi @ljharb @EvNaverniouk , Do we have any other issue here for merge the PR ?

PR #906, opened 2016-10-12
added option to automatically search for imports and add them

fab1an:
This uses the `find` command to automatically find and insert missing JSX imports.

Config: `"react/jsx-no-undef": [2, "src"]`
Result when missing `MyComponent`: `import MyComponent from "some/folder/MyComponent";`

When it doesn't find a file named like the Component it just writes `import MyComponent from ;`

### DIFF
--- a/lib/rules/jsx-no-undef.js
+++ b/lib/rules/jsx-no-undef.js
@@ -26,6 +26,7 @@ module.exports = {
       category: 'Possible Errors',
       recommended: true
     },
+    fixable: 'code',
     schema: []
   },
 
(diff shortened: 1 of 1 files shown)


lencioni:
You might want to check out [ImportJS](https://github.com/galooshi/import-js), which is an editor plugin built to do this type of thing.

fab1an:
@lencioni Doesn't support my editor TextMate, unfortunately.

fab1an:
There is already a plugin doing auto-import, but it doesn't work with JSX: https://github.com/kriszyp/eslint-plugin-auto-import

lencioni:
It might not be too hard to add an integration for TextMate. I'd be happy to point you in the right direction and review it if you want to go down that route.

fab1an:
@lencioni Thank you, but not necessary. I've written my share of TextMate plugins and this fix works for me. Maybe the functionality would be better added in the `auto-import` plugin anyway, so I've opened an issue there: https://github.com/kriszyp/eslint-plugin-auto-import/issues/2

PR #909, opened 2016-10-12
Support tab indentation

arperry:
The autofix for this rule substitutes spaces even when the code is normally indented with tabs. This change accommodates tabs, spaces, and even mixed whitespace (we won't judge).


arperry:
Tests added.

lencioni:
Thanks!

PR #911, opened 2016-10-14
Mark deprecated rules as deprecated in their metadata

randycoulman:
Over on [eslint-find-rules](https://github.com/sarbbottam/eslint-find-rules), there has been some discussion about being able to detect deprecated rules in order to [not report them as unused](https://github.com/sarbbottam/eslint-find-rules/issues/172) and to [report when they are still being used](https://github.com/sarbbottam/eslint-find-rules/issues/188).

In order to implement those features, there needs to be a way to detect that a rule has been deprecated.  Core ESLint rules that are deprecated are marked as such in their metadata.

This PR adds the same `deprecated` flag to the metadata of the deprecated rules in this plugin.

It might be possible to make use of this metadata in `index.js` to automatically determine which rules are deprecated, rather than duplicating the information in two places.  However, my attempts at this were not that clean and didn’t seem worth it.  I could take another look if you think it worth it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,12 +1,6 @@
 'use strict';
 
-var deprecatedRules = {
-  'no-comment-textnodes': require('./lib/rules/no-comment-textnodes'),
-  'require-extension': require('./lib/rules/require-extension'),
-  'wrap-multilines': require('./lib/rules/wrap-multilines')
-};
-
-var rules = {
+var allRules = {
   'jsx-uses-react': require('./lib/rules/jsx-uses-react'),
   'no-multi-comp': require('./lib/rules/no-multi-comp'),
   'prop-types': require('./lib/rules/prop-types'),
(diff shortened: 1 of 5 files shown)


tests/index.js
```diff
@@ -23,6 +23,20 @@ describe('all rule files should be exported by the plugin', function() {
   });
 });
 
+describe('deprecated rules', function() {
+  it('marks all deprecated rules as deprecated', function() {
```
ljharb: I wonder if we should just use the `deprecated` property to dynamically build up the `deprecatedRules` list?

ljharb: lol i see this in your OP now. yeah i think it might be worth it - `getAllRules().filter(x => x.deprecated)` seems like it should be a viable approach

randycoulman: @ljharb OK, here's what I came up with: 33aa4d5.  Let me know what you think.

index.js
```diff
@@ -58,32 +52,45 @@ var rules = {
...
+  return result;
 }
 
+var activeRules = filterRules(allRules, function(rule) {
+  return !rule.meta.deprecated;
+});
+var activeRulesConfig = configureAsError(activeRules);
+
+var deprecatedRules = filterRules(allRules, function(rule) {
+  return rule.meta.deprecated;
```
ljharb: this seems like the `react/` prefix won't actually be there for the deprecated rules?

ljharb: That seems odd to me but I believe you that the behavior won't be changed by this PR - so while I think all the exported rule name should have the prefix, that clearly is out of scope of this PR. Thanks for clarifying!

ljharb:
This LGTM. Let's get multiple collabs' eyes on it before this gets merged, however.

index.js
ljharb: hm, that's a good point. thanks

yannickcr:
Merged, thanks!

PR #914, opened 2016-10-17
fix #884

### LINKED ISSUE #884: no-unused-prop-types - cannot detect usage within componentWillReceiveProps
I have a component where I'm only referencing a prop within `componentWillReceiveProps(nextProps)` and `react/no-unused-prop-types` is not picking it up.
It's probably a bad practice that I need to revisit on my end, but wondering if you guys think it's something the rule should handle.

marcgrr:
The most straightforward way of fixing #884 would be to add a third condition to `isPropTypesUsage` that checks if it's a `nextProps` or `prevProps` in a lifecycle method. If I did this, then I'd also have to add `nextProps` and `prevProps` to `DIRECT_PROPS_REGEX`. I didn't want to do that because it duplicates logic between `DIRECT_PROPS_REGEX` and `isPropAttributeName`.

So this PR removes the need for `DIRECT_PROPS_REGEX`.

The main idea is that before this PR, the `MemberExpression` callback caught direct access (`props.foo`) and ThisExpression access (`this.props.foo`) at different "levels" in the AST. This meant that `getPropertyName` had to have some logic to fix them up to be on the same level. I added a `Identifer` callback that catches direct access at the same level as `MemberExpression` catches ThisExpression access so that `getPropertyName` no longer needs that logic.

This change uncovered something that I think is a bug in Components.js that made one of the existing no-unused-prop-types tests start failing, so I fixed that too.

### DIFF
--- a/tests/lib/rules/prop-types.js
+++ b/tests/lib/rules/prop-types.js
@@ -1277,6 +1277,9 @@ ruleTester.run('prop-types', rule, {
         '  return names.map((name) => {',
         '    return <div>{name}</div>;',
         '  });',
+        '}',
+        'Hello.propTypes = {',
+        '  names: React.PropTypes.string',
         '}'
       ].join('\n'),
       parser: 'babel-eslint'
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -11,12 +11,6 @@ var Components = require('../util/Components');
 var variable = require('../util/variable');
 
 // ------------------------------------------------------------------------------
-// Constants
-// ------------------------------------------------------------------------------
-
-var DIRECT_PROPS_REGEX = /^props\s*(\.|\[)/;
-
-// ------------------------------------------------------------------------------
 // Rule Definition
 // ------------------------------------------------------------------------------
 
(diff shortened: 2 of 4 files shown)


tests/lib/rules/prop-types.js
```diff
@@ -1277,6 +1277,9 @@ ruleTester.run('prop-types', rule, {
         '  return names.map((name) => {',
         '    return <div>{name}</div>;',
         '  });',
+        '}',
+        'Hello.propTypes = {',
+        '  names: React.PropTypes.string',
```
yannickcr: Here `Hello` is not a component (it returns an Array, so no `propTypes` should be specified)

modosc:
any update here? would be happy to help out if needed.

marcgrr:
Sorry, I never found the time to address the change requests and to see this through to merging. And I probably won't any time soon. If you want to completely take this over and see it through to merging, feel free!

modosc:
sure!

marcgrr:
Is there something I can do to let you "take over" this PR? Or maybe should I close this one and let you open another one?

modosc:
yeah, that makes the most sense to me.

PR #923, opened 2016-10-22
add props to lifecycle sort group

calebmer:
When using Flow, you’ll often include types for your props like so:

``` js
  props: {
    a: number,
    b: number,
  }

  state: {
    c: number,
    d: number,
  } = {
    c: 1,
    d: 2,
  }

  ...
}
```

In such a case you will also want the `props` type definition on top of the component. This PR adds the `props` to the `sort-comp` `lifecycle` rule so that by default the `props` property will be sorted to the top. Adding a `props` property is generally not desirable in any other case but this one.

### DIFF
--- a/lib/rules/sort-comp.js
+++ b/lib/rules/sort-comp.js
@@ -87,6 +87,7 @@ module.exports = {
       groups: {
         lifecycle: [
           'displayName',
+          'props',
           'propTypes',
           'contextTypes',
           'childContextTypes',
--- a/docs/rules/sort-comp.md
+++ b/docs/rules/sort-comp.md
@@ -9,7 +9,7 @@ When creating React components it is more convenient to always follow the same o
 With default configuration the following organisation must be followed:
 
   1. static methods and properties
-  2. lifecycle methods: `displayName`, `propTypes`, `contextTypes`, `childContextTypes`, `mixins`, `statics`,`defaultProps`, `constructor`, `getDefaultProps`, `getInitialState`, `state`, `getChildContext`, `componentWillMount`, `componentDidMount`, `componentWillReceiveProps`, `shouldComponentUpdate`, `componentWillUpdate`, `componentDidUpdate`, `componentWillUnmount` (in this order).
+  2. lifecycle methods: `displayName`, `props`, `propTypes`, `contextTypes`, `childContextTypes`, `mixins`, `statics`,`defaultProps`, `constructor`, `getDefaultProps`, `getInitialState`, `state`, `getChildContext`, `componentWillMount`, `componentDidMount`, `componentWillReceiveProps`, `shouldComponentUpdate`, `componentWillUpdate`, `componentDidUpdate`, `componentWillUnmount` (in this order).
   3. custom methods
   4. `render` method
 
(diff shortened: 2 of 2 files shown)


lib/rules/sort-comp.js
```diff
@@ -87,6 +87,7 @@ module.exports = {
       groups: {
         lifecycle: [
           'displayName',
+          'props',
```
ljharb: I'm a big -1 on including this in the default. Outside of Flow, you would *never* want to declare the `props` instance property anywhere, nor mutate it or assign to it directly.

calebmer: This approach is recommended by the [Flow documentation](https://flowtype.org/docs/react.html). I agree that outside of Flow you would never ever want to do this, and that’s why I think it’s fine to make this addition. If you’re ever assigning to `props` (even in Flow you aren’t assigning), it will result in a bug. However, when you do whether it be for TypeScript, Flow, or any other arbitrary type checker or JavaScript intelligence tool there is an idiomatic place to put the `props` attribute, the top of the component class.

I’m also going to cc @thejameskyle on this one to see if he wants to make a comment.

ljharb: If someone is *not* using Flow, and they did in fact have `props = {}` as a class property (setting aside that nobody would ever do that in React), would this sorting category affect it?

thejameskyle:
Why would this be an unidiomatic way of typing props? Other than the fact that props are meant to be read-only you are typing an object that does exist.

ljharb:
@thejameskyle in React, prior to Flow, props were typed with `propTypes`. I do accept that typing it this way makes more sense since it's directly using the property name, but at the moment, having types at all is unidiomatic, let alone when there's no associated value (ie, no default prop). Is that allowed in Flow? (specifying the default value along with the type)

thejameskyle:
`propTypes` is something separate because it lives in value-space and needs to be a separate property name to make sense in JavaScript. Flow is able to merge values from both `props` and `defaultProps` fine and is totally idiomatic in terms of how classes work.

Class property types in Flow with values look like this:

```js
class Foo {
  bar: boolean = true;
}
```

ljharb:
Thanks for the Flow explanations (I don't use it yet). I'm convinced that support for this belongs in the plugin - i'm just not convinced it belongs in the default ordering. I'm content to be overruled by another collab, however.

thejameskyle:
Seems like you need the ability to differentiate values from types. It seems like it's currently treating both as one in the same which, as you pointed out, is not the case.

```js
class Component { props: {...} } // ok
class Component { props = {...} } // bad
```

calebmer:
What if we added another rule where assignment of props was treated as an error? That way, practically, this sort would only affect types.

idris:
I think this can be closed now that there is a `type-annotations` group.

```js
"react/sort-comp": [1, {
  order: [
    'type-annotations',
    'static-methods',
    'lifecycle',
    'everything-else',
    'render',
  ],
}]
```

ljharb:
Closing per https://github.com/yannickcr/eslint-plugin-react/pull/923#issuecomment-312373840

PR #925, opened 2016-10-24
Fix incomplete recommended configs

MoOx:
This change allow you to extends "plugin:react/{something}" without having to do something more.
If you just extends one config, you get errors about missing rules definition.

### DIFF
--- a/index.js
+++ b/index.js
@@ -86,6 +86,9 @@ module.exports = {
   rules: exportedRules,
   configs: {
     recommended: {
+      plugin: [
+        'react'
+      ],
       parserOptions: {
         ecmaFeatures: {
           jsx: true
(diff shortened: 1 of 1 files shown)


ljharb:
I'm not sure I understand - you extend configs, not rules or plugins. `react/jsx-indent` is a rule, not a config.

In other words, the only two configs that this plugin provides are "recommended" and "all", and that's intentional.

MoOx:
I am not sure to follow your response.
The idea here is to be able to "just" do

```js
{ 
  extends: "plugin:react/recommended"
}
```

But currently you need to do


```js
{ 
  "extends": "plugin:react/recommended",
  "plugins": [ "react" ]
}
```

Otherwise your get error about unknown rules. It's pretty weird to register rules without adding definitions for those rules.
This issue is a follow up on https://github.com/MoOx/phenomic/issues/838

yannickcr:
Merged, thanks!

PR #936, opened 2016-11-01
Add forbidPatterns to forbid-component-props rule

burabure:
Adds a forbidPatterns option that takes an array of regexp strings so it's possible to forbid the use of things like `data-*` props

### DIFF
--- a/docs/rules/forbid-component-props.md
+++ b/docs/rules/forbid-component-props.md
@@ -1,6 +1,6 @@
 # Forbid certain props on Components (forbid-component-props)
 
-By default this rule prevents passing of [props that add lots of complexity](https://medium.com/brigade-engineering/don-t-pass-css-classes-between-components-e9f7ab192785) (`className`, `style`) to Components. This rule only applies to Components (e.g. `<Foo />`) and not DOM nodes (e.g. `<div />`). The list of forbidden props can be customized with the `forbid` option.
+By default this rule prevents passing of [props that add lots of complexity](https://medium.com/brigade-engineering/don-t-pass-css-classes-between-components-e9f7ab192785) (`className`, `style`) to Components. This rule only applies to Components (e.g. `<Foo />`) and not DOM nodes (e.g. `<div />`). The list of forbidden props can be customized with the `forbid` or `forbidPatterns` options.
 
 ## Rule Details
 
@@ -35,10 +35,18 @@ The following patterns are not considered warnings:
 
 ```js
 ...
-"forbid-component-props": [<enabled>, { "forbid": [<string>] }]
+"forbid-component-props": [<enabled>, { "forbid": [<string>], "forbidPatterns": [<string>], ignoreDomNodes: <boolean> }]
 ...
 ```
 
 ### `forbid`
 
 An array of strings, with the names of props that are forbidden. The default value of this option is `['className', 'style']`.
+
+### `forbidPatterns`
+
+An array of strings, with the patterns of props that are forbidden; you can use `*` as a wildcard. e.g: `['data-*', '*-foo-*']`
+
+### `ignoreDomNodes`
+
+Whether to check forbidden props on DOM nodes (div, input, span, etc...). Defaults to true
--- a/docs/rules/forbid-component-props.md
+++ b/docs/rules/forbid-component-props.md
@@ -35,7 +35,7 @@ The following patterns are not considered warnings:
 
 ```js
 ...
-"forbid-component-props": [<enabled>, { "forbid": [<string>], "forbidPatterns": [<string>], ignoreDomNodes: <boolean> }]
+"forbid-component-props": [<enabled>, { "forbid": [<string>], "forbidPatterns": [<string>], checkDomNodes: <boolean>, ignoreComponents: <boolean> }]
 ...
 ```
 
@@ -47,6 +47,10 @@ An array of strings, with the names of props that are forbidden. The default val
 
 An array of strings, with the patterns of props that are forbidden; you can use `*` as a wildcard. e.g: `['data-*', '*-foo-*']`
 
-### `ignoreDomNodes`
+### `checkDomNodes`
 
-Whether to check forbidden props on DOM nodes (div, input, span, etc...). Defaults to true
+Whether to check forbidden props on DOM nodes (div, input, span, etc...). Defaults to false.
+
+### `ignoreComponents`
+
+Whether to skip checking forbidden props on Components (div, input, span, etc...). Defaults to false.
--- a/lib/rules/forbid-component-props.js
+++ b/lib/rules/forbid-component-props.js
@@ -1,8 +1,11 @@
 /**
  * @fileoverview Forbid certain props on components
  * @author Joe Lencioni
+ * @author Nicolas Fernandez <@burabure>
  */
 'use strict';
+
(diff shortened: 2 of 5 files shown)


docs/rules/forbid-component-props.md
```diff
@@ -35,10 +35,14 @@ The following patterns are not considered warnings:
...
 ...
 ```
 
 ### `forbid`
 
 An array of strings, with the names of props that are forbidden. The default value of this option is `['className', 'style']`.
+
+### `forbidPatterns`
+
+An array of regular expressions as strings, with the patterns of props that are forbidden. e.g: `['^data-']`
```
ljharb: Accepting user string input into regexes is a bad idea. I'd prefer we accept a limited subset, like just *

lib/rules/forbid-component-props.js
```diff
@@ -38,11 +45,31 @@ module.exports = {
 
   create: function(context) {
 
+    function somePatternMatches(patterns, subject) {
+      return patterns.reduce(function (acc, pattern) {
+        if (RegExp(pattern).test(subject)) {
```
ljharb: This line is problematic, especially creating the same regex more than once.

ljharb (changes requested):
I like the idea of forbidPatterns. I very much dislike the idea of feeding the string into RegExp.

lib/rules/forbid-component-props.js
burabure: we're iterating through the array fo patterns, so each pattern is transformed to a regex only once

docs/rules/forbid-component-props.md
burabure: done!

lib/rules/forbid-component-props.js
ljharb: Good call on that part :-) passing user input into `RegExp` is still an issue.

burabure: fixed that, I think this review is showing the older version of the code because I squashed

docs/rules/forbid-component-props.md
```diff
@@ -35,10 +35,18 @@ The following patterns are not considered warnings:
...
 
 ### `forbid`
 
 An array of strings, with the names of props that are forbidden. The default value of this option is `['className', 'style']`.
+
+### `forbidPatterns`
+
+An array of strings, with the patterns of props that are forbidden; you can use `*` as a wildcard. e.g: `['data-*', '*-foo-*']`
+
+### `ignoreDomNodes`
```
ljharb: i like this - we should also add `ignoreComponents` that defaults to false.

docs/rules/forbid-component-props.md
```diff
@@ -35,10 +35,18 @@ The following patterns are not considered warnings:
...
+
+Whether to check forbidden props on DOM nodes (div, input, span, etc...). Defaults to true
```
ljharb: having options where absence and `false` are not the same is a very bad idea - it'd be better to rename this to `checkDOMNodes` and default it to `false`.

lib/util/escapeRegExp.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * Escapes RegExp special characters from a string
```
ljharb: I understand what you were trying to do here, but I'm explicitly saying that regular expression patterns *never belong in an eslint config*. This is also the position of the eslint core team, fwiw.

ljharb (changes requested):

lib/util/escapeRegExp.js
burabure: I understand but right now what i'm doing is enabling users to use `*` as a wildcard only, the user won't be able to use a regexp in the pattern; only `*` as a wildcard. If you're suggesting dropping the regexp implementation detail altogether could you suggest how to handle cases like: `'*-foo-*'`, `'*-foo-*-bar'`, `'*-foo-*bar*-baz*'` and so on in a performant manner?

docs/rules/forbid-component-props.md
burabure: I Agree, I'll refactor that.

lib/util/escapeRegExp.js
burabure: or maybe you're suggesting that we should only support `*userInput*` and don't provide the user with a way to use wildcards?

ljharb: Wildcards of "*" seem very reasonable - but general regex syntax, implied or otherwise, is dangerous. eslint uses ` git-ignore` iirc to match patterns, but that (and glob/minimatch) tend to apply to filenames, not strings.

burabure: just cheked how eslint does it, and they just straight take regexes from users https://github.com/eslint/eslint/blob/660e0918933e6e7fede26bc675a0763a6b357c94/lib/rules/line-comment-position.js#L63

I'll commit the other changes and push, I think the current implementation is safe enough (just wildcard `*`)

ljharb: woof, ok. thanks

docs/rules/forbid-component-props.md
burabure: done!

burabure:
all changes done.

docs/rules/forbid-component-props.md
```diff
@@ -35,10 +35,22 @@ The following patterns are not considered warnings:
...
+
+An array of strings, with the patterns of props that are forbidden; you can use `*` as a wildcard. e.g: `['data-*', '*-foo-*']`
+
+### `checkDomNodes`
+
+Whether to check forbidden props on DOM nodes (div, input, span, etc...). Defaults to false.
+
+### `ignoreComponents`
+
+Whether to skip checking forbidden props on Components (div, input, span, etc...). Defaults to false.
```
ljharb: these examples are DOM nodes, not Components

lib/rules/forbid-component-props.js
```diff
@@ -25,31 +28,68 @@ module.exports = {
...
+
+      return false;
     }
 
     return {
       JSXAttribute: function(node) {
         var tag = node.parent.name.name;
-        if (tag && tag[0] !== tag[0].toUpperCase()) {
-          // This is a DOM node, not a Component, so exit.
+        var isComponent = tag && tag[0] === tag[0].toUpperCase();
```
ljharb: i think your code here is making the false assumption that there's only two kinds of nodes, DOM and Component. nodes can also be strings, numbers, arrays… are we sure this is the right way to check "is dom node" and "is component"? I feel like we have functions for this in the codebase already.

ljharb (changes requested):

docs/rules/forbid-component-props.md
```diff
@@ -35,10 +35,22 @@ The following patterns are not considered warnings:
...
+
+### `forbidPatterns`
+
+An array of strings, with the patterns of props that are forbidden; you can use `*` as a wildcard. e.g: `['data-*', '*-foo-*']`
+
+### `checkDomNodes`
+
+Whether to check forbidden props on DOM nodes (div, input, span, etc...). Defaults to false.
+
+### `ignoreComponents`
```
ljharb: https://github.com/yannickcr/eslint-plugin-react/pull/850#issuecomment-258027762

PR #937, opened 2016-11-02
Add autofix for jsx-first-prop-new-line

shioju:


ljharb:
This appears to be based on #886?

lib/rules/jsx-first-prop-new-line.js
```diff
@@ -15,14 +17,49 @@ module.exports = {
...
 
-    schema: [{
-      enum: ['always', 'never', 'multiline', 'multiline-multiprop']
-    }]
+    schema: [
+      {
+        enum: ['always', 'never', 'multiline', 'multiline-multiprop']
+      },
+      {
+        oneOf: [{
```
ljharb: indent should not be passed separately - it should just use 4 spaces (the `indent` default) and the individual indent rules should handle autofixing them.

shioju:
there's another PR for the same autofix

PR #939, opened 2016-11-02
add mount/unmount rule

Bonuspunkt:
i'm having a legacy codebase where i partially introduced React.
The Lifetime of a React component is determined by its container which is managed by the other codebase.
this rule helps me find find missing unmounts.

### DIFF
--- a/index.js
+++ b/index.js
@@ -55,7 +55,8 @@ var allRules = {
   'no-children-prop': require('./lib/rules/no-children-prop'),
   'no-comment-textnodes': require('./lib/rules/no-comment-textnodes'),
   'require-extension': require('./lib/rules/require-extension'),
-  'wrap-multilines': require('./lib/rules/wrap-multilines')
+  'wrap-multilines': require('./lib/rules/wrap-multilines'),
+  'mount-unmount': require('./lib/rules/mount-unmount')
 };
 
 function filterRules(rules, predicate) {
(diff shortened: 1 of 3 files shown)


ljharb:
It seems like a very uncommon use case to unmount components - most code I've seen mounts a React component and then never touches it again.

Bonuspunkt:
=> its a special use case that does not match the vision of this project?

i just need to know if it will get merged or i need to move it to an other repo.

ljharb:
@Bonuspunkt can you explain the use case a bit more where you need to unmount?

Bonuspunkt:
yeah of course, as example i'm having a custom dijit widget where i embed a React Component
```js
declare("MyWidget", [_WidgetBase], {
    startup: function() {
        this.inherited(arguments);
        this.renderFilter();
    },
    renderFilter: function() {
        ReactDOM.render(
            React.createElement(Filter, {
                // ...
            }),
            this.domNode
        );
    },
    destroy: function() {
        ReactDOM.unmountComponentAtNode(this.domNode);
        this.inherited(arguments);
    }
});
```
then i could use `MyWidget` in my other existing dijit widgets.
when a widget is created/destroyed, it calls startup/destroy on its descents widgets.
ref. [Understanding _Widget](https://dojotoolkit.org/documentation/tutorials/1.6/understanding_widget/)

the only other use case i can come up with is modal dialogs like [react-portal](https://github.com/tajo/react-portal) or [react-modal](https://github.com/reactjs/react-modal) but seems like they don't use ReactDOM.render

ljharb:
I think that that is a great idea for a custom rule for "dijit" users, but is definitely not a common case for React users. Even in your example, there's no requirement that the unmount happen unless "destroy" is called - that's very specific to dijit, i think.

PR #946, opened 2016-11-07
detect used props in jsx

### LINKED ISSUE #885: no-unused-prop-types sometimes fails to detect props in functions on stateless components
Here is a minimal example that triggers this:

``` jsx
const PagingBlock = function(props) {
  return (
    <span>
      <a onClick={() => props.previousPage()}/>
      <a onClick={() => props.nextPage()}/>
    </span>
 );
};

PagingBlock.propTypes = {
  nextPage: React.PropTypes.func.isRequired,
  previousPage: React.PropTypes.func.isRequired,
};
```

The error message is:

```
  18:17  error  'previousPage' PropType is defined but prop is never used  react/no-unused-prop-types
```

Notes:
- Converting this to a React.createClass with static propTypes fixes the issue
- Removing either `<a/>` tag produces the correct lint error, the false error only occurs when both are present
- The error is notified for the first tag (switching `nextPage` and `previousPage` inside the render function will make the error text say `nextPage` is defined but not used

BarryThePenguin:
fixes #885

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -362,12 +362,13 @@ function componentRule(rule, context) {
         var isFunction = /Function/.test(node.type); // Functions
         var isMethod = node.parent && node.parent.type === 'MethodDefinition'; // Classes methods
         var isArgument = node.parent && node.parent.type === 'CallExpression'; // Arguments (callback, etc.)
+        var isJSX = node.parent && node.parent.type === 'JSXExpressionContainer';
         // Stop moving up if we reach a class or an argument (like a callback)
         if (isClass || isArgument) {
           return null;
         }
         // Return the node if it is a function that is not a class method
-        if (isFunction && !isMethod) {
+        if (isFunction && !isMethod && !isJSX) {
           return node;
         }
         scope = scope.upper;
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -362,13 +362,14 @@ function componentRule(rule, context) {
         var isFunction = /Function/.test(node.type); // Functions
         var isMethod = node.parent && node.parent.type === 'MethodDefinition'; // Classes methods
         var isArgument = node.parent && node.parent.type === 'CallExpression'; // Arguments (callback, etc.)
-        var isJSX = node.parent && node.parent.type === 'JSXExpressionContainer';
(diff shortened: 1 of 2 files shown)


lib/util/Components.js
```diff
@@ -347,12 +347,13 @@ function componentRule(rule, context) {
         var isFunction = /Function/.test(node.type); // Functions
         var isMethod = node.parent && node.parent.type === 'MethodDefinition'; // Classes methods
         var isArgument = node.parent && node.parent.type === 'CallExpression'; // Arguments (callback, etc.)
+        var isJSX = node.parent && node.parent.type === 'JSXExpressionContainer';
         // Stop moving up if we reach a class or an argument (like a callback)
         if (isClass || isArgument) {
           return null;
         }
         // Return the node if it is a function that is not a class method
```
lencioni: This comment should probably be updated.

lib/util/Components.js
```diff
@@ -347,12 +347,13 @@ function componentRule(rule, context) {
         var isFunction = /Function/.test(node.type); // Functions
         var isMethod = node.parent && node.parent.type === 'MethodDefinition'; // Classes methods
         var isArgument = node.parent && node.parent.type === 'CallExpression'; // Arguments (callback, etc.)
+        var isJSX = node.parent && node.parent.type === 'JSXExpressionContainer';
```
lencioni: I might be misunderstanding what this is checking for here, but I'm not sure this variable name makes sense. I think `isJSXExpressionContainerExpression` would be more accurate.

Additionally, it might be nice to include a comment with an example of what type of code this identifies.

lencioni: Also, does this bug exist for folks not using JSX? (i.e. `React.createElement`)

```js
React.createElement("a", { onClick: function onClick() {
    return props.previousPage();
  } });
```

lib/util/Components.js
BarryThePenguin: Updated

lib/util/Components.js
BarryThePenguin: The problem is that functions inside a `JSXExpressionContainer` are bing mistaken for stateless components.

For example: `<button onClick={() => props.handleClick()}></button>`
where `{() => props.handleClick()}` is the`JSXExpressionContainer`

Issue #945, opened 2016-11-06
jsx-indent wrong works with ternary operator


exeto:
**Version:** 6.6.0

<img width="692" alt="2016-11-06 23 49 28" src="https://cloud.githubusercontent.com/assets/4079046/20041540/c7c4a066-a47b-11e6-8863-620656ebbb6e.png">

omgaz:
ditto for the ternary operator being inline with the logical if/brackets

```
{ isFoo ?
  <div>lorem</div>
:
  <div>ipsum</div>
}
```

lukeapage:
yes my biggest remaining problem is with terneries and brackets e.g. when brackets are used, usually indentation goes up by 1, but within a ternary, it currently does not.

Kovensky:
This was caused by a fix to allow this kind of formatting:

```js
cond ?
  <div/> :
  <span/>
```

Before 6.6.0, `jsx-indent` would have required:

```js
cond ?
  <div/> :
    <span/>
```

But this breaks the also common style with `:` on its own line :/

lukeapage:
examples formatted how 6.6.0 expects vs expectation
```
        return isDisabled ?
            content :
            (
            <Touchable onTap={this.handleTap}>
                {content}
            </Touchable>
        );
```
but I'd expect
```
        return isDisabled ?
            content :
            (
                <Touchable onTap={this.handleTap}>
                    {content}
                </Touchable>
            );
```
next one
```
    return secondaryTitle ?
        (
            <div>
                <p>{primaryTitle}</p>
                <p>{secondaryTitle}</p>
            </div>
        ) : (
        <p>{primaryTitle}</p>
        );
```
but I would expect
```
    return secondaryTitle ?
        (
            <div>
                <p>{primaryTitle}</p>
                <p>{secondaryTitle}</p>
            </div>
        ) : (
            <p>{primaryTitle}</p>
        );
```

voxpelli:
Makes it really hard when combined with the `operator-linebreak`, which eg. comes with the `standard` package, as that doesn't allow you to move your `:` either. But mostly it's just bad to have the linting tool suddenly change it's mind about the linting and encourage a bad coding style rather than a good one, so hoping for a swift fix.

This apparently was fix of #901 and #897 that was made as a response to another bug around this that was introduced in `6.4.0`.

chrisvasz:
Code like this also produces a linter error now (not previously):

```
{condition ? <Hello /> : (
  <World />
)}
[...]

PR #950, opened 2016-11-07
Fix jsx-indent bug on multiline ternary

voxpelli:
This PR is an attempt at fixing #945.

I'm no master at ESLint or it's parser, but adding some new tests and looking at the AST and figuring out what it would take to make the new test pass while not breaking the tests from  https://github.com/yannickcr/eslint-plugin-react/commit/d88ee379079cee18e3bc718dddc7dfc1c91bee1c gave the following solution.

I also ran this version of the module against my large react project and all errors that occurred with `6.6.0` was then gone.

### DIFF
--- a/lib/rules/jsx-indent.js
+++ b/lib/rules/jsx-indent.js
@@ -193,7 +193,8 @@ module.exports = {
         node.parent &&
         node.parent.parent &&
         node.parent.parent.type === 'ConditionalExpression' &&
-        node.parent.parent.alternate === node.parent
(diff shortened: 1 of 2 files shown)


Kovensky:
This (unfortunately) won't deal with this, still requiring that the `<span/>` start on column 2:

```jsx
cond
  ?
    <div/>
  :
    <span/>
```

Though not supporting even the case where you use parens is certainly the bigger evil.

yannickcr:
Merged, thanks!

Issue #851, opened 2016-09-18
Destructuring values from props is not recognized by proptypes inside PropTypes.shape


GreenGeorge:
This is similar to #782 which actually works properly for me, what doesn't work is when the proptypes are inside shape() :

```javascript
const Person = ({ person }) =>
  <h1>{ person.name }</h1>;
```

```javascript
Person.propTypes = {
  person = React.PropTypes.shape({
    name: React.PropTypes.string, 
    // eslint error 'person.name' PropType is defined but prop is never used
    // error also pops up on stateful components
  }),
};
```

It works fine when i dont destructure
```javascript
const Person = props =>
  <h1>{ props.person.name }</h1>;  // no eslint error
```

batjko:
Just found the same issue.
But it still won't work if I replace the destructuring of the function argument, e.g.:
![image](https://cloud.githubusercontent.com/assets/7287076/18874797/0a5858b8-84bb-11e6-9304-716fc4ea4859.png)

(Before that, I just used `function SideMenuItems({menuItems})` instead. Same lint errors)

EvNaverniouk:
The `no-unused-prop-types` rule does not support `shape` props at the moment as such detection is very difficult. If you use shape props, I recommend setting the `skipShapeProps` option to `true` on the rule.

Duplicate of #819. Will track further discussion of `shape` props there.

Issue #819, opened 2016-09-11
no-unused-prop-types does not recognize use of prop inside Array.map()


smsenesac:
In the following example, no-unused-prop-types will flag a non-use of arrayOfObjWithFoo.*.foo:

``` jsx
    static PropTypes = {
      arrayOfObjWithFoo: PropTypes.arrayOf(
        PropTypes.shape({ foo: PropTypes.string.isRequired })
      ).isRequired
    };

    render() {
      const fooItems = this.props.arrayOfObjWithFoo.map(objWithFoo => {
        <Item foo={objWithFoo.foo} />
      });
    }
```

EvNaverniouk:
Duplicate of #871. 

The `no-unused-prop-types` rule does not support `shape` props at the moment as such detection is very difficult. If you use shape props, I recommend setting the `skipShapeProps` option to `true` on the rule.

ljharb:
Again, this is still a bug. If it doesn't support shape props, it should not warn on them at all.

EvNaverniouk:
If you set the `skipShapeProps` option to true, it should stop warning on all shape props. Is that not working?

ljharb:
I'm saying that if that option is required for the rule to be usable when using shapes, then the rule should never check shapes.

EvNaverniouk:
@ljharb That's a valid point. I'll look into whether or not it makes sense to set `skipShapeProps` to `true` as a default.

merlinpatt:
I agree that `skipShapeProps` should default to true if it can't properly detect usage of props in `shape`.

The current behavior contradicts the [forbid-prop-types](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/forbid-prop-types.md) rule since that encourages the use of `shape` instead of `object`.

ljharb:
@EvNaverniouk any progress here?

EvNaverniouk:
@ljharb Just took a few minutes now to look into this but wasn't able to get far. This issue is talking about 2 things:

1. **The code snippet @smsenesac originally posted is supposedly throwing an error when it shouldn't.**
I cannot reproduce the failure. I added a test for this here: https://github.com/EvNaverniouk/eslint-plugin-react/commit/15e2fd8448e415b1a3961bcbbe62e3e9cdb93580 but the test isn't failing for me. So I wonder if maybe this has already been fixed in another commit.
2. **Setting `skipShapeProps` to `true` by default.**
[...]

PR #953, opened 2016-11-08
skipShapeProps should be enabled by default until it works properly

everdimension:
The option `no-unused-prop-types` doesn't work properly for shape proptypes in lots of popular cases: #851, #819 and other cases, like when used outside of the component class.

This causes very annoying inconviences.

So currently the option only works without problems if the `skipShapeProps` is set to `true`. Therefore it's best to skip linting shape props by default until the mentioned problems are fixed and setting this option to false actually becomes useful.

### DIFF
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -47,9 +47,10 @@ module.exports = {
 
   create: Components.detect(function(context, components, utils) {
 
+    var defaults = {skipShapeProps: true};
     var sourceCode = context.getSourceCode();
-    var configuration = context.options[0] || {};
-    var skipShapeProps = configuration.skipShapeProps || false;
+    var configuration = Object.assign({}, defaults, context.options[0] || {});
+    var skipShapeProps = configuration.skipShapeProps;
     var customValidators = configuration.customValidators || [];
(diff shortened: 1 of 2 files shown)


ljharb:
Actually I think this is probably a good change. @yannickcr?

everdimension:
@ljharb yea, I planned to fix the failing checks and then reopen the pr. Perhaps a bit later today.

everdimension:
@ljharb, @yannickcr  Updated the PR

yannickcr:
Seems the rule is not working properly if the`skipShapeProps` option is set to `false`, so is it useful to keep this option?

everdimension:
@yannickcr Well, it does work for simple cases (which are in tests) so it might be enough for someone's writing style.
And hopefully it will once work just as expected for all other cases, too.

yannickcr:
@everdimension I see. But in this case passing this option to `true` by default is a breaking change, so it should be done in a major release.

ljharb:
@yannickcr why is it a breaking change? Setting that default causes *fewer* errors, so by eslint's semver policy that's just a patch.

yannickcr:
@ljharb Ho, you're right. LGTM then :)

PR #956, opened 2016-11-14
Fix `jsx-tag-spacing` crash when options object isn't passed

daltones:
Fixes #955

PR #957, opened 2016-11-14
[Fix] Never call Object prototype builtins directly

### LINKED ISSUE #955: jsx-tag-spacing crashes if options object isn't passed
This simple configuration makes `jsx-tag-spacing` crash:
```
'react/jsx-tag-spacing': 1
```

A quick fix is:
```
'react/jsx-tag-spacing': [1, {}]
```

Error:
```
TypeError: Error while loading rule 'react/jsx-tag-spacing': Cannot read property 'hasOwnProperty' of undefined
    at Object.create ([...]/node_modules/eslint-plugin-react/lib/rules/jsx-tag-spacing.js:208:60)
```

ljharb:
Fixes #955.

I'll worry about removing all the for..in loops (which should also never be used) in a future PR.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -152,6 +152,7 @@
         "max-len": [2, 120],
         "max-params": 0,
         "max-statements": 0,
-        "no-plusplus": 0
+        "no-plusplus": 0,
+        "no-prototype-builtins": 2
     }
 }
--- a/index.js
+++ b/index.js
@@ -1,5 +1,7 @@
 'use strict';
 
+var has = require('has');
+
 var allRules = {
   'jsx-uses-react': require('./lib/rules/jsx-uses-react'),
   'no-multi-comp': require('./lib/rules/no-multi-comp'),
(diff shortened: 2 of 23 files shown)


yannickcr:
This PR go way beyond fixing #955, I'd prefer to stay with a simple fix like #956 for now.

ljharb:
We can certainly add that simple fix, but this is a systemic problem by the bad practice of calling `.hasOwnProperty` directly - which this PR nips in the bud. Is this not a change you want to make?

yannickcr:
Well, I am not against rethinking some rules (for `hasOwnProperty` or `for...in` like you mentioned) but the use of `hasOwnProperty` here is not the reason of the problem here (the problem was the direct use of `context.options[0]` without checking it's existence before).

Also I don't like the idea to add some additional dependencies to the plugin for such things.

So for now I'd prefer to simply fix #955 then discuss about the other changes in another PR/Issue :)

ljharb:
OK - now that #956 is merged, let's discuss it :-)

I do agree that (as I discovered when adding a test for it, which #956 did not do) that `hasOwnProperty` wasn't the problem here.

However, I'm not sure why adding an additional dependency is a problem; it's transparent to users, and this is a dev tool, not a production tool. Using prototype builtins directly is a huge footgun, and having `has` available, combined with the lint rule, ensures that developers won't run into it.

ljharb:
@yannickcr any update on making this change?

ljharb:
This is now updated - I'm hoping to merge it soon, please review.

yannickcr:
Ok for me 👍

PR #958, opened 2016-11-15
"jsx-indent" -> "react/jsx-indent"

Jorundur:


Jorundur:
There are some things I don't understand in this rule. The rule's description is 
`"jsx-indent": [<enabled>, 'tab'|<number>]`
but there is no further description on `<enabled>`.

I would assume that it's a boolean, either 0 or 1, but in every single example `<enabled>` is 2, e.g.:

> // 2 spaces indentation
// [2, 2]

> // tab indentation
// [2, 'tab']

> // no indentation
// [2, 0]

What does that mean?

_Furthermore_, we can see this contradiction:

> **The following patterns are considered warnings:**
...
// tab indentation
// [2, 'tab']
\<App\>
&nbsp;&nbsp;\<Hello /\>
\</App\>
... 
**The following patterns are not warnings:**
// tab indentation
// [2, 'tab']
\<App\>
&nbsp;&nbsp;\<Hello /\>
\</App\>

These patterns are identical and yet one is considered a warning and the other not.

ljharb:
"enabled" refers to 0/1/2 or off/warn/error, just like every eslint rule in the ecosystem and core is configured.

If this change were to be made (adding "react/"), we'd have to do it on all the rules - as this is how they're all currently documented.

Jorundur:
@ljharb Thanks for the clarification regarding `<enabled>`. 

Regarding the addition of `react/` on all rules, you're right, that would be quite the hassle.
However, I might suggest that for consistency you then do the opposite thing and _remove_ `react/` from  [jsx-filename-extension](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-filename-extension.md) and [require-extension](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-extension.md) that currently include the lines:
> ```
"rules": {
  "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
}
```

and 
>```
"rules": {
  "react/require-extension": [1, { "extensions": [".js", ".jsx"] }],
}
```
Lastly, what about the contradiction I mentioned, is there a reason for it or am I misunderstanding something?

ljharb:
Because tabs vs spaces are very hard to convey in markdown format, all the examples use spaces, but have a comment directly above them describing the intended indentation.

PR #965, opened 2016-11-19
New: Add require-default-props rule (fixes #528)

### LINKED ISSUE #528: Rule request: require-default-props
I'd like a rule that requires an explicit `defaultProp` to be present for every non-required prop (in `propTypes` or otherwise).

vitorbal:
Implements the new `require-default-props` rule. (#528)

~~**This is a work in progress**: support for flow annotations hasn't been added yet, but the code is already considerably big, so I thought I could get the PR up already to get some early feedback on implementation, docs, missing test cases, etc.~~ flow annotation support has now been added.

This is my first time contributing to eslint-plugin-react, so I hope I didn't do something horribly wrong in the implementation of the rule 😟

### DIFF
--- a/lib/util/variable.js
+++ b/lib/util/variable.js
@@ -29,8 +29,7 @@ function findVariable(variables, name) {
  * Contain a patch for babel-eslint to avoid https://github.com/babel/babel-eslint/issues/21
  *
  * @param {Object} context The current rule context.
- * @param {Array} name The name of the variable to search.
- * @returns {Boolean} True if the variable was found, false if not.
+ * @returns {Array} The variables list
  */
 function variablesInScope(context) {
   var scope = context.getScope();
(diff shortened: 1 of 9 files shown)


ljharb:
Your changes look great, thanks!

lib/util/variable.js
```diff
@@ -29,8 +29,7 @@ function findVariable(variables, name) {
  * Contain a patch for babel-eslint to avoid https://github.com/babel/babel-eslint/issues/21
  *
  * @param {Object} context The current rule context.
- * @param {Array} name The name of the variable to search.
- * @returns {Boolean} True if the variable was found, false if not.
+ * @returns {Array} The variables list
```
vitorbal: Hope this is okay, sneaked in a fix for the wrong jsdoc for this function.

vitorbal:
@ljharb @yannickcr alright, i pushed some new commits that add support for flow annotations to the rule. I tried to come up with the biggest number of test cases I could think of, but unfortunately I don't have any big repositories that use react and flow to test the rule against.

Are there any opensource projects that use both react and flow that we could test the rule against to catch some potential bugs? @ljharb is there any chance you could give the rule a test-drive at Airbnb?

ljharb:
@vitorbal Airbnb does not yet use Flow; I don't have any projects I can test flow support on.

vitorbal:
I added some extra tests for flow annotation use cases. Unfortunately I couldn't find any codebases using flow to test the rule on. However, I'm willing to react fast to any bugs found after this rule gets published, FWIW.

yannickcr:
This is really great, thanks!

PR #978, opened 2016-11-25
Add no-array-index-key rule

lencioni:
This rule will warn when using the array index as the `key`. As a first
pass, I decided to implement this for `Array.prototype.map`, since that
is likely the most common case, but it theoretically could be expanded
to find other cases so I kept the naming more generic.

Like many rules, this one is imperfect and is prone to some false
positives and negatives. For instance, if someone defines a `.map()`
function on an object that doesn't have the same signature as
`Array.prototype.map`, this will likely end up warning in those cases.
However, I think the value of this rule outweighs its hypothetical
drawbacks.

### DIFF
--- a/docs/rules/no-array-index-key.md
+++ b/docs/rules/no-array-index-key.md
@@ -15,6 +15,10 @@ things.map((thing, index) => (
   <Hello key={index} />
 ));
 
+things.map((thing, index) => (
+  React.cloneElement(thing, { key: index })
+));
+
 things.forEach((thing, index) => {
   otherThings.push(<Hello key={index} />);
 });
@@ -41,6 +45,10 @@ The following patterns are not considered warnings:
 ```jsx
 things.map((thing) => (
   <Hello key={thing.id} />
+));
+
+things.map((thing) => (
+  React.cloneElement(thing, { key: thing.id })
 ));
 
 things.forEach((thing) => {
--- a/docs/rules/no-array-index-key.md
+++ b/docs/rules/no-array-index-key.md
@@ -1,0 +1,23 @@
+# Prevent usage of Array index in keys
+
+Warn if an element uses an Array index in its `key`.
+
+The `key` is used by React to [identify which items have changed, are added, or are removed and should be stable](https://facebook.github.io/react/docs/lists-and-keys.html#keys).
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```jsx
+things.map((thing, index) => (
+  <Hello key={index} />
+));
+```
+
+The following patterns are not considered warnings:
+
+```jsx
+things.map((thing) => (
+  <Hello key={thing.id} />
+));
+```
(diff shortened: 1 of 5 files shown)


docs/rules/no-array-index-key.md
```diff
@@ -0,0 +1,23 @@
+# Prevent usage of Array index in keys
+
+Warn if an element uses an Array index in its `key`.
```
EvNaverniouk: I think adding a short description of why you might want to enable this rule would be good. Something like:

```
It's a bad idea to use the array index since it doesn't uniquely identify your elements. In cases where the array is sorted, the index will be changed, even though the element representing that index may be the same -- resulting in unnecessary render cycles.
```

lib/rules/no-array-index-key.js
```diff
@@ -0,0 +1,183 @@
...
+
+    function getMapIndexParamName(node) {
+      var callee = node.callee;
+      if (callee.type !== 'MemberExpression') {
+        return null;
+      }
+      if (callee.property.type !== 'Identifier') {
+        return null;
+      }
+      if (callee.property.name !== 'map') {
```
EvNaverniouk: I would add `forEach` and `filter` in here as well. It's easy enough to support it for those rare cases where users would iterate over elements in different ways.

EvNaverniouk:
Nice. Would be really great if this could also support `React.cloneElement`, ala:

```
[].map((item, i) => {
    return React.cloneElement(someChild, {
        key: i
    })
})
```

docs/rules/no-array-index-key.md
lencioni: Sure, I'll expand this document a little. I intended my next sentence to provide the reasoning, but perhaps it needs to be more explicit.

lib/rules/no-array-index-key.js
lencioni: I can see `forEach` for sure, but `filter` would be quite unusual. It should be easy enough to include though and I don't see any harm in it, so I'll add these.

lencioni:
All feedback has been addressed! PTAL

docs/rules/no-array-index-key.md
```diff
@@ -0,0 +1,73 @@
...
+
+things.forEach((thing, index) => {
+  otherThings.push(<Hello key={index} />);
+});
+
+things.filter((thing, index) => {
+  otherThings.push(<Hello key={index} />);
+});
+
+things.some((thing, index) => {
```
ljharb: what about `reduce`, `reduceRight`, `find`?

lencioni:
> Also, what about the .forEach method of a Map or Set?

@ljharb What do you mean by this question? Are you referring to the name of the rule and documentation mentioning arrays?

ljharb:
@lencioni i mean, `aMapOrSet.forEach(x => { somethingElse.push(<div key={x} />); })` or similar. Will these be caught as well, since there's no way to know if `aMapOrSet` is an array or not?

lencioni:
@ljharb Yes, these will be caught as well for precisely the reason you mention. Likewise, so will any instance of any class that has one of these methods on it. This is what I was referring to about potential false positives in my commit message.

lencioni:
@ljharb I believe I have addressed your feedback.

PR #979, opened 2016-11-25
jsx-sort-props noSortAlphabetically option

### LINKED ISSUE #541: jsx-sort-props with out alphabetical sort
I apologize for posting a question here, but is there any way to use `jsx-sort-props` with out alphabetical sort? 

I really really want the `callbacksLast` rule but have no desire to enforce alphabetical prop order? Is this currently possible or would it require a source change?

### LINKED ISSUE #786: Proposal: Allow react/sort-prop-types to only sort by callbacks and required properties and not alphabetically
## Proposal

Add another option to `sort-prop-types` called `noSortAlphabetically` that will allow disabling sorting the props by alphabet when the user wants it.

I would be happy to create a pull request.

markus101:
Adds a `noSortAlphabetically ` option to `jsx-sort-props` which will allow sorting of callbacks last without forcing alphabetical ordering.

Closes #541
Closes #786

### DIFF
--- a/lib/rules/jsx-sort-props.js
+++ b/lib/rules/jsx-sort-props.js
@@ -40,6 +40,10 @@ module.exports = {
         },
         ignoreCase: {
           type: 'boolean'
+        },
+        // Whether alphabetical sorting should be enforced
+        noSortAlphabetically: {
+          type: 'boolean'
         }
       },
       additionalProperties: false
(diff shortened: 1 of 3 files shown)


markus101:
Not sure if anyone has taken a look at this yet, it'd be great to get this added if it's acceptable. If I should make any changes, please anything let me know.

lib/rules/jsx-sort-props.js
```diff
@@ -114,7 +119,7 @@ module.exports = {
             }
           }
 
-          if (currentPropName < previousPropName) {
+          if (!noSortAlphabetically && currentPropName < previousPropName) {
```
ljharb: This should be using localeCompare, not <, but that's out of scope for this PR

PR #980, opened 2016-11-26
#961 support flow variance syntax

### LINKED ISSUE #961: react/prop-types incompatible with flow variance syntax
Flow introduced variance syntax in 0.34.0.  This seems to have introduced a bug when component prop-types are explicitly flagged.

This code is flagged as incorrect:

```
import React from 'react';

type Props = {
  +value: string;
};

const Component = (props: Props) => (
  <div data-value={props.value} />
);
```

image

Versions:

```
"babel-core": "^6.18.2",
"babel-eslint": "^7.1.0",
"eslint": "~3.6.0",
"eslint-plugin-react": "^6.7.1",
"eslint-plugin-flowtype": "^2.25.0",
```

ajhyndman:
Fixes #961 

It looks like getKeyValue was naively grabbing the first token for `'ObjectTypeProperty'`, which was sometimes a '+'.  There's probably a more robust solution, but I've just opted to account for `'+'` and `'-'` as possible tokens that might be encountered.

### DIFF
--- a/lib/rules/prop-types.js
+++ b/lib/rules/prop-types.js
@@ -271,8 +271,11 @@ module.exports = {
      */
     function getKeyValue(node) {
       if (node.type === 'ObjectTypeProperty') {
-        var tokens = context.getFirstTokens(node, 1);
-        return tokens[0].value;
+        var tokens = context.getFirstTokens(node, 2);
+        return (tokens[0].value === '+' || tokens[0].value === '-'
+          ? tokens[1].value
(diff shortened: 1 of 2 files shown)


yannickcr:
LGTM, can you rebase your branch and resolve the conflicts? Then I think we should be good to go.

ajhyndman:
Cool, updated.

yannickcr:
Merged, thanks!

PR #981, opened 2016-11-28
Add no-common-typos rule

koenpunt:
I had it quite some time that I incorrectly used `dangerouslySetInnerHtml` instead of `dangerouslySetInnerHTML`. 
To prevent certain typos I introduce a rule for which you can set a pattern and the corrected string.

I can imagine this might not only apply to attributes, in which case this rule could be extended to also validate props and other variables.

**Todo:**
- [ ] Documentation

### DIFF
--- a/lib/rules/no-common-typos.js
+++ b/lib/rules/no-common-typos.js
@@ -18,9 +18,6 @@ var PATTERN_REGEXP = new RegExp('^/(.+)/(.+)?$');
  * @return {RegExp}
  */
 function parsePattern(pattern) {
-  if (pattern instanceof RegExp) {
-    return pattern;
-  }
   if (PATTERN_REGEXP.exec(pattern)) {
     return new RegExp(RegExp.$1, RegExp.$2);
   }
--- a/lib/rules/no-common-typos.js
+++ b/lib/rules/no-common-typos.js
@@ -40,10 +40,7 @@ module.exports = {
         type: 'object',
         properties: {
           pattern: {
-            oneOf: [
-              { type: 'string' },
-              { type: 'object' }
-            ]
+            type: 'string'
           },
           correct: {
             type: 'string'
--- a/lib/rules/no-common-typos.js
+++ b/lib/rules/no-common-typos.js
@@ -1,0 +1,94 @@
+/**
+ * @fileoverview Warn for common attribute typos.
+ * @author Koen Punt
+ */
+'use strict';
+
+var Components = require('../util/Components');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+var PATTERN_REGEXP = new RegExp('^/(.+)/(.+)?$');
+
+/**
+ * Convert pattern string to a regular expression.
+ * @param {String|RegExp} string regex
+ * @return {RegExp}
+ */
+function parsePattern(pattern) {
+  if (pattern instanceof RegExp) {
+    return pattern;
+  }
+  if (PATTERN_REGEXP.exec(pattern)) {
+    return new RegExp(RegExp.$1, RegExp.$2);
+  }
+  return new RegExp('^' + pattern + '$');
+}
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Warn for common attribute typos.',
+      category: 'Possible Errors',
+      recommended: false
+    },
+    schema: [{
+      type: 'object',
+      rules: {
+        type: 'array',
+        items: {
+          type: 'object',
+          properties: {
+            pattern: {
+              type: 'string'
+            },
+            correct: {
+              type: 'string'
+            }
+          },
+          additionalProperties: false
+        }
+      }
+    }],
+    fixable: 'code'
+  },
+
+  create: Components.detect(function(context) {
+
+    var rules = context.options || [];
+
+    /**
+     * Checks if we are using an incorrect cased dangerouslySetInnerHTML attribute.
+     * @param {ASTNode} node The AST node being checked.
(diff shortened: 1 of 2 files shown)


lib/rules/no-common-typos.js
```diff
@@ -0,0 +1,91 @@
...
+
+var PATTERN_REGEXP = new RegExp('^/(.+)/(.+)?$');
+
+/**
+ * Convert pattern string to a regular expression.
+ * @param {String|RegExp} string regex
+ * @return {RegExp}
+ */
+function parsePattern(pattern) {
+  if (pattern instanceof RegExp) {
```
ljharb: since configs come from JSON and YAML, when would it be an instanceof RegExp?

lib/rules/no-common-typos.js
```diff
@@ -0,0 +1,91 @@
...
+        node.name &&
+        node.name.name !== correct &&
+        regex.test(node.name.name)
+      );
+    }
+
+    return {
+      JSXAttribute: function(node) {
+        rules.forEach(function(rule) {
+          var regex = parsePattern(rule.pattern);
```
ljharb: i'm very concerned about using full user-provided regex patterns without sanitizing them.

tests/lib/rules/no-common-typos.js
```diff
@@ -0,0 +1,64 @@
...
+  valid: [{
+    code: [
+      'var Hello = React.createClass({',
+      '  render: function() {',
+      '    return <div dangerouslySetInnerHTML={{__html: "Hello World"}}/>;',
+      '  }',
+      '});'
+    ].join('\n'),
+    options: options,
+    parser: 'babel-eslint',
```
ljharb: why does this test need to use babel-eslint? all of these examples should use the default parser.

lib/rules/no-common-typos.js
koenpunt: Because config can be JavaScript too:

> Configuration Files - use a JavaScript, JSON or YAML
> http://eslint.org/docs/user-guide/configuring

lib/rules/no-common-typos.js
koenpunt: Why's that? To protect users from stupidity and accidental rogue expressions?

tests/lib/rules/no-common-typos.js
koenpunt: No idea, just used a different rule as boilerplate.

koenpunt:
Also, I'm not sure about the schema.

Currently it requires patterns like so:

```js
"no-common-typos": [{
  pattern: /somepattern/i,
  correct: "SomePattern"
}]
```

But maybe a format like this is better:

```js
"no-common-typos": {
  "SomePattern": /somepattern/i,
  "OtherPattern": /otherpattern/i
}
```

lib/rules/no-common-typos.js
ljharb: Yes, exactly that.

lib/rules/no-common-typos.js
ljharb: Then `instanceof` is not reliable (due to cross-realm issues) and so we should be converting it to a string and parsing it.

ljharb: Either way, the schema should require it be a string, since rule schemas don't permit JS-only values, afaik.

tests/lib/rules/no-common-typos.js
ljharb: The default parser, but when `ecmaVersion` set to 6 or higher, and `jsx` set to true.

It should use the default parser because that's the important one it works with - if you also want to test that it works with babel-eslint that's fine, but that's secondary.

lib/rules/no-common-typos.js
koenpunt: With fixing not enabled this shouldn't be an issue IMO

ljharb: users using a shared config that then run with `--fix` makes it an issue.

ljharb: I think that rules should not have ways of configuring them that don't work in all three contexts - in other words, not a regex ever, only a string (or other formats that can work in JSON or YAML). This is also eslint core's policy.

koenpunt: I changed it already so that regular expressions are always treated as string. Updated my previous comment to reflect that.

ljharb: I'm saying that the user of eslint should be protected against a malicious or accidental config, that might come from a plugin or a shared config.

Which means that the person who writes the pattern can't be trusted to mark it as safe - and neither can the user - which means the rule itself has to ensure that nothing unsafe can possibly happen.

koenpunt: For example:

```js
"no-common-typos": [{
  "case-sensitive": {
    "somepattern": "SomePattern"
  },
  "exact": {
    "colour": "color"
  }
}]

ljharb: That's certainly one way to do it - or alternatively, `[{ pattern: "SomePattern": caseInsensitive: true }]` perhaps?

lencioni:
Would it be more useful to have a rule that knows what all of the DOM element props can be, and if you use something that isn't one of those it warns? That would have the advantages of working for everyone out of the box, and being specific for each individual node type.

ljharb:
It's also worth noting that React itself does this for all DOM elements in React v15.3 and later.

jseminck:
This can probably be closed now that we have the `no-typos` rule? Any specific cases that we want to add/support can go to that rule instead.

ljharb:
Indeed, this is a bit duplicated by #1209.

Happy to reopen if it can be adapted.

PR #982, opened 2016-11-29
Indicate that jsx-indent is fixable (README)

captbaritone:
The rule's docs specify that this rule is fixable, but the readme did not.


lencioni:
Thanks!

PR #984, opened 2016-11-30
Fix `plugin` typo in configs

jamischarles:
ESLint looks for the `plugins` config property. Currently, if you extend `plugin:react/recommended` you still need to add `plugins: ['react']`.

This commit fixes the typo, so you no longer need to add the plugins prop when you extend `recommended` or `all` configs.

Before:
```
 1:1  error  Definition for rule 'react/no-find-dom-node' was not found          react/no-find-dom-node
 1:1  error  Definition for rule 'react/display-name' was not found              react/display-name
 1:1  error  Definition for rule 'react/jsx-no-undef' was not found              react/jsx-no-undef
```

After:
```
 28:21  error  'items' is missing in props validation      react/prop-types
 28:27  error  'items.map' is missing in props validation  react/prop-types
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -94,7 +94,7 @@ module.exports = {
   rules: allRules,
   configs: {
     recommended: {
-      plugin: [
+      plugins: [
         'react'
       ],
       parserOptions: {
@@ -120,7 +120,7 @@ module.exports = {
       }
     },
     all: {
-      plugin: [
+      plugins: [
         'react'
       ],
       parserOptions: {


ljharb:
@jamischarles what version of eslint are you using? I thought both `plugin` and `plugins` were supported.

jamischarles:
Using 3.7 I believe. I can take a look at the source tomorrow and verify.

ljharb:
It would be good to confirm whether this worked in some versions of eslint, or if it never worked.

jamischarles:
@ljharb As far as I can tell this never worked. I went back as far as 2.x: https://github.com/eslint/eslint/blob/2.x/lib/config/config-file.js#L504  

I was unable to find any code that references plugin. The only place `plugin` is singular is the cli: option `  --plugin [String]            Specify plugins`

ljharb:
Thanks, good enough for me.

jamischarles:
Thx 👍

PR #989, opened 2016-12-02
Don't treat async functions as components

taion:
This is just a silly little thing – there's no way for async functions to be components.

### DIFF
--- a/tests/lib/rules/prop-types.js
+++ b/tests/lib/rules/prop-types.js
@@ -1361,6 +1361,30 @@ ruleTester.run('prop-types', rule, {
       ].join('\n'),
       options: [{skipUndeclared: false}],
       parserOptions: parserOptions
+    }, {
+      // Async functions can't be components.
+      code: [
+        'var Hello = async function(props) {',
+        '  return <div>Hello {props.name}</div>;',
+        '}'
+      ].join('\n'),
+      parser: 'babel-eslint'
+    }, {
+      // Async functions can't be components.
+      code: [
+        'async function Hello(props) {',
+        '  return <div>Hello {props.name}</div>;',
+        '}'
+      ].join('\n'),
+      parser: 'babel-eslint'
+    }, {
+      // Async functions can't be components.
+      code: [
+        'var Hello = async (props) => {',
+        '  return <div>Hello {props.name}</div>;',
+        '}'
+      ].join('\n'),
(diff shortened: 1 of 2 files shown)


ljharb:
This seems good, but please note that in React Fiber, this probably won't be the case, as you'll be able to use a promise for a component as if it were a component.

tests/lib/rules/prop-types.js
```diff
@@ -1361,6 +1361,30 @@ ruleTester.run('prop-types', rule, {
...
       options: [{skipUndeclared: false}],
       parserOptions: parserOptions
+    }, {
+      // Async functions can't be components.
+      code: [
+        'var Hello = async function(props) {',
+        '  return <div>Hello {props.name}</div>;',
+        '}'
+      ].join('\n'),
+      parser: 'babel-eslint'
```
ljharb: there's no need to use babel-eslint here - let's test this with eslint's default parser, which fully support `async function` now.

taion: I was following the convention down the file in

```js
    }, {
      code: [
        'var Hello = function(props) {',
        '  return <div>Hello {props.name}</div>;',
        '}'
      ].join('\n'),
      parser: 'babel-eslint',
      errors: [{
        message: '\'name\' is missing in props validation'
      }]
    }, {
      code: [
        'function Hello(props) {',
        '  return <div>Hello {props.name}</div>;',
        '}'
      ].join('\n'),
      parser: 'babel-eslint',
      errors: [{
        message: '\'name\' is missing in props validation'
      }]
    }, {
      code: [
        'var Hello = (props) => {',
        '  return <div>Hello {props.name}</div>;',
        '}'
      ].join('\n'),
      parser: 'babel-eslint',
      errors: [{
        message: '\'name\' is missing in props validation'
      }]
    }, {
```

Should those be changed as well?

taion:
A promise for a component or a promise for an element?

ljharb:
@taion a promise for a component, is my understanding.

taion:
I'm not sure what you mean. A promise for a component _class_? These aren't that, and there are no component instances involved here anyway.

These would be component classes that try to render a promise for elements, which wouldn't be the same as the parent of one of these components trying to render a promise for elements.

yannickcr:
Merged, thanks!

PR #1005, opened 2016-12-19
Update jsx-closing-bracket-location.md

yooungt13:
the examples of 'after-props' and 'props-aligned' is wrong,  so it should change position them


yannickcr:
Thanks!

PR #1011, opened 2017-01-02
Check for createElement being called from React

### LINKED ISSUE #996: False positive with display-name rule
Seems like this: `document.createElement` makes the rule believe you're using `React.createElement`:

https://dxr.mozilla.org/mozilla-central/source/devtools/client/webconsole/new-console-output/test/fixtures/serviceContainer.js#16

jomasti:
Fixes #996. This will prevent document.createElement triggering a
false positive for react/display-name.

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -255,7 +255,20 @@ function componentRule(rule, context) {
         node[property] &&
         node[property].type === 'JSXElement'
       ;
+      var destructuredReactCreateElement = function () {
+        var variables = variableUtil.variablesInScope(context);
+        var variable = variableUtil.getVariable(variables, 'createElement');
+        if (variable !== null) {
+          var map = variable.scope.set;
+          // eslint-disable-next-line no-undef
+          if (map instanceof Map && map.has('React')) {
+            return true;
+          }
+        }
+        return false;
+      };
       var returnsReactCreateElement =
+        destructuredReactCreateElement() ||
         node[property] &&
         node[property].callee &&
         node[property].callee.object &&
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -258,10 +258,9 @@ function componentRule(rule, context) {
       var destructuredReactCreateElement = function () {
         var variables = variableUtil.variablesInScope(context);
         var variable = variableUtil.getVariable(variables, 'createElement');
-        if (variable !== null) {
+        if (variable) {
           var map = variable.scope.set;
-          // eslint-disable-next-line no-undef
-          if (map instanceof Map && map.has('React')) {
+          if (map.has('React')) {
             return true;
           }
         }
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -258,6 +258,8 @@ function componentRule(rule, context) {
       var returnsReactCreateElement =
         node[property] &&
         node[property].callee &&
+        node[property].callee.object &&
+        node[property].callee.object.name === 'React' &&
         node[property].callee.property &&
         node[property].callee.property.name === 'createElement'
       ;
--- a/lib/util/variable.js
+++ b/lib/util/variable.js
@@ -11,16 +11,9 @@
  * @returns {Boolean} True if the variable was found, false if not.
  */
 function findVariable(variables, name) {
-  var i;
-  var len;
-
-  for (i = 0, len = variables.length; i < len; i++) {
-    if (variables[i].name === name) {
-      return true;
-    }
-  }
-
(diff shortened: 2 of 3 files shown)


lib/util/Components.js
```diff
@@ -258,6 +258,8 @@ function componentRule(rule, context) {
       var returnsReactCreateElement =
         node[property] &&
         node[property].callee &&
+        node[property].callee.object &&
+        node[property].callee.object.name === 'React' &&
```
ljharb: won't this prevent `import { createElement } from 'react';` from being detected?

jomasti: You're right. I found a way to check for this by using the ModuleScope of the Variable when importing `createElement` like that. However, it involves using Map, and it doesn't seem like I can get away with that with the project setup.

jomasti: I'll just temporarily add in a disable comment so I can feel good about pushing it up. I'll remove it after receiving feedback.

lib/util/Components.js
```diff
@@ -255,9 +255,24 @@ function componentRule(rule, context) {
         node[property] &&
         node[property].type === 'JSXElement'
       ;
+      var destructuredReactCreateElement = function () {
+        var variables = variableUtil.variablesInScope(context);
+        var variable = variableUtil.getVariable(variables, 'createElement');
+        if (variable !== null) {
+          var map = variable.scope.set;
+          // eslint-disable-next-line no-undef
+          if (map instanceof Map && map.has('React')) {
```
ljharb: There's no need to use Map if the keys are strings; you can just use a normal object.

lib/util/variable.js
```diff
@@ -24,6 +24,25 @@ function findVariable(variables, name) {
...
+  var i;
+  var len;
+
+  for (i = 0, len = variables.length; i < len; i++) {
+    if (variables[i].name === name) {
+      return variables[i];
+    }
+  }
+
+  return null;
```
ljharb:
```js
return variables.find(function (variable) { return variable.name === name; });
```

tests/lib/rules/display-name.js
```diff
@@ -548,5 +561,16 @@ ruleTester.run('display-name', rule, {
     errors: [{
       message: 'Component definition is missing display name'
     }]
+  }, {
+    code: [
+      'import React, { createElement } from "react";',
```
ljharb: Let's also add cases for `var { createElement } = React;` and `var createElement = React.createElement;`.

Probably too difficult to cover `var c = React.createElement;` tho.

lib/util/variable.js
jomasti: I prefer this, so I'll go with it. I was mimicking the `findVariable` function already there.

ljharb: might as well update both then ;-)

lib/util/Components.js
jomasti: I might be misunderstanding, but that is what it returns when I pull it out of the ModuleScope. Wouldn't I need to convert it to an object to use it like one?

ljharb: oops, sorry, if that's what it's doing then this is correct :-)

ljharb: altho i'm confused about the `instanceof Map` check - when will that be false? eslint requires node 4 or later.

lib/util/variable.js
```diff
@@ -11,16 +11,21 @@
...
 
-  return false;
+/**
+ * Find and return a particular variable in a list
+ * @param {Array} variables The variables list.
+ * @param {Array} name The name of the variable to search.
+ * @returns {Object} Variable if the variable was found, null if not.
+ */
+function getVariable(variables, name) {
+  return variables.find(function (variable) {
```
lencioni: This seems to have broken the build for Node < 4. We have a PR that removes support for Node < 4 (#1038) that will un-break this. @ljharb do you think we will merge this soon or should I quick fix this line? I'm in favor of merging your PR now.

ljharb: No, we absolutely must fix this first, since otherwise it's a semver-major change, and we need to publish master *before* a semver-major line.

lencioni: Sounds good. I'll put up a PR shortly.

ljharb: No worries; I pushed up https://github.com/yannickcr/eslint-plugin-react/commit/e332b08694de6271674bdf2182f03b399306ba82 to fix it

lencioni: Oh nevermind looks like you fixed it in e332b08

Issue #922, opened 2016-10-20
Add onlyImported flag for jsx-no-undef


nkt:
There is component named `Text` in my project. ESLint doesn't tell me that `Text` is `undefined` because it's existing class in the [DOM API](https://developer.mozilla.org/en-US/docs/Web/API/Text).
I suggest add rule which deny use components defined globally:

``` json
{
  "rules": {
    "react/jsx-no-undef": ["error", { "onlyImported": true }]
  }
}
```

``` js
import Text from './Text';

<Text /> // ok
```

``` js
<Text /> // react/jsx-no-undef: 'Text' component not found in current scope.
```

ljharb:
This is a great solution to this problem for React.

alexzherdev:
Fixed in #1013 ?

PR #1013, opened 2017-01-05
react/jsx-no-undef allowGlobals option

jomasti:
This should allow for the use case outlined in #922.

With the `disallowGlobals` option enabled, the global scope will be ignored
when checking for defined components. Without this option enabled,
Pascal case global variables will be allowed when checking for defined
components.

### DIFF
--- a/docs/rules/jsx-no-undef.md
+++ b/docs/rules/jsx-no-undef.md
@@ -10,12 +10,47 @@ The following patterns are considered warnings:
 <Hello name="John" />;
 ```
 
+```jsx
+// will ignore Text in the global scope and warn
+var Hello = React.createClass({
+  render: function() {
+    return <Text>Hello</Text>;
+  }
+});
+module.exports = Hello;
+```
+
+
 The following patterns are not considered warnings:
 
 ```jsx
 var Hello = require('./Hello');
 
 <Hello name="John" />;
+```
+
+## Rule Options
+
+```js
+...
+"react/jsx-no-undef": [<enabled>, { "allowGlobals": <boolean> }]
+...
+```
+
+### `allowGlobals`
+
+When `true` the rule will consider the global scope when checking for defined Components.
+
+The following patterns are considered okay and do not cause warnings:
+
+```jsx
+var Text = require('./Text');
+var Hello = React.createClass({
+  render: function() {
+    return <Text>Hello</Text>;
+  }
+});
+module.exports = Hello;
 ```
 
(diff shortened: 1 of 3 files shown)


tests/lib/rules/jsx-no-undef.js
```diff
@@ -56,6 +56,29 @@ ruleTester.run('jsx-no-undef', rule, {
       '}'
     ].join('\n'),
     parserOptions: parserOptions
+  }, {
+    code: '/*eslint no-undef:1*/ var React; React.render(<Text />);',
```
jomasti: I was following the tests that were already there. I removed it.

docs/rules/jsx-no-undef.md
```diff
@@ -18,6 +18,52 @@ var Hello = require('./Hello');
...
+...
+```
+
+### `importedOnly`
+
+When `true` the rule will ignore the global scope when checking for defined Components.
+
+The following patterns are considered okay and do not cause warnings:
+
+```js
```
lencioni: nit: this should be `jsx` (here and elsewhere)

docs/rules/jsx-no-undef.md
```diff
@@ -18,6 +18,52 @@ var Hello = require('./Hello');
 <Hello name="John" />;
 ```
 
+## Rule Options
+
+```js
+...
+"jsx-no-undef": [<enabled>, { "importedOnly": <boolean> }]
```
lencioni: nit: In #1071 we are adding `react/` to lines like this. It would be nice to get that in here if we touch this again.

docs/rules/jsx-no-undef.md
```diff
@@ -18,6 +18,52 @@ var Hello = require('./Hello');
...
+...
+```
+
+### `importedOnly`
```
lencioni: Why make this an option instead of having it be the default behavior?

ljharb: That seems like it'd be a breaking change - an option now, flipping the default later, seems safer

lencioni: Yeah, that makes sense. I'll open an issue for that.

lencioni: #1072

ljharb:
@jomasti @lencioni any objections to this being merged with my latest changes?

lencioni:
@ljharb go for it

Issue #1017, opened 2017-01-10
Rule `require-default-props` with Flow type definition assignment


wyze:
Using v6.9.0 with code like the following:

```js
// @flow

import type { Message } from '../types'

type Props = Message

const Component = ({ message }: Props) => (
  <div>{message}</div>
)

export default Component
```

Crashes with:

```
Cannot read property 'filter' of undefined
TypeError: Cannot read property 'filter' of undefined
    at getPropTypesFromTypeAnnotation ($PWD/node_modules/eslint-plugin-react/lib/rules/require-default-props.js:178:29)
    at Object.handleStatelessComponent ($PWD/node_modules/eslint-plugin-react/lib/rules/require-default-props.js:288:42)
    at EventEmitter.updatedRuleInstructions.(anonymous function) ($PWD/node_modules/eslint-plugin-react/lib/util/Components.js:581:75)
    at emitOne (events.js:101:20)
    at EventEmitter.emit (events.js:188:7)
    at NodeEventGenerator.enterNode ($PWD/node_modules/eslint/lib/util/node-event-generator.js:39:22)
    at CodePathAnalyzer.enterNode ($PWD/node_modules/eslint/lib/code-path-analysis/code-path-analyzer.js:607:23)
    at CommentEventGenerator.enterNode ($PWD/node_modules/eslint/lib/util/comment-event-generator.js:97:23)
    at Controller.enter ($PWD/node_modules/eslint/lib/eslint.js:928:36)
    at Controller.__execute ($PWD/node_modules/estraverse/estraverse.js:397:31)
```

I have a fix for this, and will submit the PR shortly.

wyze:
Also, in debugging this, it looks like it is not specific to importing the type. It will also throw the same error if you do something like this:

```js
// @flow

type Message = {
  message: string,
}
type Props = Message

const Component = ({ message }: Props) => (
  <div>{message}</div>
)

export default Component
```

wyze:
Closing as the PR to fix was merged and released in 6.10.0.

PR #1018, opened 2017-01-11
Fix `require-default-props` rule when using Flow type from assignment

wyze:
This fixes the issue I reported in #1017.

The condition I added is hit when code like the following is used:

```js
type Props = Message
```

This just takes the right side of the annotation (`Message`) and tries to find that variable in scope and then continues on.

### DIFF
--- a/lib/rules/require-default-props.js
+++ b/lib/rules/require-default-props.js
@@ -141,7 +141,12 @@ module.exports = {
       switch (node.typeAnnotation.type) {
         case 'GenericTypeAnnotation':
           var annotation = resolveGenericTypeAnnotation(node.typeAnnotation);
-          properties = annotation ? annotation.properties : [];
+
+          if (annotation && annotation.id) {
+            annotation = findVariableByName(annotation.id.name);
+          }
+
+          properties = annotation ? (annotation.properties || []) : [];
           break;
 
         case 'UnionTypeAnnotation':
--- a/tests/lib/rules/require-default-props.js
+++ b/tests/lib/rules/require-default-props.js
@@ -676,6 +676,39 @@ ruleTester.run('require-default-props', rule, {
         '}'
       ].join('\n'),
       parser: 'babel-eslint'
+    },
+    {
+      code: [
+        'import type ImportedProps from "fake";',
+        'type Props = ImportedProps;',
+        'function Hello(props: Props) {',
+        '  return <div>Hello {props.name.firstname}</div>;',
+        '}'
+      ].join('\n'),
+      parser: 'babel-eslint'
+    },
+    // don't error when variable is not in scope
+    {
+      code: [
+        'import type { ImportedType } from "fake";',
+        'type Props = ImportedType;',
+        'function Hello(props: Props) {',
+        '  return <div>Hello {props.name.firstname}</div>;',
+        '}'
+      ].join('\n'),
+      parser: 'babel-eslint'
+    },
+    // make sure error is not thrown with multiple assignments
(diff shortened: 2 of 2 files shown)


wyze:
Hey @EvNaverniouk, can you check the failing case I added? Seems a bit off as I needed to provide `errors: []` because it is under `invalid` key in the test suite, but it doesn't actually output any errors.

tests/lib/rules/require-default-props.js
```diff
@@ -1718,6 +1728,17 @@ ruleTester.run('require-default-props', rule, {
...
+    {
+      code: [
+        'import type { ImportedType } from "fake";',
+        'type Props = ImportedType;',
+        'function Hello(props: Props) {',
+        '  return <div>Hello {props.name.firstname}</div>;',
+        '}'
+      ].join('\n'),
+      parser: 'babel-eslint',
+      errors: []
```
EvNaverniouk: We should update this to return errors so we can test for it, perhaps in the event that the type isn't defined. Maybe something like this:

```
function Hello (props: UndefinedVariable) {
    return <div>Hello {props.name.firstname}</div>;
}
```

wyze: When you say "update this", are you referring to the rule, to throw an error if it can't find a variable? If so, I think that should be a separate PR.

If you are referring to the test case when you say "update this", then when making the requested changes, the invalid test still passes with `errors: []` so I am not sure there is much difference between what you are suggesting and what I added.

EvNaverniouk: Perhaps I'm misunderstanding what the scope of the changes in this PR. I was under the impression that it adds support for Flow types as valid prop assignments via variables. Your original test case covers the fact the the rule properly catches such valid Flow type assignments without error.

What I'm asking for is to add a test case that when a Flow type is defined, but the Flow type is variable is unassigned. I believe this ESLint rule should still throw an error in that case.

ljharb: That's covered by the no-undef rules, is it not?

yannickcr: Yes, this should be covered by no-undef. The require-default-props rule should just try its bests to resolve the props and the defaultProps, and give up if it cannot do it.

If this test case do not return any error then it should be in the "valid" array.

lib/rules/require-default-props.js
```diff
@@ -141,6 +141,11 @@ module.exports = {
       switch (node.typeAnnotation.type) {
         case 'GenericTypeAnnotation':
           var annotation = resolveGenericTypeAnnotation(node.typeAnnotation);
+
+          if (annotation && annotation.id) {
+            annotation = findVariableByName(annotation.id.name);
+          }
+
           properties = annotation ? annotation.properties : [];
```
yannickcr: You should also update the next line to check for `annotation.properties` existence, or the rule will keep crashing if there is multiple reassignments

```js
properties = annotation && annotation.properties ? annotation.properties : [];
```

ljharb: What would be a test case that would trigger that?

wyze: Well, technically, my fix prevented this, but if for some reason you do multiple assignments, it would trigger the original error again:

```js
import type ImportedProps from "fake";
type NestedProps = ImportedProps;
type Props = NestedProps;
function Hello(props: Props) {
  return <div>Hello {props.name.firstname}</div>;
}
```

cr0cK:
up!

PR #1025, opened 2017-01-13
show context in pure component for clarity

netpoetica:
ESLint will throw a very unclear complaint to somebody who converts a regular React component into a Pure Component (stateless) as instructed in this doc, if contextTypes are defined. Because there is no "this.context" accessible, it's unclear to someone that they need to use `props, context` arguments.

This will just help clarify and probably save a couple of people a few minutes of research if they come from an ESLint error. Cheers!

### DIFF
--- a/docs/rules/prefer-stateless-function.md
+++ b/docs/rules/prefer-stateless-function.md
@@ -27,7 +27,11 @@ var Hello = React.createClass({
 The following pattern is not considered a warning:
 
 ```jsx
-const Foo = function(props) {
+const Foo = function(props, context) {
+  const {
+    location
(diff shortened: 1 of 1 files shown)


ljharb:
Just a terminology nit:  a pure component is one that has no state and depends only on props treated as immutable - an SFC might be impure and a Component might be pure. There's `extends PureComponent` as well which indicates to React that your class-based component is pure.

netpoetica:
Thanks! Yes, I agree with you - people should refer to the docs. But, in a pinch, someone who comes here first won't get stuck and open a ticket for you later ;) cheers, thanks for clarifying

PR #1038, opened 2017-01-21
[Breaking] drop node < 4 support

ljharb:
I've added some commits recently which fix support down to node 0.10 - **those must be published in the 6.x line prior to merging this.**

Merging this will be a semver-major change - it drops < node 4 support, and drops support for eslint v2. In the process, we'll be able to use any syntax and API that node 4 supports, yay!

I also enabled rules requiring template literals instead of concatenation, and autofixed them, both because I think it's a superior approach, and because it illustrates some of the new syntax we'll get to use :-)

(I'll manually keep this rebased and updated as time goes by, so ideally it will always be ready to merge when v7 comes along)


yannickcr:
Wow! Great! Thanks for all this work. I'll try to review this PR tomorrow.

ljharb:
master must absolutely be published with v6, to fix the incorrect breaking changes in it, before this is merged and v7 is ready.

ljharb:
@yannickcr please don't release v7 before merging all the v7-tagged PRs; I'll update this one shortly and merge it today.

yannickcr:
Ho, I was more in the idea the release the v7 to be React 15.5.0 compatible and to postpone theses breaking changes for the v8.

But I'm ok to do all of this in a single major release.

ljharb:
@yannickcr ah, i guess that would work too; the v7.0.0 milestone wasn't edited to v8 so that didn't come across. I think if we're doing a breaking change we might as well get them all in tho - I would have expected react 15.5 compatibility to be non-breaking, and to go in v6.

yannickcr:
> the v7.0.0 milestone wasn't edited to v8 so that didn't come across
yeah, my bad 😐 
> I think if we're doing a breaking change we might as well get them all in tho
right, it will avoid people some pain to only have one migration to do
> I would have expected react 15.5 compatibility to be non-breaking, and to go in v6
I think this is unavoidable with the changes to `createClass` and the warnings added to `no-deprecated`

ljharb:
Couldn't we pivot on the react version in settings tho?

ljharb:
I figured the deprecation would only apply to 15.5 and later.

PR #1039, opened 2017-01-23
Add option style to jsx-wrap-multilines rule

dropfen:
Option style can be either "always" or "never". Default is "always", means wrap jsx in parentheses. Never disallows parentheses around jsx.

### DIFF
--- a/docs/rules/jsx-wrap-multilines.md
+++ b/docs/rules/jsx-wrap-multilines.md
@@ -1,10 +1,23 @@
 # Prevent missing parentheses around multiline JSX (jsx-wrap-multilines)
 
-Wrapping multiline JSX in parentheses can improve readability and/or convenience. It optionally takes a second parameter in the form of an object, containing places to apply the rule. By default, `"declaration"`, `"assignment"`, and `"return"` syntax is checked, but these can be explicitly disabled. Any syntax type missing in the object will follow the default behavior (become enabled).
+Wrapping multiline JSX in parentheses can improve readability and/or convenience*. It optionally takes a second parameter in the form of an object, containing places to apply the rule. By default, `"declaration"`, `"assignment"`, and `"return"` syntax is checked, but these can be explicitly disabled. Any syntax type missing in the object will follow the default behavior (become enabled).
+
+* however, the default style option is "always", this means use parentheses. But you can also set it to "never" and disallow parentheses around your jsx if want to.
 
 **Fixable:** This rule is automatically fixable using the `--fix` flag on the command line.
 
 ## Rule Details
+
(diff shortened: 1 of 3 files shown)


docs/rules/jsx-wrap-multilines.md
```diff
@@ -1,11 +1,24 @@
 # Prevent missing parentheses around multiline JSX (jsx-wrap-multilines)
 
-Wrapping multiline JSX in parentheses can improve readability and/or convenience. It optionally takes a second parameter in the form of an object, containing places to apply the rule. By default, `"declaration"`, `"assignment"`, and `"return"` syntax is checked, but these can be explicitly disabled. Any syntax type missing in the object will follow the default behavior (become enabled).
+Wrapping multiline JSX in parentheses can improve readability and/or convenience*. It optionally takes a second parameter in the form of an object, containing places to apply the rule. By default, `"declaration"`, `"assignment"`, and `"return"` syntax is checked, but these can be explicitly disabled. Any syntax type missing in the object will follow the default behavior (become enabled).
```
ljharb: Using a `*` here won't translate to a footnote below - I think perhaps leaving this paragraph alone, and adding a second paragraph that describes "never", would be better.

docs/rules/jsx-wrap-multilines.md
```diff
@@ -1,11 +1,24 @@
...
+```js
+"react/jsx-wrap-multilines": ["error", {
+  "style": "always|never", // default "always"
+  "assignment": bool, // default true
+  "declaration": bool, // default true
+  "return": bool, // default true
+}]
+```
+
 The following patterns are considered warnings:
```
ljharb: we'll need to add "considered warnings" and "not considered warnings" examples for the "never" option as well.

docs/rules/jsx-wrap-multilines.md
```diff
@@ -1,11 +1,24 @@
...
 
 **Fixable:** This rule is automatically fixable using the `--fix` flag on the command line.
 
 ## Rule Details
 
+Options
+
+```js
+"react/jsx-wrap-multilines": ["error", {
+  "style": "always|never", // default "always"
```
ljharb: I don't think adding an object property is the best schema here - I think we want to be able to do:
```
["error", "never", {
  …
}]


["error", "always", {
  …
}]


["error", { // implies "always"
  …
}]
```

dropfen: Yeah.. agree, I'll try to do it tomorrow, with the docs stuff.

golopot:
Superseded by #1475

PR #1055, opened 2017-01-31
Add forbid-foreign-prop-types rule

### LINKED ISSUE #696: Rule proposal: disallow referencing another component's propTypes
Some people might want to use [babel-plugin-transform-react-remove-prop-types](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types) to remove propTypes from their components in production builds, as an optimization. This transform changes code like:

``` jsx
const Baz = () => (
  <div />
);

Baz.propTypes = {
  foo: React.PropTypes.string
};
```

into:

``` jsx
const Baz = () => (
  <div />
);
```

This works, but might end up breaking things in production if you import a component and then reference that component's propTypes. Instead, the propTypes should be separately exported in this case and referenced directly from the export.

Bad:

``` jsx
import { pick } from 'lodash';

import SomeComponent from './SomeComponent';

export default function AnotherComponent(props) {
  const passedProps = pick(props, Object.keys(SomeComponent.propTypes));
  return (
    <SomeComponent {...passedProps} />
  );
};
```

Good:

``` jsx
import { pick } from 'lodash';

import { propTypes: someComponentPropTypes }, SomeComponent from './SomeComponent';

export default function AnotherComponent(props) {
  const passedProps = pick(props, Object.keys(someComponentPropTypes));
  return (
    <SomeComponent {...passedProps} />
  );
};
```

This rule would make the aforementioned Babel plugin safe to use.

iancmyers:
People may want to use babel-plugin-transform-react-remove-prop-types to remove propTypes from their components in production builds, as an optimization. The `forbib-foreign-prop-types` rule forbids using another component's prop types unless they are explicitly imported/exported, which makes that optimization less prone to error.

I've included the use-cases described in #696 as tests. I'd be happy to add additional test cases if recommended.

Fixes #696

### DIFF
--- a/docs/rules/forbid-foreign-prop-types.md
+++ b/docs/rules/forbid-foreign-prop-types.md
@@ -1,0 +1,30 @@
+# Forbid foreign propTypes (forbid-foreign-prop-types)
+
+This rule forbids using another component's prop types unless they are explicitly imported/exported. This allows people who want to use [babel-plugin-transform-react-remove-prop-types](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types) to remove propTypes from their components in production builds, to do so safely.
+
+In order to ensure that imports are explicitly exported it is recommended to use the ["named" rule in eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/named.md) in conjunction with this rule.
+
+## Rule Details
+
+This rule checks all objects and ensures that the `propTypes` property is not used.
+
+The following patterns are considered warnings:
+
+```js
+import SomeComponent from './SomeComponent';
+SomeComponent.propTypes;
+
+var { propTypes } = SomeComponent;
+
+SomeComponent['propTypes'];
+```
+
+The following patterns are not considered warnings:
+
+```js
+import SomeComponent, {propTypes as someComponentPropTypes} from './SomeComponent';
+```
+
+## When not to use
+
+This rule aims to make a certain production optimization, removing prop types, less prone to error. This rule may not be relevant to you if you do not wish to make use of this optimization.
--- a/lib/rules/forbid-foreign-prop-types.js
+++ b/lib/rules/forbid-foreign-prop-types.js
@@ -1,0 +1,63 @@
+/**
+ * @fileoverview Forbid using another component's propTypes
+ * @author Ian Christian Myers
+ */
+'use strict';
+
+var find = require('array.prototype.find');
+
+// ------------------------------------------------------------------------------
+// Constants
+// ------------------------------------------------------------------------------
+
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Forbid using another component\'s propTypes',
+      category: 'Best Practices',
+      recommended: false
+    }
+  },
+
+  create: function(context) {
+    // --------------------------------------------------------------------------
+    // Helpers
+    // --------------------------------------------------------------------------
+
+    function isLeftSideOfAssignment(node) {
+      return node.parent.type === 'AssignmentExpression' && node.parent.left === node;
+    }
+
+    return {
+      MemberExpression: function(node) {
+        if (!node.computed && node.property && node.property.type === 'Identifier' &&
+            node.property.name === 'propTypes' && !isLeftSideOfAssignment(node) ||
+            node.property && node.property.type === 'Literal' &&
+            node.property.value === 'propTypes' && !isLeftSideOfAssignment(node)) {
+          context.report({
+            node: node.property,
+            message: 'Using another component\'s propTypes is forbidden'
+          });
+        }
+      },
+
+      ObjectPattern: function(node) {
+        var propTypesNode = find(node.properties, function(property) {
+          return property.type === 'Property' && property.key.name === 'propTypes';
+        });
+
+        if (propTypesNode) {
+          context.report({
+            node: propTypesNode,
+            message: 'Using another component\'s propTypes is forbidden'
+          });
+        }
+      }
+    };
+  }
+};
--- a/tests/lib/rules/forbid-foreign-prop-types.js
+++ b/tests/lib/rules/forbid-foreign-prop-types.js
@@ -1,0 +1,156 @@
+/**
+ * @fileoverview Tests for forbid-foreign-prop-types
+ */
+'use strict';
+
+// -----------------------------------------------------------------------------
+// Requirements
+// -----------------------------------------------------------------------------
+
+var rule = require('../../../lib/rules/forbid-foreign-prop-types');
+var RuleTester = require('eslint').RuleTester;
+
+var parserOptions = {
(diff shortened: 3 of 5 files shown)


docs/rules/forbid-foreign-prop-types.md
```diff
@@ -0,0 +1,28 @@
+# Forbid foreign propTypes (forbid-foreign-prop-types)
+
+This rule forbids using another component's prop types unless they are explicitly imported/exported. This allows people who want to use babel-plugin-transform-react-remove-prop-types to remove propTypes from their components in production builds, to do so safely.
```
lencioni: It would be nice to link to the babel plugin here.

docs/rules/forbid-foreign-prop-types.md
```diff
@@ -0,0 +1,28 @@
...
+
+var { propTypes } = SomeComponent;
+
+SomeComponent['propTypes'];
+```
+
+The following patterns are not considered warnings:
+
+```js
+import {propTypes: someComponentPropTypes}, SomeComponent from './SomeComponent';
```
lencioni: Unfortunately, even importing like this does not ensure that `propTypes` is explicitly exported from the module (and not just a property on the exported object).

What do you think about making this rule additionally ensure that propTypes is explicitly exported when being imported as well? Alternatively, if there is a rule in eslint-plugin-import that covers this, we should mention it here.

Also, this line should be 

```js
import SomeComponent, {propTypes as someComponentPropTypes} from './SomeComponent';
```

iancmyers: @lencioni Looks like eslint-plugin-import has the `named` rule which would ensure that named imports are also named exports: https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/named.md

lib/rules/forbid-foreign-prop-types.js
```diff
@@ -0,0 +1,52 @@
...
+            node.property && node.property.type === 'Literal' && node.property.value === 'propTypes') {
+          context.report({
+            node: node.property,
+            message: 'Using another component\'s propTypes is forbidden'
+          });
+        }
+      },
+
+      ObjectPattern: function(node) {
+        var propTypesNode = node.properties.find(function(property) {
```
ljharb: eslint-plugin-react still supports node 0.10, so please use `array.prototype.find` here instead (after this is merged, i'll update #1038 to switch back to `.find`)

tests/lib/rules/forbid-foreign-prop-types.js
```diff
@@ -0,0 +1,102 @@
...
+
+var parserOptions = {
+  ecmaVersion: 6,
+  ecmaFeatures: {
+    experimentalObjectRestSpread: true,
+    jsx: true
+  }
+};
+
+require('babel-eslint');
```
ljharb: none of the code in this test requires object rest/spread or babel-eslint; could these tests just stick with espree?

lib/rules/forbid-foreign-prop-types.js
```diff
@@ -0,0 +1,52 @@
...
+    }
+  },
+
+  create: function(context) {
+
+    return {
+      MemberExpression: function(node) {
+        if (node.property && node.property.type === 'Identifier' && node.property.name === 'propTypes' ||
+            node.property && node.property.type === 'Literal' && node.property.value === 'propTypes') {
+          context.report({
```
ljharb: it'd be helpful if the error could include the position of the warning, so that editors can jump the cursor right to it

docs/rules/forbid-foreign-prop-types.md
lencioni: This is great! Will you please add a note about this and a link to the rule in the documentation?

tests/lib/rules/forbid-foreign-prop-types.js
```diff
@@ -0,0 +1,102 @@
...
+var ruleTester = new RuleTester();
+ruleTester.run('forbid-foreign-prop-types', rule, {
+
+  valid: [{
+    code: 'import { propTypes } from "SomeComponent";',
+    parser: 'babel-eslint'
+  }, {
+    code: 'import { propTypes as someComponentPropTypes } from "SomeComponent";',
+    parser: 'babel-eslint'
+  }],
```
lencioni: It might be worth adding some valid cases like:

```js
const foo = propTypes
foo(propTypes)
foo + propTypes
const foo = [propTypes]
const foo = { propTypes }
```

And maybe even junk like this (note the lowercase):

```js
foo.propTypes
foo['propTypes']
const { propTypes } = foo
```

ljharb: I'm not sure why lowercase matters - I agree that cases where it's an object or array literal should be valid, but `class foo extends React.Component` should have the same restrictions (someone could always `import Foo from './foo'` and use it just fine)

ljharb: I think this plugin already has some component detection logic we could reuse?

lencioni: I was thinking lowercase should perhaps be allowed because in React, components must be uppercase. Lowercase means DOM element. But this distinction is unlikely to matter much here. Reusing component detection logic sounds good to me.

oliviertassinari:
@iancmyers Thanks for working on that!
I have tried [an iteration](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types/pull/81) at the babel plugin with [no luck](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types/issues/80).

react-native is encouraging the following pattern. We might need an exception.
```jsx
import {View} from 'react-native';

class CustomView extends React.Component {
  static propTypes = {
    style: View.propTypes.style
  };
  render() {
    return (
      <View style={this.props.style}>..</View>
    );
  }
}
```

ljharb:
That's just static class properties (which are a stage 2 proposal) - definitely those should be covered as well, but since it's just sugar for `CustomView.propTypes = …`, it shouldn't be difficult.

iancmyers:
@lencioni @ljharb I've taken another pass at this and addressed much of your feedback.

lib/rules/forbid-foreign-prop-types.js
```diff
@@ -0,0 +1,63 @@
...
+    }
+  },
+
+  create: function(context) {
+    // --------------------------------------------------------------------------
+    // Helpers
+    // --------------------------------------------------------------------------
+
+    function isLeftSideOfAssignment(node) {
+      return node.parent.type === 'AssignmentExpression' && node.parent.left === node;
```
ljharb: will a node always have a `.parent` property?

iancmyers: I believe that `MemberExpression` and `ObjectPattern` will always at least have a parent of `Program`.

lencioni:
@iancmyers thanks for your contribution!

@oliviertassinari Thanks for raising this scenario. I'm going to merge this as it is, and we can follow up with an improvement to address that if we need to. One thought is to allow referencing propTypes on objects imported from packages, but we can discuss the details in a separate issue or PR.

PR #1058, opened 2017-02-01
[Fix] jsx-indent with tabs

### LINKED ISSUE #1057: jsx-indent with tabs fails to error/fix
Here's [a repo](https://github.com/kentcdodds/jsx-indent-bug) that demonstrates this issue.

**Versions**:

- `eslint`: 3.14.1
- `eslint-plugin-react`: 6.9.0

**Config**:

```javascript
{
  "plugins": [
    "react"
  ],
  "parserOptions": {
    "ecmaVersion": 6,
    "ecmaFeatures": {
      "jsx": true
    }
  },
  "rules": {
    "react/jsx-indent": ["error", "tab"],
    "indent": ["error", "tab"],
  }
}
```

**Input**:

```javascript
const React = require('react')
module.exports = MyComponent

function MyComponent(props) {
  return (
    <div
      className="foo-bar"
      id="thing"
    >
      Hello world!
    </div>
  )
}
```

**lint output**:

```
<path>/jsx-indent-bug/index.js
  5:3  error  Expected indentation of 1 tab but found 2 spaces     indent
  6:5  error  Expected indentation of 1 tab character but found 0  react/jsx-indent
```

**`--fix`ed file**:

```javascript
const React = require('react')
module.exports = MyComponent

function MyComponent(props) {
	return (
		<div
      className="foo-bar"
      id="thing"
    >
      Hello world!
    </div>
	)
}
```

After the `--fix`, there are no more linting errors. However this is clearly not indenting lines 7-11 properly.

I'll start working on a PR. I'm pretty sure I can make a failing test, but I'm not certain how I'd fix the issue. Any tips appreciated!

kentcdodds:
This fixes the `--fix` for `jsx-indent`. Specifically my issue was with tabs (Fixes #1057), but this fixes the same issues with spaces as well.

Highlevel important bits:
1. Adds fix for when the ending of a tag (`>`) is on its own line. This will now lint and fix properly.
2. Adds fix for linting and fixing Literal strings inside JSX
3. Adds configuration option called `indentLogicalExpressions` which will allow for indenting the right side of logical expressions (documented)
4. Adds fix for linting and fixing the indentation of the closing tag.

### DIFF
--- a/lib/rules/jsx-indent.js
+++ b/lib/rules/jsx-indent.js
@@ -46,6 +46,14 @@ module.exports = {
       }, {
         type: 'integer'
       }]
+    }, {
+      type: 'object',
+      properties: {
+        indentLogicalExpressions: {
+          type: 'boolean'
+        }
+      },
+      additionalProperties: false
     }]
   },
 
@@ -56,6 +64,7 @@ module.exports = {
     var extraColumnStart = 0;
     var indentType = 'space';
     var indentSize = 4;
+    var indentLogicalExpressions = false;
 
     var sourceCode = context.getSourceCode();
 
@@ -67,24 +76,25 @@ module.exports = {
         indentSize = context.options[0];
         indentType = 'space';
       }
+      if (context.options[1]) {
+        indentLogicalExpressions = context.options[1].indentLogicalExpressions || false;
+      }
     }
 
     var indentChar = indentType === 'space' ? ' ' : '\t';
 
     /**
      * Responsible for fixing the indentation issue fix
-     * @param {ASTNode} node Node violating the indent rule
+     * @param {Boolean} rangeToReplace is used to specify the range
+     *   to replace with the correct indentation.
      * @param {Number} needed Expected indentation character count
      * @returns {Function} function to be executed by the fixer
      * @private
      */
-    function getFixerFunction(node, needed) {
+    function getFixerFunction(rangeToReplace, needed) {
       return function(fixer) {
         var indent = Array(needed + 1).join(indentChar);
-        return fixer.replaceTextRange(
-          [node.start - node.loc.start.column, node.start],
-          indent
-        );
+        return fixer.replaceTextRange(rangeToReplace, indent);
       };
     }
 
@@ -93,46 +103,38 @@ module.exports = {
      * @param {ASTNode} node Node violating the indent rule
      * @param {Number} needed Expected indentation character count
      * @param {Number} gotten Indentation character count in the actual node/code
-     * @param {Object} loc Error line and column location
+     * @param {Array} rangeToReplace is used in the fixer.
+     *   Defaults to the indent of the start of the node
+     * @param {Object} loc Error line and column location (defaults to node.loc
      */
-    function report(node, needed, gotten, loc) {
+    function report(node, needed, gotten, rangeToReplace, loc) {
       var msgContext = {
         needed: needed,
         type: indentType,
         characters: needed === 1 ? 'character' : 'characters',
         gotten: gotten
       };
+      rangeToReplace = rangeToReplace || [node.start - node.loc.start.column, node.start];
 
-      if (loc) {
-        context.report({
-          node: node,
-          loc: loc,
-          message: MESSAGE,
-          data: msgContext,
-          fix: getFixerFunction(node, needed)
-        });
-      } else {
-        context.report({
-          node: node,
-          message: MESSAGE,
-          data: msgContext,
-          fix: getFixerFunction(node, needed)
-        });
-      }
+      context.report({
+        node: node,
+        loc: loc || node.loc,
+        message: MESSAGE,
+        data: msgContext,
+        fix: getFixerFunction(rangeToReplace, needed)
+      });
     }
 
     /**
-     * Get node indent
-     * @param {ASTNode} node Node to examine
-     * @param {Boolean} byLastLine get indent of node's last line
-     * @param {Boolean} excludeCommas skip comma on start of line
-     * @return {Number} Indent
+     * Get the indentation (of the proper indentType) that exists in the source
+     * @param {String} src the source string
+     * @param {Boolean} byLastLine whether the line checked should be the last
+     *   Defaults to the first line
+     * @param {Boolean} excludeCommas whether to skip commas in the check
+     *   Defaults to false
+     * @return {Number} the indentation of the indentType that exists on the line
      */
-    function getNodeIndent(node, byLastLine, excludeCommas) {
-      byLastLine = byLastLine || false;
-      excludeCommas = excludeCommas || false;
-
-      var src = sourceCode.getText(node, node.loc.start.column + extraColumnStart);
+    function getIndentFromString(src, byLastLine, excludeCommas) {
       var lines = src.split('\n');
       if (byLastLine) {
         src = lines[lines.length - 1];
@@ -154,7 +156,24 @@ module.exports = {
     }
 
     /**
-     * Checks node is the first in its own start line. By default it looks by start line.
+     * Get node indent
+     * @param {ASTNode} node Node to examine
+     * @param {Boolean} byLastLine get indent of node's last line
+     * @param {Boolean} excludeCommas skip comma on start of line
+     * @return {Number} Indent
+     */
+    function getNodeIndent(node, byLastLine, excludeCommas) {
+      byLastLine = byLastLine || false;
+      excludeCommas = excludeCommas || false;
+
+      var src = sourceCode.getText(node, node.loc.start.column + extraColumnStart);
+
+      return getIndentFromString(src, byLastLine, excludeCommas);
+    }
+
+    /**
+     * Checks if the node is the first in its own start line. By default it looks by start line.
+     * One exception is closing tags with preceeding whitespace
      * @param {ASTNode} node The node to check
      * @return {Boolean} true if its the first in the its start line
      */
@@ -165,8 +184,9 @@ module.exports = {
       } while (token.type === 'JSXText' && /^\s*$/.test(token.value));
       var startLine = node.loc.start.line;
       var endLine = token ? token.loc.end.line : -1;
+      var whitespaceOnly = token ? /\n\s*$/.test(token.value) : false;
 
-      return startLine !== endLine;
+      return startLine !== endLine || whitespaceOnly;
     }
 
   /**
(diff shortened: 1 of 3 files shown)


lib/rules/jsx-indent.js
```diff
@@ -218,41 +220,47 @@ module.exports = {
       }
     }
 
+    function getOpeningElementIndent(node) {
```
kentcdodds: Needed to move this because the `JSXClosingElement` needs to get the opening element's indent as it _should be_ not as it is.

kentcdodds:
Getting pretty close on this one!

lib/rules/jsx-indent.js
```diff
@@ -93,46 +95,32 @@ module.exports = {
...
-    function report(node, needed, gotten, loc) {
+    function report(node, needed, gotten, isEndOfTag) {
       var msgContext = {
         needed: needed,
         type: indentType,
         characters: needed === 1 ? 'character' : 'characters',
         gotten: gotten
       };
 
-      if (loc) {
```
kentcdodds: I noticed that no calls to `report` were using `loc` anyway, so I could remove `loc` and replace it with the new `isEndOfTag`

lib/rules/jsx-indent.js
```diff
@@ -93,46 +91,35 @@ module.exports = {
...
+    function report(node, needed, gotten, rangeToReplace, loc) {
       var msgContext = {
         needed: needed,
         type: indentType,
         characters: needed === 1 ? 'character' : 'characters',
         gotten: gotten
       };
+      rangeToReplace = rangeToReplace || [node.start - node.loc.start.column, node.start];
 
-      if (loc) {
```
kentcdodds: Based on the ESLint docs, if `loc` isn't provided, it defaults to `node.loc` so I figured we could just do the same and save ourselves an if statement :)

lib/rules/jsx-indent.js
```diff
@@ -218,41 +226,78 @@ module.exports = {
...
+        prevToken = sourceCode.getNodeByRangeIndex(prevToken.start);
+        while (prevToken.parent && prevToken.parent.type !== 'ConditionalExpression') {
+          prevToken = prevToken.parent;
+        }
+      }
+      prevToken = prevToken.type === 'JSXExpressionContainer' ? prevToken.expression : prevToken;
+
+      var parentElementIndent = getNodeIndent(prevToken);
+      if (prevToken.type === 'JSXElement') {
+        parentElementIndent = getOpeningElementIndent(prevToken.openingElement);
```
kentcdodds: Note the recurssion here.

tests/lib/rules/jsx-indent.js
```diff
@@ -505,22 +529,22 @@ ruleTester.run('jsx-indent', rule, {
       '  );',
       '}'
     ].join('\n'),
-    // The detection logic only thinks <App> is indented wrong, not the other
-    // two lines following. I *think* because it incorrectly uses <App>'s indention
-    // as the baseline for the next two, instead of the realizing the entire three
-    // lines are wrong together. See #608
-    /* output: [
+    output: [
```
kentcdodds: Woot! These changes fix this case and we can now auto-fix this!

tests/lib/rules/jsx-indent.js
```diff
@@ -212,17 +214,6 @@ ruleTester.run('jsx-indent', rule, {
     ].join('\n'),
     parserOptions: parserOptions
   }, {
-    // Literals indentation is not touched
```
kentcdodds: I think that Literal indentation should be touched as it doesn't make a difference in the resulting HTML (different for JSXExpressions, but that doesn't apply here).

tests/lib/rules/jsx-indent.js
```diff
@@ -51,14 +51,16 @@ ruleTester.run('jsx-indent', rule, {
...
-    code: [
-      '  <App>',
-      '<Foo />',
-      '  </App>'
-    ].join('\n'),
-    options: [-2],
-    parserOptions: parserOptions
+  // }, {
+  // should we put effort in making this work?
+  // who in their right mind would do this?
```
kentcdodds: Question stands... It would require more logic in the fixer and I don't think it's worth the complexity.

kentcdodds:
Pinging those who have touched this rule in one way or another: @yannickcr, @Jorundur, @eelyafi, @jayphelps, @petersendidit, @voxpelli

Thanks friends!

kentcdodds:
Don't merge this yet, just found a bug with the following code:

```javascript
<div>
  {
    foo &&
      <div>
        <div></div>
      </div>
  }
</div>
```

We're getting a bad report on the middle-most `div`. Looking into fixing that.

kentcdodds:
Just noticed that this is actually by design (a few valid tests showing this). I'm going to add a config option to enable this style.

kentcdodds:
Ok, I've updated the PR with documentation for the new configuration option `indentLogicalExpressions`

kentcdodds:
Oh, and just an argument for the `indentLogicalExpressions` pattern is the `language-babel` plugin for Atom (761,282 downloads) has an option to auto-indent JSX and this is how it indents things. Also I personally prefer this indentation style FWIW :)

kentcdodds:
I published my branch as `@kentcdodds/eslint-plugin-react` if anyone wants to try it out.

kentcdodds:
I've verified that this works on my work project :+1:

jayphelps:
Is this multiple fixes to separate indention cases?

kentcdodds:
> Is this multiple fixes to separate indention cases?

```
<ng-sheepishly>Yes, it is...</ng-sheepishly>
```

Originally, it was to fix a single issue with `--fix`, but I noticed that in my project we also indent multi-line LogicalExpressions so I slipped that in. I'd definitely prefer to just merge things together (as they are now), but can create a separate PR if people really feel strongly about this...

jayphelps:
It's just a bit hard for me to follow with the various changes for separate things.

kentcdodds:
Understood @jayphelps, if others feel this way, then I'll split things up.

lib/rules/jsx-indent.js
```diff
@@ -218,41 +238,82 @@ module.exports = {
...
+      }
+
+      if (isRightInLogicalExp(node) && indentLogicalExpressions) {
```
kentcdodds: This is the only relevant bit for the `indentLogicalExpressions` option.

jayphelps:
Could you perhaps edit the original post to include a summary of the changed behaviors?

lib/rules/jsx-indent.js
```diff
@@ -218,41 +238,82 @@ module.exports = {
...
+        }
+      }
+    }
+
+    /**
+     * Gets what the JSXOpeningElement's indentation should be
+     * @param {ASTNode} node The JSXOpeningElement
+     * @return {Number} the number of indentation characters it should have
+     */
+    function getOpeningElementIndent(node) {
```
kentcdodds: This function came from the `JSXOpeningElement` function below. I extracted it to make use of it else where. The only changes from what was there before is the two if statements before the `var indent = (` assignment.

kentcdodds:
Updated, let me know if more details would be helpful. Thanks @jayphelps!

lib/rules/jsx-indent.js
```diff
@@ -218,41 +238,82 @@ module.exports = {
...
+      return parentElementIndent + indent;
+    }
+
     return {
       JSXOpeningElement: function(node) {
         var prevToken = sourceCode.getTokenBefore(node);
         if (!prevToken) {
           return;
         }
-        // Use the parent in a list or an array
```
kentcdodds: This code was simply refactored to `getOpeningElementIndent`

lib/rules/jsx-indent.js
```diff
@@ -165,8 +184,9 @@ module.exports = {
       } while (token.type === 'JSXText' && /^\s*$/.test(token.value));
       var startLine = node.loc.start.line;
       var endLine = token ? token.loc.end.line : -1;
+      var whitespaceOnly = token ? /\n\s*$/.test(token.value) : false;
```
lencioni: This probably doesn't matter, but I think this regex won't match the first line? Should it be `/^\s*$/` instead? I'm also not sure what `token.value` represents, so perhaps the `\n` is necessary. (see also the regex on line 184)

If these end up being the same, it might not be a bad idea to move them to a constant at the module level.

lib/rules/jsx-indent.js
lencioni: Thanks for providing this context. This is helpful for reviewers.

lib/rules/jsx-indent.js
```diff
@@ -218,41 +238,82 @@ module.exports = {
...
+    function getOpeningElementIndent(node) {
+      var prevToken = sourceCode.getTokenBefore(node);
+      if (!prevToken) {
+        return 0;
+      }
+      // Use the parent in a list or an array
+      if (prevToken.type === 'JSXText' || prevToken.type === 'Punctuator' && prevToken.value === ',') {
+        prevToken = sourceCode.getNodeByRangeIndex(prevToken.start);
+        prevToken = prevToken.type === 'Literal' ? prevToken.parent : prevToken;
+      // Use the first non-punctuator token in a conditional expression
```
lencioni: nit: I find this style of commenting to be pretty awkward. I think this would be best if it was moved inside the conditional branch. Don't worry about this unless you end up touching this again, especially since you didn't write it.

tests/lib/rules/jsx-indent.js
lencioni: I don't think this is worth supporting.

lib/rules/jsx-indent.js
kentcdodds: I just tried out changing it to `^` and it results in an additional extraneous error in one of the test cases. Specifically:

```javascript
function MyComponent(props) {
	return (
    <div
			className="foo-bar"
			id="thing"
    >
      Hello world!
    </div>
	)
}
```

Because `token.value` is actually everything between the `>` and the `</div>` (which includes a `\n`)

lib/rules/jsx-indent.js
kentcdodds: I was in the area and updated the commenting style. I also find it confusing as it is.

tests/lib/rules/jsx-indent.js
kentcdodds: Great, I went ahead and removed the commented-out code.

PR #1064, opened 2017-02-06
Detect base class expressions wrapping `React.Component`

rosston:
e.g., treat `class Foo extends specialWrapper(React.Component) {}` as an ES6 React component.

I'm looking to use expressions as base classes in a project (a la [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#Mix-ins)), but eslint-plugin-react doesn't properly detect components in that case.

This PR simply ignores any parentheses/function call wrappers around `(React.)?(Pure)?Component`. I'm absolutely open to other approaches. I also came up with a version that allows a custom component class regex shared ESLint setting (although I think it suffers from ignoring pure components) here: https://github.com/yannickcr/eslint-plugin-react/compare/master...appropos:custom-component-regex

I'd be happy to add tests for my change here. I just wasn't sure where to put them, since this change affects multiple rules.

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -172,7 +172,7 @@ function componentRule(rule, context) {
       if (!node.superClass) {
         return false;
       }
-      return new RegExp('^(' + pragma + '\\.)?(Pure)?Component$').test(sourceCode.getText(node.superClass));
+      return new RegExp('^(.*\\()?(' + pragma + '\\.)?(Pure)?Component\\)?$').test(sourceCode.getText(node.superClass));
     },
 
     /**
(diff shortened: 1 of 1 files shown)


ljharb:
It's not idiomatic to extend anything but `React.Component` directly in react components, so this absolutely should not be default behavior.

If this is a pattern you insist on using, then perhaps we may want to add support for a config setting for "componentWrappers", or "components" - ie, things that, when extended, are still considered components - but all that said, this isn't a good pattern to be using.

rosston:
Yeah, I'll just close this.

I was specifically looking for a way to auto-bind instance methods while keeping the effectiveness of `prefer-stateless-function` (putting something along the lines of `autoBind(this)` <sup>1</sup> in your constructor quiets the `prefer-stateless-function` warning even if you have no event handlers/props/state). An MDN-style "mix-in" was simply a way to inject some extra code in the constructor without quieting `prefer-stateless-function`.

Anyway, your comment pushed me to find another way, something along the lines of `@autobind` from [core-decorators](https://github.com/jayphelps/core-decorators.js).

[1] https://github.com/cassiozen/React-autobind

PR #1066, opened 2017-02-09
Skip shape props in flow

Agontuk:
Fixes shape type props issue for flow.

### DIFF
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -411,6 +411,9 @@ module.exports = {
           }
           return true;
         case 'ObjectTypeAnnotation':
+          if (skipShapeProps) {
+            return true;
+          }
           var shapeTypeDefinition = {
             type: 'shape',
             children: []


Agontuk:
I forgot to run the unit test, I'll check and try to update as soon as possible.

Agontuk:
There's some test for flow which give error for unused shape prop types. Example:
```js
    props: {
        name: {
            unused: string
        }
    }

    render () {
        return <div>Hello {this.props.name.lastname}</div>;
    }
}
```
But in my case, even if I use the prop it gives me error. Here's a example of my component:
```js
type Props = {
    position: {
        x: number,
        y: number
    }
};

class Table extends PureComponent {
    props: Props

    render() {
        const { position } = this.props;

        return (
            <div style={ { left: position.x, top: position.y } }>Hello World</div>
        );
    }
}
```

This gives me error `position.x prop is defined but never used`. Since currently there's a issue with detecting shape props, should I remove those test cases ? Or just use `eslint-disable-line` from my end ?

ljharb:
@Agontuk hi! this PR still needs some tests; any interest in adding some?

Agontuk:
@ljharb sorry for the late response, unfortunately I'm not using flow for quite a while. So it's difficult for me to add the required test cases. I hope someone can pick it up and make the required changes.

PR #1082, opened 2017-02-21
Fix sort-comp ClassExpression

### LINKED ISSUE #1076: sort-comp does not detect component declaration starting with `return`
Similar to #174 a component is not detected if the declaration begins with `return` (Such as, in our case, where we're wrapping the component inside a HOC).  Other react rules appear to be applied correctly to these cases, just not `sort-comp`.

Example code (does not show `sort-comp` warning):
```
let makeComp = () => {
	return class MyComponent extends React.Component {
		componentDidUpdate () { }
		componentDidMount () {}
		constructor () { super(); }
	};
};
```

Example code (correctly shows `sort-comp` warning):
```
let makeComp = () => {
		componentDidUpdate () { }
		componentDidMount () {}
		constructor () { super(); }
	};
	return MyComponent;
};
```

As another note:  The react lifecycle methods are out-of-order and yet generate no warning.  Not sure why they are not, maybe that's another ticket.

webOS101:
Fixes #1076. Also adds test.

### DIFF
--- a/lib/rules/sort-comp.js
+++ b/lib/rules/sort-comp.js
@@ -291,6 +291,7 @@ module.exports = {
      */
     function getComponentProperties(node) {
       switch (node.type) {
+        case 'ClassExpression':
         case 'ClassDeclaration':
           return node.body.body;
         case 'ObjectExpression':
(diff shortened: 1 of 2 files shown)


webOS101:
To be clear, you're asking for non-react component class tests?  I will add those.

Issue #985, opened 2016-11-30
Rule proposal: no-useless-scu


nkt:
`shouldComponentUpdate` method is useless when component extending `PureComponent`.

ljharb:
It's worth noting that I think it will actually *work* - it just means there's no point in it extending PureComponent then. I'm 👍 on this rule.

jseminck:
I'll take a look at this rule, since it was just recently requested again in https://github.com/yannickcr/eslint-plugin-react/issues/1217

It looks pretty easy to add:

* Validate the MethodDefinition name
* Validate that the method is declared in a PureComponent (for which a helper already exists)

Probably will need some help with the actual rule name and errors + docs. I suppose `no-useless-scu` is not a good name, as we want to warn the user not to extend `React.PureComponent` when they want to override `shouldComponentUpdate`?

jseminck:
Nevermind. There is a PR already. Missed it.

alexzherdev:
Fixed via #1084.

PR #1084, opened 2017-02-22
Add react/no-redundant-should-component-update

jomasti:
This will check for `shouldComponentUpdate` defined in a class that extends React.PureComponent.

In regards to the name: I went with what is in the proposal, but perhaps something more informative can be found. Also, I realized that the functions dealing with component properties are duplicated in a few spots, but since there were some differences, I figured they could be refactored out at some other point. That's ultimately up to the reviewers, though.

This PR implements the rule proposed in #985.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,7 @@ Finally, enable all of the rules that you would like to use.  Use [our preset](#
 * [react/no-find-dom-node](docs/rules/no-find-dom-node.md): Prevent usage of `findDOMNode`
 * [react/no-is-mounted](docs/rules/no-is-mounted.md): Prevent usage of `isMounted`
 * [react/no-multi-comp](docs/rules/no-multi-comp.md): Prevent multiple component definition per file
+* [react/no-redundant-should-component-update](docs/rules/no-redundant-should-component-update.md): Prevent usage of `shouldComponentUpdate` when extending React.PureComponent
 * [react/no-render-return-value](docs/rules/no-render-return-value.md): Prevent usage of the return value of `React.render`
 * [react/no-set-state](docs/rules/no-set-state.md): Prevent usage of `setState`
 * [react/no-string-refs](docs/rules/no-string-refs.md): Prevent using string references in `ref` attribute.
--- a/docs/rules/no-redundant-should-component-update.md
+++ b/docs/rules/no-redundant-should-component-update.md
@@ -1,0 +1,64 @@
+# Prevent usage of shouldComponentUpdate when extending React.PureComponent (react/no-redundant-should-component-update)
+
+Warns if you have `shouldComponentUpdate` defined when defining a component that extends React.PureComponent.
+While having `shouldComponentUpdate` will still work, it becomes pointless to extend PureComponent.
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```jsx
+class Foo extends React.PureComponent {
+  shouldComponentUpdate() {
+    // do check
+  }
+
+  render() {
+    return <div>Radical!</div>
+  }
+}
+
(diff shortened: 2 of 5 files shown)


docs/rules/no-useless-scu.md
```diff
@@ -0,0 +1,64 @@
...
+While having `shouldComponentUpdate` will still work, it becomes pointless to extend PureComponent.
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```jsx
+class Foo extends React.PureComponent {
+  shouldComponentUpdate() {
+    return true;
```
ljharb: is the "return true" important? if not, i'd maybe put a comment here or something to indicate "any code can go here" (same throughout)

lib/rules/no-useless-scu.js
```diff
@@ -0,0 +1,94 @@
...
+     */
+    function getPropertyName(node) {
+      // Special case for class properties
+      // (babel-eslint does not expose property name so we have to rely on tokens)
+      if (node.type === 'ClassProperty') {
+        var tokens = context.getFirstTokens(node, 2);
+        return tokens[1] && tokens[1].type === 'Identifier' ? tokens[1].value : tokens[0].value;
+      }
+
+      return node.key.name;
```
ljharb: does `node` always have a `key` property?

lib/rules/no-useless-scu.js
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Checks for violation of rule
+     * @param {ASTNode} node The AST node being checked.
+     */
+    function checkForViolation(node) {
+      if (utils.isPureComponent(node)) {
+        var hasScu = hasShouldComponentUpdate(node);
+        if (hasScu) {
+          var className = node.id.name;
```
ljharb: does `node` always have an `id` property?

docs/rules/no-useless-scu.md
jomasti: Yeah, I was thinking this could come up when writing it. I'll update.

lib/rules/no-useless-scu.js
jomasti: From what I can see, yes. It's only ever checking properties from the class. Unless there's some other possible node that I'm not aware of for a class.

lib/rules/no-useless-scu.js
jomasti: I'll update this.

lencioni:
I think "redundant" might be more specific than "useless". What do folks think?

Also, I'm not sure if we should abbreviate shouldComponentUpdate as scu. I think it might be best if the rule was named `react/no-redundant-should-component-update`.

lib/rules/no-useless-scu.js
ljharb: Might there be one in the future? It would probably be good to be defensible here.

ljharb:
I agree with @lencioni's naming comments.

jomasti:
I've updated the code based on the feedback given. I'll squash at the end after everything checks out.

ljharb:
@jomasti what do you think about my v14/v13/v15.0/v15.1 question?

jomasti:
@ljharb It makes sense. Would you expect a test with settings for a lower version? All I can think of would be using the syntax with a lower version and having it not catch the error. It doesn't seem useful, though.

ljharb:
@jomasti hm, that's a fair point - considering it'd just "do nothing" in a React where PureComponent also does nothing, it's probably not worth doing.

ljharb:
I'll merge this tomorrow if there's no objections.

PR #1086, opened 2017-02-23
prefer-stateless-function w/ decorators

### LINKED ISSUE #1034: prefer-stateless-function with decorators
i am using decorators, decorators work just with class so i must not use stateless function
```javascript
@dragDropContext(HTML5Backend)
class ToDoList extends Component {
  render() {
    const { data } = this.props;
    return (
      <div>
        {data.map(((todo, i) => <SingleToDo
          data={todo}
        />))}
      </div>
    );
  }
}
```
and in this case i have error prefer-stateless-function

benstepp:
Currently prefer-stateless-function warns when using a decorated class
over a stateless function. The decorator syntax only works with classes,
so it makes sense not to warn in this case.

Fixes #1034


nwwells:
_bump_ @yannickcr or @lencioni?

lencioni:
Thanks!

PR #1087, opened 2017-02-23
Add new rule 'default-props-match-prop-types'

### LINKED ISSUE #1022: New rule: `no-invalid-default-props`
I'd like a rule that errors if anything is in `defaultProps` that is not also in `propTypes`.

This came up with the following code:
```js
Component.propTypes = {
  fooValue: PropTypes.string,
};
Component.defaultProps = {
  foo: '',
};
```

In other words, it was a partially completed prop rename.

This rule should also look up a propTypes variable (`Component.propTypes = propTypes`) defined elsewhere in the file.

webOS101:
Fixes #1022 

There's currently one caveat that I haven't addressed: If there are **no** `propTypes` at all no warnings are issued. I didn't distinguish between that case and the case where they are not able to be determined (e.g.: external import).  Also, this rule may conflict with `require-default-props` as I am re-using the same component members, though testing did not reveal any problems using the two rules together.

### DIFF
--- a/lib/rules/default-props-match-prop-types.js
+++ b/lib/rules/default-props-match-prop-types.js
@@ -5,11 +5,10 @@
  */
 'use strict';
 
-var has = require('has');
-var find = require('array.prototype.find');
-var Components = require('../util/Components');
-var variableUtil = require('../util/variable');
-var annotations = require('../util/annotations');
+const has = require('has');
+const Components = require('../util/Components');
+const variableUtil = require('../util/variable');
+const annotations = require('../util/annotations');
 
 // ------------------------------------------------------------------------------
 // Rule Definition
@@ -36,8 +35,8 @@ module.exports = {
 
   create: Components.detect(function(context, components, utils) {
 
-    var configuration = context.options[0] || {};
-    var allowRequiredDefaults = configuration.allowRequiredDefaults || false;
+    const configuration = context.options[0] || {};
+    const allowRequiredDefaults = configuration.allowRequiredDefaults || false;
 
     /**
      * Get properties name
@@ -52,7 +51,7 @@ module.exports = {
       // Special case for class properties
       // (babel-eslint@5 does not expose property name so we have to rely on tokens)
       } else if (node.type === 'ClassProperty') {
-        var tokens = context.getFirstTokens(node, 2);
+        const tokens = context.getFirstTokens(node, 2);
         return tokens[1] && tokens[1].type === 'Identifier' ? tokens[1].value : tokens[0].value;
       }
       return '';
@@ -73,7 +72,7 @@ module.exports = {
      * @returns {Boolean}       `true` if the node is a defaultProps declaration, `false` if not
      */
     function isDefaultPropsDeclaration(node) {
-      var propName = getPropertyName(node);
+      const propName = getPropertyName(node);
       return (propName === 'defaultProps' || propName === 'getDefaultProps');
     }
 
@@ -92,9 +91,7 @@ module.exports = {
      * @returns {ASTNode|null} Return null if the variable could not be found, ASTNode otherwise.
      */
     function findVariableByName(name) {
-      var variable = find(variableUtil.variablesInScope(context), function(item) {
-        return item.name === name;
-      });
+      const variable = variableUtil.variablesInScope(context).find((item) => item.name === name);
 
       if (!variable || !variable.defs[0] || !variable.defs[0].node) {
         return null;
@@ -151,7 +148,7 @@ module.exports = {
      * @returns {Object[]}        Array of PropType object representations, to be consumed by `addPropTypesToComponent`.
      */
     function getPropTypesFromObjectExpression(objectExpression) {
-      var props = objectExpression.properties.filter(function(property) {
+      const props = objectExpression.properties.filter(function(property) {
         return property.type !== 'ExperimentalSpreadProperty';
       });
 
@@ -170,11 +167,11 @@ module.exports = {
      * @returns {Object[]}     Array of PropType object representations, to be consumed by `addPropTypesToComponent`.
      */
     function getPropTypesFromTypeAnnotation(node) {
-      var properties;
+      let properties;
 
       switch (node.typeAnnotation.type) {
         case 'GenericTypeAnnotation':
-          var annotation = resolveGenericTypeAnnotation(node.typeAnnotation);
+          let annotation = resolveGenericTypeAnnotation(node.typeAnnotation);
 
           if (annotation && annotation.id) {
             annotation = findVariableByName(annotation.id.name);
@@ -184,7 +181,7 @@ module.exports = {
           break;
 
         case 'UnionTypeAnnotation':
-          var union = resolveUnionTypeAnnotation(node.typeAnnotation);
+          const union = resolveUnionTypeAnnotation(node.typeAnnotation);
           properties = union.reduce(function(acc, curr) {
             if (!curr) {
               return acc;
@@ -203,14 +200,14 @@ module.exports = {
           break;
       }
 
-      var props = properties.filter(function(property) {
+      const props = properties.filter(function(property) {
         return property.type === 'ObjectTypeProperty';
       });
 
       return props.map(function(property) {
         // the `key` property is not present in ObjectTypeProperty nodes, so we need to get the key name manually.
-        var tokens = context.getFirstTokens(property, 1);
-        var name = tokens[0].value;
+        const tokens = context.getFirstTokens(property, 1);
+        const name = tokens[0].value;
 
         return {
           name: name,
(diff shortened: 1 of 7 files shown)


lib/rules/no-invalid-default-props.js
```diff
@@ -0,0 +1,624 @@
...
+     * @returns {Boolean}       `true` if the node is a propTypes declaration, `false` if not
+     */
+    function isPropTypesDeclaration(node) {
+      return getPropertyName(node) === 'propTypes';
+    }
+
+    /**
+     * Checks if the Identifier node passed in looks like a defaultProps declaration.
+     * @param   {ASTNode}  node The node to check. Must be an Identifier node.
+     * @returns {Boolean}       `true` if the node is a defaultProps declaration, `false` if not
```
ljharb: I don't think any of these spaces are needed; do we typically visually align these comments in other rules?

lencioni:
Thanks for the contribution! I haven't had time to review the code here yet, but I wanted to mention that I think `no-invalid-default-props` isn't quite the right name for this rule. To me that name implies that the default props are going to be checked against the propTypes validators.

I'm not quite sure what a good name might be. Maybe something along the lines of `no-extra-default-props`?

Also, I'm not entirely sure this makes sense as a separate rule instead of enhancing the `prop-types` rule to include this. @ljharb what do you think?

lib/rules/no-invalid-default-props.js
```diff
@@ -0,0 +1,624 @@
...
+     */
+    function isDefaultPropsDeclaration(node) {
+      return (getPropertyName(node) === 'defaultProps' || getPropertyName(node) === 'getDefaultProps');
```
ljharb: Let's cache the property name rather than calling the function twice.

lib/rules/no-invalid-default-props.js
```diff
@@ -0,0 +1,624 @@
...
+          } else {
+            addDefaultPropsToComponent(component, getDefaultPropsFromObjectExpression(expression));
+          }
+
+          return;
+        }
+
+        // e.g.:
+        // MyComponent.propTypes.baz = React.PropTypes.string;
+        if (node.parent.type === 'MemberExpression' && node.parent.parent.type === 'AssignmentExpression') {
```
ljharb: Will `node.parent` always have a `.parent` node?

lib/rules/no-invalid-default-props.js
```diff
@@ -0,0 +1,624 @@
...
+
+        if (!node.static) {
+          return;
+        }
+
+        if (!node.value) {
+          return;
+        }
+
+        var isPropType = getPropertyName(node) === 'propTypes';
```
ljharb: Let's cache the property name here rather than calling the function thrice.

ljharb:
Hmm, that makes sense that "invalid" isn't the best name. `extra` works, I suppose - or `no-superfluous-default-props`, or some other synonym?

ljharb:
I do think it belongs as a separate rule since it warns on defaults, not prop types.

lib/rules/no-invalid-default-props.js
webOS101: Happy to make changes, but this code is nearly identical to `require-default-props`.  Do you want me to update the source there, too?

webOS101:
`no-unnecessary-default-props`?  `no-unneeded-default-props`?  `no-useless-default-props`?

lib/rules/no-invalid-default-props.js
ljharb: in that case (that our rules already are inconsistent) then don't worry about it; it can be cleaned up in a separate PR.

ljharb:
A defaultProp for a required prop is arguably not "useless" since it may serve as documentation (not that I would ever use the option that allows it) so I don't think that's the right word.

`unnecessary` could work too (along with `extra` or `superfluous`).

`no-unused-default-props` would be good but there's no safe way to tell if the *user* uses it, only if React will use it.

Let's keep bikeshedding on the name, might as well pick the best one now.

lib/rules/no-invalid-default-props.js
webOS101: I think it's unlikely you wouldn't have a parent node in the case of a `MemberExpression` whose parent is a `MemberExpression` and belongs to a Component.  But, I will admit I'm not intimately familiar with this code.

ljharb: I don't know much of anything about AST nodes, I just want to avoid a crash in the future when a new node type comes out

webOS101:
@lencioni A rule that validates the types of the `defaultProps` seems very useful!

webOS101:
I addressed some of the concerns raised in the review, but understand we're still searching for a good name.

webOS101:
Anyone had any good ideas on the naming?  Nothing particularly new has come to me.  Maybe `no-stray-default-props`?

ljharb:
What about `default-props-match-prop-types`?

webOS101:
That seems more in keeping with @lencioni 's checker that would validate types.  `no-undefined-default-props`?  `no-unexpected-default-props`?  `only-meaningful-default-props`?

ljharb:
I think `default-props-match-prop-types` is good, and checking for matching types would be enabled in that rule via an additional option :-)

webOS101:
I'm curious how accurate we could be in matching the types.  Certainly simple types would be easy, more complicated ones could easily become problematic.  If you like the new name a lot, I can update the PR to the new name.

ljharb:
I think matching the types will likely only be possible when using the built-in `PropTypes` validators, or else it'd have to actually pull in and evaluate any custom prop types - but that's a discussion for that separate PR.

I think "match" doesn't necessarily connote "types match", which is why I don't think having the option to match types is a blocker. I mainly just want to get this in, and I think that's a good enough name ¯\\\_(ツ)_/¯

webOS101:
Updated the name of the rule and rebased on master.

lib/rules/default-props-match-prop-types.js
```diff
@@ -0,0 +1,627 @@
+/**
+ * @fileOverview Enforce all defaultProps are defined in propTypes
+ * @author Vitor Balocco
+ * @author Roy Sutton
+ */
+'use strict';
+
+var has = require('has');
+var find = require('array.prototype.find');
```
ljharb: this is no longer necessary, since master now only supports node 4+

lib/rules/default-props-match-prop-types.js
```diff
@@ -0,0 +1,627 @@
...
+      return propTypeExpression.type === 'MemberExpression' && propTypeExpression.property.name === 'isRequired';
+    }
+
+    /**
+     * Find a variable by name in the current scope.
+     * @param  {string} name Name of the variable to look for.
+     * @returns {ASTNode|null} Return null if the variable could not be found, ASTNode otherwise.
+     */
+    function findVariableByName(name) {
+      var variable = find(variableUtil.variablesInScope(context), function(item) {
```
ljharb: so this can be `const variable = variableUtil.variablesInScope(context).find(item => item.name === name);`

lib/rules/default-props-match-prop-types.js
```diff
@@ -0,0 +1,627 @@
...
+
+    /**
+     * Extracts a DefaultProp from an ObjectExpression node.
+     * @param   {ASTNode} objectExpression ObjectExpression node.
+     * @returns {Object|string}            Object representation of a defaultProp, to be consumed by
+     *                                     `addDefaultPropsToComponent`, or string "unresolved", if the defaultProps
+     *                                     from this ObjectExpression can't be resolved.
+     */
+    function getDefaultPropsFromObjectExpression(objectExpression) {
+      var hasSpread = find(objectExpression.properties, function(property) {
```
ljharb: and this

lib/rules/default-props-match-prop-types.js
```diff
@@ -0,0 +1,627 @@
...
+
+      addPropTypesToComponent(component, getPropTypesFromTypeAnnotation(node.typeAnnotation, context));
+    }
+
+    function isPropTypeAnnotation(node) {
+      return (getPropertyName(node) === 'props' && !!node.typeAnnotation);
+    }
+
+    function propFromName(propTypes, name) {
+      return find(propTypes, function (prop) {
```
ljharb: and this can be `return propTypes.find(prop => prop.name === name);`

lib/rules/default-props-match-prop-types.js
webOS101: What's sad is I originally implemented with `Array.find()` but realized this didn't match with the current style.

webOS101:
Updated all the `var` declarations as well.

webOS101:
Anything else needed for this?

ljharb:
Sorry for the delay. I'll give this another day or so, and if there's no objections, I'll merge it.

webOS101:
Whoops, looks like I didn't add the rule to the README.md in the root.  Should I create a separate PR for that?

ljharb:
@webOS101 yes, thanks, that'd be great!

webOS101:
https://github.com/yannickcr/eslint-plugin-react/pull/1239

PR #1089, opened 2017-02-23
Bug fix for false positives with no-multi-comp

### LINKED ISSUE #1088: False positives on "Declare only one React component per file" (react/no-multi-comp)
```js
import test from "jest-ava-api"
import React, { createElement } from "react"
import { createRenderer } from "react-addons-test-utils"

const renderer = (...args) => { // I guess this is detected as a component
  const render = createRenderer()
  render.render(...args)
  return render.getRenderOutput()
}

test("what", () => {
  return {
    isActive: () => false, // what?
  }
})
```

react/no-multi-comp is detecting 2 components here since I get a warning on the last () => false.
Both are wrong.

Ps: the code does not make sense as it is as I removed some part to make the debugging more easy.

Not that if I does not import createElement, the error seems to disappear  (but I need this...)

benstepp:
Previously, when createElement was destructured from React, it would
cause any function defined in the file to be flagged as a React
Component. This change makes it such that the function must call
createElement to be considered a component.

Fixes #1088 

---
Source of bug:
https://github.com/yannickcr/eslint-plugin-react/commit/416deff06d90b100f991bfbf86f7484061dca0d2#diff-bafb3355a409c4ba076a9e3d609ad65bR294

Note: Not sure if this is the best place for this test. I feel like component detection testing should have its own place where it happens. It seems like you have to run the whole test suite to see if Components are still being detected correctly. Thankfully the whole test suite runs fairly quickly.

### DIFF
--- a/tests/lib/rules/no-multi-comp.js
+++ b/tests/lib/rules/no-multi-comp.js
@@ -10,6 +10,7 @@
 
 var rule = require('../../../lib/rules/no-multi-comp');
 var RuleTester = require('eslint').RuleTester;
+var assign = require('object.assign');
 
 var parserOptions = {
   ecmaVersion: 6,
@@ -89,6 +90,21 @@ ruleTester.run('no-multi-comp', rule, {
     options: [{
       ignoreStateless: true
     }]
+  }, {
+    // multiple non-components
+    code: [
+      'import React, { createElement } from "react"',
+      'const helperFoo = () => {',
+      '  return true;',
+      '};',
+      'function helperBar() {',
+      '  return false;',
+      '};',
+      'function RealComponent() {',
+      '  return createElement("img");',
+      '};'
+    ].join('\n'),
+    parserOptions: assign({sourceType: 'module'}, parserOptions)
   }],
 
   invalid: [{
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -237,7 +237,7 @@ function componentRule(rule, context) {
      * @returns {Boolean} True if React.createElement called
      */
     isReactCreateElement: function(node) {
-      return (
+      var calledOnReact = (
         node &&
         node.callee &&
         node.callee.object &&
@@ -245,6 +245,17 @@ function componentRule(rule, context) {
         node.callee.property &&
         node.callee.property.name === 'createElement'
       );
+
+      var calledDirectly = (
+        node &&
+        node.callee &&
+        node.callee.name === 'createElement'
+      );
+
+      if (this.hasDestructuredReactCreateElement()) {
+        return calledDirectly || calledOnReact;
+      }
+      return calledOnReact;
     },
 
     /**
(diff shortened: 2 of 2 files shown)


tests/lib/rules/no-multi-comp.js
```diff
@@ -89,6 +89,21 @@ ruleTester.run('no-multi-comp', rule, {
...
+      '  return true;',
+      '};',
+      'function helperBar() {',
+      '  return false;',
+      '};',
+      'function RealComponent() {',
+      '  return createElement("img");',
+      '};'
+    ].join('\n'),
+    parser: 'babel-eslint'
```
ljharb: This test case doesn't need babel-eslint - let's use the default parser whenever possible.

benstepp: I used `babel-eslint` over `sourceType: 'module'` because it was way more popular in the codebase for handling import statements. (git grep sourceType)

Do you think it would be better to apply the module sourceType to the entire test file, just the single test using object.assign, or using `const createElement = require('react').createElement;` in the actual test? 

Maybe a chore issue to clean out all of babel-eslint and replace it with sourceType module. eslint's default parser could handle a vast majority of the tests.

ljharb: The sourceType is required for anything using the default parser, but it's unnecessary for anything using babel-eslint.

Ideally, every test case that uses babel-eslint would also have a corresponding test case using the default parser, and babel-eslint tests would only be for features that eslint core can't handle (like flow types, or proposals)

ljharb: Specifically I think it should be applied to this specific test, and if we want to make things consistent holistically, that can be a separate PR.

benstepp: Sounds good. I went ahead and squashed a change with `sourceType: 'module'`.

arian:
This fixes another false positive for the display-name:

```diff
diff --git a/tests/lib/rules/display-name.js b/tests/lib/rules/display-name.js
index 3f8db25..83a3772 100644
--- a/tests/lib/rules/display-name.js
+++ b/tests/lib/rules/display-name.js
@@ -349,6 +349,17 @@ ruleTester.run('display-name', rule, {
     parser: 'babel-eslint'
   }, {
     code: [
+      'import React, {createElement} from "react";',
+      'const SomeComponent = (props) => {',
+      '  const {foo, bar} = props;',
+      '  return someComponentFactory({',
+      '    onClick: () => foo(bar("x"))',
+      '  });',
+      '};'
+    ].join('\n'),
+    parser: 'babel-eslint'
+  }, {
+    code: [
       'import React, {Component} from "react";',
       'function someDecorator(ComposedComponent) {',
       '  return class MyDecorator extends Component {',
```

I wrote a test for it, but it seems this pull request already fixed it in master. For the tagged version 6.10.0 this test fails.

ljharb:
If you'd like to add your test in a PR, that'd be great :-)

PR #1091, opened 2017-02-24
Fix defaultProps detection when ObjectExpression

webOS101:
`defaultProps` defined as objects in `React.CreateClass` were
not being recorded causing false positives. Added test
case. Added corresponding ES6 version test as none was present.


webOS101:
whoops, the bug was in my understanding of how `defaultProps` works in `React.CreateClass`

PR #1103, opened 2017-03-07
Add no-unused-state

wbinnssmith:
Link to significant commit [is here](https://github.com/wbinnssmith/eslint-plugin-react/commit/1e5b989cf7299962e62e65574241b7288120dd61)

This adds a new rule, react/no-unused-state, which discovers state fields in a React component and warns if any of them are never read. It was developed internally at Facebook by @rjbailey and has been in use throughout Facebook's JS for a couple months now. It was written against a modern version of node, so has been rebased on @ljharb's branch dropping support for node < 4.

It currently supports es2015 classes extending `React.Component`  (no support currently for `React.createClass()`) and can detect when state is as `this.state = {}`, assigning state in a property initializer, and when calling `this.setState()`.

### DIFF
--- a/lib/rules/no-unused-state.js
+++ b/lib/rules/no-unused-state.js
@@ -9,6 +9,8 @@
 
 'use strict';
 
+var Components = require('../util/Components');
+
 // Descend through all wrapping TypeCastExpressions and return the expression
 // that was cast.
 function uncast(node) {
@@ -31,17 +33,25 @@ function getName(node) {
   return null;
 }
 
-function isMethodDefinitionWithName(node, name, isStatic) {
-  isStatic = isStatic || false;
-  return (
-    node.type === 'MethodDefinition' &&
-    node.static === isStatic &&
-    getName(node.key) === name
-  );
-}
-
 function isThisExpression(node) {
   return uncast(node).type === 'ThisExpression';
+}
+
+function getInitialClassInfo() {
+  return {
+    // Set of nodes where state fields were defined.
+    stateFields: [],
+
+    // Set of names of state fields that we've seen used.
+    usedStateFields: [],
+
+    // Names of local variables that may be pointing to this.state. To
+    // track this properly, we would need to keep track of all locals,
+    // shadowing, assignments, etc. To keep things simple, we only
+    // maintain one set of aliases per method and accept that it will
+    // produce some false negatives.
+    aliases: null
+  };
 }
 
 module.exports = {
@@ -54,7 +64,7 @@ module.exports = {
     schema: []
   },
 
-  create: function(context) {
+  create: Components.detect(function(context, components, utils) {
     // Non-null when we are inside a React component ClassDeclaration and we have
     // not yet encountered any use of this.state which we have chosen not to
     // analyze. If we encounter any such usage (like this.state being spread as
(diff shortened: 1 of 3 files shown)


wbinnssmith:
I'm happy to port down to node 0.10 to get this merged sooner, but would rather #1038 be merged :)

ljharb:
Yes, please do that. It will take much longer if it needs to wait for a major release, and this PR's diff will be cleaner.

wbinnssmith:
Rebased and ported back to node 0.10

How do you feel about the lack of support for `React.createClass()`?

ianobermiller:
> It currently supports es2015 classes extending React.Component

To clarify, it actually supports ANY class that has a `render` method, no matter what it extends from.

wbinnssmith:
Addressed the feedback and implemented createClass support. Components util made quick work of it -- awesome.

lib/rules/no-unused-state.js
```diff
@@ -0,0 +1,338 @@
...
+    var classInfo = null;
+
+    // Returns true if the given node is possibly a reference to `this.state`.
+    function isStateReference(node) {
+      node = uncast(node);
+
+      var isDirectStateReference =
+        node.type === 'MemberExpression' &&
+        isThisExpression(node.object) &&
+        node.property.name === 'state';
```
ljharb: will every `node` have a `property` property? If not, the `.name` could crash when encountering unexpected node types.

wbinnssmith: every `MemberExpression` should, yeah, which should be limited by the prior condition. cc @rjbailey

rjbailey: Yep, every `MemberExpression` should have a `property` AFAIK. It may not have a `name` (if it's computed), but it won't crash.

If we used `getName(node.property) === 'state'` instead we'd also handle the computed case (so we'd detect uses like `this['state']`). Looks like I forgot to do that here :)

ljharb:
@yannickcr @lencioni @EvNaverniouk ping for review

danielheyman:
Any update on this rule?

ljharb:
@wbinnssmith would you mind freshly rebasing this? after that, i'll give it a rereview, and then merge ASAP if there's no objections from other collabs.

wbinnssmith:
@ljharb no problem :)

wbinnssmith:
okay @ljharb, just rebased, cleaned things up, and moved some stuff back to node 4 compatible features (where this all started 😂)

tests/lib/rules/no-unused-state.js
```diff
@@ -0,0 +1,438 @@
...
+const parserOptions = {
+  ecmaVersion: 6,
+  ecmaFeatures: {
+    jsx: true
+  }
+};
+
+const eslintTester = new RuleTester({
+  parserOptions: parserOptions,
+  parser: 'babel-eslint'
```
ljharb: is babel-eslint needed for all these test cases? It'd be ideal to use the default parser whenever possible.

wbinnssmith:
I can split the tests up using multiple RuleTesters, separating those that parse js compatible with espree + es2015, and those that require babylon with support for experimental features enabled (property initializers and object spread, mainly). Does that sound useful?

ljharb:
You don't need to create a new RuleTester instance - the constructor argument is just the default parser options; each valid/invalid example can also provide its own parser options.

wbinnssmith:
@ljharb that's exactly what I ended up doing :)

ping @yannickcr @lencioni @EvHaus

wbinnssmith:
ping! can @yannickcr @lencioni or @EvHaus look at this please?

ljharb:

I'll merge within the next day or two, if no objections, after that's done.

wbinnssmith:
done.

wbinnssmith:
🎉 Thanks for your help throughout the process, @ljharb!

ljharb:
Whoops, this needs documentation. @wbinnssmith, if you could file a PR to fix #1351, that'd be great!

PR #1105, opened 2017-03-08
false positive in `no-unused-prop-types`

mathieudutour:
I added a test case to show a false positive of the rule `no-unused-prop-types`.
It happens when using both `nextProps` and `this.props` in a lifecycle method.
The `prop-types` rule does complain (as it should) when not specifying a propType for `foo`.

I tried to dive into the code and find the difference (and understand why those 2 rules aren't based on the same "parser") but I got lost.


mathieudutour:
fixed in #1106

PR #1106, opened 2017-03-09
Fix nextProps false positive in no-unused-prop-types

### LINKED ISSUE #1079: nextProps aren't handled by no-unused-props
I have the following (simplified) piece of code:

```js
  componentWillReceiveProps(nextProps) {
    if (!nextProps.error) {
      console.log('oh noez');
    }
  }
```

This triggers the following ESLint error:

```
  89:27  error  'error' is missing in props validation  react/prop-types
```

However, if I do add `error` to `propTypes`, another error is triggered:

```js
  33:10  error  'error' PropType is defined but prop is never used  react/no-unused-prop-types
```

I think the two rules should cooperate.

I'm using the following relevant packages:

- eslint@3.15.0
- eslint-config-airbnb@14.1.0
- eslint-plugin-react@6.10.0

Kerumen:
Fixes #1079

I did the same comparisons as the `prop-types` rules (https://github.com/yannickcr/eslint-plugin-react/blob/b64648521f9a7949eef4cd1a210768e906b4f3d1/lib/rules/prop-types.js#L130) for the `no-unused-prop-types` rule.
This way they both handle nextProps identically.

I also borrowed the test case from #1105.

### DIFF
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -18,6 +18,7 @@ var annotations = require('../util/annotations');
 // ------------------------------------------------------------------------------
 
 var DIRECT_PROPS_REGEX = /^props\s*(\.|\[)/;
+var DIRECT_NEXT_PROPS_REGEX = /^nextProps\s*(\.|\[)/;
 
 // ------------------------------------------------------------------------------
 // Rule Definition
@@ -78,6 +79,24 @@ module.exports = {
     }
 
     /**
+     * Check if we are in a class constructor
+     * @return {boolean} true if we are in a class constructor, false if not
+     **/
+    function inComponentWillReceiveProps() {
+      var scope = context.getScope();
+      while (scope) {
+        if (
+          scope.block && scope.block.parent &&
+          scope.block.parent.key && scope.block.parent.key.name === 'componentWillReceiveProps'
+        ) {
+          return true;
+        }
+        scope = scope.upper;
+      }
+      return false;
+    }
+
+    /**
      * Checks if we are using a prop
      * @param {ASTNode} node The AST node being checked.
      * @returns {Boolean} True if we are using a prop, false if not.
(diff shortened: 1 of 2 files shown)


lib/rules/no-unused-prop-types.js
```diff
@@ -17,7 +17,7 @@ var annotations = require('../util/annotations');
 // Constants
 // ------------------------------------------------------------------------------
 
-var DIRECT_PROPS_REGEX = /^props\s*(\.|\[)/;
+var DIRECT_PROPS_REGEX = /^(props|nextProps)\s*(\.|\[)/;
```
ljharb: this seems to suggest that there will be places that `props` and `nextProps` are interchangeable - that doesn't seem right to me.

tests/lib/rules/no-unused-prop-types.js
```diff
@@ -1461,6 +1461,21 @@ ruleTester.run('no-unused-prop-types', rule, {
...
+        '    bar: PropTypes.bool',
+        '  }',
+        '  componentWillReceiveProps (nextProps) {',
+        '    if (nextProps.foo) {',
+        '      doSomething(this.props.bar);',
+        '    }',
+        '  }',
+        '}'
+      ].join('\n'),
+      parser: 'babel-eslint'
```
ljharb: there's no need to use babel-eslint here; let's just use normal `Hello.propTypes = `, and use the default parser

lib/rules/no-unused-prop-types.js
Kerumen: I did exactly like the [`prop-types` rules](https://github.com/yannickcr/eslint-plugin-react/blob/b64648521f9a7949eef4cd1a210768e906b4f3d1/lib/rules/prop-types.js#L20). It means that this one is wrong too.

What do you suggest? This regex is only used one time.

Kerumen: I pushed a try. I split the regex into 2 variables. It leads to more conditions. Tell me if you prefer this version.

vinch:
+1

PR #1112, opened 2017-03-14
Fix required prop types for classes with annotated methods.

ethanjgoldberg:
The `react/prop-types` rule incorrectly considers  flow type
annotations on class methods with arguments named `props` as
prop type declarations for the class, even though flow itself
gives them no such weight.

When a react component is defined as a class extending `React.Component`,
and it defines a method that takes an annotated argument named `props`,
then, even though flow still thinks the props have `any` type, the
`prop-types` rule will not produce 'missing in props validation' errors.

A pattern that gets used a lot in React components is as follows:
```javascript
type Props = {...};
class Comp extends React.Component {
  constructor(props: Props) {
    super(props);
    // do something...
  }

  // ...
}
```
which makes this especially nefarious.

### DIFF
--- a/lib/rules/prop-types.js
+++ b/lib/rules/prop-types.js
@@ -868,7 +868,12 @@ module.exports = {
 
       ArrowFunctionExpression: handleStatelessComponent,
 
-      FunctionExpression: handleStatelessComponent,
+      FunctionExpression: function(node) {
+        if (node.parent.type === 'MethodDefinition') {
+          return;
+        }
+        handleStatelessComponent(node);
+      },
 
       MemberExpression: function(node) {
         var type;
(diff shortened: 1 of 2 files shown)


ethanjgoldberg:
I've rebased this to fix the merge conflicts and bring the branch up to date -- could we get it merged this time?

ljharb:
@yannickcr it'd be great to get a release out with this fix in it; any chance you'll have time soon?

ethanjgoldberg:
@ljharb, I think this one is actually fixed by the latest flow ([first change listed here](https://github.com/facebook/flow/blob/master/Changelog.md#0460)) -- might be irrelevant now or possibly even wrong. I haven't tested the new flow, but I suspect this should be reverted.

ljharb:
We need to support both versions of flow, so reverting it wouldn't work by itself

PR #1078, opened 2017-02-18
Improve title in void-dom-elements-no-children.md

omerzach:
Add the name of the rule in parentheses at the end of Markdown title like in other documentation files.

PR #1071, opened 2017-02-14
docs(prefer-es6-class): add missing prefix

adnasa:
figured...

PR #1121, opened 2017-03-21
Improve title in void-dom-elements-no-children.md

omerzach:
Add the name of the rule in parentheses at the end of Markdown title like in other documentation files.

### DIFF
--- a/docs/rules/void-dom-elements-no-children.md
+++ b/docs/rules/void-dom-elements-no-children.md
@@ -1,4 +1,4 @@
-# Prevent void DOM elements (e.g. `<img />`, `<br />`) from receiving children
+# Prevent void DOM elements (e.g. `<img />`, `<br />`) from receiving children (react/void-dom-elements-no-children)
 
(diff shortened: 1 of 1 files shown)


ljharb:
Replaces #1078

docs/rules/void-dom-elements-no-children.md
```diff
@@ -1,4 +1,4 @@
-# Prevent void DOM elements (e.g. `<img />`, `<br />`) from receiving children
+# Prevent void DOM elements (e.g. `<img />`, `<br />`) from receiving children (react/void-dom-elements-no-children)
```
yannickcr: We did not put the `react/` prefix in the other titles, I think it's not necessary to put it here either

ljharb: This was my request, per https://github.com/yannickcr/eslint-plugin-react/pull/1078#discussion_r101881565

yannickcr: Ho, did now saw your comment and #1071. Ok then :)

PR #1122, opened 2017-03-23
Fix ignorePureComponents when using class expressions.

dreid:
This allows the use of a React.PureComponent when used in an expression context for `prefer-stateless-function` with `ignorePureComponents`.

Example:

```js
const Foo = class extends React.PureComponent {
  render() {
  }
};
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,47 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
-## [6.9.0] - 2016-01-08
+## [6.10.0] - 2017-02-16
+### Added
+* Add [`forbid-foreign-prop-types`][] rule ([#696][] @iancmyers)
+* Add [`void-dom-elements-no-children`][] rule ([#709][] @lencioni)
+* Add [`forbid-elements`][] rule ([#887][] @kentor)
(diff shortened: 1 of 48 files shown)


ljharb:
Should this only apply when the [react version setting](https://github.com/yannickcr/eslint-plugin-react#configuration) is >= v15.3?

ljharb:
(also, why use an expression in that example and not `class Foo extends …`?

dreid:
> (also, why use an expression in that example and not class Foo extends …?

It's just a minimal obviously correct syntax where ignorePureComponents doesn't work right.

This also fixes usages of non-decorator wrapper functions (and maybe `export default class extends…`?) 

In practice this example probably shouldn't be allowed but that should probably be enforced by a more general eslint rule. :)

dreid:
I'm not sure about the react version setting.

ljharb:
I'm not super concerned about people on v15.0, 15.1, or 15.2, but someone on v0.14 shouldn't get a pass from using PureComponent.

dreid:
Sorry, I might be misunderstanding you, are you proposing that ignorePureComponents shouldn't do anything on < 0.15.3? There do not appear to be (but maybe I missed them) any existing version constraints on ignorePureComponents.

If so, should that be done in this PR? It seems like it's a larger change that would require more discussion.

dreid:
Is there anything else I need to do for this? I noticed there were some conflicts so I resolved them.

This was approved almost a month ago and hasn't been merged so I just wanted to make sure this wasn't waiting on some action from me.

yannickcr:
Thanks for your work! And sorry for the delay :)

PR #1132, opened 2017-03-29
remove quotes when considering the name of a proptype

ethanjgoldberg:
The added test illustrates the failing case. Proptypes with quotes
were referred to by the quoted name, instead of by the bare name,
for flow types.

### DIFF
--- a/lib/rules/prop-types.js
+++ b/lib/rules/prop-types.js
@@ -305,7 +305,7 @@ module.exports = {
      * @param {string} the identifier to strip
      */
     function stripQuotes(string) {
-      if (string[0] === '\'' || string[0] === '"') {
+      if (string[0] === '\'' || string[0] === '"' && string[0] === string[string.length - 1]) {
         return string.slice(1, string.length - 1);
       }
       return string;
--- a/lib/rules/prop-types.js
+++ b/lib/rules/prop-types.js
@@ -301,6 +301,17 @@ module.exports = {
     }
 
     /**
+     * Removes quotes from around an identifier.
+     * @param {string} the identifier to strip
+     */
+    function stripQuotes(string) {
+      if (string[0] === '\'' || string[0] === '"') {
+        return string.slice(1, string.length - 1);
+      }
+      return string;
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


lib/rules/prop-types.js
```diff
@@ -301,6 +301,17 @@ module.exports = {
     }
 
     /**
+     * Removes quotes from around an identifier.
+     * @param {string} the identifier to strip
+     */
+    function stripQuotes(string) {
+      if (string[0] === '\'' || string[0] === '"') {
```
ljharb: What about backticks?

ethanjgoldberg: [Flow doesn't seem to like them](https://flowtype.org/try/#0C4TwDgpgBACgTgezAZygXigbwFBSgAwGMEBbMAGwmAgH58AuKAOwFcSAjCOAGmwF8A3EA)

ljharb: Backticked strings require computed property brackets:
```js
{ [`complete?`]: number }
```
Does this not work with flow?

ethanjgoldberg: Oh wild.

It doesn't seem to work with flow in type declarations, and this makes sense to me. Like you say, flow would have to execute the code to compute the key name, since it might have `${}`s.

The lint rule also doesn't work for computed property names in normal `propTypes:` declaration. Maybe something could be done for simple expressions, like backticks without interpolation. But I'm curious if anyone has a good use case for dynamic proptype checking anyway.

ljharb: Will this string always have matching quotes? or would it be better to do an unconditional replace that looks for a starting `'`/`"` and a matching token at the end?

ethanjgoldberg: I really can't think of a way to get a key name to start with a quote, without wrapping it in quotes.

I agree it feels a bit hacky. I'll work on getting a fix this weekend.

ethanjgoldberg: @ljharb, well, it took a bit more than a weekend, but i've updated the function as you suggest.

Issue #807, opened 2016-09-06
Enhance jsx-sort-props to allow for non-alphabetically sorted groupings


ericmatthys:
Is there any interest in including a rule like the following in this plugin?

https://github.com/ericmatthys/eslint-plugin-jsx-grouped-sort-props

This is a rule that allows sorting of jsx props in groups without requiring an alphabetical sort. My motivation was wanting specific props sorted first (e.g. `key` and `ref`), then all other non-callback props in any order, then all callback props in any order at the end. The rule allows specific prop names in an array or various pre-configured groups (e.g. `html`, `reserved`, `callback`). If there is a spread, it validates the prop order before and after the spread independently. For the use case above, the configuration would look like:

``` js
[
  'key',
  'ref',
  { group: 'other' },
  { group: 'callback' }
]
```

**Valid:**

``` jsx
<Button
  ref="btn"
  isLoading={true}
  onClick={onClick}
/>
```

``` jsx
<Button
  ref="btn"
  onClick={onClick}
  {...props}
  isLoading={true}
/>
```

**Invalid:**

``` jsx
<Button
  isLoading={true}
  ref="btn"
  onClick={onClick}
/>
```

``` jsx
<Button
  ref="btn"
  onClick={onClick}
  isLoading={true}
/>
```

``` jsx
<Button
  ref="btn"
  {...props}
  onClick={onClick}
  isLoading={true}
/>
```

ljharb:
I'd also want the ability to require spread props appear first or last.

Daniel15:
Sounds like an enhanced version of [jsx-sort-props](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/jsx-sort-props.js). There's already a `callbacksLast` configuration option.

ericmatthys:
The main difference regarding `callbacksLast` is that `jsx-sort-props` is meant to enforce alphabetical sorting, while this rule is not.

lencioni:
I agree with @Daniel15. This seems like it would be a good enhancement to `jsx-sort-props`.

lencioni:
Related: #786

yannickcr:
Definitely interested if we can found a way to merge it with the existing `jsx-sort-props` rule.

merlinpatt:
My use case for this is that I typically want to put certain props first or last and I don't care about what's in between.

[...]

PR #1134, opened 2017-03-31
Add a `reservedFirst` option to the `jsx-sort-props` rule

MatthewHerbst:
This PR is in response to reading some of the comments in #807 plus real-world needs by my company: we always put the `key` prop first, and then alphabetize the rest of them.

First time writing an ESLint rule option, so please let me know what could be better.

Thinking about this option, I intentionally placed it before `shorthandFirst` so that it trumps that option. This seemed appropriate to me based on thinking about props as reserved or not, and considering that there are no reserved shorthand props.

For "reserved," I am considering all props that are specifically related to React. So for example, DOM properties are not reserved since they are related to JSX, not React.

-------------------------------------------------------

This adds a new option, `reservedFirst`, to the `jsx-sort-props` rule. This option can be enabled with a boolean, or an array configuration. Using the array configuration allows customizing the list of reserved props that are checked.

### DIFF
--- a/docs/rules/jsx-sort-props.md
+++ b/docs/rules/jsx-sort-props.md
@@ -28,7 +28,8 @@ The following patterns are considered okay and do not cause warnings:
   "shorthandFirst": <boolean>,
   "shorthandLast": <boolean>,
   "ignoreCase": <boolean>,
-  "noSortAlphabetically": <boolean>
+  "noSortAlphabetically": <boolean>,
+  "reservedFirst": <boolean>|<array<string>>,
 }]
 ...
 ```
@@ -73,6 +74,26 @@ When `true`, alphabetical order is not enforced:
 
 ```jsx
 <Hello tel={5555555} name="John" />
+```
+
+### `reservedFirst`
+
+This can be a boolean or an array option.
+
+When `reservedFirst` is defined, React reserved props (`children`, `dangerouslySetInnerHTML` - **only for DOM components**, `key`, and `ref`) must be listed before all other props, but still respecting the alphabetical order:
+
+```jsx
+<Hello key={0} ref="John" name="John">
+  <div dangerouslySetInnerHTML={{__html: 'ESLint Plugin React!'}} ref="dangerDiv" />
+</Hello>
+```
+
+If given as an array, the array's values will override the default list of reserved props. **Note**: the values in the array may only be a **subset** of React reserved props.
+
+With `reservedFirst: [2, ["key"]]`, the following will not warn:
+
+```jsx
+<Hello key={'uuid'} name="John" ref="ref" />
 ```
 
 ## When not to use
--- a/lib/rules/jsx-sort-props.js
+++ b/lib/rules/jsx-sort-props.js
@@ -4,6 +4,7 @@
  */
 'use strict';
 
+var elementType = require('jsx-ast-utils/elementType');
 var propName = require('jsx-ast-utils/propName');
 
 // ------------------------------------------------------------------------------
@@ -12,6 +13,72 @@ var propName = require('jsx-ast-utils/propName');
 
 function isCallbackPropName(name) {
   return /^on[A-Z]/.test(name);
+}
+
+var COMPAT_TAG_REGEX = /^[a-z]|\-/;
+function isDOMComponent(node) {
+  var name = elementType(node);
+
+  // Get namespace if the type is JSXNamespacedName or JSXMemberExpression
+  if (name.indexOf(':') > -1) {
+    name = name.substring(0, name.indexOf(':'));
+  } else if (name.indexOf('.') > -1) {
+    name = name.substring(0, name.indexOf('.'));
+  }
+
+  return COMPAT_TAG_REGEX.test(name);
+}
+
+var RESERVED_PROPS_LIST = [
+  'children',
+  'dangerouslySetInnerHTML',
+  'key',
+  'ref'
+];
+
+function isReservedPropName(name, list) {
+  return list.indexOf(name) >= 0;
+}
+
+/**
+ * Checks if the `reservedFirst` option is valid
+ * @param {Object} context The context of the rule
+ * @param {Boolean|Array<String>} reservedFirst The `reservedFirst` option
+ * @return {?Function} If an error is detected, a function to generate the error message, otherwise, `undefined`
+ */
+// eslint-disable-next-line consistent-return
+function validateReservedFirstConfig(context, reservedFirst) {
+  if (reservedFirst) {
+    if (Array.isArray(reservedFirst)) {
+      // Only allow a subset of reserved words in customized lists
+      // eslint-disable-next-line consistent-return
+      var nonReservedWords = reservedFirst.filter(function(word) {
+        if (!isReservedPropName(word, RESERVED_PROPS_LIST)) {
+          return true;
+        }
+      });
+
+      if (reservedFirst.length === 0) {
+        return function(decl) {
+          context.report({
+            node: decl,
+            message: 'A customized reserved first list must not be empty'
+          });
+        };
+      } else if (nonReservedWords.length > 0) {
+        return function(decl) {
+          context.report({
+            node: decl,
+            message: 'A customized reserved first list must only contain a subset of React reserved props.' +
+              ' Remove: {{ nonReservedWords }}',
+            data: {
+              nonReservedWords: nonReservedWords.toString()
+            }
+          });
+        };
+      }
+    }
+  }
 }
 
 module.exports = {
@@ -44,6 +111,9 @@ module.exports = {
         // Whether alphabetical sorting should be enforced
         noSortAlphabetically: {
           type: 'boolean'
+        },
+        reservedFirst: {
+          type: ['array', 'boolean']
         }
       },
       additionalProperties: false
@@ -58,9 +128,19 @@ module.exports = {
     var shorthandFirst = configuration.shorthandFirst || false;
     var shorthandLast = configuration.shorthandLast || false;
     var noSortAlphabetically = configuration.noSortAlphabetically || false;
+    var reservedFirst = configuration.reservedFirst || false;
+    var reservedFirstError = validateReservedFirstConfig(context, reservedFirst);
+    var reservedList = Array.isArray(reservedFirst) ? reservedFirst : RESERVED_PROPS_LIST;
 
     return {
       JSXOpeningElement: function(node) {
+        // `dangerouslySetInnerHTML` is only "reserved" on DOM components
+        if (reservedFirst && !isDOMComponent(node)) {
+          reservedList = reservedList.filter(function(prop) {
+            return prop !== 'dangerouslySetInnerHTML';
+          });
+        }
+
         node.attributes.reduce(function(memo, decl, idx, attrs) {
           if (decl.type === 'JSXSpreadAttribute') {
             return attrs[idx + 1];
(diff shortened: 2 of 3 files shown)


docs/rules/jsx-sort-props.md
```diff
@@ -75,6 +79,34 @@ When `true`, alphabetical order is not enforced:
...
+
+```jsx
+<Hello key={0} ref="John" name="John" />
+```
+
+If given as an object, the option accepts two optional sub-options:
+
+#### `reservedList`
+
+Use this to optionally override the default list of reserved props.
```
ljharb: i think the defaults should go here, not on line 86

lib/rules/jsx-sort-props.js
```diff
@@ -14,6 +14,19 @@ function isCallbackPropName(name) {
   return /^on[A-Z]/.test(name);
 }
 
+var RESEVED_PROPS_LIST = [
```
ljharb: RESEVED → RESERVED

lib/rules/jsx-sort-props.js
```diff
@@ -14,6 +14,19 @@ function isCallbackPropName(name) {
...
+  'children',
+  'dangerouslySetInnerHTML',
+  'htmlFor',
+  'key',
+  'ref',
+  'style'
+];
+
+function isReservedPropName(name, list) {
```
ljharb: this is just "is in list" if it takes the list argument; it'd probably be better to define this function so that `reservedList` is available via closure, and this function would take only the `name`?

lib/rules/jsx-sort-props.js
```diff
@@ -44,6 +57,17 @@ module.exports = {
         // Whether alphabetical sorting should be enforced
         noSortAlphabetically: {
           type: 'boolean'
+        },
+        reservedFirst: {
```
ljharb: can we ensure that `additionalProperties` are disallowed on the object form? (not sure if line 73 covers this or not)

lib/rules/jsx-sort-props.js
MatthewHerbst: I was playing with making it a "default" option if `list` wasn't specified; would you prefer that?

Regarding your general comment, I think it's important to allow the user to customize this list since some people may only want certain "reserved" props first. For example, maybe we want `key` first but all the rest to be alphabetized. I'm open to not allowing that (or, maybe better, to allowing overriding the list with only a subset of the reserved props), but I think without this option this  option won't be as useful.

```
function isReservedProp(name, list) {
  var listToUse = list || RESERVED_PROPS_LIST;
  return list.indexOf(name) >= 0;
}
```

lib/rules/jsx-sort-props.js
MatthewHerbst: Woops! Thanks

lib/rules/jsx-sort-props.js
MatthewHerbst: I believe that it (line 73) does do that. http://eslint.org/docs/developer-guide/working-with-rules

docs/rules/jsx-sort-props.md
MatthewHerbst: I'm open to that; I was hoping to let people know what the rule is doing without making them read all the options.

MatthewHerbst:
As alluded to in one of my reply-comments, would your concern be alleviated if the customized list could only be a subset of the full list rather than accepting arbitrary values? I do think having some customization of the option is important for people who might want to use it.

Noted on `style`/`htmlFor`, I'll update, thanks.

lib/rules/jsx-sort-props.js
ljharb: I can see an argument for allowing a custom sorting of the reserved list - but by calling it "reserved", that means it's not something the user can change. In other words, "foo" isn't considered "reserved" by React, thus it should *never* be considered "reserved" by anything else.

ljharb:
Yes, I think that if the list must always contain a non-empty subset of the default list, that would alleviate my concerns there (although I don't really see the utility).

lib/rules/jsx-sort-props.js
MatthewHerbst: (Part of making this trivial functionality a function was just to match the style of there also being an `isCallbackPropName` function.)

MatthewHerbst:
@ljharb I've updated based on the discussion. Now the customized reserved list may only be a subset of the actual list. The length of the check got me to move it into its own function to make the core `create` function easier to read. Let me know your thoughts; thanks for being flexible on this, really appreciated

docs/rules/jsx-sort-props.md
```diff
@@ -75,6 +79,36 @@ When `true`, alphabetical order is not enforced:
 <Hello tel={5555555} name="John" />
 ```
 
+### `reservedFirst`
+
+This can be a boolean or an object option.
+
+When `reservedFirst` is defined, React reserved props (`children`, `dangerouslySetInnerHTML`, `key`, `ref`) must be listed before all other props, but still respecting the alphabetical order:
```
ljharb: actually now that i think of it; isn't `dangerouslySetInnerHTML` a special prop only on DOM nodes?

I think you may want separate lists for DOM nodes, and custom components.

docs/rules/jsx-sort-props.md
```diff
@@ -75,6 +79,36 @@ When `true`, alphabetical order is not enforced:
...
+
+With `reservedList: ["key"]`, the following will not warn:
+
+```jsx
+<Hello key={'uuid'} name="John" ref="ref" />
+```
+
+#### `noSortReservedAlphabetically`
+
+When `true`, alphabetical sorting of reserved props is not enforced, but alphabetical sorting of other props still is, unless `noSortAlphabetically` is set to `true` as well.
```
ljharb: to confirm: if `noSortAlphabetically` is set to `true`, then this option is ignored?

MatthewHerbst: No, right now setting `noSortAlphabetically` will allow non-reserved to be listed non-alphabetically, but reserved would still need to be alphabetical. Should setting `noSortAlphabetically` allow reserved to be non-alphabetical as well?

docs/rules/jsx-sort-props.md
MatthewHerbst: Hmmm. Well, all DOM nodes can take the other props, so I think this is ok; I don't actually know what happens if you set `dangerouslySetInnerHTML` on a composite component. Either way, I think users not putting `dangerouslySetInnerHTML` on non-DOM components is something left for a different rule check.

docs/rules/jsx-sort-props.md
ljharb: I would think that `noSortAlphabetically` would be a universal setting

ljharb: although, if it is, that probably makes this setting a bit more complex.

Can you elaborate on the use case for wanting alphabetical sorting, but not wanting it to apply universally? (or the inverse)

docs/rules/jsx-sort-props.md
ljharb: It does nothing on a composite component; what I mean is, it's a normal (non-reserved) prop on them.

docs/rules/jsx-sort-props.md
MatthewHerbst: Oh, I was just trying to match the other options, wasn't trying to create new functionality. I'll update so that setting `noSortAlphabetically` overrides `noSortReservedAlphabetically`

docs/rules/jsx-sort-props.md
MatthewHerbst: Is it? I guess my version of "reserved" are things that have special meaning in React, which I feel like this kind of does, even though it mimicks "real" DOM functionality

MatthewHerbst: (I don't care either way though imho)

docs/rules/jsx-sort-props.md
ljharb: I'm still questioning the need for different sorting here tho - can you elaborate on the use case for wanting alphabetical sorting, but not wanting it to apply universally? (or the inverse)

MatthewHerbst: I think it was a pre-mature optionality (I think I just made that phrase up). I'll remove it - I guess I just was trying to allow over-customization, but maybe someone else can add that in the future if they need it and can present a compelling reason.

MatthewHerbst: Plus, actually, removing that significantly simplifies the option, since now it can be `<boolean>|<array>` rather than `<boolean>|{...}`

MatthewHerbst: Updated; it's much less complex now, good catch calling me out on that!

docs/rules/jsx-sort-props.md
```diff
@@ -75,6 +76,24 @@ When `true`, alphabetical order is not enforced:
 <Hello tel={5555555} name="John" />
 ```
 
+### `reservedFirst`
+
+This can be a boolean or an array option.
+
+When `reservedFirst` is defined, React reserved props (`children`, `dangerouslySetInnerHTML`, `key`, `ref`) must be listed before all other props, but still respecting the alphabetical order:
```
ljharb: I still think `dangerouslySetInnerHTML` should only be considered "reserved" on DOM nodes, even if the schema has a single list.

ljharb: I also don't see any tests that exercise this prop - please add tests that cover every "reserved prop" :-)

MatthewHerbst: I'll add the tests for them all shortly :)

Regarding [`dangerouslySetInnerHTML`](https://facebook.github.io/react/docs/dom-elements.html#dangerouslysetinnerhtml), from [playing around with a fiddle](https://jsfiddle.net/69z2wepo/75289/), it seems to have no affect at all on composite components - so it seems to be treated as a "normal" props. I'll add the DOM node check for it.

(Maybe there should be another rule to ensure people don't use `dangerouslySetInnerHTML` on composite components? Seems like something that could easily cause bugs for people who expect it to work.)

MatthewHerbst: @ljharb tried working on this last night, ran into a small problem: what's the best way to identify if an AST node is a DOM element or not? I saw that `void-dom-elements-no-children` just creates a map of the ones it cares about, but for this we would care about all of them. Checking if the name starts with an upper-case letter would be a good guess, but would be wrong sometimes. It's not clear to me if `node.type` or if something in `node.attributes` can be used for this.

If I do just need to hardcode the list, that would mean we wouldn't catch custom DOM elements - is that ok?

ljharb: Hmm - React treats all lowercased names as a DOM element, I think; but either way I think `node.type` will only be a string when it's a DOM element.

MatthewHerbst: > React treats all lowercased names as a DOM element


I don't believe that's correct, you can name an imported composite component whatever you want and use that within JSX no problem; there's actually a rule in this repo to discourage that.

`jsx-pascal-case` (live-edit:) I realized that this rule must have the ability to determine user defined components! [I'll utilize the method found in that rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/jsx-pascal-case.js).

~~[Here's]( https://github.com/facebook/react/blob/master/src/test/ReactTestUtils.js?source=c#L155) how React Test Utils does it (it comes down to check [`ReactElement.isValidElement`](https://github.com/facebook/react/blob/b1b4a2fb252f26fe10d29ba60d85ff89a85ff3ec/src/isomorphic/classic/element/ReactElement.js#L376)). I'll work on it tomorrow and see what we have available on `node`.~~

MatthewHerbst:
@ljharb I've fully updated the changes based on what we've discussed, let me know your thoughts, thanks!

MatthewHerbst:
*bump @lencioni @yannickcr @EvNaverniouk if any of you have some spare moments, looking for some love <3 thanks!

MatthewHerbst:
@ljharb do all 4 reviewers on the PR need to approve it?

ljharb:
@MatthewHerbst no, not necessarily.

Issue #913, opened 2016-10-16
Nested proptypes in FlowType causes RangeError: Maximum call stack size exceeded


willlma:
I'm guessing this is causing an infinite loop. Not sure is this belongs here or in the [Flow repo](https://github.com/facebook/flow/) (or if I'm just doing things wrong). I've got a nested array that looks like this:

```
type Note = {text: string, children?: Note[]}
type Props = {
  notes: Note[]
};
```

Which is throwing `RangeError: Maximum call stack size exceeded`

Here's the full stack (absolute path to project abbreviated as `~`):

```
    at getKeyValue (~/node_modules/eslint-plugin-react/lib/rules/prop-types.js:289:25)
    at iterateProperties (~/node_modules/eslint-plugin-react/lib/rules/prop-types.js:308:21)
    at buildTypeAnnotationDeclarationTypes (~/node_modules/eslint-plugin-react/lib/rules/prop-types.js:441:11)
    at buildTypeAnnotationDeclarationTypes (~/node_modules/eslint-plugin-react/lib/rules/prop-types.js:433:20)
    at buildTypeAnnotationDeclarationTypes (~/node_modules/eslint-plugin-react/lib/rules/prop-types.js:472:28)
    at ~/node_modules/eslint-plugin-react/lib/rules/prop-types.js:442:54
    at iterateProperties (~/node_modules/eslint-plugin-react/lib/rules/prop-types.js:311:11)
    at buildTypeAnnotationDeclarationTypes (~/node_modules/eslint-plugin-react/lib/rules/prop-types.js:441:11)
    at buildTypeAnnotationDeclarationTypes (~/node_modules/eslint-plugin-react/lib/rules/prop-types.js:433:20)
    at buildTypeAnnotationDeclarationTypes (~/node_modules/eslint-plugin-react/lib/rules/prop-types.js:472:28)
```

I've stuck with a simple `type Props = { notes: Object[] }` for now

ljharb:
See also #1032.

jetpacmonkey:
Is there any good workaround here? I'm still seeing this issue.

ljharb:
#1138 / #1377 / #1412 should help, but this might still require further fixes.

jetpacmonkey:
It looks like #1138 did address this issue, but it was closed in favor of #1377, which, as far as I can tell, does not.

jetpacmonkey:
I opened #1653 based off of the commit in #1138 that fixed recursive proptype definitions.

PR #1138, opened 2017-04-03
Add further support for Flow prop types.

### LINKED ISSUE #453: Support Flow Prop/State syntax
Is it possible for you to support this:
`class MyComponent extends React.Component<void, Props, void> {`

which does the same thing as this:

```
  props: Props
```

Also described [here](https://www.reddit.com/r/reactjs/comments/44yqge/fyi_proptypes_are_on_their_way_out/czuflhe).

phpnode:
This adds support for constructions like this:

    type Props = {name: string};
    class X extends React.Component<void, Props, void> {
      ...
    }

Which is the preferred way to declare prop types using Flow.

The less preferred syntax is still supported:

    class Z extends React.Component {
      props: Props;
    }

Fixes #453, #913

### DIFF
--- a/lib/rules/prop-types.js
+++ b/lib/rules/prop-types.js
@@ -484,7 +484,7 @@ module.exports = {
      * Creates the representation of the React props type annotation for the component.
      * The representation is used to verify nested used properties.
      * @param {ASTNode} annotation Type annotation for the props class property.
-     * @param {Array} seen Keeps track of annotations we've already seen.
+     * @param {Set<ASTNode>} seen Keeps track of annotations we've already seen.
      * @return {Object|Boolean} The representation of the declaration, true means
      *    the property is declared without the need for further analysis.
      */
@@ -492,13 +492,13 @@ module.exports = {
       if (seen === void 0) {
         // Keeps track of annotations we've already seen to
         // prevent problems with cyclic types.
-        seen = [];
+        seen = new Set();
       }
-      if (seen.indexOf(annotation) > -1) {
+      if (seen.has(annotation)) {
         // this must be a recursive type annotation, just accept anything.
         return true;
       }
-      seen.push(annotation);
+      seen.add(annotation);
       switch (annotation.type) {
         case 'GenericTypeAnnotation':
           if (typeScope(annotation.id.name)) {
--- a/lib/rules/prop-types.js
+++ b/lib/rules/prop-types.js
@@ -484,14 +484,25 @@ module.exports = {
      * Creates the representation of the React props type annotation for the component.
      * The representation is used to verify nested used properties.
      * @param {ASTNode} annotation Type annotation for the props class property.
+     * @param {Array} seen Keeps track of annotations we've already seen.
      * @return {Object|Boolean} The representation of the declaration, true means
      *    the property is declared without the need for further analysis.
      */
-    function buildTypeAnnotationDeclarationTypes(annotation) {
+    function buildTypeAnnotationDeclarationTypes(annotation, seen) {
+      if (seen === void 0) {
+        // Keeps track of annotations we've already seen to
+        // prevent problems with cyclic types.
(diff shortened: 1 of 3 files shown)


ljharb:
Why "void, Props, void"? props are the first argument to the constructor; what's the first void for?

phpnode:
@ljharb the first parameter is the default properties, e.g.

```js
type Props = {
  name: string
};
const defaultProps = {
  name: 'Foo'
};

type State = {
  foo: string
}

class Y extends React.Component<typeof defaultProps, Props, State> {
  state = { foo: 'bar' };
  static defaultProps = defaultProps;
  //  ...
}
```

ljharb:
can you explain a bit about how Flow handles that syntax in a generic way, such that I'd use it on a non-React component?

phpnode:
@ljharb I guess there are two things here, the type parameters and `$Diff`.

## Type Parameters
```js
// So this is a generic container
// `T` can be anything
class Container<T> {
  items: T[];
  
  push (...args: T[]) {
    this.items.push(...args);
    return this;
  }
}

// When extending we can pass the type parameter
// to the super class, making the type more specialized
class StringContainer extends Container<string> {
  
}

const c = new Container();
c.push(true, 123, 'this is fine'); // all good

const sc = new StringContainer();
sc.push('this is fine'); 
sc.push(false); // flow error
```

## $Diff<A, B>
It takes two (object) types and returns the properties that exist in one and not the other, hopefully this is helpful:

```js
type APIOptions = {
  hostname: string,
  port: number
};

const defaultOptions = {
  port: 80  
};


// Note: I don't think $Subtype<> should really be required,
// and it's probably a flow bug. You may be able to omit it.
type RequiredOptions = $Subtype<$Diff<APIOptions, typeof defaultOptions>>;

class API {
  opts: APIOptions;

  constructor (options: RequiredOptions) {
	this.opts = Object.assign({}, defaultOptions, options);
  }

  getURL () {
    return this.opts.hostname + ':' + this.opts.port;
  }
}

const api = new API({
  hostname: 'foo'
});

console.log(api.getURL());

```

phpnode:
@ljharb @yannickcr friendly ping - this is working locally on quite a large codebase so quite confident it works, anything else I need to do before it can be merged?

lib/rules/prop-types.js
```diff
@@ -445,14 +484,25 @@ module.exports = {
...
+     * @param {Array} seen Keeps track of annotations we've already seen.
      * @return {Object|Boolean} The representation of the declaration, true means
      *    the property is declared without the need for further analysis.
      */
-    function buildTypeAnnotationDeclarationTypes(annotation) {
+    function buildTypeAnnotationDeclarationTypes(annotation, seen) {
+      if (seen === void 0) {
+        // Keeps track of annotations we've already seen to
+        // prevent problems with cyclic types.
+        seen = [];
```
ljharb: we support node 4+ now, so let's use a `Set` instead of an array here

EvNaverniouk:
LGTM. We use Flow a lot so I can definitely keep a close eye on this once it gets released to see if it causes any issues, but the test coverage looks good already.

keithkml:
Very excited for this to launch!

igorline:
Waiting for this to be merged!

ljharb:
@phpnode would you mind rebasing this on top of latest master? #1236 may also be related/overlap with this PR.

MoOx:
Maybe not relevant since #1377 has been merged.

ljharb:
@jseminck can you confirm that #1377 made this PR obsolete?

jseminck:
I was not aware of this PR!

Yes. It should cover the same functionality as this PR and more ( the new syntax ). This PR contains more test cases though. I will add them as well to make sure they all pass. 

I think they will because my PR has a similar implementation as the prop-types rule and these cases are covered there.

Issue #1113, opened 2017-03-14
Rule idea: no-direct-mutation-props


ianschmitz:
A component is never meant to modify its own props (https://facebook.github.io/react/docs/components-and-props.html#props-are-read-only).

I suggest we had a `no-mutation-props` rule to check for mutations of a components own props. It looks like we could likely copy the logic used in [no-direct-mutation-state.js](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/no-direct-mutation-state.js).

viankakrisna:
maybe just go ahead and send a pr?

ljharb:
Seems reasonable. However, this rule would have to warn *any* time the entire props object is passed to a function (which is a bad practice anyways) because that makes it impossible (in most cases) to determine if it's being mutated.

pgangwani:
How can I achieve "no-mutation-props" in my eslint config ?

ljharb:
@pgangwani #1416 isn’t merged yet.

ljharb:
You can read the PR to find out the challenges.

Really, stopping mutation isn’t something that can be done with static analysis - you can do it in your tests more reliably.

PR #1145, opened 2017-04-08
Add no-mutation-props rule

ianschmitz:
PR for #1113.

Note the change in /index.js to the recommended rules. Not sure if this rule should be part of the recommended set or not?

@ljharb This does not currently warn any time the entire props object is passed to a function. It currently has the same checks as the `no-direct-mutation-state` rule.


ianschmitz:
I'm not super familiar with MobX, but it may be worth thinking about how this rule may impact those users?

lib/rules/no-mutation-props.js
```diff
@@ -0,0 +1,89 @@
...
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Prevent direct mutation of this.props',
+      category: 'Possible Errors',
+      recommended: true
```
ljharb: Changing which rules are recommended is semver-major, so that definitely shouldn't be done in this PR.

lib/rules/no-mutation-props.js
```diff
@@ -0,0 +1,89 @@
...
+
+  create: Components.detect(function (context, components, utils) {
+
+    /**
+     * Checks if the component is valid
+     * @param {Object} component The component to process
+     * @returns {Boolean} True if the component is valid, false if not.
+     */
+    function isValid(component) {
+      return Boolean(component && !component.mutateProps);
```
ljharb: let's use `‼x` instead of `Boolean(x)` to coerce to booleans - and in this case, `‼component && !component.mutateProps` is sufficient

lib/rules/no-mutation-props.js
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /**
+     * Reports this.props mutations for a given component
+     * @param {Object} component The component to process
+     */
+    function reportMutations(component) {
+      var mutation;
+      for (var i = 0, j = component.mutations.length; i < j; i++) {
+        mutation = component.mutations[i];
```
ljharb: can this use `component.mutations.forEach` instead of a loop?

lib/rules/no-mutation-props.js
```diff
@@ -0,0 +1,89 @@
...
+        context.report({
+          node: mutation,
+          message: 'Do not mutate props.'
```
ljharb: it might be nice if this error message held more information - in the message, or by providing source locations.

lib/rules/no-mutation-props.js
```diff
@@ -0,0 +1,89 @@
...
+    return {
+      AssignmentExpression: function (node) {
+        var item;
+
+        if (!node.left || !node.left.object || !node.left.object.object) {
+          return;
+        }
+
+        item = node.left.object;
+        while (item.object.property) {
```
ljharb: what if `item` doesn't have a `.object` property?

lib/rules/no-mutation-props.js
```diff
@@ -0,0 +1,89 @@
...
+          components.set(node, {
+            mutateProps: true,
+            mutations: mutations
+          });
+        }
+      },
+
+      'Program:exit': function () {
+        var list = components.list();
+        for (var component in list) {
```
ljharb: let's use `Object.keys(list).forEach` if it's an object, and if it's an array, `list.forEach` - `for..in` should be avoided.

index.js
```diff
@@ -117,6 +118,7 @@ module.exports = {
         'react/jsx-uses-vars': 2,
         'react/no-deprecated': 2,
         'react/no-direct-mutation-state': 2,
+        'react/no-mutation-props': 2,
```
ljharb: (same here - adding rules to the recommended set is semver-major, so let's remove this)

ianschmitz:
Thanks for the quick review @ljharb!

I've made the changes you requested.

docs/rules/no-mutation-props.md
```diff
@@ -0,0 +1,19 @@
...
+## Rule Details
+
+This rule is aimed to forbid the use of mutating `this.props`.
+
+The following patterns are considered warnings:
+
+```jsx
+var Hello = React.createClass({
+  render: function() {
+    this.props.name = this.props.name.toUpperCase();
```
ljharb: What about things like `this.props.list.push('foo')`, or other array mutators? What about `delete this.props.name`, or `Object.defineProperty(this.props, something)`, or `Object.assign(this.props, something)`?

docs/rules/no-mutation-props.md
```diff
@@ -0,0 +1,19 @@
+# Prevent mutation of this.props (no-mutation-props)
+
+NEVER mutate `this.props`, as all React components must act like pure functions with respect to their props.
```
Daniel15: "NEVER" in all-caps is a bit strong, maybe change it to "Never"? :)

docs/rules/no-mutation-props.md
Daniel15: ^^ Further to that, you probably also want to catch this:
```
const {list} = this.props;
list.push('foo');
```

docs/rules/no-mutation-props.md
ljharb: I would be fine with making it stronger, but there's no such thing as "super caps"

joeybaker:
@ianschmitz I hope you don't mind. I used your fork as a base and added a bunch more fixes. https://github.com/yannickcr/eslint-plugin-react/pull/1416

@ljharb @Daniel15 I'd love to get that PR looked at, and if it's good, we can port nearly all of it to the `no-direct-mutation-state` rule to dramatically improve it.

docs/rules/no-mutation-props.md
joeybaker: I've got a PR submitted that builds on @ianschmitz's work and catches this and many other things as well :) https://github.com/yannickcr/eslint-plugin-react/pull/1416

ianschmitz:
Not at all @joeybaker. Sorry, I've fell out of touch with this PR 😢 

Thanks for helping out everyone!

ianschmitz:
Closing in favor of the updated PR over at #1416. Thanks @joeybaker !

ljharb:
In the future it'd be preferable to ask maintainers to add your commits to the original PR; replacement PRs are clutter.

joeybaker:
@ljharb Sounds good. I'm happy for you to do that and close the new PR :)

ljharb:
Nah at this point it's open, we'll continue with that :-)

Issue #1144, opened 2017-04-08
React 15.5: Support `react-create-class` and `prop-types`


ljharb:
To support React 15.5, we need the following changes asap:

 - [x] Everywhere we support `React.createClass`, we also need to support an `import` or `require` of `react-create-class`
 - [x] if we support `PropTypes.*` anywhere, we also need to support an `import` or `require` of `prop-types`
 - [x] we need to make `no-deprecated` warn on React.createClass and React.PropTypes usage and recommends the packages as well. It should apply to all versions of React since it's safe to migrate to the standalone packages on all versions.

yannickcr:
For the third point we already have the [`no-deprecated`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-deprecated.md) rule that can be used.

EvNaverniouk:
I can take care of the `no-unused-prop-types` rule. Will look into it early next week.

EvNaverniouk:
After review, looks like the `no-unused-prop-types` and `prop-types` rules are not affected by the React 15.5 changes as they don't make any assumptions about where the `PropTypes` object comes from. No changes needed there.

kentor:
In the README there's this section for the eslint config:
```
{
  "settings": {
    "react": {
      "createClass": "createClass", // Regex for Component Factory to use, default to "createClass"
      "pragma": "React",  // Pragma to use, default to "React"
      "version": "15.0" // React version, default to the latest React stable release
    }
  }
}
```

Changing `"createClass": "createClass"` to `"createClass": "createReactClass"` worked for the `sort-comp` rule for component detection. Haven't tested other rules...

ljharb:
@EvNaverniouk I think it's a flaw in the `prop-types` rule in particular that any validator named "shape", for example, assumes the semantics of `PropTypes.shape` - but you're right that that means it's not a blocker for React 15.5.

@kentor if that's the case, then perhaps we should default `createClass` to `createReactClass` when the `version` is 15.5+?

yannickcr:
React 15.5.0 support seems to be complete.

* Component factory now default to `createReactClass` (you still got the possibility to change it to `createClass` using the shared settings).
* PropTypes detection is not affected by the changes in React 15.5.0.
* `React.PropTypes` and `React.createClass` were added to `no-deprecated`.

I will publish a RC for the 7.0.0 (since there is some breaking changes here), you can test it with:

```
npm install eslint-plugin-react@next
```

[...]

PR #1148, opened 2017-04-11
Add new deprecations from 15.5.0

Calyhre:
Added more deprecations rules regarding last react update.

But it does not seem to match patterns like this:
```javascript
import { PropTypes, createClass } from 'react';
```

Any idea on how we can match those too?

### DIFF
--- a/lib/rules/no-deprecated.js
+++ b/lib/rules/no-deprecated.js
@@ -66,6 +66,7 @@ module.exports = {
       deprecated.MemberExpression['Perf.getMeasurementsSummaryMap'] = ['15.0.0', 'Perf.getWasted'];
       // 15.5.0
       deprecated.MemberExpression['React.createClass'] = ['15.5.0', 'the npm module create-react-class'];
+      deprecated.MemberExpression['React.PropTypes'] = ['15.5.0', 'the npm module prop-types'];
 
       return deprecated;
     }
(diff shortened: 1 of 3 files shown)


ljharb:
(linking to #1144)

yannickcr:
Actually destructuring is not handled for any deprecation in this rule. Mainly because it was not something common at the time it was written. So I think it should not be a blocker for this PR.

But, since it is now a very common pattern, I'm working on a patch to add destructuring support to the whole rule.

lib/rules/no-deprecated.js
```diff
@@ -64,6 +64,9 @@ module.exports = {
       deprecated.MemberExpression['Perf.printDOM'] = ['15.0.0', 'Perf.printOperations'];
       deprecated.MemberExpression['ReactPerf.getMeasurementsSummaryMap'] = ['15.0.0', 'ReactPerf.getWasted'];
       deprecated.MemberExpression['Perf.getMeasurementsSummaryMap'] = ['15.0.0', 'Perf.getWasted'];
+      // 15.5.0
+      deprecated.MemberExpression['React.createClass'] = ['15.5.0', 'the npm module create-react-class'];
+      deprecated.MemberExpression['React.PropTypes'] = ['15.5.0', 'the npm module prop-types'];
```
yannickcr: Please use the `pragma` variable instead of the hardcoded `React` string (you can take the other deprecation as example).

Calyhre: Pragma would match if there is other thing than `React.*` used right?
I don't think it should match if this is from another object than `React`.

yannickcr: `pragma` will match other thing than `React` only if the user has customized it to match something else (with a comment or using the shared settings).

ljharb:
@yannickcr i'm fine with not handling destructuring in this PR (especially if you're working on adding it asap), but named imports should be handled, no?

lib/rules/no-deprecated.js
Calyhre: Got it 👍  PR updated

yannickcr:
@ljharb like destructuring, it's something that was not handled by the rule, so I think it should be done separately from this PR.

PR #1149, opened 2017-04-12
Prepend "react/" to rule options documentation code

Adzz:
Using `"jsx-pascal-case": ...` leads to
```shell
1:1  error  Definition for rule 'jsx-pascal-case' was not found  jsx-pascal-case
```

We need `"react/jsx-pascal-case": ...` as shown in the [Readme](https://github.com/yannickcr/eslint-plugin-react/blob/master/README.md)

### DIFF
--- a/docs/rules/display-name.md
+++ b/docs/rules/display-name.md
@@ -29,7 +29,7 @@ var Hello = React.createClass({
 
 ```js
 ...
-"display-name": [<enabled>, { "ignoreTranspilerName": <boolean> }]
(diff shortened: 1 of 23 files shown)


Adzz:
Good question, I didn't check! I can have a look see for you if you like

Adzz:
There's a few, let me do this proper and add them here, or would you prefer a different PR?

ljharb:
@Adzz here would be great, thanks :-) please rebase them all down to a single commit.

PR #1155, opened 2017-04-19
Prevent adjacent inline elements not separated by whitespace.

SeanHayes:
FYI, this is my first time writing an eslint rule.

I wrote this because I want to be warned when I have adjacent inline elements without any whitespace, as it affects the way the page looks when unstyled, and may affect the way search engines see the page.

Let me know if you think any changes are needed. If this is something you're interested in merging I'll add the related docs.

### DIFF
--- a/lib/rules/no-adjacent-inline-elements.js
+++ b/lib/rules/no-adjacent-inline-elements.js
@@ -1,0 +1,115 @@
+/**
+ * @fileoverview Prevent adjacent inline elements not separated by whitespace.
+ * @author Sean Hayes
+ */
+
+'use strict';
+
+// ------------------------------------------------------------------------------
+// Helpers
+// ------------------------------------------------------------------------------
+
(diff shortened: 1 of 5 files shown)


lib/rules/no-adjacent-inline-elements.js
```diff
@@ -0,0 +1,76 @@
...
+  ' label select textarea').split(' ');
+var nbsp = '&nbsp;';
+
+function isInline(node) {
+  if (node.type === 'Literal') {
+    // Regular whitespace will be removed.
+    var raw = node.raw.trim();
+    // To properly separate inline elements, each end of the literal will need
+    // a &nbsp;.
+    return !(raw.startsWith(nbsp) && raw.endsWith(nbsp));
```
ljharb: also node 0.10 doesn't have startsWith or endsWith, so we'll want to use `.slice(0, 1)` and `.slice(-1)` here

PR #1156, opened 2017-04-20
Add rule for disallowing instantiation of classes in JSX props

Daniel15:
Instantiating a class in the `render` method of a React component means that a brand new instance will be created every single time the component is re-rendered. This is bad for performance, as it will result in the garbage collector being invoked way more than is necessary. This issue is in a similar category as `jsx-no-bind`.

### DIFF
--- a/docs/rules/jsx-no-new.md
+++ b/docs/rules/jsx-no-new.md
@@ -1,0 +1,21 @@
+# Prevent instantiation of classes in JSX attributes (jsx-no-new)
+
+Instantiating a class in the `render` method of a React component means that a brand new instance will be created every single time the component is re-rendered. This is bad for performance, as it will result in the garbage collector being invoked way more than is necessary.
+
+## Rule Details
+
+The following patterns are considered warnings:
+```jsx
+<Foo bar={new HelloWorld()} />
+<Foo bar={new HelloWorld()}></Foo>
+```
+
+The following patterns are not considered warnings:
+```jsx
+<Foo bar={helloWorld} />
+<Foo bar={helloWorld}></Foo>
+```
+
+## When Not To Use It
+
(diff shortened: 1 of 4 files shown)


ljharb:
I can't conceive of how this could ever become a problem. Why are people doing this in the first place?

The issue with jsx-no-bind is primarily that `.bind` is slow, and only secondarily that creating any new functions in the render path prevents `===` optimizations and increases GC (this turns out to not be that bad for performance in actual practice).

If we want to handle other patterns, what about creating new arrays and objects in the render path? What about function calls that do this and return them into the render path?

Daniel15:
> Why are people doing this in the first place?

That is what I'd love to know 😛  It's a pattern I stumbled upon today - I found ~100 places in our codebase that were doing something similar to this. A few of them were from people that didn't have much experience writing React code. I posted an internal group post and someone said "This is such an easy mistake to make when you're somewhat new to react"

>  prevents === optimizations and increases GC (this turns out to not be that bad for performance in actual practice).

Every little bit counts when you have a large app 😃  The app I deal with at work has over 300 Flux stores and probably thousands or tens of thousands of components in total. It's death by a thousand cuts if you have little GC-unfriendly patterns like this all over the place. Things like `bind` in render are particularly bad as an entire subtree could re-render unnecessarily.

It's fine if this is not merged, as I can just add it to our internal lint rules (we've already forked a few of the rules in this repo to adjust them to better suit our use cases). I just thought other people might find use in it.

ljharb:
I can't decide if I'm -1 or +1 on it; I think a generic configurable rule that tries to prevent any new creation of things in the render path might be more useful? but that also might be too hard to achieve.

Daniel15:
I don't care enough about this any more to push it through.

PR #1166, opened 2017-05-03
New Rule: No import of propTypes from react package

Slumber86:
This rule forbids using propTypes from react package, since is
deprecated in 15.5

Bonus: yarn.lock


Slumber86:
Some errors, re-post later

ljharb:
Please be sure to absolutely not include a yarn lockfile; we don't use yarn, and non-apps shouldn't have lockfiles.

Slumber86:
No problem, realized that this rule is rendundant with `no-deprecated`, will not repen, you can delete.
I opened #1167

PR #1167, opened 2017-05-03
Update docs for deprecated prop-types

Slumber86:


docs/rules/no-deprecated.md
```diff
@@ -22,6 +22,8 @@ React.createClass({ /* Class object */ });
 const propTypes = {
   foo: PropTypes.bar,
 };
+
+import {React PropTypes} from 'react';
```
ljharb: This is invalid syntax. Do you mean `import React, { PropTypes } from 'react'`?

Slumber86: Missed a comma, 
` import {React, PropTypes} from 'react'; `

ljharb: That would import a named import called "React", which doesn't exist.

The syntax I supplied is the correct one.

docs/rules/no-deprecated.md
```diff
@@ -22,6 +22,8 @@ React.createClass({ /* Class object */ });
 const propTypes = {
   foo: PropTypes.bar,
 };
+
+import {React, PropTypes} from 'react';
```
ljharb: this is `const { React, PropTypes } = require('react');` which is invalid - `React.React` isn't a thing.

PR #1173, opened 2017-05-06
Update jsx-filename-extension.md

luftywiranda13:
* Trailing comma is causing .json file stops working
* Trailing comma is supported in .eslintrc.js only


yannickcr:
Thanks!

Issue #857, opened 2016-09-21
react/jsx-curly-spacing not working


tleunen:
For some reasons, I can't get the jsx-curly-spacing rule to report errors.

After enabling it (using the default airbnb settings), I set several use case and nothing is reported.

The config has the following setting so all these cases should be reported, but they are not in my case.

```
'react/jsx-curly-spacing': ['error', 'never', { allowMultiline: true }],
```

```
{ something}
{something }
{ something }
```

yannickcr:
The `jsx-curly-spacing` rule apply only to jsx attributes.

tleunen:
Oh I see... I'll better read the doc next time.
But do you think it should also be applied to all jsx curly values?

```jsx
<Comp>
  {something}
</Comp>
```

yannickcr:
ESLint rules does not apply to `{something}` here ? If not we should support it in this rule imho.

ljharb:
I'm relatively sure we have a rule that enforces this spacing in our airbnb config; it's different from object curly spacing because it's not object curlies. Not sure off the top of my head which rule tho.

tleunen:
From my testing, the rule is not applied for the example I posted. I checked all the [jsx rules](https://github.com/yannickcr/eslint-plugin-react#jsx-specific-rules) and couldn't find another rule which check this.

luokuning:
Same issue :-)

Carlows:
Same issue, is there any other rule that works for this?
```jsx
<Comp>
  {something}
</Comp>
```

ljharb:
I think that perhaps an option to `jsx-curly-spacing` that enabled checking of child curlies would be appropriate here?

zimme:
Also this doesn't work.
```
<Button
  active={prevButtonStatus}
  // I expect the following to error because spacing around prevButtonText
  text={<span>{ prevButtonText }</span>}
  onClickHandler={expandButtonHandler}
  className={classNames('btn', 'btn-dots')}
/>
```

Kronuz:
Is there currently no rule to support this?

ljharb:
@Kronuz no. https://github.com/yannickcr/eslint-plugin-react/issues/857#issuecomment-255827739

fatfisz:
I'd like to work on this during the weekend :)

Should this be an expansion of `react/jsx-curly-spacing` or a completely new rule?

ljharb:
[...]

PR #1177, opened 2017-05-07
Enable curly spacing check for children

fatfisz:
This is a feature discussed in #857.

### TODO:
- [x] move all the bugfixes to separate branches
- [x] merge the master after the fixes are merged
- [x] add more tests for attributes and children
- [x] add more tests for config inheritance
- [x] restore the old config behavior alongside the new one
- [x] update the README

### DIFF
--- a/docs/rules/jsx-curly-spacing.md
+++ b/docs/rules/jsx-curly-spacing.md
@@ -1,4 +1,4 @@
-# Enforce or disallow spaces inside of curly braces in JSX attributes. (react/jsx-curly-spacing)
+# Enforce or disallow spaces inside of curly braces in JSX attributes and expressions. (react/jsx-curly-spacing)
 
 While formatting preferences are very personal, a number of style guides require or disallow spaces between curly braces.
 
@@ -6,26 +6,26 @@ While formatting preferences are very personal, a number of style guides require
 
 ## Rule Details
 
-This rule aims to maintain consistency around the spacing inside of JSX attributes.
+This rule aims to maintain consistency around the spacing inside of JSX attributes and expressions inside element children.
 
 It either requires or disallows spaces between those braces and the values inside of them.
 
-### Options
+## Rule Options
 
 There are two main options for the rule:
 
-* `"always"` enforces a space inside of curly braces
-* `"never"` disallows spaces inside of curly braces (default)
+* `{"when": "always"}` enforces a space inside of curly braces
+* `{"when": "never"}` disallows spaces inside of curly braces (default)
 
-Depending on your coding conventions, you can choose either option by specifying it in your configuration:
+There are also two properties that allow specifying how the rule should work with the attributes (`attributes`) and the expressions (`children`). The possible values are:
 
-```json
-"react/jsx-curly-spacing": [2, "always"]
-```
+* `true` enables checking for the spacing using the options (default for `attributes`), e.g. `{"attributes": false}` disables checking the attributes
+* `false` disables checking for the spacing (default for `children`, for backward compatibility), e.g. `{"children": true}` enables checking the expressions
+* an object containing options that override the global options, e.g. `{"when": "always", "children": {"when": "never"}}` enforces a space inside attributes, but disallows spaces inside expressions
 
-#### never
+### never
 
-When `"never"` is set, the following patterns are considered warnings:
+When `{"when": "never"}` is set, the following patterns are considered warnings:
 
 ```jsx
 <Hello name={ firstname } />;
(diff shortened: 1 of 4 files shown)


fatfisz:
I think I'll add a few tests that have both attributes and children set to check that the defaults are being properly set.

lib/rules/jsx-curly-spacing.js
```diff
@@ -30,34 +32,87 @@ module.exports = {
     fixable: 'code',
 
     schema: [{
-      enum: SPACING_VALUES
```
ljharb: this seems like it's removing the old schema, which makes it a breaking change - the idea would be to allow all existing configuration possibilities to continue to work, but provide an alternative that people could migrate to.

Specifically, this also means that (separate from bugfixes) zero existing tests should be modified, and should all continue to pass.

fatfisz: Ok, will do.

fatfisz:
I mistakenly removed the branch -.-

ljharb:
Seems like with #1179 and #1180 merged, this needs a rebase?

fatfisz:
Sure.

ljharb:
@fatfisz please rebase on the command line so there's no merge commits

fatfisz:
Rebased.

Just curious, why do you want to avoid merge commits? It's less time-consuming to solve the conflicts once; also the history is being lost.

fatfisz:
I've added back the old tests (avoiding duplication) and restored the old config behavior. Should I leave any comments about what is from where?

ljharb:
Git logs aren't history, they're change logs - nobody ever needs to know how many times you merged master. Also, merge conflict resolutions in a rebase are preserved for future cherry pickers; conflicts resolved in a merge potentially force developers to re-resolve the same conflicts. It also generally takes precisely the same amount of time to solve a conflict in a merge as it does in a rebase.

lib/rules/jsx-curly-spacing.js
```diff
@@ -32,7 +32,52 @@ module.exports = {
...
+      definitions: {
```
ljharb: i'm not familiar with the "definitions" rule schema key - what version of eslint added this?

tests/lib/rules/jsx-curly-spacing.js
```diff
@@ -27,171 +27,1669 @@ var parserOptions = {
...
+    ].join('\n'),
+    options: [{children: false}]
+  }, {
+    code: '<App>{ bar }</App>;',
+    options: [{children: false}]
+  }, {
+    code: '<App>{ { bar: true, baz: true } }</App>;',
+    options: [{children: false}]
+  }, {
+    code: '<App foo={bar} />;',
```
ljharb: It seems like some of the `<App foo={bar} />` tests have been removed - for any code pattern, we need explicit tests for a) no options, b) all explicit options on the old (current) schema, c) all explicit options on the new schema, d) autofixes

fatfisz:
I was asking because in my experience it takes much longer to do a rebase, especially when there are many commits already. E.g. yesterday merge took < 5 mins, but the rebase took over half an hour. But the argument about cherry-picking speaks to me.

Just for the future reference - would it be ok to squash the commits next time while rebasing?

lib/rules/jsx-curly-spacing.js
fatfisz: I've just read in the ESLint docs that it's using JSON schema - then went over to the JSON schema reference page ([this one is very good](https://spacetelescope.github.io/understanding-json-schema/)) and saw that this technique there.

After a little investigation, it should be supported since the schema validation itself was added: https://github.com/eslint/eslint/commit/fd8fbe721781ce26957ad0f0b3ffbb33201ac387. According to the changelog, the version was 0.22.0, released on 2015-05-30.

The version of is-my-json-valid used then is 2.10.0, which already handled refs: https://github.com/mafintosh/is-my-json-valid/blob/21aa9b8916a2efb3172d17d2d3d1b517668834c7/test/json-schema-draft4/ref.json

Btw. "definitions" is just a convention; the ref can reference any path in the schema.

tests/lib/rules/jsx-curly-spacing.js
fatfisz: I changed a few of them to include children too - they should be at the top. I did not remove any tests, just enhanced a few. The tests without the options should mostly be at the top of the list.

Should I split them to test attributes and children separately?

ljharb:
Commits should ideally be atomic and thus each represent a conceptual change. Naive squashing is fine, but it can cause two changes to be combined where they should be separate.

tests/lib/rules/jsx-curly-spacing.js
ljharb: Yes please :-) when existing test cases are utterly untouched, it adds confidence that a change is not breaking.

fatfisz: Ok, now all tests that are currently on the master branch are also added here.

lib/rules/jsx-curly-spacing.js
ljharb: Wow, awesome. I had no idea.

fatfisz:
Sorry, forgot about updating the docs. Will do that around tomorrow, I'm pumped out after work today.

fatfisz:
Took a bit more time, but here it is.

fatfisz:
Would it be possible for some other collaborators to review this?

fatfisz:
It seems that resolving conflicts through GitHub results in a merge. Will rebase instead.

docs/rules/jsx-curly-spacing.md
```diff
@@ -135,6 +221,41 @@ When `"never"` is used and `objectLiterals` is `"always"`, the following pattern
 
 Please note that spacing of the object literal curly braces themselves is controlled by the built-in [`object-curly-spacing`](http://eslint.org/docs/rules/object-curly-spacing) rule.
 
+### Shorthand options
```
lencioni: Do we see value in supporting these shorthand options indefinitely, or might we want to deprecate them to simplify in a major release? If we plan to deprecate them eventually, it might be worth noting that here so people can avoid using them now.

ljharb: I'd say we might want to deprecate them (string options are horrible for extensibility), but there's not really any pressing need to.

ljharb:
(The appveyor failure appears to be an npm install fluke, so i'm going to merge as-is)

fatfisz:
Thanks a lot!

102:
@ljharb is there any ETA for this to be released?

ljharb:
@102 https://github.com/yannickcr/eslint-plugin-react/issues/1251#issuecomment-307773556

PR #1180, opened 2017-05-07
Fix the newline with object literals bug

fatfisz:
From comment https://github.com/yannickcr/eslint-plugin-react/issues/857#issuecomment-299698993:

> Found another bug:
>
> ```
> <App foo={
> { bar: true, baz: true }
> } />;
> ```
> This would get 2 errors using the default settings: `There should be no space after '{'` and `There should be no space before '}'`.
>
> This is because conditions for the object literal spacing were written so that the second one was never reached:
>
> ```
>           if (sourceCode.isSpaceBetweenTokens(first, second)) {
>             reportNoBeginningSpace(node, first);
>           } else if (!config.allowMultiline && isMultiline(first, second)) {
>             reportNoBeginningNewline(node, first, config);
>           }
> ```
> If the first condition fails, `isMultiline` will return `false` - there is no newline.
>
> Will fix this and add appropriate tests.

### DIFF
--- a/tests/lib/rules/jsx-curly-spacing.js
+++ b/tests/lib/rules/jsx-curly-spacing.js
@@ -29,8 +29,21 @@ ruleTester.run('jsx-curly-spacing', rule, {
   valid: [{
     code: '<App foo={bar} />;'
   }, {
+    code: [
+      '<App foo={',
+      '{ bar: true, baz: true }',
+      '} />;'
+    ].join('\n')
+  }, {
     code: '<App foo={bar} />;',
     options: ['never']
+  }, {
+    code: [
+      '<App foo={',
+      '{ bar: true, baz: true }',
+      '} />;'
+    ].join('\n'),
+    options: ['never', {spacing: {objectLiterals: 'never'}}]
   }, {
     code: '<App foo={ bar } />;',
     options: ['always']
@@ -41,12 +54,19 @@ ruleTester.run('jsx-curly-spacing', rule, {
     code: '<App foo={{ bar:baz }} />;',
     options: ['never']
   }, {
+    code: [
+      '<App foo={',
+      '{ bar: true, baz: true }',
+      '} />;'
+    ].join('\n'),
+    options: ['never']
+  }, {
     code: '<App foo={ {bar:baz} } />;',
     options: ['always']
   }, {
     code: [
       '<App foo={',
-      'bar',
+      '{ bar: true, baz: true }',
       '} />;'
     ].join('\n'),
     options: ['always']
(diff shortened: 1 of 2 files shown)


tests/lib/rules/jsx-curly-spacing.js
```diff
@@ -45,13 +60,21 @@ ruleTester.run('jsx-curly-spacing', rule, {
...
+    options: ['never'],
+    parserOptions: parserOptions
+  }, {
     code: '<App foo={ {bar:baz} } />;',
     options: ['always'],
     parserOptions: parserOptions
   }, {
     code: [
       '<App foo={',
-      'bar',
```
ljharb: can we keep this test case as well?

tests/lib/rules/jsx-curly-spacing.js
```diff
@@ -217,6 +248,62 @@ ruleTester.run('jsx-curly-spacing', rule, {
...
+      '} />;'
+    ].join('\n'),
+    output: '<App foo={ { bar: true, baz: true } } />;',
+    options: ['always', {allowMultiline: false, spacing: {objectLiterals: 'always'}}],
+    errors: [{
+      message: 'There should be no newline after \'{\''
+    }, {
+      message: 'There should be no newline before \'}\''
+    }],
+    parserOptions: parserOptions
```
ljharb: Now that we support node 4+, all these can just be `parserOptions` (only the new examples; we'll change all the rest in their own commit)

tests/lib/rules/jsx-curly-spacing.js
fatfisz: This test is identical with the one just below it, so I suggest it stays removed.

tests/lib/rules/jsx-curly-spacing.js
fatfisz: I'm not sure if having only a part the code conform to the new convention is a good idea, especially since this convention can be applied automatically. How about I make a separate PR which replaces all occurences of `parserOptions: parserOptions` with `parserOptions` everywhere?

ljharb: Sure, that's fine, I suppose.

tests/lib/rules/jsx-curly-spacing.js
fatfisz: (or is there any reason for the duplicated test?)

fatfisz:
I've rebased this. There's only the unresolved problem of the duplicate test left.

tests/lib/rules/jsx-curly-spacing.js
ljharb: good call, i didn't realize it was duplicated

PR #1192, opened 2017-05-12
FIX: wrong syntax in docs for prop-types rule

markus-willems:
This fixes the syntax in the docs for the prop-types rule: "prop-types" (wrong) instead of "react/prop-types" (correct)


markus-willems:
I will go through the docs and report back.

markus-willems:
I went through every file. Found one that needed that fix. A lot of those files are missing the "Rule Options" section. I guess that's wanted/normal?

ljharb:
That's expected when there's no options.

PR #1195, opened 2017-05-15
fix: void-dom-elements-no-children

oliviertassinari:
For some reason, we were using the following pattern that eslint **don't like**:
```js
const props = {}
React.createElement('img', props)
```


yannickcr:
Thanks!

PR #1202, opened 2017-05-17
Allow for stricter no-literals rule

deecewan:
Hey,

I was going to make a separate plugin to do this, but I feel like it can fit quite nicely here.

I've added an option to disallow *all* string literals in components.  We translate all text inside our components, so don't want any strings in the JSX, as they then cannot be translated.

Strings are still allowed inside of function calls inside of `JSXExpressionContainer`s so that translation systems requiring a string key can still be used.


lib/rules/jsx-no-literals.js
```diff
@@ -1,13 +1,16 @@
...
  * @author Caleb Morris
+ * @author David Buchan-Swanson
  */
 'use strict';
 
 // ------------------------------------------------------------------------------
 // Rule Definition
 // ------------------------------------------------------------------------------
 
+var NO_STRINGS = 'no-strings';
```
ljharb: const

lib/rules/jsx-no-literals.js
```diff
@@ -25,27 +30,38 @@ module.exports = {
...
         node: node,
-        message: 'Missing JSX expression container around literal string'
+        message: message
       });
     }
 
+    function getValidation(node) {
+      const standard = !/^[\s]+$/.test(node.value) &&
+          node.parent &&
+          node.parent.type.indexOf('JSX') !== -1 &&
```
ljharb: Will `node.parent` always have a `type` property that's a string or an array?

tests/lib/rules/jsx-no-literals.js
```diff
@@ -202,6 +219,24 @@ ruleTester.run('jsx-no-literals', rule, {
       ].join('\n'),
       parser: 'babel-eslint',
       errors: [{message: 'Missing JSX expression container around literal string'}]
+    }, {
+      code: `
+        <Foo bar="test">
+          {'Test'}
+        </Foo>
+      `,
+      parser: 'babel-eslint',
```
ljharb: Please also add identical test cases that use the default parser.

lib/rules/jsx-no-literals.js
```diff
@@ -17,6 +20,8 @@ module.exports = {
     },
 
     schema: [{
+      enum: [NO_STRINGS]
```
ljharb: Rather than add a string option, let's add this to the options object, for future extensibility.

lib/rules/jsx-no-literals.js
deecewan: This was the existing validation. I just removed part of it if you enable the option.

docs/rules/jsx-no-literals.md
```diff
@@ -1,11 +1,12 @@
...
 
-In JSX when using a literal string you can wrap it in a JSX container `{'TEXT'}`.
-This rules requires that you wrap all literal strings.
-Prevents any odd artifacts of highlighters if your unwrapped string contains an enclosing character like `'` in contractions and enforces consistency.
+There are a couple of scenarios where you want to avoid string literals in JSX.  Either to enforce consistency and reducing strange behaviour, or for enforcing that literals aren't kept in JSX so they can be translated.
 
 ## Rule Details
 
+In JSX when using a literal string you can wrap it in a JSX container `{'TEXT'}`. This rules by default requires that you wrap all literal strings.
+Prevents any odd artifacts of highlighters if your unwrapped string contains an enclosing character like `'` in contractions and enforces consistency.
```
ljharb: ftr tho, this shouldn't ever happen, since straight quotes are typographically incorrect - and this would be a bug in the highlighter. I'm not sure this is worth mentioning.

docs/rules/jsx-no-literals.md
```diff
@@ -18,6 +19,40 @@ The following patterns are not considered warnings:
...
+### Options
+
+There is only one option:
+
+* `noStrings` - Enforces no string literals used as children, wrapped or unwrapped.
+
+To use, you can specify like the following:
+
+```json
+"react/jsx-no-literals": [{"noStrings": false}]
```
ljharb: would this not be `true`?

docs/rules/jsx-no-literals.md
deecewan: this is the original text, essentially, with some modifications because of the different structure. If what you're saying is true, then this rule has/had no purpose.

docs/rules/jsx-no-literals.md
deecewan: uhhhh, because i had a mind blank.

docs/rules/jsx-no-literals.md
ljharb: It still has many purposes; but dealing with quotes isn't one of them imo :-)

Fine to leave the text as-is.

deecewan:
thanks for the feedback. what else would you like for this one?

yannickcr:
Thanks!

PR #1209, opened 2017-05-19
Add rule no static typos

jseminck:
My attempt to add this rule. It seemed pretty basic... so here it is! I am open to new ideas and improvements to the rule. Please let me know!

Issue: https://github.com/yannickcr/eslint-plugin-react/issues/1189

Currently it only checks for`static` class properties. And only for `defaultProps`, `contextTypes` and `propTypes`. It should be trivial to add new properties as well though.

Because it only checks for static properties, it means that having `PropTypes` as a non-static class property or a method definition is still allowed. Although honestly I do not know why anyone would do that? Perhaps the rule can be extended to also warn in case `propTypes` is not declared as a `static` class property?

The rule only works for React Components. Within classes that are not React components, anything is still possible.


docs/rules/no-static-typos.md
```diff
@@ -0,0 +1,59 @@
+# Prevent casing typos when declaring static propTypes, contextType and defaultProps
```
ljharb: Please add ` (react/no-static-typos)` on this line.

docs/rules/no-static-typos.md
```diff
@@ -0,0 +1,59 @@
...
+
+Ensure no casing typos were made declaring static class properties
+
+**Fixable:** This rule is automatically fixable using the `--fix` flag on the command line.
```
ljharb: This shouldn't be fixable - it's not safe to do so.

It's a great rule to prevent people doing this by accident - but autofixing these can cause actual behavior changes in production. eslint core's policy is to never allow autofixes that could break code.

docs/rules/no-static-typos.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Rule Details
+
+This rule checks whether the declared static class properties related to React components
+do not contain any typos. It currently makes sure that the following class properties have
+no casing typos:
+
+* propTypes
+* contextTypes
+* defaultProps
```
ljharb: what about `childContextTypes`?

docs/rules/no-static-typos.md
jseminck: Ok, I will remove it. 

I was actually not aware of this policy. So only stylistic issues should be fixed? Adding removing lines, spaces, semi-colons?

FWIW, The actual help for `--fix` actually says the following: `This option instructs ESLint to try to fix as many issues as possible.`

Considering the words "as many issues as possible" - I thought that every rule which we know how to fix should be auto fixable.

ljharb: It's not that it has to just be stylistic - but generally eslint errs on the side of caution to avoid breaking code.

If this rule warns, it'll only be for two reasons: 1) a typo, which is totally autofixable - and 2) a badly but intentionally chosen name, and the user needs to know to manually rename it.

jseminck:
Tackled your comments. I was not aware of `childContextTypes` - it is not something I have had to use before. If there are any properties that should be added please let me know. 

It's pretty trivial to add them - just have to make sure the test cases are added as well.

Also I wanted to ask: as I was working on this, I thought that it might be worth it to detect small spelling mistakes as well? We could easily detect 1-letter spelling mistakes with Levenshtein distance algorithm.

ljharb:
Detecting spelling mistakes seems like a slippery slope.

jseminck:
The default parser throws an error. I assume that is the case because the class properties proposal is still stage-0 and it is not supported by espree (which is the default parser). 

This is the error that the test also gives: https://astexplorer.net/#/gist/da19b693a7a3d48bcc3fb249180b5295/a5875bd2c21133f64ce9f397ff82adb5e37da26a

Please correct me if I am wrong, would need guidance what exactly needs to be changed. :)

I did remove the `require("babel-eslint");` at the top of the test file. That seems like it was not needed.

ljharb:
Indeed, I'm talking about assigning it outside the `class` - this should apply to the currently-standard `Foo.propTypes = {}` as well as the not-yet-standard class properties form.

jseminck:
I added tests and support for assignment outside the class.

But I cannot get it to work when assigning on a stateless function component: I am not sure how to detect whether the function is a stateless function component or just some random function yet.

```js
      function MyComponent(props) {
        return <div>{props.name}</div>;
      }

      QuestionList.PrOpTyPeS = {
        name: PropTypes.string.isRequired,
      };
```

ljharb:
We have `Components.detect` used elsewhere in the project which should be able to handle that for you.

jseminck:
I spent quite a bit of time on this but no luck so far. I cannot figure out how to properly detect a stateless function component so that the following becomes invalid:

```js
function MyComponent() { return (<div>{this.props.myprop}</div>) };
MyComponent.pRoPTyPeS = { myProp: PropTypes.string }
```

But the following is valid (since it is not a stateless component):

```js
function SomeRandomFunction() { return {} };
MyComponent.pRoPTyPeS = { }
```

I tried using `components.isReturningJSX` but didn't get anywhere.

I added all the test cases, maybe someone else has any idea?

ljharb:
`Components.detect`?

jseminck:
So I finally managed to fix this. I also rebased everything unto the latest master.

I ended up extending `Components.findReturnStatement` to also work when passing in ` FunctionDeclaration` node. Previously, it would not work for this specific node.

This way, I could now use `utils.isReturningJSX` to properly detect a functional stateless component compared to a normal function..

I tried various other approaches from the existing `Components` API but nothing wold make the tests pass as I expected them to.

README.md
```diff
@@ -100,6 +100,7 @@ Finally, enable all of the rules that you would like to use.  Use [our preset](#
 * [react/no-redundant-should-component-update](docs/rules/no-redundant-should-component-update.md): Prevent usage of `shouldComponentUpdate` when extending React.PureComponent
 * [react/no-render-return-value](docs/rules/no-render-return-value.md): Prevent usage of the return value of `React.render`
 * [react/no-set-state](docs/rules/no-set-state.md): Prevent usage of `setState`
+* [react/no-static-typos](docs/rules/no-static-typos.md): Prevent casing typos when declaring static class properties.
```
lencioni: In addition to casing, it would be nice to flag also anything with a small levenshtein distance. That could be done as a separate PR.

And, if you are interested, it would be really cool to have a similar rule for detecting typos in lifecycle methods: https://github.com/yannickcr/eslint-plugin-react/issues/843

I wonder if this rule should have a more general name like `react/no-typos` and make it work also for lifecycle methods?

lib/rules/no-static-typos.js
```diff
@@ -0,0 +1,60 @@
...
+    schema: []
+  },
+
+  create: Components.detect(function(context, components, utils) {
+    function reportErrorIfCasingTypo(node, propertyName) {
+      STATIC_CLASS_PROPERTIES.forEach(function(CLASS_PROP) {
+        if (CLASS_PROP.toLowerCase() === propertyName.toLowerCase() && CLASS_PROP !== propertyName) {
+          context.report({
+            node: node,
+            message: 'Typo in static class property declaration'
```
lencioni: It seems like this might be pretty straightforward to write a fixer for. Would you be interested in adding that?

ljharb: I don't think a fixer is safe here; there's no way to know if it's intentional or not.

README.md
jseminck: I suggested this also earlier in the PR: https://github.com/yannickcr/eslint-plugin-react/pull/1209#issuecomment-303517683

I will look into the similar issue regarding lifecycle methods. Would be useful to have it there as well, as I never remember if it was `prevProps`, `nextProps`, `prevState` or `nextState` (which one and in which order).

lencioni: > as I never remember if it was prevProps, nextProps, prevState or nextState

To be clear, my suggestion is to apply this rule to lifecycle method names, not their arguments. e.g. `componentWillMount`. A rule that enforces naming of lifecycle method arguments would be interesting, but I think that is definitely a different rule than typo detection.

ljharb: Even if we're planning on extending it in the future but not yet, then we should rename the rule to be more generic.

jseminck: Agreed. I will rename to `no-typos` and update the docs accordingly. 

We can then add casing typos for lifecycle methods in another PR. And we can look into detecting small Levenshtein distances in a separate PR as well. 

Sounds good?

ljharb: Sounds great!

jseminck:
Could you please elaborate a bit more what you mean exactly?

I added a test case for this, as this was currently causing a crash:

```
      class First extends React.Component {}
      First["prop" + "Types"] = {};'
      First["context" + "Types"] = {};'
      First["childContext" + "Types"] = {};'
      First["default" + "Props"] = {};
```

And yeah - this is not supported. E.g. `First["deFauLT" + "prOps"] = {}` will not error.

ljharb:
I expect, from the implementation, that computed property names are ignored. I'm asking for tests to ensure they *are* ignored - to prevent them accidentally being considered in the future, and to ensure that considering them is a breaking change.

jseminck:
I understood why we want the test cases. I just wasn't 100% sure that I understood correctly what you meant with "computed static property names".  

I understood this was used for:
```js
const myObj = {
  ["myKey-" + getIndex()]: "myValue"
}
```

But I don't see how this syntax can be used to declare these properties.  The tests I added are basically what this would transpile to:

```js
const myObj = {}
myObj["myKey-" + getIndex()]: "myValue
```

I hope that's all that is needed to finalize this, as I'd like to start focussing on the next steps. :)

Issue #1206, opened 2017-05-18
Rule proposal: indentation of closing jsx tag 


rsolomon:
Not to be confused with the closing bracket, I would like to enforce the location of a closing jsx tag when its children also live on a new line.

For example, the following would fail:

```
<MyComponent
  prop1="foo"
  prop2="bar">
  {content}</MyComponent>
```

The following would pass:

```
<MyComponent
  prop1="foo"
  prop2="bar">
  {content}
</MyComponent>
```

This would also pass:

```
<MyComponent prop1="foo">{content}</MyComponent>
```

ljharb:
This sounds great - essentially, "must be on the same line, or must be on its own line that matches the indentation of the opening tag"?

`jsx-closing-tag-location`, perhaps.

ljharb:
(Done in #1215)

PR #1215, opened 2017-05-20
Add indentation rule for closing tag of multi-line jsx

rsolomon:
For issue #1206:

This rule validates that the closing tag for a multi-line jsx component matches the indentation of the opening tag.

### DIFF
--- a/tests/lib/rules/jsx-closing-tag-location.js
+++ b/tests/lib/rules/jsx-closing-tag-location.js
@@ -13,7 +13,6 @@ const RuleTester = require('eslint').RuleTester;
 const parserOptions = {
   sourceType: 'module',
   ecmaFeatures: {
-    experimentalObjectRestSpread: true,
     jsx: true
   }
 };
--- a/tests/lib/rules/jsx-closing-tag-location.js
+++ b/tests/lib/rules/jsx-closing-tag-location.js
@@ -9,9 +9,8 @@
 // ------------------------------------------------------------------------------
 
 const rule = require('../../../lib/rules/jsx-closing-tag-location');
-const {RuleTester} = require('eslint');
+const RuleTester = require('eslint').RuleTester;
 const parserOptions = {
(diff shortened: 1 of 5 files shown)


tests/lib/rules/jsx-closing-tag-location.js
```diff
@@ -0,0 +1,56 @@
...
+'use strict';
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+const rule = require('../../../lib/rules/jsx-closing-tag-location');
+const {RuleTester} = require('eslint');
+const parserOptions = {
+  ecmaVersion: 8,
```
ljharb: These tests don't need an ecmaVersion specified, nor rest/spread enabled

rsolomon: Oops! Will fix.

rsolomon:
@ljharb I think I've addressed your concerns. The one situation this doesn't fix is the following:

```
<App
  foo="bar">baz</App>
```

It will fix it to:

```
<App
  foo="bar">baz
</App>
```

When ideally it would be fixed to:

```
<App
  foo="bar">
  baz
</App>
```

I opted not to handle it in this rule because I imagine multiline children location should be handled in a separate rule entirely (maybe it already is?)

ljharb:
I agree with you that it should be handled in a separate rule.

rsolomon:
Pardon my ignorance (this is the first eslint rule I've written), but isn't that what the `output` attributes of the `invalid` tests are for?

HHogg:
@rsolomon Would this rule also cover the situation below? 

```jsx
// Considered a warning 
<App>
  { condition && <Component>
    ...
  </Component> }
</App>

// Not considered a warning 
<App>
  { condition && (
    <Component>
      ...
    </Component>    
  ) }
</App>
```

PR #1216, opened 2017-05-22
Don't issue errors for `<a/>` with target=_blank and non-external URLs

gfx:
`<a href="relative/path" target="_blank"/>`and `<a href="/absolute/path" target="_blank"/>` are considered safe because the link is placed in the same host, so the rule only issues for URLs with protocols.

### DIFF
--- a/lib/rules/jsx-no-target-blank.js
+++ b/lib/rules/jsx-no-target-blank.js
@@ -33,8 +33,21 @@ module.exports = {
           var relFound = false;
           var attrs = node.parent.attributes;
           for (var idx in attrs) {
-            if (attrs[idx].name && attrs[idx].name.name === 'rel') {
-              var tags = attrs[idx].value.type === 'Literal' && attrs[idx].value.value.toLowerCase().split(' ');
+            if (!Object.prototype.hasOwnProperty.call(attrs, idx)) {
+              continue;
+            }
+            var attr = attrs[idx];
+            if (!attr.name) {
+              continue;
+            }
+            if (attr.name.name === 'href') {
+              if (attr.value.type === 'Literal' && !/^\w+:/.test(attr.value.value)) {
+                // it's safe because it is not an external link (i.e. doesn't start with a protocol)
+                return;
+              }
+            }
+            if (attr.name.name === 'rel') {
+              var tags = attr.value.type === 'Literal' && attr.value.value.toLowerCase().split(' ');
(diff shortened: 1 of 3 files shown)


lib/rules/jsx-no-target-blank.js
```diff
@@ -33,8 +33,21 @@ module.exports = {
           var relFound = false;
           var attrs = node.parent.attributes;
           for (var idx in attrs) {
-            if (attrs[idx].name && attrs[idx].name.name === 'rel') {
-              var tags = attrs[idx].value.type === 'Literal' && attrs[idx].value.value.toLowerCase().split(' ');
+            if (!Object.prototype.hasOwnProperty.call(attrs, idx)) {
```
gfx: the guard is for [guard-for-in](http://eslint.org/docs/rules/guard-for-in).

ljharb: Since we have node 4+ now, let's just use `Object.keys().forEach` - then no guard is needed.

gfx: in `forEach()`, only there's only`return` available to control loops, but this loop needs `return`, `continue` and `breaks`, so `forEach()` is not suitable.

ljharb: In that case I'd prefer to rewrite it to use `.filter` in lieu of the continues, and `.find` in lieu of the breaks and returns.

However, to keep this PR small, we can stick with this guard (which should have been here in the first place).

lib/rules/jsx-no-target-blank.js
```diff
@@ -33,8 +33,21 @@ module.exports = {
...
+              continue;
+            }
+            var attr = attrs[idx];
+            if (!attr.name) {
+              continue;
+            }
+            if (attr.name.name === 'href') {
+              if (attr.value.type === 'Literal' && !/^\w+:/.test(attr.value.value)) {
+                // it's safe because it is not an external link (i.e. doesn't start with a protocol)
```
ljharb: `//google.com/foo` is an external link - please add a test case ensuring that [protocol-less URLs](https://www.paulirish.com/2010/the-protocol-relative-url/) still treated as external.

gfx: Right!

Fixed 664e5df

lib/rules/jsx-no-target-blank.js
gfx: IMO the conditions are too complex so re-writing them with ES2015 features is useful.

Refactored in 89c33b6

PR #1226, opened 2017-05-27
Fix void-dom-elements-no-children crash

kokobeware:
`React.createElement()` currently crashes the linter with the following error and linting doesn't happen:

image

```
events.js:106 Cannot read property 'value' of undefined TypeError: Cannot read property 'value' of undefined
```

Of course ``React.createElement()`` will cause a runtime error, but it shouldn't crash the linter.


tests/lib/rules/void-dom-elements-no-children.js
```diff
@@ -54,6 +54,10 @@ ruleTester.run('void-dom-elements-no-children', rule, {
     }, {
       code: 'React.createElement("img");'
     }, {
+      code: 'React.createElement();'
```
ljharb: I feel like this shouldn't be valid; altho maybe this rule shouldn't be handling it.

kokobeware: Yup, a different rule that checks for required parameters to React methods would be good. Maybe it could even give a hint on the order and nature of the parameters without needing to go to the docs.

yannickcr:
Thanks!

PR #1229, opened 2017-05-30
Add fix for prefer-stateless-function

### LINKED ISSUE #1096: Add fix to prefer-stateless-function
Would be awesome to be able to autofix this. Anyone wanna do it?! :) [as seen on twitter](https://twitter.com/ljharb/status/836654700967579648)

RiddleMan:
Fixes #1096.
I've managed to accomplish this fix! Yay!
Also, I had tested this on my codebase and worked nicely.

I've added one module for indentation detection because I'd like to keep original formatting of the elements.
Unfortunately, This solution works only for ES6 components. This code does not handle ES5 ones, but I created a possibility to add later fix for those too.

I'm open for the suggestions :)


lib/rules/prefer-stateless-function/index.js
```diff
@@ -221,8 +223,16 @@ module.exports = {
           property.kind === 'constructor' &&
           isRedundantSuperCall(property.value.body.body, property.value.params)
         ;
+        var isStatic = property.static;
         var isRender = name === 'render';
-        return !isDisplayName && !isPropTypes && !contextTypes && !isUselessConstructor && !isRender;
+        return !(
```
ljharb: i kind of feel like the prior form was more readable; a bunch of "nots" seems easier to understand to me than a bunch of positives, negated

ljharb:
(This is awesome, btw, thanks!)

lib/rules/prefer-stateless-function/fixer.js
```diff
@@ -0,0 +1,132 @@
+'use strict';
+
+var detectIndent = require('detect-indent');
```
ljharb: all these `var`s should probably be let/const

docs/rules/prefer-stateless-function.md
```diff
@@ -2,6 +2,8 @@
 
 Stateless functional components are simpler than class based components and will benefit from future React performance optimizations specific to these components.
 
+**Fixable:** This rule is automatically fixable using the `--fix` flag on the command line.
```
RiddleMan: Yes, you are right. We should postpone this line until the rule is fully fixable.

lib/rules/prefer-stateless-function/index.js
RiddleMan: Agree ;) I've saved couple characters, but the readability of previous one is much better.

tests/lib/rules/prefer-stateless-function.js
```diff
@@ -513,6 +513,289 @@ ruleTester.run('prefer-stateless-function', rule, {
...
+        message: 'Component should be written as a pure function'
+      }]
+    }, {
+      // should work without render function for default parser
+      code: [
+        'class Foo extends Component {',
+        '}'
+      ].join('\n'),
+      output: [
+        'let Foo = (props) => {};'
```
ljharb: Why does the fixer output an arrow function for SFCs?

This will run afoul of, for example, the airbnb style guide, which only allows named functions as SFCs. If we instead defaulted to a normal named function here, other rules would be able to convert it to an SFC.

Also, converting from a `class` declaration to an expression isn't always a safe change; the only safe replacement for `class` is a `function` declaration (which will be an expression when in expression position).

lib/rules/prefer-stateless-function/fixer.js
RiddleMan: Yeah, I've made it before you've dropped support for node 0.10. Agree.

tests/lib/rules/prefer-stateless-function.js
RiddleMan: Yep!

lib/rules/prefer-stateless-function/fixer.js
```diff
@@ -90,7 +86,7 @@ function ruleFixer(sourceCode, componentNode, utils) {
    * @returns {string} stateless component body
    */
   function getComponentText(name, transformedBody) {
-    return `let ${name} = (props) => ${transformedBody};`;
+    return `function ${name}(props) ${transformedBody};`;
```
ljharb: This probably shouldn't include the semicolon either, unless the original text did

RiddleMan: Good catch! I have to take a deep breath before commit ;)

RiddleMan:
@ljharb Yes, I think it's doable by me. Let me look into this in the next week.
Do you think, that we should merge what we have here and then add support for createClass?
Or should I add fixer for that in a scope of this PR?

For me both options are ok.

ljharb:
Since this needs to wait for other collabs to review before merging, I'd say we'll play it by ear.

tests/lib/rules/prefer-stateless-function.js
```diff
@@ -513,6 +513,289 @@ ruleTester.run('prefer-stateless-function', rule, {
...
+        '  }',
+        '}'
+      ].join('\n'),
+      output: [
+        'function Foo(props) {',
+        '  const {foo, bar} = props;',
+        '  ',
+        '  return <div>{props.test}</div>;',
+        '}',
+        'Foo.propTypes={foo: PropTypes.func};',
```
jseminck: I would like to see an empty line between the closing `}` and the `Foo.propTypes` assignment. It's a personal taste though. What do others think?

jseminck:
I ran this on my code-base which only uses class syntax, no stateless function components at all, and it seems to work pretty well. I found one issue where the `prefer-stateless-function` rule doesn't detect well, I will create a ticket for that.

I added a tiny comment about a syntactical preference, but it's very personal: https://github.com/yannickcr/eslint-plugin-react/pull/1229/files#r119805697

tests/lib/rules/prefer-stateless-function.js
ljharb: That seems like something another rule should be handling.

deecewan:
This doesn't seem to handle the case of `context`.

For instance,

```jsx
    render() {
        <div>{this.context.thing}</div>
    }
}
```

could be re-written as

```jsx
function MyComponent(props, context) {
  return (
      <div>{context.thing}</div>
  );
}
```

I'd like to see a test case that uses `context`, even if it doesn't support fixing context.

ljharb:
@RiddleMan if you're able to rebase this, it'd be great to land it :-)

golopot:
I rebased this. https://github.com/golopot/eslint-plugin-react/tree/pr-1229 . (Commits are squashed and the three files are merged to one.)

ljharb:
@RiddleMan would you mind checking the "allow edits" box on the RHS of the PR?

RiddleMan:
@ljharb I'm worried that I don't have such checkbox on the RHS. The explanation is simple... accidentally I removed my fork :/ Can we just create a new one with golopot's changes? Or anything you find fit.

ljharb:
@RiddleMan oof, that's unfortunate. you could try recreating the fork and see what happens?

ljharb:
@golopot i suppose we'll need to make a new PR. Mind creating one from your branch?

ljharb:
ping @golopot

ljharb:
I've rebased this and added the correct `output` results here: https://github.com/ljharb/eslint-plugin-react/tree/pr-1229. Test results are here: https://travis-ci.com/github/ljharb/eslint-plugin-react/builds/190413744.

Help is welcome.

ljharb:
It'd be great if someone wants to file a new PR that solves this; but unfortunately this one is unrecoverable.

PR #1232, opened 2017-05-31
Support lifecycle methods with nextProps/prevProps in no-unused-prop-types

jseminck:
https://github.com/yannickcr/eslint-plugin-react/issues/1213 https://github.com/yannickcr/eslint-plugin-react/issues/1142#issuecomment-304737468

This makes the no-unused-prop-types work when a prop is used in React lifecycle components other than `componentWillReceiveProps` using the "dot-operator". E.g:

```
   if (nextProps.foo) { // This will now be detected
       ...
  }
```

There is one thing to keep in mind: under the current implementation the rule will only work if the parameter is named `nextProps`, `prevProps` or `props`. This logic already existed and I did not want to touch it, though I thought it was kind of odd. Do we want to keep it that way? 

Perhaps we could introduce a rule that validates the parameter names in lifecycle methods - sometimes it's confusing anyway whether `shouldComponentUpdate` has `prevProps` or `nextProps` as the parameter, and what is the order of `nextProps` or `nextState`. Anyway, this would be for another issue.

### DIFF
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -16,8 +16,10 @@ var annotations = require('../util/annotations');
 // Constants
 // ------------------------------------------------------------------------------
 
-var DIRECT_PROPS_REGEX = /^props\s*(\.|\[)/;
-var DIRECT_NEXT_PROPS_REGEX = /^nextProps\s*(\.|\[)/;
+const DIRECT_PROPS_REGEX = /^props\s*(\.|\[)/;
+const DIRECT_NEXT_PROPS_REGEX = /^nextProps\s*(\.|\[)/;
+const DIRECT_PREV_PROPS_REGEX = /^prevProps\s*(\.|\[)/;
+const LIFE_CYCLE_METHODS = ['componentWillReceiveProps', 'shouldComponentUpdate', 'componentWillUpdate', 'componentDidUpdate'];
 
 // ------------------------------------------------------------------------------
 // Rule Definition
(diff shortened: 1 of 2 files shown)


lib/rules/no-unused-prop-types.js
```diff
@@ -78,15 +80,16 @@ module.exports = {
...
      **/
-    function inComponentWillReceiveProps() {
+    function inLifeCycleMethod() {
       var scope = context.getScope();
       while (scope) {
         if (
           scope.block && scope.block.parent &&
-          scope.block.parent.key && scope.block.parent.key.name === 'componentWillReceiveProps'
+          scope.block.parent.key &&
+            LIFE_CYCLE_METHODS.includes(scope.block.parent.key.name)
```
ljharb: arrays don't have `.includes` until node 6; we need to use `indexOf` or https://npmjs.com/array-includes here.

jseminck:
Used `.indexOf()`. Thanks for that, I did not notice that or even think about it.

I also fixed the invalid test cases, as the lines and column values were incorrect. Hopefully CI should be green now. Sorry for the problems!

https://npmjs.com/array-includes would be nice to use, actually. There's a lot of places that use `indexOf()` and they all use it differently: `!== -1`, `> -1`, `>= 0`. However I am not comfortable making the decision of adding a new dependency.

ljharb:
Adding a new dep is fine imo, but it should definitely be in a separate PR :-)

jseminck:
Is there any change needed in the docs for this? It's for example important that the parameter names are correct (only works with `nextProps`, `props` and `prevProps`)

ljharb:
Sure, it's worth making an additional PR for docs.

PR #1236, opened 2017-06-02
Register prop-types when set through Flow SuperTypeParameters

gpeal:
This picks up prop types when set using flow [SuperTypeParameters ](https://flow.org/en/docs/frameworks/react/) such as:
```
type Props = { /* ... */ };

class MyComponent extends React.Component<void, Props, void> {
  ...
}
```

@ljharb This is modified from the property annotation checks. However, I couldn't test the wrapped `TypeAnnotation` or `NullableTypeAnnotation` resolution checks.

### DIFF
--- a/lib/rules/prop-types.js
+++ b/lib/rules/prop-types.js
@@ -151,6 +151,21 @@ module.exports = {
       return false;
     }
 
+/**
+ * Checks if we are declaring a props as a generic type in a flow-annotated class.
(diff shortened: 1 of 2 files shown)


ljharb:
I'm going to merge this today if there's no objections.

ljharb:
@yannickcr any chance of a release soon, to get this fix in?

Issue #969, opened 2016-11-22
jsx-no-duplicate-props: TypeError: name.toLowerCase is not a function


sompylasar:
I use `eslint-plugin-react@6.3.0`

The latest `master` as of today contains the same code that fails: https://github.com/yannickcr/eslint-plugin-react/blob/21f684a738af7a6f2b1ec9517b5040162a750789/lib/rules/jsx-no-duplicate-props.js#L48
```
TypeError: name.toLowerCase is not a function
    at /__CENSORED__/node_modules/eslint-plugin-react/lib/rules/jsx-no-duplicate-props.js:48:25
    at Array.forEach (native)
    at EventEmitter.JSXOpeningElement (/__CENSORED__/node_modules/eslint-plugin-react/lib/rules/jsx-no-duplicate-props.js:40:25)
    at emitOne (events.js:101:20)
    at EventEmitter.emit (events.js:188:7)
    at NodeEventGenerator.enterNode (/__CENSORED__/node_modules/eslint/lib/util/node-event-generator.js:40:22)
    at CodePathAnalyzer.enterNode (/__CENSORED__/node_modules/eslint/lib/code-path-analysis/code-path-analyzer.js:608:23)
    at CommentEventGenerator.enterNode (/__CENSORED__/node_modules/eslint/lib/util/comment-event-generator.js:97:23)
    at Controller.enter (/__CENSORED__/node_modules/eslint/lib/eslint.js:898:36)
    at Controller.__execute (/__CENSORED__/node_modules/estraverse/estraverse.js:397:31)
```

The issue happens when I (accidentally) put a colon after a prop name, for example:
```js
      <SomeComponent
        minValue={5000}
        maxValue={750000}
        someProp:   //<<<< The colon here triggers the error.
        value={this.state.value}
        onChange={this._handleChange}
      />
```

ljharb:
That's invalid jsx syntax - I don't think it's a reasonable expectation that rules should work when your syntax is invalid.

In other words, even if you disabled that rule, eslint itself should be dying on the syntax.

sompylasar:
Right, but it should not throw, i.e. it should just somehow ignore the input if the rule does not apply.

ljharb:
That's a fair point; that the rule should be silent when there's a parse error. Although, that seems like something eslint itself should be handling perhaps?

sompylasar:
[...]

PR #1237, opened 2017-06-04
#969 fix: exception for jsx-no-duplicate-props

marcelmokos:
jsx-no-duplicate-props is causing error. 
TypeError: name.toLowerCase is not a function. 
When <Element {...props}> is used.


ljharb:

I'll merge within the next day or so, if no objections, after that's done.

PR #1239, opened 2017-06-06
Update README to include default-props-match-prop-types rule

webOS101:
Added bonus typo fix.


webOS101:
Noticed that there are still instructions for eslint 1 in the README.  Is that even still supported?

PR #1250, opened 2017-06-11
[New] `jsx-boolean-value`: add inverse option for always/never

### LINKED ISSUE #1249: Extend "jsx-boolean-value" to support optional "ignore" list
It would be great if we could configure `jsx-boolean-value` to ignore a list of props.

### Example
```js
{
  'react/jsx-boolean-value': ['error', 'never', { ignore: ['value']}]
}
```

This allows the following expression to be valid

```jsx
<Toggle value={true} />
```

And if you wonder why I'm not using a prop like `toggled` instead of `value`, it's because we want to have a set of components that support the same prop.

I'm more than happy to support the idea with a PR.

ljharb:
Fixes #1249

### DIFF
--- a/docs/rules/jsx-boolean-value.md
+++ b/docs/rules/jsx-boolean-value.md
@@ -6,27 +6,29 @@
 
 ## Rule Details
 
-This rule takes one argument. If it is `"always"` then it warns whenever an attribute is missing its value. If `"never"` then it warns if an attribute has a `true` value. The default value of this option is `"never"`.
+This rule takes two arguments. If the first argument is `"always"` then it warns whenever an attribute is missing its value. If `"never"` then it warns if an attribute has a `true` value. The default value of this option is `"never"`.
 
-The following patterns are considered warnings when configured `"never"`:
+The second argument is optional: if provided, it must be an object with a `"never"` property (if the first argument is `"always"`), or an `"always"` property (if the first argument is `"never"`). This property’s value must be an array of strings representing prop names.
+
+The following patterns are considered warnings when configured `"never"`, or with `"always", { "never": ["personal"] }`:
 
 ```jsx
 var Hello = <Hello personal={true} />;
 ```
 
-The following patterns are not considered warnings when configured `"never"`:
+The following patterns are not considered warnings when configured `"never"`, or with `"always", { "never": ["personal"] }`:
 
 ```jsx
 var Hello = <Hello personal />;
 ```
 
-The following patterns are considered warnings when configured `"always"`:
+The following patterns are considered warnings when configured `"always"`, or with `"never", { "always": ["personal"] }`:
 
 ```jsx
 var Hello = <Hello personal />;
 ```
 
-The following patterns are not considered warnings when configured `"always"`:
+The following patterns are not considered warnings when configured `"always"`, or with `"never", { "always": ["personal"] }`:
 
 ```jsx
 var Hello = <Hello personal={true} />;
--- a/lib/rules/jsx-boolean-value.js
+++ b/lib/rules/jsx-boolean-value.js
@@ -17,6 +17,16 @@ const exceptionsSchema = {
 const ALWAYS = 'always';
 const NEVER = 'never';
 
+const errorData = new WeakMap();
+function getErrorData(exceptions) {
+  if (!errorData.has(exceptions)) {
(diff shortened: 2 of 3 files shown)


ljharb:
ping for reviews :-)

tests/lib/rules/jsx-boolean-value.js
```diff
@@ -28,13 +28,18 @@ var ruleTester = new RuleTester({parserOptions});
 ruleTester.run('jsx-boolean-value', rule, {
   valid: [
     {code: '<App foo />;', options: ['never']},
+    {code: '<App foo />;', options: ['always', {never: ['foo']}]},
```
jseminck: Should we also cover the `always` part of these test cases? E.g. checking that `bar={true}` is still valid when the config is `['always', {never: ['foo']}]`.

ljharb: sure, why not

ljharb:
@yannickcr @lencioni @EvNaverniouk if you have a chance to review this that'd be appreciated :-)

lib/rules/jsx-boolean-value.js
```diff
@@ -17,44 +26,78 @@ module.exports = {
...
+    const configObject = context.options[1] || {};
+    const exceptions = new Set((configuration === ALWAYS ? configObject[NEVER] : configObject[ALWAYS]) || []);
+    const exceptionProps = Array.from(exceptions, (name) => `\`${name}\``).join(', ');
+    const exceptionsMessage = exceptions.size > 0 ? ` for the following props: ${exceptionProps}` : '';
+    const data = {exceptionsMessage: exceptionsMessage};
 
-    var NEVER_MESSAGE = 'Value must be omitted for boolean attributes';
-    var ALWAYS_MESSAGE = 'Value must be set for boolean attributes';
+    const NEVER_MESSAGE = 'Value must be omitted for boolean attributes{{exceptionsMessage}}';
+    const ALWAYS_MESSAGE = 'Value must be set for boolean attributes{{exceptionsMessage}}';
```
lencioni: It might be nice to move some of this into a helper function that takes `exceptions` as an argument, so this code only has to run in the event of a warning.

lib/rules/jsx-boolean-value.js
```diff
@@ -17,44 +26,78 @@ module.exports = {
...
-                  return fixer.insertTextAfter(node, '={true}');
-                }
-              });
+      JSXAttribute(node) {
+        const propName = node.name && node.name.name;
+        const value = node.value;
+        const isException = exceptions.has(propName);
+
+        const isAlways = configuration === ALWAYS ? !isException : isException;
+        const isNever = configuration === NEVER ? !isException : isException;
```
lencioni: I find these two ternaries less clear than if/else. I had to work through the logic here, which slowed me down. It might be worth converting.

docs/rules/jsx-boolean-value.md
```diff
@@ -6,27 +6,29 @@
 
 ## Rule Details
 
-This rule takes one argument. If it is `"always"` then it warns whenever an attribute is missing its value. If `"never"` then it warns if an attribute has a `true` value. The default value of this option is `"never"`.
+This rule takes two arguments. If the first argument is `"always"` then it warns whenever an attribute is missing its value. If `"never"` then it warns if an attribute has a `true` value. The default value of this option is `"never"`.
 
-The following patterns are considered warnings when configured `"never"`:
+The second argument is optional: if provided, it must be an object with a `"never"` property (if the first argument is `"always"`), or an `"always"` property (if the first argument is `"never"`). This property’s value must be an array of strings representing prop names.
+
+The following patterns are considered warnings when configured `"never"`, or with `"always", { "never": ["personal"] }`:
```
lencioni: I'm on the fence about whether this should be `never`/`always`, or something more general like `exceptions` or `except`. As it is written here, it is possible to have both `never` and `always` which would be a little weird. I think it is fine, but I think a unified option might make the documentation a little clearer and the code a small amount simpler. What do you think?

ljharb: To me, "exceptions" means "these things are not checked" - the semantic here is, everything is forced to something: some things are forced to always, and some to never.

lencioni: Ah I see. That makes sense.

ljharb:
@lencioni updated per your comments; i'll merge tomorrow if there's no other issues.

PR #1253, opened 2017-06-12
Add support for propTypes wrapped in a function

dustinsoftware:
PropTypes would previously get ignored if wrapped in forbidExtraProps.

### DIFF
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -44,6 +44,13 @@ module.exports = {
         },
         skipShapeProps: {
           type: 'boolean'
+        },
+        propWrapperFunctions: {
+          type: 'array',
+          items: {
+            type: 'string'
+          },
+          uniqueItems: true
         }
       },
       additionalProperties: false
@@ -56,6 +63,8 @@ module.exports = {
     var configuration = Object.assign({}, defaults, context.options[0] || {});
     var skipShapeProps = configuration.skipShapeProps;
     var customValidators = configuration.customValidators || [];
+    var propWrapperFunctions = new Set(configuration.propWrapperFunctions || []);
+
     // Used to track the type annotations in scope.
     // Necessary because babel's scopes do not track type annotations.
     var stack = null;
@@ -721,6 +730,15 @@ module.exports = {
             return;
           }
           ignorePropsValidation = true;
+          break;
+        case 'CallExpression':
+          if (
+            propWrapperFunctions.has(propTypes.callee.name) &&
+            propTypes.arguments && propTypes.arguments[0]
+          ) {
+            markPropTypesAsDeclared(node, propTypes.arguments[0]);
+            return;
+          }
           break;
         case null:
           break;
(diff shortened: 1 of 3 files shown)


dustinsoftware:
Is this what you had in mind?

lib/rules/no-unused-prop-types.js
```diff
@@ -44,6 +44,9 @@ module.exports = {
         },
         skipShapeProps: {
           type: 'boolean'
+        },
+        extraPropsFunc: {
```
ljharb: i think this would be better as an array of unique strings - that way multiples could be provided.

Let's also rename it to "propWrapperFunctions"

dustinsoftware:
Ok, I think it's ready for another pass :)

lib/rules/no-unused-prop-types.js
```diff
@@ -45,8 +45,11 @@ module.exports = {
         skipShapeProps: {
           type: 'boolean'
         },
-        extraPropsFunc: {
-          type: 'string'
+        propWrapperFunctions: {
+          type: 'array',
+          items: {
```
ljharb: please add `uniqueItems: true` so that jsonschema enforces uniqueness

dustinsoftware:
👍 rebased. Thanks for the quick feedback.

lib/rules/no-unused-prop-types.js
```diff
@@ -722,6 +729,16 @@ module.exports = {
           }
           ignorePropsValidation = true;
           break;
+        case 'CallExpression':
+          if (
+            configuration.propWrapperFunctions &&
+            configuration.propWrapperFunctions.indexOf(propTypes.callee.name) !== -1 &&
```
lencioni: It might be worth converting the propWrapperFunctions to a structure with constant time lookup, like a Set, to keep this path fast.

ljharb: that's a really great point. this should definitely be converted to a Set.

dustinsoftware: Ok, updated

lib/rules/no-unused-prop-types.js
```diff
@@ -56,6 +63,9 @@ module.exports = {
     var configuration = Object.assign({}, defaults, context.options[0] || {});
     var skipShapeProps = configuration.skipShapeProps;
     var customValidators = configuration.customValidators || [];
+    var propWrapperFunctions = configuration.propWrapperFunctions ?
+      new Set(configuration.propWrapperFunctions) : new Set();
```
ljharb: `var propWrapperFunctions = new Set(configuration.propWrapperFunctions || []);`

dustinsoftware: Dargh, thanks...should have caught that one

ljharb:
@dustinsoftware are there other rules that could benefit from this information?

If so, maybe we should move it to "settings", and read from it in multiple rules?

mqklin:
Has it been published?

ljharb:
Nope, not yet.

mqklin:
@ljharb still, doesn't work for me :(
image
repro: https://github.com/mqklin/eslint-proptypes-bug

ljharb:
@mqklin because you haven't explicitly defined it as a propWrapperFunction in your eslintrc.

Please file a new issue if you still have questions instead of commenting on a merged PR.

mqklin:
@ljharb thank you very much! I missed this.

PR #1260, opened 2017-06-14
[New] `jsx-max-depth`: validate a specific depth for JSX

### LINKED ISSUE #1219: Check for jsx's depth
From #1217 

We want to restrict jsx to a depth of 3 or 4, this is something like restricting nested block in common js code

This should be off by default, for those who want to restrict their depth, use `'jsx-max-depth': [2, 4]` to enable it

chriswong:
Fixes #1219

### DIFF
--- a/docs/rules/jsx-max-depth.md
+++ b/docs/rules/jsx-max-depth.md
@@ -1,0 +1,84 @@
+# Validate JSX maximum depth (react/jsx-max-depth)
+
+This option validates a specific depth for JSX.
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```jsx
+<App>
+  <Foo>
+    <Bar>
+      <Baz />
+    </Bar>
+  </Foo>
+</App>
+
+```
+
+## Rule Options
+
+It takes an option as the second parameter which can be a positive number for depth count.
+
+```js
+...
+"react/jsx-no-depth": [<enabled>, { "max": <number> }]
+...
+```
+
+The following patterns are considered warnings:
+
+```jsx
+// [2, { "max": 1 }]
+<App>
+  <Foo>
+    <Bar />
+  </Foo>
+</App>
+
+// [2, { "max": 1 }]
+const foobar = <Foo><Bar /></Foo>;
+<App>
+  {foobar}
+</App>
+
+// [2, { "max": 2 }]
+<App>
+  <Foo>
+    <Bar>
+      <Baz />
+    </Bar>
+  </Foo>
+</App>
+```
+
+The following patterns are not warnings:
+
+```jsx
+
+// [2, { "max": 1 }]
+<App>
+  <Hello />
+</App>
+
+// [2,{ "max": 2 }]
+<App>
+  <Foo>
+    <Bar />
+  </Foo>
+</App>
+
+// [2, { "max": 3 }]
+<App>
+  <Foo>
+    <Bar>
+      <Baz />
+    </Bar>
+  </Foo>
+</App>
+```
+
+## When not to use
+
+If you are not using JSX then you can disable this rule.
--- a/lib/rules/jsx-max-depth.js
+++ b/lib/rules/jsx-max-depth.js
@@ -1,0 +1,146 @@
+/**
+ * @fileoverview Validate JSX maximum depth
+ * @author Chris<wfsr@foxmail.com>
+ */
+'use strict';
+
+const has = require('has');
+const variableUtil = require('../util/variable');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Validate JSX maximum depth',
+      category: 'Stylistic Issues',
+      recommended: false
+    },
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          max: {
+            type: 'integer',
+            minimum: 0
+          }
+        },
+        additionalProperties: false
+      }
+    ]
+  },
+  create: function(context) {
+    const MESSAGE = 'Expected the depth of nested jsx elements to be <= {{needed}}, but found {{found}}.';
+    const DEFAULT_DEPTH = 2;
+
+    const option = context.options[0] || {};
+    const maxDepth = has(option, 'max') ? option.max : DEFAULT_DEPTH;
+
+    function isJSXElement(node) {
+      return node.type === 'JSXElement';
+    }
+
+    function isExpression(node) {
+      return node.type === 'JSXExpressionContainer';
+    }
+
+    function hasJSX(node) {
+      return isJSXElement(node) || isExpression(node) && isJSXElement(node.expression);
+    }
+
+    function isLeaf(node) {
+      const children = node.children;
+
+      return !children.length || !children.some(hasJSX);
+    }
+
+    function getDepth(node) {
+      let count = 0;
+
+      while (isJSXElement(node.parent) || isExpression(node.parent)) {
+        node = node.parent;
+        if (isJSXElement(node)) {
+          count++;
+        }
+      }
+
+      return count;
+    }
+
+
+    function report(node, depth) {
+      context.report({
+        node: node,
+        message: MESSAGE,
+        data: {
+          found: depth,
+          needed: maxDepth
+        }
+      });
+    }
+
+    function findJSXElement(variables, name) {
+      function find(refs) {
+        let i = refs.length;
+
+        while (--i >= 0) {
+          if (has(refs[i], 'writeExpr')) {
+            const writeExpr = refs[i].writeExpr;
+
+            return isJSXElement(writeExpr)
+              && writeExpr
+              || writeExpr.type === 'Identifier'
+              && findJSXElement(variables, writeExpr.name);
+          }
+        }
+
+        return null;
+      }
+
+      const variable = variableUtil.getVariable(variables, name);
+      return variable && variable.references && find(variable.references);
+    }
+
+    function checkDescendant(baseDepth, children) {
+      children.forEach(node => {
+        if (!hasJSX(node)) {
+          return;
+        }
+
+        baseDepth++;
+        if (baseDepth > maxDepth) {
+          report(node, baseDepth);
+        } else if (!isLeaf(node)) {
+          checkDescendant(baseDepth, node.children);
+        }
+      });
+    }
+
+    return {
+      JSXElement: function(node) {
+        if (!isLeaf(node)) {
+          return;
+        }
+
+        const depth = getDepth(node);
+        if (depth > maxDepth) {
+          report(node, depth);
+        }
+      },
+      JSXExpressionContainer: function(node) {
+        if (node.expression.type !== 'Identifier') {
+          return;
+        }
+
+        const variables = variableUtil.variablesInScope(context);
+        const element = findJSXElement(variables, node.expression.name);
+
+        if (element) {
+          const baseDepth = getDepth(node);
+          checkDescendant(baseDepth, element.children);
+        }
+      }
+    };
+  }
+};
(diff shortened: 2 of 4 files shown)


docs/rules/jsx-max-depth.md
```diff
@@ -0,0 +1,78 @@
...
+
+```
+
+## Rule Options
+
+It takes an option as the second parameter which can be a positive number for depth count.
+
+```js
+...
+"react/jsx-no-depth": [<enabled>, <number>]
```
ljharb: let's make an initial schema that's extensible; in other words, `{ "max": <number> }` instead of just the number.

lib/rules/jsx-max-depth.js
```diff
@@ -0,0 +1,71 @@
...
+    schema: [{
+      type: 'integer'
+    }]
+  },
+
+  create: function(context) {
+
+    var MESSAGE = 'JSX Element are nested too deeply ({{depth}}).';
+
+    var maxDepth = context.options[0] || 3;
```
ljharb: in new code, let's use const/let over var

lib/rules/jsx-max-depth.js
```diff
@@ -0,0 +1,71 @@
...
+      recommended: false
+    },
+    schema: [{
+      type: 'integer'
+    }]
+  },
+
+  create: function(context) {
+
+    var MESSAGE = 'JSX Element are nested too deeply ({{depth}}).';
```
ljharb: let's also include the current max length in the message, like max-len does.

tests/lib/rules/jsx-max-depth.js
```diff
@@ -0,0 +1,70 @@
...
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+var rule = require('../../../lib/rules/jsx-max-depth');
+var RuleTester = require('eslint').RuleTester;
+
+var parserOptions = {
+  ecmaVersion: 8,
```
ljharb: let's set this to `6`, since we shouldn't need ES2017 for this rule

tests/lib/rules/jsx-max-depth.js
```diff
@@ -0,0 +1,70 @@
...
+  sourceType: 'module',
+  ecmaFeatures: {
+    experimentalObjectRestSpread: true,
```
ljharb: similarly, this feature isn't required here

docs/rules/jsx-max-depth.md
chriswong: Done.

lib/rules/jsx-max-depth.js
chriswong: Done.

lib/rules/jsx-max-depth.js
chriswong: Done.

tests/lib/rules/jsx-max-depth.js
chriswong: Done.

tests/lib/rules/jsx-max-depth.js
chriswong: Done.

lib/rules/jsx-max-depth.js
```diff
@@ -0,0 +1,151 @@
...
+  meta: {
+    docs: {
+      description: 'Validate JSX maximum depth',
+      category: 'Stylistic Issues',
+      recommended: false
+    },
+    schema: [{
+      oneOf: [
+        {
+          type: 'integer',
```
ljharb: I'd prefer to not even allow this as an option, if possible - meaning, to only have the object schema.

lib/rules/jsx-max-depth.js
```diff
@@ -0,0 +1,151 @@
...
+        message: MESSAGE,
+        data: {
+          gotten: depth,
+          needed: maxDepth
+        }
+      });
+    }
+
+    function findJSXElement(variables, name) {
+      let find = function(refs) {
```
ljharb: `const` - or a function declaration

lib/rules/jsx-max-depth.js
```diff
@@ -0,0 +1,151 @@
...
+        let i = refs.length;
+
+        while (--i >= 0) {
+          if ('writeExpr' in refs[i]) {
```
ljharb: please use the `has` package rather than `in`, to avoid traversing the prototype chain.

lib/rules/jsx-max-depth.js
```diff
@@ -0,0 +1,151 @@
...
+            let writeExpr = refs[i].writeExpr;
```
ljharb: `const`

lib/rules/jsx-max-depth.js
```diff
@@ -0,0 +1,151 @@
...
+
+            return isJSXElement(writeExpr)
+            && writeExpr
```
ljharb: these chained lines should be indented one level further in

lib/rules/jsx-max-depth.js
```diff
@@ -0,0 +1,151 @@
...
+            || writeExpr.type === 'Identifier'
+            && findJSXElement(variables, writeExpr.name);
+          }
+        }
+
+        return null;
+      };
+
+      let variable = variableUtil.getVariable(variables, name);
```
ljharb: `const`

lib/rules/jsx-max-depth.js
```diff
@@ -0,0 +1,151 @@
...
+          report(node, depth);
+        }
+      },
+      JSXExpressionContainer: function(node) {
+        if (node.expression.type !== 'Identifier') {
+          return;
+        }
+
+        let variables = variableUtil.variablesInScope(context);
+        let element = findJSXElement(variables, node.expression.name);
```
ljharb: both of these should be `const`

lib/rules/jsx-max-depth.js
```diff
@@ -0,0 +1,151 @@
...
+
+        if (element) {
+          let baseDepth = getDepth(node);
```
ljharb: `const`

lib/rules/jsx-max-depth.js
```diff
@@ -0,0 +1,151 @@
...
+    function isExpression(node) {
+      return node.type === 'JSXExpressionContainer';
+    }
+
+    function hasJSX(node) {
+      return isJSXElement(node) || isExpression(node) && isJSXElement(node.expression);
+    }
+
+    function isLeaf(node) {
+      let children = node.children;
```
ljharb: `const`

lib/rules/jsx-max-depth.js
```diff
@@ -0,0 +1,151 @@
...
+
+      ]
+    }]
+  },
+  create: function(context) {
+    const MESSAGE = 'Expected the depth of JSX Elements nested should be {{needed}} but found {{gotten}}.';
+    const DEFAULT_DEPTH = 3;
+
+    let option = context.options[0] || DEFAULT_DEPTH;
+    let maxDepth = (typeof option === 'number' ? option : option.max) || DEFAULT_DEPTH;
```
ljharb: `const`

Also, what happens if I pass a value of `0`? This looks like it will always set 0 to 3. Please add some tests for values of `0` - ie, no nesting allowed.

lib/rules/jsx-max-depth.js
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+      ]
+    }]
+  },
+  create: function(context) {
+    const MESSAGE = 'Expected the depth of JSX Elements nested should be {{needed}} but found {{gotten}}.';
+    const DEFAULT_DEPTH = 3;
+
+    let option = context.options[0] || DEFAULT_DEPTH;
```
ljharb: `const`

lib/rules/jsx-max-depth.js
```diff
@@ -0,0 +1,151 @@
...
+              minimum: 0
+            }
+          }
+        }
+
+      ]
+    }]
+  },
+  create: function(context) {
+    const MESSAGE = 'Expected the depth of JSX Elements nested should be {{needed}} but found {{gotten}}.';
```
ljharb: "gotten" seems like an odd word choice; maybe "found"?

lib/rules/jsx-max-depth.js
chriswong: Done.

lib/rules/jsx-max-depth.js
chriswong: Done.

lib/rules/jsx-max-depth.js
chriswong: Done.

lib/rules/jsx-max-depth.js
chriswong:
```js
          max: {
            type: 'integer',
            minimum: 1
          }
```
`0` is no a valid value now.

lib/rules/jsx-max-depth.js
chriswong: Done.

lib/rules/jsx-max-depth.js
chriswong: Done. Changed to function declaration.

lib/rules/jsx-max-depth.js
chriswong: Done.

lib/rules/jsx-max-depth.js
chriswong: Done.

lib/rules/jsx-max-depth.js
chriswong: Done.

lib/rules/jsx-max-depth.js
chriswong: Done.

lib/rules/jsx-max-depth.js
chriswong: Done.

lib/rules/jsx-max-depth.js
chriswong: Done.

lib/rules/jsx-max-depth.js
ljharb: Why isn't zero a valid value? I'd think prohibiting nesting entirely is a totally valid option.

lib/rules/jsx-max-depth.js
```diff
@@ -0,0 +1,146 @@
...
+      category: 'Stylistic Issues',
+      recommended: false
+    },
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          max: {
+            type: 'integer',
+            minimum: 1
```
ljharb: i think this should be made to work with `0`.

otakustay: Since this rule is called "max-depth", I think it should detect the "depth" but not "nesting depth" of jsx, a depth of 0 means "no element here" which is absolutely useless

ljharb: If that's true, then a `max` of `1` would permit `<div />` but forbid `<div><span /><div>`; is that what the current PR's docs and test cases indicate?

ljharb: Specifically, I want it to be configurable so I can forbid *any nesting whatsoever* - a depth of `0` means "flat", so in fact `<div />` has a depth of zero.

chriswong: OKey.

jseminck:
Is there anything stopping this rule from moving forward? This seems pretty useful to me! 😄 

It probably needs a rebase though, at least one file has a conflict.

ljharb:
Happy to rereview after a rebase.

jseminck:
@chriswong any chance you could rebase this?

forabi:
@ljharb I rebased this on master. How should I submit the changes?

ljharb:
Certainly not with a duplicate PR, which I can never delete from my git history :-( Posting a link to the branch would be more than sufficient.

ljharb:
I've rebased this properly and pushed the changes up to here and to #1676.

PR #1264, opened 2017-06-19
New `boolean-prop-naming` rule for enforcing the naming of booleans

EvNaverniouk:
At my workplace, we prefer to name our boolean props as "isSomething" or "hasWhatever". This is an ESLint rule which allows devs to specify a Regex pattern for how they want their boolean props named.

Tested with createReactClass, ES6 components, stateless components and Flowtype.

I support, in the future, this rule can be expanded to support naming rules for ANY props, but that felt like overkill for the time being. Hopefully this rule will be useful as is.

### DIFF
--- a/docs/rules/boolean-prop-naming.md
+++ b/docs/rules/boolean-prop-naming.md
@@ -1,0 +1,67 @@
+# Enforces consistent naming for boolean props (react/boolean-prop-naming)
+
+Allows you to enforce a consistent naming pattern for props which expect a boolean value.
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```jsx
+var Hello = createReactClass({
+  propTypes: {
+    enabled: PropTypes.bool
+  },
+  render: function() { return <div />; };
+});
+```
+
+The following patterns are not considered warnings:
+
+```jsx
+var Hello = createReactClass({
+  propTypes: {
+    isEnabled: PropTypes.bool
+  },
+  render: function() { return <div />; };
+});
+```
+
+## Rule Options
+
+```js
+...
+"react/boolean-prop-naming": [<enabled>, { "propTypeNames": Array<string>, "rule": <string> }]
+...
+```
+
+### `propTypeNames`
+
+The list of prop type names that are considered to be booleans. By default this is set to `['bool']` but you can include other custom types like so:
+
+```jsx
+"react/boolean-prop-naming": ["error", { "propTypeNames": ["bool", "mutuallyExclusiveTrueProps"] }]
+```
+
+### `rule`
+
+The RegExp pattern to use when validating the name of the prop. The default value for this option is set to: `"^(is|has)[A-Z]([A-Za-z0-9]?)+"` to enforce `is` and `has` prefixes.
+
+For supporting "is" and "has" naming (default):
+
+- isEnabled
+- isAFK
+- hasCondition
+- hasLOL
+
+```jsx
+"react/boolean-prop-naming": ["error", { "rule": "^(is|has)[A-Z]([A-Za-z0-9]?)+" }]
+```
+
+For supporting "is" naming:
+
+- isEnabled
+- isAFK
+
+```jsx
+"react/boolean-prop-naming": ["error", { "rule": "^is[A-Z]([A-Za-z0-9]?)+" }]
+```
--- a/lib/rules/boolean-prop-naming.js
+++ b/lib/rules/boolean-prop-naming.js
@@ -1,0 +1,241 @@
+/**
+ * @fileoverview Enforces consistent naming for boolean props
+ * @author Evgueni Naverniouk
+ */
+'use strict';
+
+const has = require('has');
+const Components = require('../util/Components');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      category: 'Stylistic Issues',
+      description: 'Enforces consistent naming for boolean props',
+      recommended: false
+    },
+
+    schema: [{
+      additionalProperties: false,
+      properties: {
+        propTypeNames: {
+          items: {
+            type: 'string'
+          },
+          minItems: 1,
+          type: 'array',
+          uniqueItems: true
+        },
+        rule: {
+          default: '^(is|has)[A-Z]([A-Za-z0-9]?)+',
+          minLength: 1,
+          type: 'string'
+        }
+      },
+      type: 'object'
+    }]
+  },
+
+  create: Components.detect(function(context, components, utils) {
+    const sourceCode = context.getSourceCode();
+    const config = context.options[0] || {};
+    const rule = config.rule ? new RegExp(config.rule) : null;
+    const propTypeNames = config.propTypeNames || ['bool'];
+
+    // Remembers all Flowtype object definitions
+    var objectTypeAnnotations = new Map();
+
+    /**
+     * Checks if node is `propTypes` declaration
+     * @param {ASTNode} node The AST node being checked.
+     * @returns {Boolean} True if node is `propTypes` declaration, false if not.
+     */
+    function isPropTypesDeclaration(node) {
+      // Special case for class properties
+      // (babel-eslint does not expose property name so we have to rely on tokens)
+      if (node.type === 'ClassProperty') {
+        const tokens = context.getFirstTokens(node, 2);
+        if (tokens[0].value === 'propTypes' || (tokens[1] && tokens[1].value === 'propTypes')) {
+          return true;
+        }
+        // Flow support
+        if (node.typeAnnotation && node.key.name === 'props') {
+          return true;
+        }
+        return false;
+      }
+
+      return Boolean(
+        node &&
+        node.name === 'propTypes'
+      );
+    }
+
+    /**
+     * Returns the prop key to ensure we handle the following cases:
+     * propTypes: {
+     *   full: React.PropTypes.bool,
+     *   short: PropTypes.bool,
+     *   direct: bool
+     * }
+     * @param {Object} node The node we're getting the name of
+     */
+    function getPropKey(node) {
+      if (node.value.property) {
+        return node.value.property.name;
+      }
+      if (node.value.type === 'Identifier') {
+        return node.value.name;
+      }
+      return null;
+    }
+
+    /**
+	 * Returns the name of the given node (prop)
+	 * @param {Object} node The node we're getting the name of
+	 */
+    function getPropName(node) {
+      // Due to this bug https://github.com/babel/babel-eslint/issues/307
+      // we can't get the name of the Flow object key name. So we have
+      // to hack around it for now.
+      if (node.type === 'ObjectTypeProperty') {
+        return sourceCode.getFirstToken(node).value;
+      }
+
+      return node.key.name;
+    }
+
+    /**
+     * Checks and mark props with invalid naming
+     * @param {Object} node The component node we're testing
+     * @param {Array} proptypes A list of Property object (for each proptype defined)
+     */
+    function validatePropNaming(node, proptypes) {
+      const component = components.get(node) || node;
+      const invalidProps = component.invalidProps || [];
+
+      proptypes.forEach(function (prop) {
+        const propKey = getPropKey(prop);
+        const flowCheck = (
+          prop.type === 'ObjectTypeProperty' &&
+          prop.value.type === 'BooleanTypeAnnotation' &&
+          rule.test(getPropName(prop)) === false
+        );
+        const regularCheck = (
+          propKey &&
+          propTypeNames.indexOf(propKey) >= 0 &&
+          rule.test(getPropName(prop)) === false
+        );
+
+        if (flowCheck || regularCheck) {
+          invalidProps.push(prop);
+        }
+      });
+
+      components.set(node, {
+        invalidProps: invalidProps
+      });
+    }
+
+    /**
+     * Reports invalid prop naming
+     * @param {Object} component The component to process
+     */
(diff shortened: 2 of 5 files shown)


docs/rules/boolean-prop-naming.md
```diff
@@ -0,0 +1,69 @@
...
+Allows you to enforce a consistent naming pattern for props which expect a boolean value.
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```jsx
+var Hello = createReactClass({
+  propTypes: {
+    enabled: PropTypes.boolean
```
ljharb: `PropTypes.bool` - `.boolean` isn't a thing (same throughout)

docs/rules/boolean-prop-naming.md
```diff
@@ -0,0 +1,69 @@
...
+
+```js
+...
+"react/boolean-prop-naming": [<enabled>, { "rule": <string> }]
+...
+```
+
+### `rule`
+
+The Regex pattern to use when validating the name of the prop. Some common examples are:
```
ljharb: `regex` or `RegExp`

docs/rules/boolean-prop-naming.md
```diff
@@ -0,0 +1,69 @@
...
+- hasCondition
+- hasLOL
+
+```jsx
+"react/boolean-prop-naming": ["error", { "rule": "^(is|has)[A-Z]([A-Za-z0-9]?)+" }]
+```
+
+## About component detection
+
+For this rule to work we need to detect React components, this could be very hard since components could be declared in a lot of ways.
```
ljharb: We should be using `Components.detect` in all our rules; is there a reason this paragraph needs to exist?

lib/rules/boolean-prop-naming.js
```diff
@@ -0,0 +1,210 @@
+/**
+ * @fileoverview Enforces consistent naming for boolean props
+ * @author Evgueni Naverniouk
+ */
+'use strict';
+
+var has = require('has');
```
ljharb: const?

tests/lib/rules/boolean-prop-naming.js
```diff
@@ -0,0 +1,337 @@
...
+// ------------------------------------------------------------------------------
+
+var ruleTester = new RuleTester({parserOptions});
+ruleTester.run('boolean-prop-naming', rule, {
+
+  valid: [{
+    // createReactClass components with PropTypes
+    code: [
+      'var Hello = createReactClass({',
+      '  propTypes: {isSomething: PropTypes.boolean},',
```
ljharb: i think all the PropType test cases are invalid - `s/boolean/bool/g`.

docs/rules/boolean-prop-naming.md
EvNaverniouk: Was copy/pasted from another rule's document. Should I remove it?

lib/rules/boolean-prop-naming.js
EvNaverniouk: Thanks. Not all rules were using ES6, wasn't sure which version of Node we're supporting. Will update accordingly.

docs/rules/boolean-prop-naming.md
EvNaverniouk: Oops. Good catch.

EvNaverniouk:
> What about custom propTypes that enforce booleans but aren't PropTypes.bool? For example, mutuallyExclusiveTrueProps from https://www.npmjs.com/package/airbnb-prop-types.
> 
> Can there be a way to customize the name of custom propTypes that are booleans?

Good idea. I've added a new `propTypeNames` configuration option.

docs/rules/boolean-prop-naming.md
ljharb: Yeah, this seems more like a project-level thing, not a rule-level thing.

docs/rules/boolean-prop-naming.md
```diff
@@ -0,0 +1,77 @@
...
+
+The list of prop type names that are considered to be booleans. By default this is set to `['bool']` but you can include other custom types like so:
+
+```jsx
+"react/boolean-prop-naming": ["error", { "propTypeNames": ["bool", "mutuallyExclusiveTrueProps"] }]
+```
+
+### `rule`
+
+The RegExp pattern to use when validating the name of the prop. Some common examples are:
```
ljharb: what's the default for this option?

lib/rules/boolean-prop-naming.js
```diff
@@ -0,0 +1,237 @@
...
+        }
+      },
+      type: 'object'
+    }]
+  },
+
+  create: Components.detect(function(context, components, utils) {
+    const sourceCode = context.getSourceCode();
+    const config = context.options[0] || {};
+    const rule = config.rule ? new RegExp(config.rule) : null;
```
ljharb: i'm really really not a fan of user-supplied regexes, and passing unsanitized input into `new RegExp` can definitely cause catastrophic backtracking and other malicious regexes to be invoked.

Can we simplify this somehow, perhaps requiring a list of prefixes?

lib/rules/boolean-prop-naming.js
```diff
@@ -0,0 +1,237 @@
...
+          return true;
+        }
+        // Flow support
+        if (node.typeAnnotation && node.key.name === 'props') {
+          return true;
+        }
+        return false;
+      }
+
+      return Boolean(
```
ljharb: usually we just do `‼x` instead of `Boolean(x)`, nbd tho

tests/lib/rules/boolean-prop-naming.js
```diff
@@ -0,0 +1,427 @@
...
+// Requirements
+// ------------------------------------------------------------------------------
+
+var rule = require('../../../lib/rules/boolean-prop-naming');
+var RuleTester = require('eslint').RuleTester;
+
+require('babel-eslint');
+
+var parserOptions = {
+  ecmaVersion: 8,
```
ljharb: this can probably be `6`/`2015`, yes?

tests/lib/rules/boolean-prop-naming.js
```diff
@@ -0,0 +1,427 @@
...
+  sourceType: 'module',
+  ecmaFeatures: {
+    experimentalObjectRestSpread: true,
```
ljharb: i don't think we're using this feature in these tests

docs/rules/boolean-prop-naming.md
EvNaverniouk: `null` is the default at the moment. It didn't seem right to enforce my own preferences on all users. Do you think is makes sense to set "(is|has)*" as the default?

docs/rules/boolean-prop-naming.md
EvNaverniouk: WIll remove.

lib/rules/boolean-prop-naming.js
EvNaverniouk: I'm not opposed to replacing the RegExp entry to just prefixes, however, that makes the rule much less powerful and less customizable. Having user-supplied regexes in ESLint rules isn't uncommon (eg. http://eslint.org/docs/rules/id-match) when specifying naming conventions for things.

tests/lib/rules/boolean-prop-naming.js
EvNaverniouk: Will fix.

tests/lib/rules/boolean-prop-naming.js
EvNaverniouk: Will fix.

docs/rules/boolean-prop-naming.md
ljharb: Since if not, the rule does nothing by default, yes.

The ideal scenario is that nobody ever needs to write a regex, except for the minority edge cases.

EvNaverniouk: Agreed. Will set a default.

lib/rules/boolean-prop-naming.js
```diff
@@ -0,0 +1,238 @@
...
+      additionalProperties: false,
+      properties: {
+        propTypeNames: {
+          items: {
+            type: 'string'
+          },
+          type: 'array'
+        },
+        rule: {
+          default: null,
```
ljharb: Can we put the default here, in the schema, and use `minLength: 1` to avoid the empty string?

lib/rules/boolean-prop-naming.js
```diff
@@ -0,0 +1,238 @@
...
+      type: 'object'
+    }]
+  },
+
+  create: Components.detect(function(context, components, utils) {
+    const DEFAULT_RULE = '^(is|has)[A-Z]([A-Za-z0-9]?)+';
+    const sourceCode = context.getSourceCode();
+    const config = context.options[0] || {};
+    const rule = new RegExp(config.rule || DEFAULT_RULE);
+    const propTypeNames = config.propTypeNames ? config.propTypeNames : ['bool'];
```
ljharb: Similarly, can this default go in the schema instead?

lib/rules/boolean-prop-naming.js
```diff
@@ -0,0 +1,238 @@
...
+    },
+
+    schema: [{
+      additionalProperties: false,
+      properties: {
+        propTypeNames: {
+          items: {
+            type: 'string'
+          },
+          type: 'array'
```
ljharb: Let's add uniqueItems: true and minItems: 1

lib/rules/boolean-prop-naming.js
EvNaverniouk: For some reason, `default` schema definition doesn't work for this option. Maybe because it's an `array` type? Specifying a `default` at the schema level here never returns anything from config object.

ljharb: i guess you can't have a default array in jsonschema.

in that case, `a ? a : b` should always be `a || b` :-)

EvNaverniouk: Updated. Thanks!

lib/rules/boolean-prop-naming.js
```diff
@@ -0,0 +1,240 @@
...
+        }
+      },
+
+      'Program:exit': function() {
+        if (!rule) {
+          return;
+        }
+
+        const list = components.list();
+        for (let component in list) {
```
ljharb: Let's avoid for..in; can we use Object.keys().forEach here instead?

EvNaverniouk: Fixed

lib/rules/boolean-prop-naming.js
```diff
@@ -0,0 +1,240 @@
...
+      const component = components.get(node) || node;
+      const invalidProps = component.invalidProps || [];
+
+      proptypes.forEach(function (prop) {
+        const propKey = getPropKey(prop);
+        if (
+          (
+            prop.type === 'ObjectTypeProperty' &&
+            prop.value.type === 'BooleanTypeAnnotation' &&
+            getPropName(prop).search(rule) < 0
```
lencioni: I think `rule.test(getPropName(prop))` might be more appropriate here (and below)

lib/rules/boolean-prop-naming.js
```diff
@@ -0,0 +1,240 @@
...
+    /**
+     * Reports invalid prop naming
+     * @param {Object} component The component to process
+     */
+    function reportInvalidNaming(component) {
+      component.invalidProps.forEach(function (propNode) {
+        const propName = getPropName(propNode);
+        context.report({
+          node: propNode,
+          message: `Prop name (${propName}) doesn't match specified rule (${config.rule})`,
```
lencioni: "specified" seems redundant in this message. I think you can remove it without losing any meaning. What do you think?

lib/rules/boolean-prop-naming.js
```diff
@@ -0,0 +1,240 @@
...
+  },
+
+  create: Components.detect(function(context, components, utils) {
+    const sourceCode = context.getSourceCode();
+    const config = context.options[0] || {};
+    const rule = config.rule ? new RegExp(config.rule) : null;
+    const propTypeNames = config.propTypeNames || ['bool'];
+
+    // Remembers all Flowtype object definitions
+    var ObjectTypeAnnonations = {};
```
lencioni: typo? `ObjectTypeAnnotations`?

Also, since this isn't instantiated, it should probably start with a lowercase letter.

And it might be a bit nicer to use a `Map` instead of a plain object.

ljharb: +1 on all 3 points, nice catch

lib/rules/boolean-prop-naming.js
EvNaverniouk: Will change to use `.exec()` actually, since (after some research) that appears to be the most performance way to execute Regex rules. Hope that's ok.

lib/rules/boolean-prop-naming.js
EvNaverniouk: Agreed. Will remove.

lib/rules/boolean-prop-naming.js
ljharb: `exec` mutates; `test` doesn't - test might be better.

EvNaverniouk: Fixed.

wmertens:
How should you name a prop that enables some feature? Normally I'd write e.g. `<Nav showHeaders />`. Should this instead be `<Nav hasHeaders />`, `<Nav isHeadered />`, …?

EvHaus:
@wmertens I would call it `hasHeader`. Keep in mind, this rule is optional and configurable. If you prefer to use `showHeaders` then you can add the "show" prefix to the rule's options.

PR #1271, opened 2017-06-24
Fix crash when rel tag is empty

### LINKED ISSUE #1269: jsx-no-target-blank crashes on empty `rel` attribute
Using `eslint@4.0.0` and `eslint-plugin-react@7.1.0`, the `jsx-no-target-blank` rule crashes on the following code:

```html
<a href="https://example.com" target="_blank" rel />
```

This can be reproduced with:

```sh
echo '<a href="https://example.com" target="_blank" rel />' | node_modules/.bin/eslint --stdin --no-eslintrc --plugin=react --parser-options='ecmaFeatures: {jsx: true}' --rule 'react/jsx-no-target-blank:error'
```

This is a very minor issue, since I don't know of a good use case for having an empty `rel` attribute. I only noticed it because a crash notification appeared in my editor.

dustinsoftware:
Fixes #1269

### DIFF
--- a/lib/rules/jsx-no-target-blank.js
+++ b/lib/rules/jsx-no-target-blank.js
@@ -26,8 +26,8 @@ function hasExternalLink(element) {
 function hasSecureRel(element) {
   return element.attributes.find(function (attr) {
     if (attr.name.name === 'rel') {
-      var tags = attr.value.type === 'Literal' && attr.value.value.toLowerCase().split(' ');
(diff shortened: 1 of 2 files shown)


dustinsoftware:
It doesn't crash :) added some more test cases

ljharb:
@dustinsoftware any chance you'd mind following up with some test cases for template literals?

PR #1272, opened 2017-06-24
Add more support for wrapped propTypes

### LINKED ISSUE #1266: Most propType rules do not support propTypes wrapped in a function
https://github.com/yannickcr/eslint-plugin-react/pull/1253 added support for `propTypes` wrapped in a function to `no-unused-prop-types`. Similar support does not exist for other `propType` rules such as `sort-prop-types`, `forbid-prop-types`, and `require-default-props`. This came up a result of https://github.com/airbnb/prop-types/issues/23.

@ljharb's suggestion to move the `propWrapperFunctions` setting from `no-unused-prop-types` to `settings` and add support for that setting in all `propType` rules seems like the right move.

Hoping that I can find time to open a PR for this sometime over the next several weeks.

dustinsoftware:
Fixes #1266

### DIFF
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -44,13 +44,6 @@ module.exports = {
         },
         skipShapeProps: {
           type: 'boolean'
-        },
-        propWrapperFunctions: {
-          type: 'array',
-          items: {
-            type: 'string'
-          },
-          uniqueItems: true
         }
       },
       additionalProperties: false
@@ -63,7 +56,7 @@ module.exports = {
     var configuration = Object.assign({}, defaults, context.options[0] || {});
     var skipShapeProps = configuration.skipShapeProps;
     var customValidators = configuration.customValidators || [];
-    var propWrapperFunctions = new Set(configuration.propWrapperFunctions || []);
+    var propWrapperFunctions = new Set(context.settings.propWrapperFunctions || []);
 
     // Used to track the type annotations in scope.
     // Necessary because babel's scopes do not track type annotations.
--- a/lib/rules/require-default-props.js
+++ b/lib/rules/require-default-props.js
@@ -22,11 +22,25 @@ module.exports = {
       category: 'Best Practices'
     },
 
-    schema: []
+    schema: [{
+      type: 'object',
+      properties: {
+        propWrapperFunctions: {
+          type: 'array',
+          items: {
+            type: 'string'
+          },
+          uniqueItems: true
+        }
+      },
+      additionalProperties: false
+    }]
   },
 
   create: Components.detect(function(context, components, utils) {
     var sourceCode = context.getSourceCode();
+    var configuration = context.options[0] || {};
+    var propWrapperFunctions = new Set(configuration.propWrapperFunctions || []);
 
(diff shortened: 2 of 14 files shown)


ljharb:
Thanks!

I think it'd be better, rather than only adding this option to individual rules, to also add it to "settings" (like the React pragma).

Basically, individual rule settings would take precedence, and lacking that, the global settings would be used.

dustinsoftware:
Ah, right after pushing this, I saw that you'd like to have it in settings *and* individual rule configuration. What case did you have in mind where it would be necessary to have the option to override it per-rule?

lib/rules/no-unused-prop-types.js
```diff
@@ -63,7 +56,7 @@ module.exports = {
     var configuration = Object.assign({}, defaults, context.options[0] || {});
     var skipShapeProps = configuration.skipShapeProps;
     var customValidators = configuration.customValidators || [];
-    var propWrapperFunctions = new Set(configuration.propWrapperFunctions || []);
+    var propWrapperFunctions = new Set(context.settings.propWrapperFunctions || []);
```
ljharb: will `context.settings` always exist and be an object?

lib/rules/require-default-props.js
```diff
@@ -105,6 +106,13 @@ module.exports = {
       if (node.type === 'Identifier') {
         return findVariableByName(node.name);
       }
+      if (
+        node.type === 'CallExpression' &&
+        propWrapperFunctions.has(node.callee.name) &&
+        node.arguments && node.arguments[0]
```
ljharb: this should probably be `node.arguments.length > 0`?

ljharb:
No use case specifically; just aiming for maximum configurability.

Let's keep it how you have it for now, and we can add the specific-rule settings later if needed.

lib/rules/require-default-props.js
dustinsoftware: It shouldn't matter, if `arguments` is empty, then `arguments[0]` will return undefined, which will cause this test to fail. It also handles the case where the first element of arguments contains something that is falsy :) 

It looks like checking the arguments array this way also matches code style in other places in this project.

lib/rules/no-unused-prop-types.js
dustinsoftware: Yes, I tested both with and without settings defined - `context.settings` will always be an object.

lib/rules/require-default-props.js
ljharb: in v8, accessing an `arguments` index that's greater than `length` deoptimizes, so it'd be a performance issue.

ljharb:
This needs to be merged before the next release, so if I don't hear any objections by Monday, I'm going to merge it.

dustinsoftware:
Pushed support for a couple more, I think that's all the major prop-type rules.

ljharb:
I wonder if, like Components.detect, we could make a PropTypes.detect function that would make this logic reusable?

dustinsoftware:
I think that could be helpful, but maybe out of scope for this PR. There's a lot of logic around finding declared prop types that is duplicated across multiple rules.

PR #1273, opened 2017-06-25
Add fixer for jsx/sort-props

Overload119:
Hi all.

I've been meaning to dabble with eslint for a while, and this was a small project I wanted to take on. We're currently using this linter on our codebase, and I thought it would be great to have a fixer for this one. 

This particular rule has so many configurations, and I didn't bother to handle all of them although it wouldn't be hard to do so. I may do that in a follow-up PR.

I tested using the following:

```
node_modules/mocha/bin/_mocha /Users/amirsharif/Projects/eslint-plugin-react/tests/lib/rules/jsx-sort-props.js
```

I also added an additional, multi-line test.

### DIFF
--- a/lib/rules/jsx-sort-props.js
+++ b/lib/rules/jsx-sort-props.js
@@ -40,6 +40,77 @@ function isReservedPropName(name, list) {
   return list.indexOf(name) >= 0;
 }
 
+function alphabeticalCompare(a, b, ignoreCase) {
+  if (ignoreCase) {
+    a = a.toLowerCase();
+    b = b.toLowerCase();
+  }
+  return a.localeCompare(b);
+}
+
+/**
+ * Create an array of arrays where each subarray is composed of attributes
+ * that are considered sortable.
+ * @param {Array<JSXSpreadAttribute|JSXAttribute>} attributes
+ * @return {Array<Array<JSXAttribute>}
+ */
+function getGroupsOfSortableAttributes(attributes) {
+  const sortableAttributeGroups = [];
+  let groupCount = 0;
+  for (var i = 0; i < attributes.length; i++) {
+    const lastAttr = attributes[i - 1];
+    // If we have no groups or if the last attribute was JSXSpreadAttribute
+    // then we start a new group. Append attributes to the group until we
+    // come across another JSXSpreadAttribute or exhaust the array.
+    if (
+      !lastAttr ||
+      (lastAttr.type === 'JSXSpreadAttribute' &&
+        attributes[i].type !== 'JSXSpreadAttribute')
+    ) {
+      groupCount++;
+      sortableAttributeGroups[groupCount - 1] = [];
+    }
+    if (attributes[i].type !== 'JSXSpreadAttribute') {
+      sortableAttributeGroups[groupCount - 1].push(attributes[i]);
+    }
+  }
+  return sortableAttributeGroups;
+}
+
+function generateFixerFunction(node, context) {
+  var sourceCode = context.getSourceCode();
+  var attributes = node.attributes.slice(0);
+  var configuration = context.options[0] || {};
+  var ignoreCase = configuration.ignoreCase || false;
+
+  // Sort props according to the context. Only supports ignoreCase.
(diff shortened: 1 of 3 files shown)


lib/rules/jsx-sort-props.js
```diff
@@ -40,6 +40,59 @@ function isReservedPropName(name, list) {
...
 }
 
+
+function alphabeticalCompare(a, b, ignoreCase) {
+  if (ignoreCase) {
+    a = a.toLowerCase();
+    b = b.toLowerCase();
+  }
+
+  if (a < b) {
```
ljharb: if they're both strings, this can be `return a.localeCompare(b);`

lib/rules/jsx-sort-props.js
```diff
@@ -40,6 +40,59 @@ function isReservedPropName(name, list) {
...
+
+function generateFixerFunction(node, context) {
+  var sourceCode = context.getSourceCode();
+  var sortedAttributes = node.attributes.slice(0);
+  var configuration = context.options[0] || {};
+  var ignoreCase = configuration.ignoreCase || false;
+
+  // Sort props according to the context. Only supports ignoreCase.
+  sortedAttributes.sort(function(a, b) {
+    // Put spread attributes at the end.
```
ljharb: autofixers are supposed to always be safe; it's not safe to automatically put spread props on the end because that might change the override order.

In other words, any element that has spread props is simply not safely autofixable - so if we still want to autofix situations without spread props, then we shouldn't be making any changes to elements with spread props.

tests/lib/rules/jsx-sort-props.js
```diff
@@ -153,15 +153,82 @@ ruleTester.run('jsx-sort-props', rule, {
...
+      errors: [expectedError],
+      output: '<App a b />;'
+    },
+    {
+      code: '<App {...this.props} b a />;',
+      errors: [expectedError],
+      output: '<App a b {...this.props} />;'
+    },
+    {
+      code: '<App c {...this.props} b a />;',
```
ljharb: for example, this might be totally valid, because `this.props` might contain `b` and/or `a`, and the dev might want to override their values.

Overload119:
Thanks for the feedback. That's a good point I never considered. I've made some changes that pass the tests and ignore the spread attributes.

tests/lib/rules/jsx-sort-props.js
```diff
@@ -192,12 +192,12 @@ ruleTester.run('jsx-sort-props', rule, {
...
     {
       code: '<App {...this.props} c="a" a="c" b="b" />;',
-      output: '<App a="c" b="b" c="a" {...this.props} />;',
+      output: '<App {...this.props} a="c" b="b" c="a" />;',
       errors: 2
     },
     {
       code: '<App d="d" b="b" {...this.props} c="a" a="c" />;',
-      output: '<App a="c" b="b" c="a" d="d" {...this.props} />;',
+      output: '<App a="c" b="b" {...this.props} c="a" d="d" />;',
```
ljharb: sorry if i wasn't clear - i'm afraid this is also unsafe. Any element with a spread prop *can not be touched in any way* that would alter overrides of the spread object. So, in `<App a="c" b="b" {...this.props} c="a" d="d" />`, you could sort the list of `[a, b]`, and the list of `[c, d]` - but a and b must always remain *before* `...this.props`, and `c` and `c` must remain after it.

Overload119:
Thanks for the quick feedback. I'm determined to get this in, let me know what you think. I changed it to consider the subgroups.

lib/rules/jsx-sort-props.js
```diff
@@ -77,30 +77,30 @@ function getGroupsOfSortableAttributes(attributes) {
   return sortableAttributeGroups;
 }
 
-function generateFixerFunction(node, context) {
-  var sourceCode = context.getSourceCode();
-  var attributes = node.attributes.slice(0);
-  var configuration = context.options[0] || {};
-  var ignoreCase = configuration.ignoreCase || false;
+const generateFixerFunction = (node, context) => {
```
ljharb: (fwiw i think this was fine to leave as a normal function)

Issue #1240, opened 2017-06-06
Use template literals in tests


jseminck:
This is a development related question. Since eslint has to support Node 4, and Node 4 seemingly has support for template literals (see http://node.green/). Is it ok to use this syntax in eslint-plugin-react? It would be much nicer to write:

```js
code: `
   component MyHugeComponent {
   ...
}`
```

Instead of the current approach - which is pretty annoying when writing big components:

```js
code: [
  'component MyHudeComponent {',
  '...',
  '}'
].join('\n\)
```

Also it seems eslint itself already uses the template literals: https://github.com/eslint/eslint/blob/master/tests/lib/rules/newline-after-var.js#L329

ljharb:
Yes, absolutely it would. The change to drop node < 4 is recent, and we just haven't gone through and updated everything to use patterns node 4 supports.

We also have to go through and convert every `var` to `const` or `let`, for example, as well as convert inline function callbacks to arrows.

jseminck:
Cool. It would be really nice to see the code-base more "modern". :) 

I have been doing some `var` to `const`/`let` changes recently. I should hopefully get around to finishing that on Thursday or Friday.

ljharb:
All the tests have been converted at this point.

PR #1274, opened 2017-06-25
ES6-ify void-dom-elements-no-children rule and test file

dfilipidisz:
I've seen in issue #1240 that using some new patterns from node 4 would be a nice addition. 

Starting out I only updated the ```void-dom-elements-no-children``` rule and test file to use:
* ```const``` and ```let``` instead of ```var```
* Arrow funtions
* Template literals instead of ```[].join()```
* Some minor unification of formatting

I've also seen a [comment](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/void-dom-elements-no-children.js#L16) in the rule file that using Set would be nice, which is also available with node 4. Can I make that conversion too?

Is this PR okay, and if so, is there a preferred way I should continue with updating?

### DIFF
--- a/lib/rules/void-dom-elements-no-children.js
+++ b/lib/rules/void-dom-elements-no-children.js
@@ -5,9 +5,9 @@
  */
 'use strict';
 
-var has = require('has');
+const has = require('has');
 
-var Components = require('../util/Components');
+const Components = require('../util/Components');
 
 // ------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


dfilipidisz:
Thanks for the review and directions on how to proceed! I'll look into how to utilize codemods for this and will keep each transformation in separate PRs.

PR #1285, opened 2017-07-05
Add support for conditional rendering to jsx-multiline-wrap

jseminck:
Main issue: https://github.com/yannickcr/eslint-plugin-react/issues/1282

I tried to make the fixer better so the output has the parens in the correct place. Probably, need to add a few more tests, but I think the current solution looks quite solid.

The rule is defaulted to `false` initially. If enabling it by default, a bigger version bump would be required?

See also some comments on another issue regarding `jsx-multiline-wrap`: https://github.com/yannickcr/eslint-plugin-react/issues/1207#issuecomment-313086986

### DIFF
--- a/docs/rules/jsx-wrap-multilines.md
+++ b/docs/rules/jsx-wrap-multilines.md
@@ -133,7 +133,7 @@ The following patterns are considered warnings when configured `{conditional: tr
 </div>
 ```
 
-The following patterns are not considered warnings when configured `{arrow: true}`.
+The following patterns are not considered warnings when configured `{conditional: true}`.
 
 ```jsx
 <div>
--- a/docs/rules/jsx-wrap-multilines.md
+++ b/docs/rules/jsx-wrap-multilines.md
@@ -6,7 +6,7 @@ Wrapping multiline JSX in parentheses can improve readability and/or convenience
 
 ## Rule Details
 
-This rule optionally takes a second parameter in the form of an object, containing places to apply the rule. By default, all the syntax listed below will be checked, but these can be explicitly disabled. Any syntax type missing in the object will follow the default behavior (become enabled).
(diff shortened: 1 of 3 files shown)


jseminck:
FYI broken build doesn't seem to be related to this branch. Currently when I try to build master I get the same error messages, even though the latest build on travis for master was successful (21hrs ago: https://travis-ci.org/yannickcr/eslint-plugin-react/builds/250125668)

docs/rules/jsx-wrap-multilines.md
```diff
@@ -118,3 +120,27 @@ var hello = () => (
...
+<div>
+  {myExpression &&
+    <div>
+      <p>Hello World</p>
+    </div>
+  }
+</div>
+```
+
+The following patterns are not considered warnings when configured `{arrow: true}`.
```
deecewan: do you mean `{conditional: true}`?

deecewan:
I know that it'd require a bigger version bump, but to me it'd make more sense to have all the rules enabled by default.

Also, while you're in there, feel like adding ternaries?

ljharb:
Enabling rules by default should wait for a separate semver-major PR, released after the non-major changes are released.

marcdavi-es:
*** sits wriggling with excitement waiting for the merge ***

ljharb:

jseminck:
Are there some problems regarding for this PR to be moved forward? 

I'd like to take a further look at this rule because I currently joined a project where we use a lot of conditional rendering and `eslint-plugin-react` doesn't always fix things as I expect it to. 😄

jseminck:
Seems like this was added in another PR, hence closing.

To achieve this functionality, use the option: `{logical: 'parens'}` or `{logical: 'parens-new-line'}`.

ljharb:
@jseminck sorry about the conflict :-)

PR #1294, opened 2017-07-11
Add lifecycle methods check to no-typos rule

haridusenadeera:
`no-typos` rule currently checks for class properties (`defaultProps`, `childContextTypes` etc.) only. 

My attempt is to make sure that react lifecycle methods also do not have any casing typos.

```js
  componentwillMount() {}
}

  ComponentWillReceiveProps() {}
}

  componentdidupdate() {}
}
```

The following patterns are not considered warnings:

```js
  componentWillMount() {}
}

  componentWillReceiveProps() {}
}

  componentDidUpdate() {}
}
```
I'm open to improvements to this rule. Thanks!

### DIFF
--- a/docs/rules/no-typos.md
+++ b/docs/rules/no-typos.md
@@ -1,17 +1,30 @@
 # Prevents common casing typos (react/no-typos)
 
-Ensure no casing typos were made declaring static class properties
+Ensure no casing typos were made declaring static class properties and lifecycle methods.
 
 ## Rule Details
 
-This rule checks whether the declared static class properties related to React components
-do not contain any typos. It currently makes sure that the following class properties have
+This rule checks whether the declared static class properties and lifecycle methods related to React components
+do not contain any typos.
+
+It currently makes sure that the following class properties have
 no casing typos:
 
 * propTypes
 * contextTypes
 * childContextTypes
 * defaultProps
+
+and the following react lifecycle methods:
+
+* componentWillMount
+* componentDidMount
+* componentWillReceiveProps
+* shouldComponentUpdate
+* componentWillUpdate
+* componentDidUpdate
+* componentWillUnmount
+
 
 The following patterns are considered warnings:
 
@@ -47,6 +60,18 @@ class MyComponent extends React.Component {
 class MyComponent extends React.Component {
   static defaultprops = {}
 }
+
+class MyComponent extends React.Component {
+  componentwillMount() {}
+}
+
+class MyComponent extends React.Component {
+  ComponentWillReceiveProps() {}
+}
+
+class MyComponent extends React.Component {
+  componentdidupdate() {}
+}
 ```
 
 The following patterns are not considered warnings:
@@ -66,5 +91,17 @@ class MyComponent extends React.Component {
 
 class MyComponent extends React.Component {
   static defaultProps = {}
+}
+
+class MyComponent extends React.Component {
+  componentWillMount() {}
+}
+
+class MyComponent extends React.Component {
+  componentWillReceiveProps() {}
+}
+
+class MyComponent extends React.Component {
+  componentDidUpdate() {}
 }
 ```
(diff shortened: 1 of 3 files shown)


lib/rules/no-typos.js
```diff
@@ -33,6 +42,17 @@ module.exports = {
       });
     }
 
+    function reportErrorIfLifecycleMethodCasingTypo(node) {
```
jseminck: I think we should rename the `reportErrorIfCasingTypo` function as well to contain something like `ifClassProperty`.

lib/rules/no-typos.js
```diff
@@ -54,6 +74,14 @@ module.exports = {
           const propertyName = node.property.name;
           reportErrorIfCasingTypo(node, propertyName);
         }
+      },
+
+      MethodDefinition: function (node) {
+        if (!utils.isES6Component(node.parent.parent) || utils.isReturningJSX(node.parent.parent)) {
```
jseminck: Do we need the check for `utils.isReturningJSX()` ? This is used to detect Stateless Function Component, which does not contain user defined lifecycle methods.

tests/lib/rules/no-typos.js
```diff
@@ -181,6 +182,49 @@ ruleTester.run('no-typos', rule, {
...
+    code: [
+      'class Hello extends React.Component {',
+      '  shouldComponentUpdate() { }',
+      '  componentWillReceiveProps() { }',
+      '  render() {',
+      '    return <div>Hello {this.props.name}</div>;',
+      '  }',
+      '}'
+    ].join('\n'),
+    parserOptions: parserOptions
```
jseminck: I guess we need tests for all lifecycle methods, not just a few.

Additionally, we'll need some valid tests for a class that is not a React component. that can still have any lifecycle method defined with any kind of casing typo.

Lastly, I do not know if we should support this kind of syntax. It seems to work ( https://jsfiddle.net/69z2wepo/82794/ ) but I have never seen anyone do it. @ljharb what do you think of this?

```js
	render() {
  }
}

Hello.prototype.componentDidMount = function() {
  console.log("component has mounted");
}

ReactDOM.render(
  <Hello name="World" />,
  document.getElementById('container')
);
```

jseminck:
Very good, thanks a lot! I was planning on doing this but haven't had time. This is some-what part of this issue: https://github.com/yannickcr/eslint-plugin-react/issues/843 (casing typos would be the first step, further the actual typos can be looked into).

I added a few initial comments.

docs/rules/no-typos.md
```diff
@@ -1,18 +1,31 @@
...
 
+and the following react lifecycle methods:
+
+* componentWillMount
+* componentDidMount
+* componentWillReceiveProps
+* shouldComponentUpdate
+* componentWillUpdate
+* componentDidUpdate
+* componentWillUnmount
```
jseminck: What about render? Though React already warns about this when running the code, but still having the error show early is nice.

```
No `render` method found on the returned component instance: you may have forgotten to define `render`.
```

lib/rules/no-typos.js
haridusenadeera: Ahh, I misunderstood the purpose of that util function. I guess we do not need to check for that then.

docs/rules/no-typos.md
haridusenadeera: Yeah I agree. I will add `render` method

haridusenadeera:
Thanks for the quick review! I've made changes according to your comments.

jseminck:
Looks very good to me. Thanks a lot for the contribution! 🎉 

I hope we can get some review from the maintainers as well. It would be great if we could release this together with the initial `no-typos` rule.

jseminck:
@haridusenadeera it seems some conflict was created after changes in master. Could you please rebase your changes and fix it? Thanks!

@ljharb could you please assign some of the core contributors as reviewer for this PR? I think it would be great if it can be considered for the next release. Thanks!

PR #1302, opened 2017-07-13
Fix false positives in no-literals

### LINKED ISSUE #1301: no-literals: only check strings
should the no-literals rule check if the `node.value` is a string? with the new `{ "noStrings": true }` option (https://github.com/yannickcr/eslint-plugin-react/pull/1202) I'm getting errors for explicitly set boolean and number props `booleanprop={true}` `numberprop={100}`

davidyorr:
This rule should only check strings.

Fixes #1301

### DIFF
--- a/lib/rules/jsx-no-literals.js
+++ b/lib/rules/jsx-no-literals.js
@@ -44,6 +44,7 @@ module.exports = {
 
     function getValidation(node) {
       const standard = !/^[\s]+$/.test(node.value) &&
+          typeof node.value === 'string' &&
           node.parent &&
           node.parent.type.indexOf('JSX') !== -1 &&
           node.parent.type !== 'JSXAttribute';
--- a/lib/rules/jsx-no-literals.js
+++ b/lib/rules/jsx-no-literals.js
@@ -64,6 +64,12 @@ module.exports = {
         if (getValidation(node)) {
           reportLiteralNode(node);
         }
+      },
+
+      TemplateLiteral: function(node) {
+        if (isNoStrings && node.parent.type === 'JSXExpressionContainer') {
+          reportLiteralNode(node);
+        }
       }
 
     };
--- a/tests/lib/rules/jsx-no-literals.js
+++ b/tests/lib/rules/jsx-no-literals.js
@@ -127,6 +127,18 @@ ruleTester.run('jsx-no-literals', rule, {
       parser: 'babel-eslint',
       options: [{noStrings: true}]
     }, {
+      code: '<Foo bar={true} />',
+      parser: 'babel-eslint',
+      options: [{noStrings: true}]
+    }, {
+      code: '<Foo bar={false} />',
+      parser: 'babel-eslint',
+      options: [{noStrings: true}]
+    }, {
+      code: '<Foo bar={100} />',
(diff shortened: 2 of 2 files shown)


lib/rules/jsx-no-literals.js
```diff
@@ -44,6 +44,7 @@ module.exports = {
 
     function getValidation(node) {
       const standard = !/^[\s]+$/.test(node.value) &&
+          typeof node.value === 'string' &&
```
ljharb: hm, i'm kind of confused - what's the typeof `node.value` if it's an object, or a function? Can you add some test cases for that?

davidyorr: it looks like the `Literal` visitor doesn't hit the node when it's an object or function.

davidyorr:
as is, template strings are not causing linting errors, which they should be. i can look into that as well

lib/rules/jsx-no-literals.js
ljharb: So when would the typeof not be "string"?

davidyorr:
```javascript
<Foo
  a={true}
  b={1}
  c={null} />
```
all of those props get hit, so typeof could be `boolean`, `number`, and when null it's an `object`

ljharb: Interesting. Arrays and objects and regexes and functions are all literals too, so it's weird they wouldn't show up. Thanks for explaining.

davidyorr: yeah i'm not sure why

davidyorr:
updated to check for template literals too

tests/lib/rules/jsx-no-literals.js
```diff
@@ -127,6 +127,48 @@ ruleTester.run('jsx-no-literals', rule, {
...
+      code: '<Foo bar={100} />',
+      parser: 'babel-eslint',
+      options: [{noStrings: true}]
+    }, {
+      code: '<Foo bar={null} />',
+      parser: 'babel-eslint',
+      options: [{noStrings: true}]
+    }, {
+      code: '<Foo bar={{}} />',
+      parser: 'babel-eslint',
```
ljharb: babel-eslint isn't needed for any of these new tests; let's please prefer the default parser whenever possible.

davidyorr: babel-eslint isn't needed for any of the tests in this entire PR, or just the last commit? how can I tell which tests need babel-eslint?

ljharb: tl;dr: if they don't generate a syntax error with the default parser, they don't need babel-eslint :-)

Generally tho, the default parser supports everything that's actually in JS. Things that aren't in JS (or aren't yet), like class properties, require babel-eslint.

davidyorr: ah okay i see thanks! i'll remove those tests

ljharb: keeping the tests is great! i just want them using the default parser :-)

davidyorr: alright deleted the babel parser tests and kept the default parser ones

Issue #1162, opened 2017-04-24
no-unused-prop-types does not trigger if class does not use any properties


Wilfred:
```js
import React from "react";

var Hello = React.createClass({
  propTypes: {
    name: React.PropTypes.string
  },
  render: function() {
    return <div>Hello Bob</div>;
  }
});
```

This is almost identical to [the first example in the docs](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unused-prop-types.md).

```bash
$ eslint --no-eslintrc --parser-options="{sourceType: \"module\", ecmaVersion: 6, ecmaFeatures: {jsx:true}}" --plugin react --rule react/no-unused-prop-types:error js/Hello.jsx
# No errors shown.
```

By contrast, this does trigger:

```js
import React from "react";

var Hello = React.createClass({
  propTypes: {
    name: React.PropTypes.string
  },
  render: function() {
    return <div>Hello Bob {this.props.foo}</div>;
  }
});
```

```bash
$ eslint --no-eslintrc --parser-options="{sourceType: \"module\", ecmaVersion: 6, ecmaFeatures: {jsx:true}}" --plugin react --rule react/no-unused-prop-types:error js/Hello.jsx

js/Hello.jsx
  5:11  error  'name' PropType is defined but prop is never used  react/no-unused-prop-types

✖ 1 problem (1 error, 0 warnings)

ljharb:
Both React.createClass and React.PropTypes are officially deprecated as of v15.5; but we should still fix this case.

DianaSuvorova:
I will work on it.

DianaSuvorova:
@ljharb , do you usually close issues after fix was merged to master or released or you follow some other process?

ljharb:
Ah, when the PR says `fixes #x` github does it for me :-)' (on merge)

Fixed in #1303.

DianaSuvorova:
Got it. Will follow that. Thanks!

PR #1303, opened 2017-07-14
no-unused-props: not ignoring components with no used props

DianaSuvorova:
This is tackling one of the many issues with the rule (issue #1162).

The problem was that the rule simply ignored the components that had no used properties. Once this condition was removed I had to fix some tests and remove some other (that were never actually worked correctly but were passing because of the component had no used props and thus was excluded from validation).

This change might fix few more false positives that users reported. I am planning to go through the list of open issues for the rule and close/fix some more after I get some feedback on my change.

### DIFF
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -175,7 +175,6 @@ module.exports = {
     function mustBeValidated(component) {
       return Boolean(
         component &&
-        component.usedPropTypes &&
         !component.ignorePropsValidation
       );
     }
@@ -234,8 +233,9 @@ module.exports = {
      * @returns {Boolean} True if the prop is used, false if not.
      */
     function isPropUsed(node, prop) {
-      for (let i = 0, l = node.usedPropTypes.length; i < l; i++) {
-        const usedProp = node.usedPropTypes[i];
+      const usedPropTypes = node.usedPropTypes || [];
+      for (let i = 0, l = usedPropTypes.length; i < l; i++) {
+        const usedProp = usedPropTypes[i];
         if (
           prop.type === 'shape' ||
           prop.name === '__ANY_KEY__' ||
@@ -895,6 +895,13 @@ module.exports = {
           default:
             break;
         }
+      },
+
+      JSXSpreadAttribute: function(node) {
+        const component = components.get(utils.getParentComponent());
+        components.set(component ? component.node : node, {
+          ignorePropsValidation: true
+        });
       },
 
       MethodDefinition: function(node) {
--- a/tests/lib/rules/no-unused-prop-types.js
+++ b/tests/lib/rules/no-unused-prop-types.js
@@ -559,26 +559,6 @@ ruleTester.run('no-unused-prop-types', rule, {
       options: [{customValidators: ['CustomValidator']}]
     }, {
       code: [
-        'class Comp1 extends Component {',
-        '  render() {',
-        '    return <span />;',
-        '  }',
-        '}',
-        'Comp1.propTypes = {',
-        '  prop1: PropTypes.number',
-        '};',
-        'class Comp2 extends Component {',
-        '  render() {',
-        '    return <span />;',
-        '  }',
-        '}',
-        'Comp2.propTypes = {',
-        '  prop2: PropTypes.arrayOf(Comp1.propTypes.prop1)',
-        '};'
-      ].join('\n'),
-      parser: 'babel-eslint'
-    }, {
-      code: [
         'const SomeComponent = createReactClass({',
         '  propTypes: SomeOtherComponent.propTypes',
         '});'
@@ -781,54 +761,6 @@ ruleTester.run('no-unused-prop-types', rule, {
         '    );',
         '  }',
         '});'
-      ].join('\n')
-    }, {
-      code: [
-        'const statelessComponent = (props) => {',
-        '  const subRender = () => {',
-        '    return <span>{props.someProp}</span>;',
-        '  };',
-        '  return <div>{subRender()}</div>;',
-        '};',
-        'statelessComponent.propTypes = {',
-        '  someProp: PropTypes.string',
-        '};'
-      ].join('\n')
-    }, {
-      code: [
-        'const statelessComponent = ({ someProp }) => {',
-        '  const subRender = () => {',
-        '    return <span>{someProp}</span>;',
-        '  };',
-        '  return <div>{subRender()}</div>;',
-        '};',
-        'statelessComponent.propTypes = {',
-        '  someProp: PropTypes.string',
-        '};'
-      ].join('\n')
-    }, {
-      code: [
-        'const statelessComponent = function({ someProp }) {',
-        '  const subRender = () => {',
-        '    return <span>{someProp}</span>;',
-        '  };',
-        '  return <div>{subRender()}</div>;',
-        '};',
-        'statelessComponent.propTypes = {',
-        '  someProp: PropTypes.string',
-        '};'
-      ].join('\n')
-    }, {
-      code: [
-        'function statelessComponent({ someProp }) {',
-        '  const subRender = () => {',
-        '    return <span>{someProp}</span>;',
-        '  };',
-        '  return <div>{subRender()}</div>;',
-        '};',
-        'statelessComponent.propTypes = {',
-        '  someProp: PropTypes.string',
-        '};'
       ].join('\n')
     }, {
       code: [
@@ -3214,6 +3146,21 @@ ruleTester.run('no-unused-prop-types', rule, {
         message: '\'foo\' PropType is defined but prop is never used',
         line: 11,
         column: 8
+      }]
+    }, { // None of the props are used issue #1162
+      code: [
+        'import React from "react"; ',
+        'var Hello = React.createReactClass({',
+        ' propTypes: {',
+        '   name: React.PropTypes.string',
+        ' },',
+        ' render: function() {',
+        '   return <div>Hello Bob</div>;',
+        '  }',
+        '});'
+      ].join('\n'),
+      errors: [{
+        message: '\'name\' PropType is defined but prop is never used'
       }]
     }
     /* , {
--- a/tests/lib/rules/no-unused-prop-types.js
+++ b/tests/lib/rules/no-unused-prop-types.js
@@ -3162,6 +3162,34 @@ ruleTester.run('no-unused-prop-types', rule, {
       errors: [{
         message: '\'name\' PropType is defined but prop is never used'
       }]
+    }, {
+      code: [
+        'class Comp1 extends Component {',
+        '  render() {',
+        '    return <span />;',
+        '  }',
+        '}',
+        'Comp1.propTypes = {',
+        '  prop1: PropTypes.number',
+        '};',
+        'class Comp2 extends Component {',
+        '  render() {',
+        '    return <span />;',
+        '  }',
+        '}',
+        'Comp2.propTypes = {',
+        '  prop2: PropTypes.arrayOf(Comp1.propTypes.prop1)',
+        '};'
+      ].join('\n'),
+      parser: 'babel-eslint',
+      errors: [{
+        message: '\'prop1\' PropType is defined but prop is never used'
+      }, {
+        message: '\'prop2\' PropType is defined but prop is never used'
+      }, {
+        message: '\'prop2.*\' PropType is defined but prop is never used'
+      }]
+
     }
     /* , {
       // Enable this when the following issue is fixed


tests/lib/rules/no-unused-prop-types.js
```diff
@@ -784,54 +764,6 @@ ruleTester.run('no-unused-prop-types', rule, {
       ].join('\n')
     }, {
       code: [
-        'const statelessComponent = (props) => {',
```
DianaSuvorova: First of all for some reason the same test were copy-pasted  4 times. I had to remove this test all together because it was never validated correctly. Current implementation is not able to detect correctly nested definition of components (subRender technically is a react component). I don't think this is a common/idiomatic react style so I assume we shouldn't support it.

tests/lib/rules/no-unused-prop-types.js
```diff
@@ -3215,6 +3148,23 @@ ruleTester.run('no-unused-prop-types', rule, {
         line: 11,
         column: 8
       }]
+    }, { // None of the props are used
+      code: [
```
DianaSuvorova: this is the test reported fro the issue

lib/util/Components.js
```diff
@@ -114,8 +114,12 @@ Components.prototype.list = function() {
     }
     const id = this._getId(this._list[j].node);
     list[j] = this._list[j];
+    if (!list[j].usedPropTypes) {
+      list[j].usedPropTypes = [];
+    }
+
     if (usedPropTypes[id]) {
-      list[j].usedPropTypes = (list[j].usedPropTypes || []).concat(usedPropTypes[id]);
```
ljharb: this change doesn't seem like an improvement; it increases the number of mutations.

tests/lib/rules/no-unused-prop-types.js
```diff
@@ -3215,6 +3148,23 @@ ruleTester.run('no-unused-prop-types', rule, {
...
+        'var Hello = createReactClass({',
+        '  propTypes: {',
+        '    unused: PropTypes.string',
```
ljharb: The test case specifically had `React.PropTypes`, and `React.createClass` - are these not necessary to trigger the bug?

DianaSuvorova: This didn't matter. But I updated the test so it is identical with the one reported.

lib/util/Components.js
DianaSuvorova: Yeah, I agree it is not exactly an improvement. 
Reason I added it - so usedPropTypes it is an empty array when there is no usedPropTypes instead of null. Otherwise anywhere we loop through it we would need to check if it is an array or not.
I can remove it and add few checks in the rule file - I don't mind either way. Let me know what is preferable.

ljharb: I see what you mean. Could you point to the other lines you'd need an `|| []` on, if you reverted these lines?

tests/lib/rules/no-unused-prop-types.js
```diff
@@ -1387,15 +1319,16 @@ ruleTester.run('no-unused-prop-types', rule, {
       code: [
         'const PagingBlock = function(props) {',
         '  return (',
-        '    <SomeChild {...props} />',
+        '    <span>{...props}</span>',
```
ljharb: Why this change?

tests/lib/rules/no-unused-prop-types.js
```diff
@@ -1387,15 +1319,16 @@ ruleTester.run('no-unused-prop-types', rule, {
...
         ' );',
         '};',
 
         'PagingBlock.propTypes = {',
         '  nextPage: PropTypes.func.isRequired,',
         '  previousPage: PropTypes.func.isRequired,',
         '};'
-      ].join('\n')
+      ].join('\n'),
+      parser: 'babel-eslint'
```
ljharb: what in this test needs babel-eslint?

tests/lib/rules/no-unused-prop-types.js
```diff
@@ -3215,6 +3148,22 @@ ruleTester.run('no-unused-prop-types', rule, {
...
+        'var Hello = React.createReactClass({',
+        ' propTypes: {',
+        '   name: React.PropTypes.string',
+        ' },',
+        ' render: function() {',
+        '   return <div>Hello Bob</div>;',
+        '  }',
+        '});'
+      ].join('\n'),
+      parser: 'babel-eslint',
```
ljharb: similarly here, what in this test needs babel-eslint? this test seems like it'd work with ES6.

tests/lib/rules/no-unused-prop-types.js
DianaSuvorova: this component uses JSX spread  few lines above.So it needs babel.

tests/lib/rules/no-unused-prop-types.js
DianaSuvorova: yes, here es6 would do.

tests/lib/rules/no-unused-prop-types.js
DianaSuvorova: to detect usage of a spread operator (that is handled in JSXSpreadChild I added to the rule) SomeChild has to be a valid react component (this is a babel-eslint thing). Originally when component was invalid, babel didn't detect any props passed to it.

lib/util/Components.js
DianaSuvorova: Just one place actually: 
lib/rules/no-unused-prop-types.js:236
`
      for (let i = 0, l = node.usedPropTypes.length; i < l; i++) {
        const usedProp = node.usedPropTypes[i];
        if (
...
`

tests/lib/rules/no-unused-prop-types.js
ljharb: jsx spread works in ES6 or above in the default parser with `parserOptions: { jsx: true }` - there's definitely no need for `babel-eslint` for it.

tests/lib/rules/no-unused-prop-types.js
ljharb: There's no need to use babel-eslint for spread props; so let's revert this change and keep it on the default parser.

DianaSuvorova: as far as I can tell spread works with es6 but jsx spread inside a component does not. Simply removing babel and adding parserOptions fails with : `  message: 'Parsing error: Unexpected token ...',` 

I am happy to try any suggestions you might have.
```js 
  {
    code: [
      'const PagingBlock = function(props) {',
      '  return (',
      '    <span>{...props}</span>',
      ' );',
      '};',

      'PagingBlock.propTypes = {',
      '  nextPage: PropTypes.func.isRequired,',
      '  previousPage: PropTypes.func.isRequired,',
      '};'
    ].join('\n'),
    parserOptions: Object.assign({}, parserOptions, {ecmaVersion: 2017})
    }
```
error:
```js
    AssertionError: Should have no errors but had 1: [ { ruleId: null,
    fatal: true,
    severity: 2,
    source: '    <span>{...props}</span>',
    message: 'Parsing error: Unexpected token ...',
    line: 3,
    column: 12 } ]
      + expected - actual

      -1
      +0
      
      at testValidTemplate (node_modules/eslint/lib/testers/rule-tester.js:415:20)
      at Context.RuleTester.it (node_modules/eslint/lib/testers/rule-tester.js:545:25)

```

tests/lib/rules/no-unused-prop-types.js
DianaSuvorova: responded above

tests/lib/rules/no-unused-prop-types.js
ljharb: I'm a bit confused… `{...props}` in a child context isn't anything; that shouldn't work. Spreading is a JSX attribute thing, and a JS iterable thing, and a JS object thing - but not a child thing.

DianaSuvorova: I guess I have to add that the reason it worked before is that the component was invalid by itself so the parser wouldn't try to parse the props of `SomeChild`. And that way the linter didn't know that props are actually used. That was ok before this change (where components with 0 used props were simply not validated). If we want to validate components with no props used, we need to make sure that where they are actually used, the code is syntactically valid. 

In other words if I keep this test as it is it would fail with `nextPage` and `PreviousPage` are not used error.

DianaSuvorova: > I'm a bit confused… {...props} in a child context isn't anything; that shouldn't work. Spreading is a JSX attribute thing, and a JS iterable thing, and a JS object thing - but not a child thing.

I guess it is getting late...  I am not sure I understand  what you mean .. But I agree regrdless😃

ljharb: `<span>{...props}</span>` is invalid jsx - if babel-eslint parses it, it's a bug in babel-eslint.

DianaSuvorova: Oh I see, good point. Let me update the test so `SomeChild` is a thing then. I will look into creating the issue with babel-eslint.

DianaSuvorova:
ok, it will take some more time. Neither babel nor es6 parser gives any indication that {...props} is being used. And it has nothing to do with component being a thing (thanks for catching that issue with invalid spread for babel ).

So this

```js
 return <SomeComponent {...props} />;
```
is parsed into:

image
note 0 children in arguments. If I use props without spread, I see children there.

What dependency do you think this problem starts from? Would it be an eslint/espree issue?

tests/lib/rules/no-unused-prop-types.js
```diff
@@ -1317,9 +1318,16 @@ ruleTester.run('no-unused-prop-types', rule, {
     }, {
       // `no-unused-prop-types` rest param props in jsx expressions - [Issue #885]
       code: [
+        'const SomeComponent = funciton(props) {',
```
ljharb: `function`

lib/util/Components.js
ljharb: I'm ambivalent; my general preference is to avoid mutations, but I don't feel super strongly about it. I'll leave it up to you :-)

tests/lib/rules/no-unused-prop-types.js
```diff
@@ -3148,24 +3148,7 @@ ruleTester.run('no-unused-prop-types', rule, {
         line: 11,
         column: 8
       }]
-    }, { // None of the props are used
-      code: [
-        'var Hello = createReactClass({',
```
ljharb: In this case, the duplicated test isn't a duplicate; one imports React, and the other doesn't, and it's intentional to check both.

tests/lib/rules/no-unused-prop-types.js
ljharb: It's not the same test; I see 4 different variations: arrow, non-arrow, and destructured, and non-destructured.

tests/lib/rules/no-unused-prop-types.js
DianaSuvorova: Ok I will bring this back.

tests/lib/rules/no-unused-prop-types.js
DianaSuvorova: will be gone. Thanks.

tests/lib/rules/no-unused-prop-types.js
DianaSuvorova: Got it. Makes sense. Is ti ok if I delete these tests though. Do you agree that this rule technically doesn't support this use case.

tests/lib/rules/no-unused-prop-types.js
DianaSuvorova: The fix was easy. Changing type from JSXChildSpread to JSXAttribute spread...Thanks for catching this. I will go read about JSXChildSpread though... maybe it is some beta feature...

lib/rules/no-unused-prop-types.js
```diff
@@ -234,8 +233,9 @@ module.exports = {
      * @returns {Boolean} True if the prop is used, false if not.
      */
     function isPropUsed(node, prop) {
-      for (let i = 0, l = node.usedPropTypes.length; i < l; i++) {
-        const usedProp = node.usedPropTypes[i];
+      const usedPropTypes = node.usedPropTypes || [];
```
DianaSuvorova: I reverted that change from Components file.

DianaSuvorova:
@ljharb sorry for so many updates. I squashed my change into 1 commit and it is ready for re-review. Thanks!

lib/rules/no-unused-prop-types.js
```diff
@@ -897,6 +897,13 @@ module.exports = {
         }
       },
 
+      JSXSpreadAttribute: function(node) {
+        const component = components.get(utils.getParentComponent());
```
ljharb: It seems super weird to me that `getParentComponent` infers the `node` rather than taking it as an argument (not your code ofc, just thought I'd comment)

tests/lib/rules/no-unused-prop-types.js
```diff
@@ -559,26 +559,6 @@ ruleTester.run('no-unused-prop-types', rule, {
...
-        'Comp1.propTypes = {',
-        '  prop1: PropTypes.number',
-        '};',
-        'class Comp2 extends Component {',
-        '  render() {',
-        '    return <span />;',
-        '  }',
-        '}',
-        'Comp2.propTypes = {',
-        '  prop2: PropTypes.arrayOf(Comp1.propTypes.prop1)',
```
ljharb: This test implies that Comp1's prop1 prop should be marked as used, because Comp2 references its propType; but in fact I think it shouldn't - could you move this test to be *invalid* instead of just deleting it?

tests/lib/rules/no-unused-prop-types.js
ljharb: I think that it ideally should support this use case, but our mechanism of detecting components may make it difficult.

tests/lib/rules/no-unused-prop-types.js
DianaSuvorova: Done.

lib/rules/no-unused-prop-types.js
DianaSuvorova: yeap

DianaSuvorova:
@lencioni, when would you be able to take a look at this PR?  This is my first PR for the repo so I am not sure what is the normal turn around here.

Thank you,
Diana

ljharb:
I'll give it another day, then I'll merge it if there's no objections by then.

tests/lib/rules/no-unused-prop-types.js
sompylasar: If you ask me, the fact that certain cases are not supported could also be tested. Of course, with a test name which clearly states "testing unsupported case" and assertions that clearly check for false positives or other signs of "not supported"-ness.

ljharb: sure, that seems reasonable too.

ljharb:
Sorry for the inconvenience; I've just merged a refactor to start using arrow functions. Would you mind rebasing latest master? I'll merge ASAP after that's done.

DianaSuvorova:
@ljharb, no problem. Should be all set.

PR #1304, opened 2017-07-15
Fix incorrect propType warning inside .map

dustinsoftware:
Opening this PR for discussion. This PR fixes this false error:

 'one' PropType is defined but prop is never used

Caused by this test case:
```
const Inner = (props) => <span>{props.innerOne} {props.innerTwo}</span>;
const Outer = (props) => {
    let team = props.names.map(() => (
        <Inner innerOne={props.one} innerTwo={props.two} />
    ));
    return <ul>{team}</ul>;
};
Outer.propTypes = {
    names: PropTypes.array,
    one: PropTypes.string,
    two: PropTypes.string
};
```

However, looks like my fix caused a few other tests to fail. I intend to push an update once I do some more investigation.

### DIFF
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -617,7 +617,7 @@ module.exports = {
           throw new Error(`${node.type} ASTNodes are not handled by markPropTypesAsUsed`);
       }
 
-      const component = components.get(utils.getParentComponent());
+      const component = components.get(utils.getParentComponent() || node, true);
       const usedPropTypes = component && component.usedPropTypes || [];
       let ignorePropsValidation = component && component.ignorePropsValidation || false;
 
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -49,11 +49,14 @@ Components.prototype.add = function(node, confidence) {
  * Find a component in the list using its node
  *
  * @param {ASTNode} node The AST node being searched.
+ * @param {Boolean} findParent True if the node's parent can be returned if the node isn't found, false if not
  * @returns {Object} Component object, undefined if the component is not found
  */
-Components.prototype.get = function(node) {
-  const id = this._getId(node);
-  return this._list[id];
+Components.prototype.get = function(node, findParent) {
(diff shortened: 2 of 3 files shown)


lib/rules/no-unused-prop-types.js
```diff
@@ -234,8 +234,7 @@ module.exports = {
      * @returns {Boolean} True if the prop is used, false if not.
      */
     function isPropUsed(node, prop) {
-      for (let i = 0, l = node.usedPropTypes.length; i < l; i++) {
-        const usedProp = node.usedPropTypes[i];
+      for (const usedProp of node.usedPropTypes) {
```
ljharb: I'd prefer to avoid `for..of` in favor of `.forEach` or `.map`

tests/lib/rules/no-unused-prop-types.js
```diff
@@ -38,6 +38,22 @@ ruleTester.run('no-unused-prop-types', rule, {
...
+        '    ));',
+        '  return <ul>{team}</ul>;',
+        '};',
+        'Outer.propTypes = {',
+        '  names: PropTypes.array,',
+        '  one: PropTypes.string,',
+        '  two: PropTypes.string',
+        '};'
+      ].join('\n'),
+      parser: 'babel-eslint'
```
ljharb: nothing about this test case should require babel-eslint; can it use the default parser instead?

lib/util/Components.js
```diff
@@ -52,8 +52,10 @@ Components.prototype.add = function(node, confidence) {
  * @returns {Object} Component object, undefined if the component is not found
  */
 Components.prototype.get = function(node) {
```
ljharb: Are there any tests for this change, btw?

dustinsoftware: I didn't see a good place to put a test for this specific change, and the logic behind `get` changes, the test I did add will fail.

dustinsoftware:
Updated

dustinsoftware:
@DianaSuvorova had a better fix in #1322 😄 This PR isn't needed anymore.

PR #1313, opened 2017-07-22
Add detection of subRender methods to prefer-stateless-function

jseminck:
Related to: https://github.com/yannickcr/eslint-plugin-react/issues/1235 - though the issue wasn't yet labelled with new rule. If we want it, it's available.

### DIFF
--- a/docs/rules/prefer-stateless-function.md
+++ b/docs/rules/prefer-stateless-function.md
@@ -11,6 +11,7 @@ This rule will check your class based React components for
 * extension of `React.PureComponent` (if the `ignorePureComponents` flag is true)
 * presence of `ref` attribute in JSX
 * the use of decorators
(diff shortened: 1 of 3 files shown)


jseminck:
Could you explain with an example? Do you mean this?

```js
render() {
  const mySubRenderFn = () => <div>{this.state.foo}</div>
  
  return (
    <div>{mySubRenderFn()}</div>
  )
}
```

ljharb:
```jsx
function shouldBeComponent() { // this should warn, because it should be a component
  return <div />;
}

function SFC(props) {
  return <div>{shouldBeComponent(props.foo)}</div>;
}
SFC.propTypes = {
  foo: PropTypes.any
};
```

ljharb:
(Related to #1229)

Let's rebase this and make sure it's still valid?

Issue #1309, opened 2017-07-20
no-unused-props unable to handle inline functions


Robinfr:
When you do the following, no-unused-props is triggered even though the props are used:

```js
const Thing = (props) => (
    <div>
        {(() => {
            if(props.enabled){
                return (
                    <span>Enabled!</span>
                )
            }
            
            return (
                <span>Disabled..</span>
            )
        })()}
    </div>
);

Thing.propTypes = {
    enabled: React.PropTypes.bool
};

Thing.defaultProps = {
    enabled: false
};

export default Thing;
```

ljharb:
It's difficult to catch every case.

In the meantime, I'd suggest always destructuring all props and state and context at the top of your functions, which has the side benefit of bypassing this problem.

Robinfr:
Hmm alright I will see if that helps. Thanks for the reply.

DianaSuvorova:
@Robinfr It looks like your case works with master as well as with v7.1.0. Which version are you using?

ljharb:
Closed in #1318.

Robinfr:
I'm using 7.1.0.. so this is a bit curious :confused:

Robinfr:
Updated example:

```js
import React from 'react';

const Thing = (props) => (
    <div>
        <div>
            {(() => {
                if (props.enabled && props.testString) {
                    return (
                        <span>Enabled!</span>
                    );
                }

                return (
                    <span>Disabled..</span>
                );
            })()}
        </div>
    </div>
);

Thing.propTypes = {
    enabled: React.PropTypes.bool,
    testString: React.PropTypes.string
};

Thing.defaultProps = {
    enabled: false
};

export default Thing;

```

In this case, you have two props, `enabled` and `testString`. You will get the following error when linting this:

```bash
'enabled' PropType is defined but prop is never used  react/no-unused-prop-types
```

DianaSuvorova:
@Robinfr, thanks for the updated example. It did uncover a bug in the rule.  I created a PR to fix it.  Stay tuned for the next release!

PR #1322, opened 2017-07-24
Issue #1309

DianaSuvorova:
Fixes issue #1309

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -10,6 +10,30 @@ const doctrine = require('doctrine');
 const variableUtil = require('./variable');
 const pragmaUtil = require('./pragma');
 
+const usedPropTypesAreEquivalent = (propA, propB) => {
+  if (propA.name === propB.name) {
+    if (!propA.allNames && !propB.allNames) {
+      return true;
+    } else if (propA.allNames.join('') === propB.allNames.join('')) {
+      return true;
+    }
+    return false;
+  }
+  return false;
+};
+
+const mergeUsedPropTypes = (propsList, newPropsList) => {
+  const propsToAdd = [];
+  newPropsList.forEach(newProp => {
+    const newPropisAlreadyInTheList = propsList.some(prop => usedPropTypesAreEquivalent(prop, newProp));
+    if (!newPropisAlreadyInTheList) {
+      propsToAdd.push(newProp);
+    }
+  });
+  return propsList.concat(propsToAdd);
+};
+
+
 /**
  * Components
  * @class
@@ -70,7 +94,16 @@ Components.prototype.set = function(node, props) {
     return;
   }
   const id = this._getId(node);
+  let copyUsedPropTypes;
+  if (this._list[id]) {
+    // usedPropTypes is an array. _extend replaces existing array with a new one which caused issue #1309.
+    // preserving original array so it can be merged later on.
+    copyUsedPropTypes = this._list[id].usedPropTypes && this._list[id].usedPropTypes.slice();
+  }
   this._list[id] = util._extend(this._list[id], props);
+  if (this._list[id] && props.usedPropTypes) {
+    this._list[id].usedPropTypes = mergeUsedPropTypes(copyUsedPropTypes || [], props.usedPropTypes);
+  }
 };
 
 /**
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -14,7 +14,7 @@ const usedPropTypesAreEquivalent = (propA, propB) => {
   if (propA.name === propB.name) {
     if (!propA.allNames && !propB.allNames) {
       return true;
-    } else if (propA.allNames.join('') === propB.allNames.join('')) {
+    } else if (Array.isArray(propA.allNames) && Array.isArray(propB.allNames) && propA.allNames.join('') === propB.allNames.join('')) {
       return true;
     }
     return false;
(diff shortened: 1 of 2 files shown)


DianaSuvorova:
@ljharb , I am done updating this PR. Thanks 😄

lib/util/Components.js
```diff
@@ -63,14 +63,50 @@ Components.prototype.get = function(node) {
  * @param {Object} props Additional properties to add to the component.
  */
 Components.prototype.set = function(node, props) {
+  const popsAreEquivalent = (propA, propB) => {
```
ljharb: s/pops/props

can this function be defined at file level, so it doesn't need to be recreated every time `set` is run?

ljharb:
(also it'd be great to have the PR title explain the change being made, and confine the issue reference to the OP description)

lib/util/Components.js
```diff
@@ -63,14 +63,50 @@ Components.prototype.get = function(node) {
...
+      return false;
+    }
+    return false;
+  };
+
+  const merge = (propsList, newPropList) => {
+    const propsToAdd = [];
+    newPropList.forEach(newProp => {
+      let newPropisAlreadyInTheList = false;
+      propsList.some(prop => {
```
jseminck: The use of `some` here looks strange. `some()` returns a boolean but it is not used.

Something like this should work too ( didn't try it though ):
`const newPropIsAlreadyInTheList = propsList.some(prop => popsAreEquivalent(prop, newProp)`

I think the `I` in `newPropisAlreadyInTheList` should be capitalised.

jseminck: WDYT of using reduce to write this function?

```
  return newPropsList.reduce((propsList, newProp) => {
    if (propsList.some(prop => popsAreEquivalent(prop, newProp)) {
      return propsList;
    }
    propsList.push(newProp);
    return propsList;
  }, propsList)
```

lib/util/Components.js
```diff
@@ -63,14 +63,50 @@ Components.prototype.get = function(node) {
...
+    if (propA.name === propB.name) {
+      if (!propA.allNames && !propB.allNames) {
+        return true;
+      } else if (propA.allNames.join('') === propB.allNames.join('')) {
+        return true;
+      }
+      return false;
+    }
+    return false;
+  };
```
jseminck: I personally find this nested if if else if quite hard to read. It can just be:

```js
return (
  propsA.name === propsB.name &&
  ((!propA.allNames && !propB.allNames) || (propA.allNames.join('') === propB.allNames.join(''))
)
```

Although it's not that simple either since it still has a lot of conditions, I find it a bit easier...: `conditionA && ((conditionD && conditionC) || conditionE)`

lib/util/Components.js
```diff
@@ -63,14 +63,50 @@ Components.prototype.get = function(node) {
...
   while (node && !this._list[this._getId(node)]) {
     node = node.parent;
   }
   if (!node) {
     return;
   }
   const id = this._getId(node);
+  let copyUsedPropTypes;
+  if (this._list[id]) {
+    copyUsedPropTypes = this._list[id].usedPropTypes && this._list[id].usedPropTypes.slice();
```
jseminck: Is it possible that `copyUsedPropTypes` is going to be assigned a value but that it will not be used?

The condition for assigning the value is `if (this._list[id])`
The condition for using the value is: `if (this._list[id] && props.usedPropTypes)`

lib/util/Components.js
```diff
@@ -10,6 +10,30 @@ const doctrine = require('doctrine');
...
+    }
+    return false;
+  }
+  return false;
+};
+
+const mergeUsedPropTypes = (propsList, newPropsList) => {
+  const propsToAdd = [];
+  newPropsList.forEach(newProp => {
+    const newPropisAlreadyInTheList = propsList.some(prop => usedPropTypesAreEquivalent(prop, newProp));
```
DianaSuvorova: @jseminck , thanks for the review.  You are correct about `some` function. Updated it to be more concise. As for `reduce` I have only used it for very simple logic. If you can rewrite this fnc, I will be happy  to try it and copy paste your code if it works :)

lib/util/Components.js
```diff
@@ -70,7 +94,16 @@ Components.prototype.set = function(node, props) {
     return;
   }
   const id = this._getId(node);
+  let copyUsedPropTypes;
+  if (this._list[id]) {
+    // usedPropTypes is an array. _extend replaces existing array with a new one which caused issue #1309.
```
DianaSuvorova: @ljharb added a comment here with some explanation for the change. LMK if there is a better place for it.

lib/util/Components.js
DianaSuvorova: it didn't happen yet. But it is a fair concern.  I will update.

lib/util/Components.js
DianaSuvorova: yeah, it is possible.

lib/util/Components.js
DianaSuvorova: Done 😄

npm-debug.log.957467166
```diff
@@ -0,0 +1,22 @@
+0 info it worked if it ends with ok
```
jseminck: This file was added by accident.

lib/util/Components.js
jseminck: I think this approach is good as well.

npm-debug.log.957467166
DianaSuvorova: Gone. Thanks!

DianaSuvorova:
@ljharb should be all set. thanks :)

Issue #1268, opened 2017-06-23
no-unused-prop-types false positive in functions that can return JSX in stateless components


jaromero:
eslint version 3.19.0
eslint-plugin-react version 7.1.0

This minimal example causes eslint to incorrectly flag `props.append` as never used:

~~~js
import React from 'react';
import PropTypes from 'prop-types';

export default function SampleComp(props) {
  function buildText(text) {
    return (
      <span>
        {text}
        {props.append}
      </span>
    );
  }

  return (
    <div>
      {buildText(props.text)}
    </div>
  );
}

SampleComp.propTypes = {
  text: PropTypes.string,
  append: PropTypes.string
};
~~~

I've noticed that the following things don't trigger the false positive:

* Calling `buildText` with a literal, e.g. `buildText('hello')`
* Calling `buildText` with a variable that doesn't ultimately come from props: `const hello = 'hello'; buildText(hello);`
* Assigning `props.append` to a variable outside of `buildText`, and using that variable instead inside the function body: `const {append} = props; function buildText(text) { /* use append here instead of props.append */ }; buildText(props.text);`
* Refactoring `buildText` to take and use the whole `props` object, and adjusting the call accordingly
* Refactoring `buildText` to **not** take any arguments and use the props directly inside the function body
  * In this case, even if you don't use _any of the props at all_, eslint doesn't complain... a bug in the other direction?
* Refactoring `buildText` to take multiple arguments and passing all prop to it accordingly: `function buildText(text, append) { /* ... */ }; buildText(props.text, props.append);`
* And obviously, returning something other than JSX from `buildText`

The following things don't affect the outcome at all:

* Assigning `props.append` to a variable inside `buildText` doesn't get rid of the eslint warning
* Using the prop in a flow control statement (`if`, `switch`, etc), even if reassigned as above
  * Not even if the statement that returns JSX is completely unreachable, e.g. `if (false) return (<p>bla</p>)`

This might be similar to #885 but I figured it was different enough to warrant its own issue, especially because of the JSX thing.

DianaSuvorova:
@jaromero  thanks for reporting this issue and documenting all the options you tried.

The issue here is that for all practical purposes  `buildText` function looks like a valid react component since it is a function and return JSX. 
Also this `react component` is defined inside another react component `SampleComp`. So the `append` props is only considered to be used inside inner react component.

To account for this case we could have considered any component declared inside another component to be a helper method.
 
[...]

PR #1326, opened 2017-07-27
[This is going to fail] nested components definition issue

DianaSuvorova:
To illustrate discussion in #1268.


ljharb:
This is by design; functions that return jsx should be SFCs.

Issue #1336, opened 2017-07-31
false postive with names starting with a number using jsx-pascal-case


mfyuce:
(Not directly about run. It is about the RegEx provided in the code.)

Using v7.1.0,
```jsx
<1MyComponent/>
<2MyComponent/>
<3MyComponent/>
```

do not flag an error (e.g Contained JSX component MyComponent must not start with a number (react/jsx-pascal-case))

ljharb:
This seems fine to me, since those are invalid identifiers, they shouldn't work in the first place.

mfyuce:
True, but the regex, that is checking the IDs, is not covering this case. But somehow, in some place, it gives some error when linting, but not (i think) using methods within this project.

 I can close the issue, if you want.

ljharb:
The jsx parser should be throwing on it because it's invalid JS. The regex for this rule shouldn't need to check it, because it should be impossible for it to receive that in the first place.

mfyuce:
That is reasonable. Thank you.

Issue #1334, opened 2017-07-31
false negative with namespacing/qualifications using jsx-pascal-case


mfyuce:
Using v7.1.0,
```jsx
<__this.MyComponent/>
<containerClass.MyComponent/>
<containerObject.MyComponent/>
```
flag an error (e.g Contained JSX component MyComponent must be in PascalCase (react/jsx-pascal-case))

ljharb:
The docs for that rule don't mention either member expressions (your example) or namespaces, so it's probably an omission.

Both should be accounted for - I believe the spirit of the rule is that `anything.PascalCased` and `anything:PascalCased` should both be allowed.

golopot:
And `anything.camelCase` is disallowed, right? Disallowing this will be a breaking change.

ljharb:
I’d call it a bugfix, not a breaking change.

yacinehmito:
I am looking to fix this. I think there is a legitimate bug in the jsx-pascal-case rule but before going further I'd need to clarify what the behaviour should be. I tried to look at a spec for how namespaces should be handled but didn't find any authoritative source.

@ljharb Could you tell me what should be allowed / not allowed from this list when the rule is enabled?

1. `__foo.myComponent`
2. `__foo.MyComponent`
3. `__foo:myComponent`
4. `__foo:MyComponent`
5. `Foo.myComponent`
6. `Foo.MyComponent`
7. `Foo:myComponent`
8. `Foo:MyComponent`

The bug by the way has to do with the fact that when there is `foo.bar` in the code, we look how `foo` is formatted, not `bar`, which seems weird if `foo` is the namespace and `bar` the component name. However this might just be an error of understanding on my end.

ljharb:
@yacinehmito i would expect 2, 4, 6, and 8 to be allowed; and the others not to be.

yacinehmito:
I also wonder about:

1. `foo:bar`
2. `foo.bar`

There is some code that suggests that components in all lowercase should always be accepted as they can be plain HTML, but I don't know where the limit lies.
I'm pretty sure `foo:bar` should be accepted. I don't know about `foo.bar`.

ljharb:
No, lowercase components can never be namespaced; i'm pretty sure neither of those are valid.

yacinehmito:
> No, lowercase components can never be namespaced.

Ok. I will update the test cases in #2637 to reflect this then.

[...]

PR #1337, opened 2017-07-31
[Fix] `jsx-pascal-case`: false negative with namespacing

mfyuce:
case #1334 - false negative with namespacing/qualifications using jsx-pascal

case #1336 - false postive with names starting with a number using jsx-pascal
[Can besaid it is fixed, but inside it is not accepted anyway.  Seems just fixed the regex]


ljharb:
#1336 is not an issue; identifiers can not start with numbers.

tests/lib/rules/jsx-pascal-case.js
```diff
@@ -52,6 +52,8 @@ ruleTester.run('jsx-pascal-case', rule, {
   }, {
     code: '<Modal.Header />'
   }, {
+    code: '<qualification.T3stComp0nent />'
```
ljharb: `T3stComp0nent` isn't in PascalCase; this specific example should fail.

ljharb:
@mfyuce any interest in completing this PR?

tests/lib/rules/jsx-pascal-case.js
mfyuce: If I change it to "T3StComp0Nent", would it be OK? What do you think?

ljharb: yes, i think that would qualify as (a very very weird) PascalCase

PR #1340, opened 2017-08-03
fix(jsx-sort-props): enforce prop sort when reservedFirst is enabled

### LINKED ISSUE #1320: react/jsx-sort-props not working
I seem to be having trouble getting this rule to work.  The documentation references the following interface:
```
"react/jsx-sort-props": [<enabled>, {
  "callbacksLast": <boolean>,
  "shorthandFirst": <boolean>,
  "shorthandLast": <boolean>,
  "ignoreCase": <boolean>,
  "noSortAlphabetically": <boolean>,
  "reservedFirst": <boolean>|<array<string>>,
}]
```

My current configuration is `"react/jsx-sort-props": [2]`.  I have omitted the options which seem to work fine for other rules in the library.  Don't get any errors for the following component:
```
<Component c="c" b="b" a="a" />
```

tagoro9:
If reservedFirst is enabled and noSortAlphabetically is off, then alphabetical order is only enforced in the reserved props but not in the other props. This PR fixes this.

If existing behavior is the desired, please ignore this PR.

Fixes #1320

### DIFF
--- a/tests/lib/rules/jsx-sort-props.js
+++ b/tests/lib/rules/jsx-sort-props.js
@@ -271,8 +271,21 @@ ruleTester.run('jsx-sort-props', rule, {
       errors: [expectedError]
     },
     {
-      code: '<App dangerouslySetInnerHTML={{__html: "EPR"}} key={2} b />',
+      code: '<App key={2} b a />',
       options: reservedFirstAsBooleanArgs,
+      output: '<App key={2} a b />',
+      errors: [expectedError]
+    },
+    {
+      code: '<App b a />',
+      options: reservedFirstAsBooleanArgs,
+      output: '<App a b />',
+      errors: [expectedError]
+    },
+    {
+      code: '<App dangerouslySetInnerHTML={{__html: "EPR"}} e key={2} b />',
+      options: reservedFirstAsBooleanArgs,
+      output: '<App key={2} b dangerouslySetInnerHTML={{__html: "EPR"}} e />',
       errors: [expectedReservedFirstError]
     },
     {
--- a/lib/rules/jsx-sort-props.js
+++ b/lib/rules/jsx-sort-props.js
@@ -40,10 +40,20 @@ function isReservedPropName(name, list) {
   return list.indexOf(name) >= 0;
 }
 
-function alphabeticalCompare(a, b, ignoreCase) {
-  if (ignoreCase) {
+function propNameCompare(a, b, options) {
+  if (options.ignoreCase) {
     a = a.toLowerCase();
     b = b.toLowerCase();
+  }
+  if (options.reservedFirst) {
+    const aIsReserved = isReservedPropName(a, options.reservedList);
+    const bIsReserved = isReservedPropName(b, options.reservedList);
+    if ((aIsReserved && bIsReserved) || (!aIsReserved && !bIsReserved)) {
+      return a.localeCompare(b);
+    } else if (aIsReserved && !bIsReserved) {
+      return -1;
+    }
+    return 1;
   }
   return a.localeCompare(b);
 }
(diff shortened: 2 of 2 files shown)


tests/lib/rules/jsx-sort-props.js
```diff
@@ -202,9 +202,19 @@ ruleTester.run('jsx-sort-props', rule, {
...
+    {
+      code: '<App b a />',
+      options: reservedFirstAsBooleanArgs,
+      errors: [expectedError]
+    },
+    {
       code: '<App dangerouslySetInnerHTML={{__html: "EPR"}} key={2} b />',
       options: reservedFirstAsBooleanArgs,
-      errors: [expectedReservedFirstError]
+      errors: [expectedReservedFirstError, expectedError]
```
ljharb: why would this suddenly start giving two errors?

tagoro9: I think that after the changes this is triggering the `expectedError` as the `||
 (!previousIsReserved && !currentIsReserved)` check and it is true when evaluating `dangerouslySetInnerHTML` and `b`.

tagoro9:
I updated the tests with the expected output, but I the fix is not taking into account the reserved words:

Expected: `<App key={2} b dangerouslySetInnerHTML={{__html: "EPR"}} />`
Actual : `<App b dangerouslySetInnerHTML={{__html: "EPR"}} key={2} />`

I'm gonna try to see if I can add a fix for this.

tests/lib/rules/jsx-sort-props.js
ljharb: We should try to avoid generating the extra error.

tests/lib/rules/jsx-sort-props.js
```diff
@@ -202,9 +202,19 @@ ruleTester.run('jsx-sort-props', rule, {
...
     },
     {
+      code: '<App key={2} b a />',
+      options: reservedFirstAsBooleanArgs,
+      errors: [expectedError]
+    },
+    {
+      code: '<App b a />',
+      options: reservedFirstAsBooleanArgs,
+      errors: [expectedError]
```
ljharb: please add `output: '<App a b />'`

tests/lib/rules/jsx-sort-props.js
```diff
@@ -202,9 +202,19 @@ ruleTester.run('jsx-sort-props', rule, {
       errors: [expectedError]
     },
     {
+      code: '<App key={2} b a />',
+      options: reservedFirstAsBooleanArgs,
+      errors: [expectedError]
```
ljharb: Please add `output: <App key={2} a b />` (assuming that's the expected output)

tagoro9:
Sorry I took so long to take a look at the change requests.

* This [test](https://github.com/yannickcr/eslint-plugin-react/pull/1340/files#diff-85e28e7b3c88362eeaef8cf73f4ad42aR284) is no longer expecting 2 errors.
* Added expected output for the new test cases.
* With the expected output the tests were failing, so I updated the `generateFixerFunction` to take into account the `reservedFirst` config and sort props accordingly. Is this something wanted?

ljharb:
Definitely wanted, thanks :-)

tests/lib/rules/jsx-sort-props.js
```diff
@@ -269,8 +269,21 @@ ruleTester.run('jsx-sort-props', rule, {
...
+    {
+      code: '<App b a />',
+      options: reservedFirstAsBooleanArgs,
+      output: '<App a b />',
+      errors: [expectedError]
+    },
+    {
       code: '<App dangerouslySetInnerHTML={{__html: "EPR"}} key={2} b />',
       options: reservedFirstAsBooleanArgs,
+      output: '<App key={2} b dangerouslySetInnerHTML={{__html: "EPR"}} />',
```
ljharb: hmm - with reservedFirst tho, the expected output should be `<App dangerouslySetInnerHTML={{__html: "EPR"}} key={2} b />`, no?

tagoro9: I think `dangerouslySetInnerHTML` is not a reserved prop name for this component. I think it is only considered reserved for DOM components and not react components. This [code](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/jsx-sort-props.js#L208) is deleting `dangerouslySetInnerHTML` for the cases where the component is not a DOM component.

ljharb: ahhhh true! good call :-) could you add an "e" prop as well, so it's clear that it's being sorted as a regular property?

tagoro9: Added.

ljharb:
Planning to merge this soon.

tobyf93:
Is there a reason why this change hasn't been merged in?

tobyf93:
Any idea on when the merge will happen @ljharb?

PR #1343, opened 2017-08-04
[no-unused-prop-types] test cases

### LINKED ISSUE #933: no-unused-prop-types false positive when prop is used in function
Here's a simplified example.
We're using flow, but I guess this will fail when using `propTypes` also.

```
type SomeComponentPropsType = {
  onSelectedPlace(query: any, place: any) => void, 
};

// This has been simplified.
const onSelectPlace = component => (query, place) => {
  component.props.onSelectPlace(query, place);
};

// Changing the above to the following doesn't work either.
const onSelectPlace = ({ props }: { props: SomeComponentPropsType }) => (query, place) => {
  props.onSelectPlace(query, place);
}

class SomeComponent extends Component {
    props: SomeComponentPropsType;

    render() {
      return (<button onChange={onSelectPlace(this)}>Select<button>);
    }
}
```

### LINKED ISSUE #1002: react/no-unused-prop-types false positive on destructured props
react/no-unused-prop-types thinks that both the onClick and children props are unused even though these are clearly being used:

~~~jsx
type Props = {
  onClick: () => void,
  children: any,
};

class Button extends Component {
  props: Props

  render () {
    const { props } = this;
    return (
      <button
        onClick={props.onClick}
      >
        {props.children}
      </button>
    );
  }
}

export default Button;
~~~

If I refer to onClick using this.props.onClick or destructure all the props from this.props it works.

### LINKED ISSUE #1008: react/no-unused-prop-types false positive
This code:
```jsx
import React from 'react';

type ObjectType = {
  usedDirectly: string;
  usedFromArray: string;
};

export class UseDirectly extends React.Component {
  props: {
    object: ObjectType;
  }

  render() {
    console.log(this.props.object.usedDirectly);
    return null;
  }
}

export class UseFromArray extends React.Component {
  props: {
    objects: Array<ObjectType>;
  }

  render() {
    console.log(this.props.objects[0].usedFromArray);
    return null;
  }
}
```

triggers this warning:
```
5:18  error  'object.usedFromArray' PropType is defined but prop is never used  react/no-unused-prop-types
```

Though strangely, this code encounters no errors:
```jsx
import React from 'react';

type ObjectType = {
  usedDirectly: string;
  usedFromArray: string;
};

export class UseDirectly extends React.Component {
  props: {
    object: any;
  }

  render() {
    console.log(this.props.object.usedDirectly);
    return null;
  }
}

export class UseFromArray extends React.Component {
  props: {
    objects: Array<ObjectType>;
  }

  render() {
    console.log(this.props.objects[0].usedFromArray);
    return null;
  }
}
```

So you don't have to go hunting/searching, the only difference between them is the `object: ObjectType` vs`object: any`. (Not sure why changing `UseDirectly` would affect the used-ness of `usedFromArray` field.)

### LINKED ISSUE #1323: no-unused-prop-types does not process intersection flow types
`no-unused-prop-types` does not currently flag unused PropTypes in flow intersection types.

In this case, it should be flagging `age`.

```
type AProps = {
  name: string;
  age: number;
};

type BProps = {
  greeting: string;
};

type Props = AProps & BProps;

  props: Props;

  render () {
    return <div>{this.props.greeting}, {this.props.name}</div>;
  }
}
```

DianaSuvorova:
fixes #933

### DIFF
--- a/tests/lib/rules/no-unused-prop-types.js
+++ b/tests/lib/rules/no-unused-prop-types.js
@@ -2012,6 +2012,23 @@ ruleTester.run('no-unused-prop-types', rule, {
         '}'
       ].join('\n'),
       parser: 'babel-eslint'
+    }, {
+      // issue #933
+      code: [
+        'type Props = {',
+        ' onMouseOver: Function,',
+        ' onClick: Function,',
+        '};',
+
+        'const MyComponent = (props: Props) => (',
+        '<div>',
+        '  <button onMouseOver={() => props.onMouseOver()} />',
+        '  <button onClick={() => props.onClick()} />',
+        '</div>',
+        ');'
+      ].join('\n'),
+      parser: 'babel-eslint',
+      options: [{skipShapeProps: false}]
     }
   ],
 


tests/lib/rules/no-unused-prop-types.js
```diff
@@ -2012,6 +2012,79 @@ ruleTester.run('no-unused-prop-types', rule, {
...
+        '    </div>',
+        '  );'
+      ].join('\n'),
+      parser: 'babel-eslint'
+    }, {
+      // issue #1008
+      code: [
+        'type ObjectType = {',
+        '  usedDirectly: string;',
+        '  usedFromArray: string;',
```
ljharb: shouldn't this example only pass when skipShapeProps is false, but fail when it's true?

If so, could we add an explicit config option?

tests/lib/rules/no-unused-prop-types.js
```diff
@@ -2012,6 +2012,79 @@ ruleTester.run('no-unused-prop-types', rule, {
...
+        'type AProps = {',
+        '  name: string;',
+        '  age: number;',
+        '};',
+
+        'type BProps = {',
+        '  greeting: string;',
+        '};',
+
+        'type Props = AProps & BProps;',
```
ljharb: ok so this should either mean "name and age and greeting", or "name and age, or greeting" - in the former case, "age" is missing, in the latter case, "greeting" and "name" are mutually exclusive. Shouldn't this be an error?

DianaSuvorova:
Whoops, I got carried away, closing too many tickets at once.
None of the cases  I added would work with `skipShapeProps: false` except for #933. Issue #1008 looks like a  legitimate error. I think the rest should be kept open if we want to fix flow props validation.  I updated a PR. thanks!

tests/lib/rules/no-unused-prop-types.js
```diff
@@ -2012,6 +2012,23 @@ ruleTester.run('no-unused-prop-types', rule, {
...
+
+        'const MyComponent = (props: Props) => (',
+        '<div>',
+        '  <button onMouseOver={() => props.onMouseOver()} />',
+        '  <button onClick={() => props.onClick()} />',
+        '</div>',
+        ');'
+      ].join('\n'),
+      parser: 'babel-eslint',
+      options: [{skipShapeProps: false}]
```
ljharb: just to clarify; this example should work identically both with and without skipShapeProps?

DianaSuvorova: yeah, it doesn't have any effect. The props are not shape.

PR #1355, opened 2017-08-10
Add "ctors" option to @no-direct-mutation-state" rule

doochik:
This option allow to check specified methods like constructor.

**Motivation**
I have BaseComponent in my project and use it in this way:
```
class BaseComponent extends React.PureComponent {
   constructor(props, context) {
        try {
            this._init(props, context)
        } catch(e) {
           // log error
        }
   }
}

class MyComponent extends BaseComponent {
    _init() {
        this.state = {
            foo: 'bar'
        };
    }
}
```
`_init` method is actual constructor for `MyComponent` and I want to check it like constructor with the "ctors" options
```
"react/no-direct-mutation-state": [<enabled>, {
  ctors: ['_init']
}]
```

### DIFF
--- a/docs/rules/no-direct-mutation-state.md
+++ b/docs/rules/no-direct-mutation-state.md
@@ -66,10 +66,10 @@ This rule can take one argument to ignore some specific props during validation.
 ```js
 ...
 "react/no-direct-mutation-state": [<enabled>, {
-  ctors: <ctors>
+  constructors: <constructors>
 }]
 ...
 ```
 
 * `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
-* `ctors`: optional array of methods name to check like constructor
+* `constructors`: optional array of methods name to check like constructor
--- a/docs/rules/no-direct-mutation-state.md
+++ b/docs/rules/no-direct-mutation-state.md
@@ -58,3 +58,18 @@ class Hello extends React.Component {
   }
 }
 ```
+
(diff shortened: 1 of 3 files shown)


docs/rules/no-direct-mutation-state.md
```diff
@@ -58,3 +58,18 @@ class Hello extends React.Component {
...
 ```
+
+## Rule Options
+
+This rule can take one argument to ignore some specific props during validation.
+
+```js
+...
+"react/no-direct-mutation-state": [<enabled>, {
+  ctors: <ctors>
```
ljharb: Please spell out "constructors" - letters don't cost money, but confusing abbreviations can.

doochik:
@ ljharb Let my clarify puprose of BaseComponent in my project. This is very tiny class which wraps in try-catch some lifecycle methods like `constructor`, `render`, etc. It has no functionality.
Basically it has same ideas like `componentDidCatch` in react@16.

At this time I can't update to esling-pligin-react@7.2.0 because you've fixed https://github.com/yannickcr/eslint-plugin-react/issues/832 :)

ljharb:
I'd recommend doing that with a babel transform to change `React.Component` to your base component, and *definitely not* putting an unidiomatic pattern in your code.

Alternatively, you could also define `componentDidCatch` when needed, and you could use that babel transform only on components where that was defined.

doochik:
Ok, I understand your point

PR #1356, opened 2017-08-11
Fix handling of identifiers in forbid-prop-types

### LINKED ISSUE #1352: `forbid-prop-types`: propWrapperFunctions does not follow identifiers
I fixed a crash from #1253 in 5ad93b4365cc42ff6a6c858fed76b7a2d0c17222.

However, see the [TODO](https://github.com/yannickcr/eslint-plugin-react/commit/5ad93b4365cc42ff6a6c858fed76b7a2d0c17222#diff-f8e412f530183ea4c0e4bbfd3c3f2b0eR408) - the rule (prop-types also) is not following the identifier reference when it's not inline. It should.

cc @dustinsoftware

dustinsoftware:
This fixes incorrect behavior when propTypes were stored in a variable and then referenced later.

Fixes #1352


tests/lib/rules/sort-prop-types.js
```diff
@@ -565,6 +565,42 @@ ruleTester.run('sort-prop-types', rule, {
     }]
   }, {
     code: [
+      'const First = (props) => <div />;',
+      'const propTypes = {',
```
ljharb: can you also (just in case) add a test with `export const propTypes = …` to this rule too?

tests/lib/rules/forbid-prop-types.js
```diff
@@ -401,12 +401,19 @@ ruleTester.run('forbid-prop-types', rule, {
...
+    ].join('\n'),
+    errors: [{message: ANY_ERROR_MESSAGE}]
+  }, {
+    code: [
+      'import { forbidExtraProps } from "airbnb-prop-types";',
       'export const propTypes = {a: PropTypes.any};',
       'export default function Component() {}',
       'Component.propTypes = forbidExtraProps(propTypes);'
     ].join('\n'),
-    // errors: [{message: ANY_ERROR_MESSAGE}], // TODO: make this pass
```
ljharb: <3

tests/lib/rules/sort-prop-types.js
dustinsoftware: Sure, just so I can understand, what behavior will that change be verifying?

ljharb: Hopefully nothing; I'm just making sure that an export declaration works just as well as a variable declaration :-)

Issue #1288, opened 2017-07-06
`react/prop-types` false positive with local variable `props` 


rhys-vdw:
Apparently a local variable called `props` will be incorrectly considered to be passed props in a functional component.

```js
function Foo() {
  const props = {}
  props.bar = 'bar'
  return <div {...props} />
}
```

```
warning  'bar' is missing in props validation               react/prop-types
```

Following is fine: 

```js
class Foo extends Component {
  render() {
    const props = {}
    props.bar = 'bar'
    return <div {...props} />
  }
}
```

ljharb:
Worth fixing, but I have no idea why you'd ever want a variable called "props" in a React component :-)

rhys-vdw:
@ljharb FYI

```
function GenericThankyou({ nextAssignmentId }) {
  const cardProps = {}
  if (nextAssignmentId == null) {
    cardProps.submitText = 'Back to the lobby'
    cardProps.onSubmit = navigateTo(RailsRoutes.lobby_path())
  } else {
    cardProps.submitText = 'Take the next test'
    cardProps.onSubmit = navigateTo(RailsRoutes.do_ordered_test_path(nextAssignmentId))
  }
  return (
    <TestStepCard {...cardProps}>
      <Title>{translate('test.complete.thank_you_1')}</Title>
      <p className='GenericThankyou-text'>
        {translate('test.complete.message_1')}
      </p>
    </TestStepCard>
  )
}
```

ljharb:
(`cardProps` seems like the better name, since `props` looks like it's the GenericThankYou props 😉)

rhys-vdw:
(agreed, but I decided to report it anyway)

ljharb:
Fixed in #1365

PR #1365, opened 2017-08-13
[prop types] fix an assignment to prop

DianaSuvorova:
fixes issue #1288

### DIFF
--- a/lib/rules/prop-types.js
+++ b/lib/rules/prop-types.js
@@ -117,6 +117,20 @@ module.exports = {
     }
 
     /**
+    * Checks if a prop is being assigned a value props.bar = 'bar'
+    * @param {ASTNode} node The AST node being checked.
+    * @returns {Boolean}
+    */
+
+    function isAssignmentToProp(node) {
+      return (
+        node.parent &&
+        node.parent.type === 'AssignmentExpression' &&
+        node.parent.left === node
+      );
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


lib/rules/prop-types.js
```diff
@@ -126,7 +140,7 @@ module.exports = {
         (utils.getParentES6Component() || utils.getParentES5Component()) &&
         node.object.type === 'ThisExpression' && node.property.name === 'props'
       );
-      const isStatelessFunctionUsage = node.object.name === 'props';
+      const isStatelessFunctionUsage = (node.object.name === 'props' && !isAssignmentToProp(node));
```
ljharb: these wrapping parens aren't necessary

DianaSuvorova:
@ljharb updated

alexzherdev:
@ljharb I stumbled upon this test case and traced to this PR. Why should it not fail? Is assignment to a prop not considered usage? I'd expect this to fail with missing prop type for `bar` prop.
```js
function Foo(props) {
  props.bar = 'bar';
  return <div {...props} />;
}
```

alexzherdev:
Ah I see your point. Although the presence of an assignment doesn't necessarily mean that this prop wasn't passed as well. But I guess it's not enough to mark this as usage.

ljharb:
Right, I don't consider it usage, or an untyped prop.

It's weird, but it should still work :-)

PR #1370, opened 2017-08-14
Fixes #1369

### LINKED ISSUE #1369: `boolean-prop-naming` triggers "Cannot read property 'type' of undefined"
Hello,

When I add this rule to my .eslintrc file Visual Studio Code triggers a popup error `Cannot read property 'type' of undefined`.

This is my rule:
`"react/boolean-prop-naming": [1, { "rule": "(^(is|has)[A-Z]([A-Za-z0-9]?)+)|(^[a-z]+(ed|able)$)" }],`

I think this is related somehow to the spread operator as the errors trigger when I type `...test`

EvHaus:
Fixed a crash in the boolean-prop-naming rule.

### DIFF
--- a/lib/rules/boolean-prop-naming.js
+++ b/lib/rules/boolean-prop-naming.js
@@ -1,6 +1,6 @@
 /**
  * @fileoverview Enforces consistent naming for boolean props
- * @author Evgueni Naverniouk
+ * @author Ev Haus
  */
 'use strict';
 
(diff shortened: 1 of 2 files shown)


lib/rules/boolean-prop-naming.js
```diff
@@ -1,6 +1,6 @@
 /**
  * @fileoverview Enforces consistent naming for boolean props
- * @author Evgueni Naverniouk
+ * @author Ev Haus
```
EvHaus: I recently had a legal name change. Nothing to see here ;)

PR #1371, opened 2017-08-14
[no-typo] rule enhancement checking for a correct prop type spelling

### LINKED ISSUE #213: Rule proposal: validate PropTypes aren't typos
Rule would check that calls to React.PropTypes.whatever are valid. For instance, I have a bad habit of typing `disabled: PropTypes.bools.isRequired` instead of `disabled: PropTypes.bool.isRequired`. I've also done `disabled: propTypes.bool.isRequired` and `disabled: propTypes.bool.isRequired()` by accident.

DianaSuvorova:
closes #213

### DIFF
--- a/lib/rules/no-typos.js
+++ b/lib/rules/no-typos.js
@@ -21,10 +21,12 @@ const LIFECYCLE_METHODS = [
   'render'
 ];
 
+const PROP_TYPES = Object.keys(require('prop-types'));
+
 module.exports = {
   meta: {
     docs: {
-      description: 'Prevent common casing typos',
+      description: 'Prevent common typos',
       category: 'Stylistic Issues',
       recommended: false
     },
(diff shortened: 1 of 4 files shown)


lib/rules/no-typos.js
```diff
@@ -21,18 +21,87 @@ const LIFECYCLE_METHODS = [
   'render'
 ];
 
+// https://github.com/facebook/prop-types/blob/master/factoryWithTypeCheckers.js#L96-L114
+const PROP_TYPES = [
```
ljharb: `Object.keys(require('prop-types'))`?

lib/rules/no-typos.js
```diff
@@ -21,18 +21,87 @@ const LIFECYCLE_METHODS = [
...
+
+    function checkValidPropObject (node) {
+      if (node.type === 'ObjectExpression') {
+        node.properties.forEach(prop => checkValidProp(prop.value));
+      }
+    }
+    /* eslint-enable no-use-before-define */
+
     function reportErrorIfClassPropertyCasingTypo(node, propertyName) {
+      if (propertyName === 'propTypes') {
```
jseminck: Should we also perform this check `contextTypes` and `childContextTypes` ?

DianaSuvorova: Good point. Thank for taking a look at this!

lib/rules/no-typos.js
DianaSuvorova: Good idea !

DianaSuvorova:
`const { bools } = PropTypes` are not supported. if it is a common use case I would prefer doing it  in a separate PR

ljharb:
It's pretty common (also `import { bool } from 'prop-types'`), but a separate PR is fine.

DianaSuvorova:
@ljharb this should be ready then.  Except for `import { bool } from 'prop-types'` && `const { bools } = PropTypes ` which will be tackled separately

PR #1372, opened 2017-08-15
Fix crash for no-typos rule

jseminck:
This is a fix for: https://github.com/yannickcr/eslint-plugin-react/issues/1353

I'm not particularly happy with how I fixed this issue, but I think it should be fine for now. At least, the rule will not crash anymore.

There's more info in the ticket + there is a related ticket on the eslint repo with additional info.

### DIFF
--- a/lib/rules/no-typos.js
+++ b/lib/rules/no-typos.js
@@ -66,6 +66,10 @@ module.exports = {
       },
 
       MemberExpression: function(node) {
+        if (node.parent.type !== 'AssignmentExpression') {
+          return;
+        }
+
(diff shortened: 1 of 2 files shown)


ljharb:
Thanks!

jseminck:
With ESLint 4.7.0 we could now get rid of this patch, as `parent` should always be available when executing rules. https://eslint.org/blog/2017/09/eslint-v4.7.0-released

However, I suppose it's not so "easy" to update our peerDependency to 4.7.0, right? That would probably require a major version bump?

PR #1374, opened 2017-08-15
Rule proposal: no-access-state-in-setstate

jaaberg:
This rule prevents the use of `this.state` within `this.setState`. It is further described in `docs/rules/no-access-state-in-setstate.md` in the PR, but to summarize: `this.setState` is async, and referring to `this.state` within it may lead to race conditions. 

This is my first eslint-plugin PR, so I am really happy to get any kind of feedback.

edit: 
This is also the recommended way doing it in the [react docs](https://facebook.github.io/react/docs/react-component.html#setstate): 
> If the next state depends on the previous state, we recommend using the updater function form, instead
> ```
> this.setState((prevState) => {
>  return {counter: prevState.quantity + 1};
>}); 
> ```

### DIFF
--- a/docs/rules/no-access-state-in-setstate.md
+++ b/docs/rules/no-access-state-in-setstate.md
@@ -1,0 +1,39 @@
+# Prevent using this.state within a this.setState (no-access-state-in-setstate)
+
+This rule should prevent usage of this.state inside setState calls.
+Such usage of this.state might result in errors when two state calls is
+called in batch and thus referencing old state and not the current
+state. An example can be an increment function:
+
+```
+function increment() {
+  this.setState({value: this.state.value + 1});
+}
+```
+
+If these two setState operations is grouped together in a batch it will
+look be something like the following, given that value is 1:
+
+```
+setState({value: 1 + 1})
+setState({value: 1 + 1})
+```
+
+This can be avoided with using callbacks which takes the previous state
+as first argument:
+
+```
+function increment() {
+  this.setState(prevState => ({value: prevState.value + 1}));
+}
+```
+
+Then react will call the argument with the correct and updated  state,
+even when things happen in batches. And the example above will be
+something like:
+
+
+```
+setState({value: 1 + 1})
+setState({value: 2 + 1})
+```
--- a/docs/rules/no-access-state-in-setstate.md
+++ b/docs/rules/no-access-state-in-setstate.md
@@ -1,7 +1,7 @@
 # Prevent using this.state within a this.setState (no-access-state-in-setstate)
 
-This rule should prevent usage of this.state inside setState calls.
-Such usage of this.state might result in errors when two state calls is
+This rule should prevent usage of `this.state` inside `setState` calls.
+Such usage of `this.state` might result in errors when two state calls are
 called in batch and thus referencing old state and not the current
 state. An example can be an increment function:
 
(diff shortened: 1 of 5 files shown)


docs/rules/no-access-state-in-setstate.md
```diff
@@ -0,0 +1,39 @@
...
+setState({value: 1 + 1})
+setState({value: 1 + 1})
+```
+
+This can be avoided with using callbacks which takes the previous state
+as first argument:
+
+```
+function increment() {
+  this.setState(prevState => ({value: prevState.value + 1}));
```
ljharb: This is good advice when the semantic is "increment", but referencing this.state inside a setState call doesn't necessarily mean that the dev doesn't want s "snapshot".

What happens with:
```js
const { value } = this.state;
this.setState({ value: value + 1 });
```
?

This is conceptually the same code, but it might be a useful way for the author to indicate that this is an intentional ordering.

jaaberg: Right now, I will report this as a warning. The same with 

```javascript
getstate() {
  return { value: this.state.counter++ };
}

increase() {
  this.setState(getState());
}
```

I could easily make it only report `this.state` directly inside `this.setState`, but I don't see why you would use `this.state` and not the `prevState`-solution which is guaranteed to be up-to-date.

Do you have an example when you actually would want to do that? 

This is from the [react docs](https://facebook.github.io/react/docs/react-component.html#setstate): 
> If the next state depends on the previous state, we recommend using the updater function form, instead
> ```
> this.setState((prevState) => {
>  return {counter: prevState.quantity + 1};
>}); 
> ```

okonet:
OMG can this please be merged soon?

FezVrasta:
I want this 🙏

docs/rules/no-access-state-in-setstate.md
```diff
@@ -0,0 +1,39 @@
+# Prevent using this.state within a this.setState (no-access-state-in-setstate)
+
+This rule should prevent usage of this.state inside setState calls.
+Such usage of this.state might result in errors when two state calls is
```
FezVrasta: `` `this.state`  `` maybe? (with code formatting)

jaaberg: Yes, I can probably do that for all `this.state` and `setState` mentions in the docs. Thanks, I will fix that first thing to the morning tomorrow.

jaaberg:
For anyone that needs this rule, I've just released it as a separate plugin for eslint [here](https://github.com/jaaberg/eslint-plugin-seeact) since this hasn't been going anywhere for a few months. 

I really hope it can get merged to this project some time, so if the maintainers has some questions, please reach out.

np-8:
What do you guys think about the case, where `this.state` is saved in a temporary variable? For example[, react/destructuring-assignment](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/destructuring-assignment.md) enforces that. Should/could `no-access-state-in-setstate` trigger in that case? 

```
const { value } = this.state;
this.setState({ value: value + 1 });
```

FezVrasta:
it should, because you are still referencing a possibly outdated value

np-8:
I was thinking that, too. For some reason I could get it triggered only by using `value: this.state.value +1` explicitly. Using eslint-plugin-react v. 7.5.1 & ESLint v.3.19.0.

ljharb:
@np-8 can you file that as a new issue?

np-8:
Sure, here you are: #1597

PR #1375, opened 2017-08-15
More robust no typos fix

jseminck:
Second attempt at fixing https://github.com/yannickcr/eslint-plugin-react/issues/1353

I think this is a better approach, but it means that currently one case is not supported. See the added test:

```js
      MyComponent.PROPTYPES = {}
      /** @extends React.Component */
      class MyComponent extends BaseComponent {}
```

I suppose that is acceptable for now, at least it is better than having the rule crash.

Also explained why the try/catch is needed and linked to relevant eslint-scope issue that needs to be implemented before it can be removed.

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -218,7 +218,8 @@ function componentRule(rule, context) {
       let comment;
       // Sometimes the passed node may not have been parsed yet by eslint, and this function call crashes.
       // Can be removed when eslint sets "parent" property for all nodes on initial AST traversal: https://github.com/eslint/eslint-scope/issues/27
-      // Remove try/catch when https://github.com/eslint/eslint-scope/issues/27 is implemented.
+      // eslint-disable-next-line no-warning-comments
+      // FIXME: Remove try/catch when https://github.com/eslint/eslint-scope/issues/27 is implemented.
       try {
         comment = sourceCode.getJSDocComment(node);
       } catch (e) {
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -218,6 +218,7 @@ function componentRule(rule, context) {
       let comment;
       // Sometimes the passed node may not have been parsed yet by eslint, and this function call crashes.
       // Can be removed when eslint sets "parent" property for all nodes on initial AST traversal: https://github.com/eslint/eslint-scope/issues/27
+      // FIXME: Remove try/catch when https://github.com/eslint/eslint-scope/issues/27 is implemented.
       try {
         comment = sourceCode.getJSDocComment(node);
       } catch (e) {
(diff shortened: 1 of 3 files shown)


lib/util/Components.js
```diff
@@ -215,7 +215,14 @@ function componentRule(rule, context) {
      * @returns {Boolean} True if the node is explicitly declared as a descendant of a React Component, false if not
      */
     isExplicitComponent: function(node) {
-      const comment = sourceCode.getJSDocComment(node);
+      let comment;
+      // Sometimes the passed node may not have been parsed yet by eslint, and this function call crashes.
+      // Can be removed when eslint sets "parent" property for all nodes on initial AST traversal: https://github.com/eslint/eslint-scope/issues/27
```
ljharb: What about changing the places calling this method to be ":exit"?

jseminck: It does not help. The `relatedComponent` (in the case below it is `MyComponent`) is not a child in the AST of the `MemberExpression` `MyComponent.PROPTYPES` - so using `:exit` does not work.

The reason is just that `MyComponent` is declared after `MyComponent.PROPTYPES` and therefore eslint has not yet parsed it and added the `parent` property.

```js
      MyComponent.PROPTYPES = {}
      /** @extends React.Component */
      class MyComponent extends BaseComponent {}
``` 

They have a task to add parent property to all nodes on initial traversal, that would fix our issue.

lib/util/Components.js
```diff
@@ -218,7 +218,7 @@ function componentRule(rule, context) {
       let comment;
       // Sometimes the passed node may not have been parsed yet by eslint, and this function call crashes.
       // Can be removed when eslint sets "parent" property for all nodes on initial AST traversal: https://github.com/eslint/eslint-scope/issues/27
-      // FIXME: Remove try/catch when https://github.com/eslint/eslint-scope/issues/27 is implemented.
```
ljharb: i think we should override the linting rule rather than avoid using "FIXME"

jseminck:
Also wrote a comment explaining the issue to the `eslint` core team: https://github.com/eslint/eslint/issues/9101#issuecomment-322966051

jseminck:
The comment is now fixed. One of the eslint maintainers gave some good ideas on how to get it working properly. I'll have to re-write the rule a bit: https://github.com/eslint/eslint/issues/9101#issuecomment-323134182 (or someone else can give it a go if they want to)

jseminck:
With ESLint 4.7.0 we could now get rid of this patch, as `parent` should always be available when executing rules. https://eslint.org/blog/2017/09/eslint-v4.7.0-released

However, I suppose it's not so "easy" to update our peerDependency to 4.7.0, right? That would probably require a major version bump?

ljharb:
Yes, it would require a major version bump.

PR #1380, opened 2017-08-17
require-default-props - fixing an edge case where linting stops early

brgibson:
This is to fix an edge case where linting stops when it finds a component without props.  Instead it should move on to the next component in the file.

### DIFF
--- a/tests/lib/rules/require-default-props.js
+++ b/tests/lib/rules/require-default-props.js
@@ -1307,6 +1307,25 @@ ruleTester.run('require-default-props', rule, {
       }]
     },
 
+    // component with no declared props followed by a failing component
+    {
+      code: [
+        'var ComponentWithNoProps = ({ bar = "bar" }) => {',
+        '  return <div>Hello {this.props.foo}</div>;',
+        '}',
+        'var Greetings = ({ foo = "foo" }) => {',
+        '  return <div>Hello {this.props.foo}</div>;',
+        '}',
+        'Greetings.propTypes = {',
+        '  foo: PropTypes.string',
+        '};'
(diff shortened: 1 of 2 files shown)


tests/lib/rules/require-default-props.js
```diff
@@ -1306,7 +1306,26 @@ ruleTester.run('require-default-props', rule, {
         column: 3
       }]
     },
-
+    //
+    // component with no pros followed by a failing component
```
jseminck: Typo.

Also is the empty `//` needed?

tests/lib/rules/require-default-props.js
```diff
@@ -1306,7 +1306,26 @@ ruleTester.run('require-default-props', rule, {
...
+    {
+      code: [
+        'var ComponentWithNoProps = ({ bar = "bar" }) => {',
```
jseminck: It reads a bit confusing. It says`component with no props` but this component actually has props: `bar` and `foo`. I guess that you actually mean; `component with no declared prop types` (since there is no `.propTypes`) for the first component.

jseminck:
Nice find! 

The fix looks good to me. I added a few comments about the tests.

tests/lib/rules/require-default-props.js
brgibson: updated the comment and fixed typo:

"component with no declared props followed by a failing component"

tests/lib/rules/require-default-props.js
brgibson: I was just following the pattern in the rest of the file, I have deleted it now though, since it does seem unnecessary 👍

PR #1384, opened 2017-08-19
[jsx-wrap-multilines] Add new nodes for wrap

evgeny-petukhov:
Add ConditionalExpression, LogicalExpression and JSXAttribute

### DIFF
--- a/docs/rules/jsx-wrap-multilines.md
+++ b/docs/rules/jsx-wrap-multilines.md
@@ -6,7 +6,7 @@ Wrapping multiline JSX in parentheses can improve readability and/or convenience
 
 ## Rule Details
 
-This rule optionally takes a second parameter in the form of an object, containing places to apply the rule. By default, all the syntax listed below will be checked, but these can be explicitly disabled. Any syntax type missing in the object will follow the default behavior (become enabled).
+This rule optionally takes a second parameter in the form of an object, containing places to apply the rule. By default, all the syntax listed below will be checked except the conditional expressions, logical expressions and JSX attributes, but these can be explicitly disabled. Any syntax type missing in the object will follow the default behavior (become enabled).
 
 There are the possible syntax available:
 
@@ -14,6 +14,9 @@ There are the possible syntax available:
 * `assignment`
 * `return`
 * `arrow`
+* `condition` (not enabled by default)
+* `logical` (not enabled by default)
+* `attr` (not enabled by default)
 
 The following patterns are considered warnings:
 
@@ -117,4 +120,63 @@ var hello = () => (
     <p>World</p>
   </div>
 );
+```
+
+The following patterns are considered warnings when configured `{condition: true}`.
+
+```jsx
+<div>
+  {foo ? <div>
+      <p>Hello</p>
+    </div> : null}
+</div>
+```
+
+The following patterns are not considered warnings when configured `{condition: true}`.
+
+```jsx
+<div>
+  {foo ? (<div>
+      <p>Hello</p>
+  </div>) : null}
+</div>
+```
+
+
+The following patterns are considered warnings when configured `{logical: true}`.
+
+```jsx
+<div>
+  {foo &&
+    <div>
+      <p>Hello World</p>
+    </div>
+  }
+</div>
+```
+
+The following patterns are not considered warnings when configured `{logical: true}`.
+
+```jsx
+<div> not
+```
+
+The following patterns are considered warnings when configured `{attr: true}`.
+
+```jsx
+<div attr={<div>
+    <p>Hello</p>
+  </div>}>
+  <p>Hello</p>
+</div>;
+```
+
+The following patterns are not considered warnings when configured `{attr: true}`.
+
+```jsx
+<div attr={(<div>
+    <p>Hello</p>
+  </div>)}>
+  <p>Hello</p>
+</div>;
 ```
--- a/docs/rules/jsx-wrap-multilines.md
+++ b/docs/rules/jsx-wrap-multilines.md
@@ -6,7 +6,7 @@ Wrapping multiline JSX in parentheses can improve readability and/or convenience
 
 ## Rule Details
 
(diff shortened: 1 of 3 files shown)


docs/rules/jsx-wrap-multilines.md
```diff
@@ -118,3 +121,62 @@ var hello = () => (
...
+      <p>Hello World</p>
+    </div>
+  }
+</div>
+```
+
+The following patterns are not considered warnings when configured `{logical: true}`.
+
+```jsx
+<div> not
```
ljharb: ?

tests/lib/rules/jsx-wrap-multilines.js
```diff
@@ -201,20 +232,6 @@ ruleTester.run('jsx-wrap-multilines', rule, {
       options: [{return: true}],
       errors: [{message: 'Missing parentheses around multilines JSX'}]
     }, {
-      code: DECLARATION_TERNARY_NO_PAREN,
```
ljharb: why do these tests need to be removed?

alanevEugene:
Fix example for valid jsx in the attribute

tests/lib/rules/jsx-wrap-multilines.js
alanevEugene: This is the same tests as tests for condition rule. I can return them if you think that they are useful. The more tests the better

ljharb: For semver-minor PRs, it makes me happier when tests are only added :-)

alanevEugene:
Return tests.
Also I've added conditions to defaults. They were already part of `declaration` and `assignment` rules. Now it is separated and becomes more common rule.

alanevEugene:
@ljharb Can you review again, please?

docs/rules/jsx-wrap-multilines.md
```diff
@@ -6,14 +6,17 @@ Wrapping multiline JSX in parentheses can improve readability and/or convenience
...
 
 There are the possible syntax available:
 
 * `declaration`
 * `assignment`
 * `return`
 * `arrow`
+* `condition`
+* `logical` (not enabled by default)
+* `attr` (not enabled by default)
```
ljharb: this should probably be "prop" - in jsx, elements have props, and "attributes" are something HTML elements have (which jsx isn't)

tests/lib/rules/jsx-wrap-multilines.js
```diff
@@ -155,14 +217,14 @@ ruleTester.run('jsx-wrap-multilines', rule, {
       code: DECLARATION_TERNARY_PAREN
     }, {
       code: DECLARATION_TERNARY_NO_PAREN,
-      options: [{declaration: false}]
+      options: [{condition: false}]
```
ljharb: why does this option need to change?

tests/lib/rules/jsx-wrap-multilines.js
```diff
@@ -155,14 +217,14 @@ ruleTester.run('jsx-wrap-multilines', rule, {
...
     }, {
       code: ASSIGNMENT_TERNARY_SINGLE_LINE
     }, {
       code: ASSIGNMENT_TERNARY_PAREN
     }, {
       code: ASSIGNMENT_TERNARY_NO_PAREN,
-      options: [{assignment: false}]
+      options: [{condition: false}]
```
ljharb: why does this option need to change?

docs/rules/jsx-wrap-multilines.md
alanevEugene: Agreed!
I've renamed to `prop`

tests/lib/rules/jsx-wrap-multilines.js
alanevEugene: This test case checked the condition in declaration. Now checking the condition is separated.

tests/lib/rules/jsx-wrap-multilines.js
alanevEugene: This test case checked the condition in assignment. Now checking the condition is separated.

tests/lib/rules/jsx-wrap-multilines.js
ljharb: Does that mean that "declaration: true" now doesn't check as much as it used to? That would make this a breaking change.

tests/lib/rules/jsx-wrap-multilines.js
ljharb: Similarly, does that mean that "assignment: true" now doesn't check as much as it used to? That would make this a breaking change.

tests/lib/rules/jsx-wrap-multilines.js
alanevEugene: Yes, it is a broken change for declarations. But now the logic of checking is better

tests/lib/rules/jsx-wrap-multilines.js
alanevEugene: Yes, it is a broken change for assignments too

tests/lib/rules/jsx-wrap-multilines.js
ljharb: Is there any way we could avoid making it breaking?

Like if `condition` is absent, it uses the old behavior, but if it's present (true or false) then it uses the new behavior?

alanevEugene: Yes, I can do it now. But it is not very good, so I suggest to do this only as temporary solution, and separate checking condition in major release

ljharb: Yes, definitely we'd want to separate that in the next major - but for now, we'd be able to add these new options as semver-minor.

alanevEugene:
Now there are only additions, so this version is fully compatible with previous.

alanevEugene:
@ljharb Is it would be useful to add warning?

ljharb:
What kind of warning?

alanevEugene:
To enable checking condition if they now use conditions in declaration or assignment

alanevEugene:
@ljharb Hello! What status of this PR? Do you need these changes in the project?

ljharb:
I'm waiting for reviews from other collaborators.

PR #1385, opened 2017-08-21
jsx-handler-names update for better error messaging

cbengtson85:
For example right now a user will see with default settings:
`<Component onhandleclick={this.handleClick} />`

"Prop key for onclick must begin with on"

Well that error doesn't help me....
The code is really looking for on[A-Z]

so now it will say in the error "on[A-Z]"

### DIFF
--- a/lib/rules/jsx-handler-names.js
+++ b/lib/rules/jsx-handler-names.js
@@ -36,8 +36,10 @@ module.exports = {
     const eventHandlerPrefix = configuration.eventHandlerPrefix || 'handle';
     const eventHandlerPropPrefix = configuration.eventHandlerPropPrefix || 'on';
 
+    const eventHandlerPropRegex = `${eventHandlerPropPrefix}[A-Z]`;
+
     const EVENT_HANDLER_REGEX = new RegExp(`^((props\\.${eventHandlerPropPrefix})|((.*\\.)?${eventHandlerPrefix}))[A-Z].*$`);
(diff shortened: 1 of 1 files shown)


lib/rules/jsx-handler-names.js
```diff
@@ -36,8 +36,10 @@ module.exports = {
     const eventHandlerPrefix = configuration.eventHandlerPrefix || 'handle';
     const eventHandlerPropPrefix = configuration.eventHandlerPropPrefix || 'on';
 
+    const eventHandlerPropRegex = `${eventHandlerPropPrefix}[A-Z]`;
```
jseminck: Can you update the tests as well? Thanks 😄

cbengtson85: Will update it this evening!

lib/rules/jsx-handler-names.js
```diff
@@ -63,7 +65,7 @@ module.exports = {
         } else if (propFnIsNamedCorrectly && !propIsEventHandler) {
           context.report({
             node: node,
-            message: `Prop key for ${propValue} must begin with '${eventHandlerPropPrefix}'`
+            message: `Prop key for ${propValue} must begin with '${eventHandlerPropRegex}'`
```
ljharb: This isn't very clear; "begin with" only makes sense when followed by a string. If it's a regex (which it shouldn't be) then it'd need something like "match pattern".

What about "Prop key for ${propValue} (matches pattern ${eventHandlerPropRegex}) must begin with ${eventHandlerPropPrefix}"?

ljharb (changes requested):

PR #1387, opened 2017-08-21
Handle UpdateExpressions in no-direct-mutation-state

### LINKED ISSUE #1386: no-direct-mutation-state rule doesn't handle UpdateExpressions
`this.state.foo++` does not trigger the `no-direct-mutation-state` rule.

`this.state.foo += 1` does - indicating this is likely just that `UpdateExpression`s aren't handled.

zpao:
This rule should also catch increment & decrement operators.

Fixes #1386

### DIFF
--- a/tests/lib/rules/no-direct-mutation-state.js
+++ b/tests/lib/rules/no-direct-mutation-state.js
@@ -70,6 +70,14 @@ ruleTester.run('no-direct-mutation-state', rule, {
       '}'
     ].join('\n')
   }, {
+    code: [
+      'class Hello extends React.Component {',
+      '  constructor() {',
+      '    this.state.foo = 1;',
+      '  }',
+      '}'
+    ].join('\n')
+  }, {
     code: `
       class OneComponent extends Component {
         constructor() {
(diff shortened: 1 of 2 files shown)


tests/lib/rules/no-direct-mutation-state.js
```diff
@@ -70,6 +70,14 @@ ruleTester.run('no-direct-mutation-state', rule, {
       '}'
     ].join('\n')
   }, {
+    code: [
+      'class Hello extends React.Component {',
+      '  constructor() {',
+      '    this.state.foo = 1;',
```
ljharb: Could we also add a test that makes `++` invalid inside the constructor?

zpao: Oh shoot, I must have missed adding that line checking another thing… Sorry.

I had `this.state.foo++` right here as a valid case - we're in the constructor and while I think it's awkward, it's legal.

ljharb: I don't see why it should be legal; it's mutating `this.state` after it's assigned.

zpao: Assignment doesn't actually matter until we're out of the `constructor` method. That's when React takes over.

But that's also a tangential change to what I wanted to make here - I'm bringing `++` to parity with `=` (you have a test case immediately above this one showing that). If you want to change how you treat assignment as well, that's fine but I think you probably want to do that distinctly.

ljharb: Fair enough.

TheSavior: This has been my favorite find in our codebase where we were assigning to state.

```
this.setState({
  foo: this.state.foo++
});
```

jaaberg: @TheSavior (cc: @zpao )

I actually have a PR proposing a rule catching all references to `this.state` used in at `this.setState` in #1374. Does this rule cover that as well? If not, I would be really happy for feedback.

zpao:
Do you need anything else from me here?

PR #1390, opened 2017-08-22
Support read-only props in Flow for no-unused-prop-types

jseminck:
Fixes https://github.com/yannickcr/eslint-plugin-react/issues/1388

Added a more generic implementation that supports all prefixes, not just `-` and `+` like it was the case in the `prop-types` rule.

Additionally, made both rules implement the same logic. Although, as described in the other PR, the logic for detecting prop-types should really be extracted out to a separate file. I'd like to tackle that separately as well though.

### DIFF
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -276,7 +276,7 @@ module.exports = {
      */
     function getKeyValue(node) {
       if (node.type === 'ObjectTypeProperty') {
-        const tokens = context.getFirstTokens(node, {count: 1, filter: tokenNode => ['Identifier', 'String'].includes(tokenNode.type)});
+        const tokens = context.getFirstTokens(node, {count: 1, filter: tokenNode => ['Identifier', 'String'].indexOf(tokenNode.type) >= 0});
         return stripQuotes(tokens[0].value);
       }
       const key = node.key || node.argument;
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -263,10 +263,7 @@ module.exports = {
      * @param {string} the identifier to strip
      */
     function stripQuotes(string) {
-      if (string[0] === '\'' || string[0] === '"' && string[0] === string[string.length - 1]) {
-        return string.slice(1, string.length - 1);
-      }
-      return string;
+      return string.replace(/^\'|\'$/g, '');
     }
 
     /**
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -265,7 +265,7 @@ module.exports = {
      */
     function getKeyValue(node) {
       if (node.type === 'ObjectTypeProperty') {
-        const tokens = context.getFirstTokens(node, {count: 1, filter: tokenNode => tokenNode.type === 'Identifier'});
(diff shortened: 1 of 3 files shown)


tests/lib/rules/no-unused-prop-types.js
```diff
@@ -2032,27 +2032,50 @@ ruleTester.run('no-unused-prop-types', rule, {
...
-        a: React.PropTypes.string,
-        ...SharedPropTypes // eslint-disable-line object-shorthand
-      };
-    `,
+        A.propTypes = {
+          a: React.PropTypes.string,
+          ...SharedPropTypes // eslint-disable-line object-shorthand
+        };
+      `,
+      parser: 'babel-eslint'
```
jseminck: Reformatted this test - that's why it's in the PR...

lib/rules/no-unused-prop-types.js
```diff
@@ -259,14 +259,25 @@ module.exports = {
     }
 
     /**
+     * Removes quotes from around an identifier.
+     * @param {string} the identifier to strip
+     */
+    function stripQuotes(string) {
+      if (string[0] === '\'' || string[0] === '"' && string[0] === string[string.length - 1]) {
```
ljharb: `return string.replace(/^(["']).*($1)$/, '')` or similar?

lib/rules/prop-types.js
```diff
@@ -346,11 +346,8 @@ module.exports = {
      */
     function getKeyValue(node) {
       if (node.type === 'ObjectTypeProperty') {
-        const tokens = context.getFirstTokens(node, 2);
-        return (tokens[0].value === '+' || tokens[0].value === '-'
-          ? tokens[1].value
-          : stripQuotes(tokens[0].value)
-        );
+        const tokens = context.getFirstTokens(node, {count: 1, filter: tokenNode => ['Identifier', 'String'].indexOf(tokenNode.type) >= 0});
```
ljharb: why would we want to strip all prefixes? only `+` and `-` exist, as far as I know.

jseminck: It seems a much safer implementation for the future. If Flow would introduce a new prefix, it should be handled automatically. (assuming that the prefix token type is similar to the types it has now).

If you think keeping a list of prefixes is better then we can go with that approach as well...

lib/rules/prop-types.js
```diff
@@ -333,10 +333,7 @@ module.exports = {
      * @param {string} the identifier to strip
      */
     function stripQuotes(string) {
```
ljharb: can this repeated function be extracted out into a shared file?

lib/rules/prop-types.js
ljharb: I'd prefer to have explicit support; I'd rather it not silently risk doing the wrong thing when an unanticipated prefix exists.

I definitely don't want the plugin to *crash* when a new prefix is found - but i might want it to warn on the prop.

lib/rules/prop-types.js
jseminck: Yeah - that's what I mentioned in the other PR: `prop-types.js` and `no-unused-prop-types` **on first sight** seem to share quite a bit of logic that deals with extracting the declared prop types for a component. And currently, the logic is not in sync (e.g. props in TypedArgument for Flow is not supported at all for `no-unused-prop-types`.

I'll try to deal with extracting as much of the shared logic as possible this weekend. Created an issue: https://github.com/yannickcr/eslint-plugin-react/issues/1393

PR #1391, opened 2017-08-23
[jsx-no-target-blank] Allow either noopener or noreferrer

ngyikp:
I use `noreferrer` for my external links which also disables `window.opener`, but the rule insists on requiring `noopener` even though it's redundant.

According to the [HTML5 Standard spec](https://html.spec.whatwg.org/multipage/links.html#link-type-noreferrer):

> For historical reasons, the _noreferrer_ keyword implies the behavior associated with the _noopener_ keyword when present on a hyperlink that creates a new browsing context. That is, `<a href="..." rel="noreferrer" target="_blank">` has the same behavior as `<a href="..." rel="noreferrer noopener" target="_blank">`.

Also fixes https://github.com/yannickcr/eslint-plugin-react/issues/776

See also https://github.com/GoogleChrome/lighthouse/issues/2482 and https://github.com/GoogleChrome/lighthouse/pull/2485

### DIFF
--- a/docs/rules/jsx-no-target-blank.md
+++ b/docs/rules/jsx-no-target-blank.md
@@ -2,9 +2,10 @@
 
 When creating a JSX element that has an `a` tag, it is often desired to have
 the link open in a new tab using the `target='_blank'` attribute. Using this
-attribute unaccompanied by `rel='noreferrer noopener'`, however, is a severe
-security vulnerability ([see here for more details](https://mathiasbynens.github.io/rel-noopener))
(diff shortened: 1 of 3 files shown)


Hypnosphi:
> The discussion at GoogleChrome/lighthouse#2485 is still unresolved

It is now: https://github.com/GoogleChrome/lighthouse/pull/2485#issuecomment-379082525

ngyikp:
Closing PR as it's already done https://github.com/yannickcr/eslint-plugin-react/pull/2043

PR #1394, opened 2017-08-23
[no-unused-state] Ignore computed property keys for no unused state rule

jackyho112:
For https://github.com/yannickcr/eslint-plugin-react/issues/1361.

This sets the no-unused-state rule to ignore computed property keys when the contents aren't literal.

I think it should cover all kinds of literals for checking unused state properties because it could happen in types other than string and could be checked. If that is undesirable, I could change it to only checking string literals. 

Other than that, let me know what other improvements you have in mind.


jackyho112:
@EvHaus 

You are welcome! Look forward to doing more.

PR #1400, opened 2017-08-27
Add support to ClassExpressions in the prop-types rule

jseminck:
I'm not super happy with the solution, perhaps others have a better idea?

It turns out that ESLint visits `ClassExpression` before it visits `TypeParameterDeclaration`, but I want it the other way around.

I tried using the `ClassExpression:exit` key, but it doesn't change anything.

Therefore, I needed a way to defer class expressions evaluations util after `TypeParameterDeclaration` has been visited, by storing the nodes in an array and looping through it in `Program:exit`.

This fixes https://github.com/yannickcr/eslint-plugin-react/issues/1376#issuecomment-325166909 and several other cases which we didn't catch before.

I'm not sure it will cover all of the issues related to Flow Props yet.

I'll add the same support to `no-unused-prop-types`  in https://github.com/yannickcr/eslint-plugin-react/issues/1393

### DIFF
--- a/lib/rules/prop-types.js
+++ b/lib/rules/prop-types.js
@@ -65,6 +65,7 @@ module.exports = {
     // Used to track the type annotations in scope.
     // Necessary because babel's scopes do not track type annotations.
     let stack = null;
+    const classExpressions = [];
 
     const MISSING_MESSAGE = '\'{{name}}\' is missing in props validation';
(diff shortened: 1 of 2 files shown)


johnhaley81:
@jseminck I just tested this out in my code and it's working great!

PR #1404, opened 2017-08-29
Add support for Flow IntersectionTypeAnnotation to prop-types and no-unused-prop-types

### LINKED ISSUE #1323: no-unused-prop-types does not process intersection flow types

### LINKED ISSUE #1364: Bug: react/prop-types fails to parse destructured Flow typing with Flow union type
Symptom: `react/prop-types` indicates that props validation is missing for all props in the component.

This occurs whenever there is a destructured Flow typing anywhere in the component, *and* props are typed in the component using a union `&`. This is reproducible on v7.2.0 and earlier. Minimally reproducible example:

```javascript
// @flow

type PropsA = {
  a: string,
};

type PropsB = {
  b: string,
};

type Props = PropsA & PropsB;

export default class Sample extends Component {
  props: Props;

  handleEvent = ({ value }: { value: string }) => {};

  render() {
    const { a, b } = this.props;

    return (
      <div>
        {a}{b}
      </div>
    );
  }
}
```

The reported errors are:

```
  21:13  error  'a' is missing in props validation                          react/prop-types
  21:16  error  'b' is missing in props validation                          react/prop-types
```

No ESLint errors are generated if `value` is not Flow-typed in the destructure:

```javascript
// @flow

type PropsA = {
  a: string,
};

type PropsB = {
  b: string,
};

type Props = PropsA & PropsB;

export default class Sample extends Component {
  props: Props;

  handleEvent = ({ value }: Object) => {};

  render() {
    const { a, b } = this.props;

    return (
      <div>
        {a}{b}
      </div>
    );
  }
}
```

Further, no errors are generated if the destructure typing is preserved, but the union typing is removed in favor of a single object:

```javascript
// @flow

type Props = {
  a: string,
  b: string,
};

export default class Sample extends Component {
  props: Props;

  handleEvent = ({ value }: { value: string }) => {};

  render() {
    const { a, b } = this.props;

    return (
      <div>
        {a}{b}
      </div>
    );
  }
}

```

jseminck:
Fixes #1364. Fixes #1323.

The following snippet creates an `IntersectionTypeAnnotation` which was not handled in both rules.

```js
type Props = PropsA & PropsB;
```

I hope the tests/implementation are speaking for themselves on this one 😄 If anyone has any advice for additional test cases please let me know. I didn't test the `React.Component<Props>` syntax because I figured these shouldn't affect that implementation, but perhaps it's worth it?

### DIFF
--- a/tests/lib/rules/default-props-match-prop-types.js
+++ b/tests/lib/rules/default-props-match-prop-types.js
@@ -692,7 +692,6 @@ ruleTester.run('default-props-match-prop-types', rule, {
           props: Props;
           static defaultProps = {
             foo: "foo",
-            bar: "bar",
           }
 
           render() {
--- a/tests/lib/rules/default-props-match-prop-types.js
+++ b/tests/lib/rules/default-props-match-prop-types.js
@@ -683,6 +683,26 @@ ruleTester.run('default-props-match-prop-types', rule, {
       parser: 'babel-eslint'
     },
     {
+      code: `
+        type PropsA = { foo?: string };
+        type PropsB = { bar?: string, fooBar: string };
+        type Props = PropsA & PropsB;
+
+        class Bar extends React.Component {
+          props: Props;
+          static defaultProps = {
+            foo: "foo",
+            bar: "bar",
+          }
+
+          render() {
+            return <div>{this.props.foo} - {this.props.bar}</div>
+          }
+        }
+      `,
+      parser: 'babel-eslint'
+    },
+    {
       code: [
         'import type Props from "fake";',
         'class Hello extends React.Component {',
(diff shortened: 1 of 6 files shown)


jseminck:
Hmm - yes, I think that `default-props-match-prop-types` doesn't support it either. Will check.

jseminck:
Added support to `default-props-match-prop-types` as well. Enjoy!

This one was a bit harder because the code is structured completely differently from the other 2 rules, even though the props detection should handle all the same cases... hopefully I can fix that issue soon. 😄

tests/lib/rules/default-props-match-prop-types.js
```diff
@@ -683,6 +683,26 @@ ruleTester.run('default-props-match-prop-types', rule, {
...
+      code: `
+        type PropsA = { foo?: string };
+        type PropsB = { bar?: string, fooBar: string };
+        type Props = PropsA & PropsB;
+
+        class Bar extends React.Component {
+          props: Props;
+          static defaultProps = {
+            foo: "foo",
+            bar: "bar",
```
ljharb: Can you add a test case where a defaultProp is omitted for a `?` prop type, and there's no error? Alternatively, you could just remove the defaultProp for "bar" here.

LINKIWI:
Can confirm that this resolves #1364. Thanks for your work!

PR #1408, opened 2017-08-31
[make-no-direct-mutation-state] Handle update expressions

jackyho112:
For https://github.com/yannickcr/eslint-plugin-react/issues/1386, so that code such as 

```this.state.foo++```

or 

```this.state.foo--```

gets warned.

Let me know if there is any improvement you have in mind or any case I haven't covered.

### DIFF
--- a/lib/rules/no-direct-mutation-state.js
+++ b/lib/rules/no-direct-mutation-state.js
@@ -63,6 +63,25 @@ module.exports = {
         });
       },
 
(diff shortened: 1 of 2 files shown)


ljharb:
How does this PR compare to #1387, which already handles #1386?

jackyho112:
@ljharb Sorry, didn't notice. My bad. Closing this. :sweat_smile:

PR #1411, opened 2017-08-31
[no-unused-state] Add ability to handle arrow function class method

jackyho112:
For https://github.com/yannickcr/eslint-plugin-react/issues/1363, so that state property usage by object destructuring assignment in a class method defined with an arrow function gets recorded. Examples:

```js
class ArrowFunctionClassMethodTest extends React.Component {
  constructor() {
    this.state = { foo: 0 };
  }

  doSomething = () => {
    const { state: { foo } } = this;

    return foo;
  }

  render() {
    return <SomeComponent />;
  }
}
```

should not be warned.

Let me know if there is anything I can do to improve this PR. Also, just to confirm, this rule doesn't care about whether the class method is used or not, right?

### DIFF
--- a/lib/rules/no-unused-state.js
+++ b/lib/rules/no-unused-state.js
@@ -54,7 +54,7 @@ function getInitialClassInfo() {
     // shadowing, assignments, etc. To keep things simple, we only
     // maintain one set of aliases per method and accept that it will
     // produce some false negatives.
-    aliases: null
+    aliases: new Set()
   };
 }
 
--- a/lib/rules/no-unused-state.js
+++ b/lib/rules/no-unused-state.js
@@ -51,10 +51,10 @@ function getInitialClassInfo() {
 
     // Names of local variables that may be pointing to this.state. To
     // track this properly, we would need to keep track of all locals,
-    // shadowing, assignments, etc. To keep things simple, we only
-    // maintain one set of aliases per method and accept that it will
-    // produce some false negatives.
-    aliases: null
+    // shadowing, assignments, etc. To keep things simple, we clear the
+    // current set of aliases after each method definition and accept
+    // that it will produce some false negatives.
+    aliases: new Set()
   };
 }
 
@@ -86,7 +86,6 @@ module.exports = {
 
       const isAliasedStateReference =
         node.type === 'Identifier' &&
-        classInfo.aliases &&
         classInfo.aliases.has(node.name);
 
       return isDirectStateReference || isAliasedStateReference;
@@ -125,10 +124,7 @@ module.exports = {
       for (const prop of node.properties) {
         if (prop.type === 'Property') {
           addUsedStateField(prop.key);
-        } else if (
-          prop.type === 'ExperimentalRestProperty' &&
-          classInfo.aliases
-        ) {
+        } else if (prop.type === 'ExperimentalRestProperty') {
           classInfo.aliases.add(getName(prop.argument));
         }
       }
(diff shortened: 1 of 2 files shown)


lib/rules/no-unused-state.js
```diff
@@ -237,6 +237,26 @@ module.exports = {
...
           addStateFields(node.value);
         }
+
+        if (
+          !node.static &&
+          node.value &&
+          node.value.type === 'ArrowFunctionExpression'
+        ) {
+          // Create a new set for this.state aliases local to this function.
+          classInfo.aliases = new Set();
```
ljharb: i'm not seeing anything that adds to this set?

If it otherwise crashes because this doesn't exist; then let's initialize it in every case.

jackyho112: This list will only be added to if the property name gets assigned to a different variable name.  

For instance,

```js
class App extends React.Component {
        state = { foo: 0 };

        doSomething = () => {
          const { state: { foo: bar }} = this;

          return bar;
        }

        render() {
          return <SomeComponent />;
        }
      }
```

will add `'bar'` to the list. I have added one more test regarding this.

I also refactored how the code deals with `classInfo.aliases` according to your feedback. We still need to clear the set after each method definition though because each alias added is local to the class method. Let me know if this looks better.

jackyho112:
@ljharb 

This list will only be added to if the property name gets assigned to a different variable name.  

For instance,

```js
class App extends React.Component {
        state = { foo: 0 };

        doSomething = () => {
          const { state: { foo: bar }} = this;

          return bar;
        }

        render() {
          return <SomeComponent />;
        }
      }
```

will add `'bar'` to the list. I have added one more test regarding this.

I also refactored how the code deals with `classInfo.aliases` according to your feedback. We still need to clear the set after each method definition though because each alias added is local to the class method. Let me know if this looks better.

lib/rules/no-unused-state.js
```diff
@@ -239,20 +235,29 @@ module.exports = {
...
 
       'MethodDefinition:exit'() {
         if (!classInfo) {
           return;
         }
-        // Forget our set of local aliases.
-        classInfo.aliases = null;
+
+        // Remove the current set of aliases local to the previous class method
+        classInfo.aliases.clear();
```
ljharb: rather than storing across files and clearing; we may want to instead assign `classInfo.aliases = new Set()` at program entry?

jackyho112:
@ljharb 

I have restored the implementation back to using `classInfo.aliases = new Set()` which is a part of the original code. However, we still need to do `classInfo.aliases = new Set()` for every class method defintion because the aliases in `classInfo.aliases` are local to each class method. Let me know if that's ok.

jackyho112:
@ljharb 

Thanks for the review! After this is merged, I can take a look at documenting no-unused-state as I saw there is an issue open for it.

jackyho112:
[](url)

jackyho112:
One test for the `no-typos` rule is failing in the builds. Not sure why. Don't think it has anything to do with node versions.

jackyho112:
@ljharb 

I think this [test](https://github.com/jackyho112/eslint-plugin-react/commit/fa890cbad6d34658fda810288c4ab7765ec9a4f0#diff-a36b384ace30edac8eacb00c170eed54R247) is failing on master. Any clue? Maybe this test should have errors in the first place?

jseminck:
Thanks @jackyho112 

This is because of the new ESLint release. They changed some internal process and a case that we didn't handle before will now work.

I can update the case to be valid, but it would only work with the latest ESLint. With older cases it will still fail. Even more: We had to write some work-around to make sure ESLint did not crash but I can't remove that code unless we update the peerDependency for ESLint.

I also commented here: https://github.com/yannickcr/eslint-plugin-react/pull/1375#issuecomment-329899585

But I think I can start a new PR and we can start/continue a discussion there.

jseminck:
Created a PR to move that test: https://github.com/yannickcr/eslint-plugin-react/pull/1437

jackyho112:
@ljharb, @EvHaus, @yannickcr, @lencioni 

Decided to adhere to the original implementation because it is the safest approach, it is easier to review and if there is a new way of defining a class method, the rule won't break immediately. 

Now, this PR just merely adds the support for a class method defined by an arrow function based on how the original code intends to and does nothing else. I hope that is ok and should at least make this PR more mergable.

jackyho112:
@EvHaus, @ljharb 

Thanks for the approvals! 😄

PR #1415, opened 2017-09-06
Fix issues with IntersectionTypeAnnotation for prop-types and no-unused-prop-types

### LINKED ISSUE #1413: 7.4.0-rc.0 flow imported type intersection error - Cannot read property 'length' of undefined in prop-types.js
We are heavy flow users and are on flow 0.54.0, so I assume this may be related to @jseminck's changes in #1400. This may be an entirely new case or separate bug.

```
 yarn list eslint-plugin-react eslint                                                                                                                                                                   ⏎
yarn list v0.27.5
├─ eslint-plugin-react@7.4.0-rc.0
└─ eslint@4.5.0
```

```
git clone https://github.com/alienfast/material-ui.git
git checkout remove-class-property-props
yarn install
yarn lint
```

```
Cannot read property 'length' of undefined
TypeError: Cannot read property 'length' of undefined
    at iterateProperties (/Users/kross/projects/material-ui/node_modules/eslint-plugin-react/lib/rules/prop-types.js:365:21)
    at propTypes.types.forEach.annotation (/Users/kross/projects/material-ui/node_modules/eslint-plugin-react/lib/rules/prop-types.js:798:13)
    at markPropTypesAsDeclared (/Users/kross/projects/material-ui/node_modules/eslint-plugin-react/lib/rules/prop-types.js:796:27)
    at markAnnotatedFunctionArgumentsAsDeclared (/Users/kross/projects/material-ui/node_modules/eslint-plugin-react/lib/rules/prop-types.js:911:7)
    at Object.handleStatelessComponent (/Users/kross/projects/material-ui/node_modules/eslint-plugin-react/lib/rules/prop-types.js:920:7)
    at Linter.updatedRuleInstructions.(anonymous function) (/Users/kross/projects/material-ui/node_modules/eslint-plugin-react/lib/util/Components.js:666:75)
    at emitOne (events.js:120:20)
    at Linter.emit (events.js:210:7)
    at NodeEventGenerator.applySelector (/Users/kross/projects/material-ui/node_modules/eslint/lib/util/node-event-generator.js:265:26)
```

It appears to be failing when visiting `IntersectionTypeAnnotation`:

prop-types_js_-_material-ui_-____projects_material-ui_

How can I help narrow this down further?

jseminck:
Closes #1413

3 cases needed to be covered:

* Mutliple intersections:

```js
type PropsA = {}
type PropsB = {}
type PropsC = {}
type PropsD = PropsA & PropsB
type Props = PropsC & PropsD
```

* Importing types. This was already handled for other cases, but not for the intersection.
* `type Props = PropsA & { foo: string }`

### DIFF
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -474,7 +474,7 @@ module.exports = {
      * The representation is used to verify nested used properties.
      * @param {ASTNode} annotation Type annotation for the props class property.
      * @param {String} parentName Name of the parent prop node.
-     * @return {Object|Boolean} The representation of the declaration, true means
+     * @return {Object} The representation of the declaration, an empty object means
      *    the property is declared without the need for further analysis.
      */
     function buildTypeAnnotationDeclarationTypes(annotation, parentName) {
@@ -483,10 +483,10 @@ module.exports = {
           if (typeScope(annotation.id.name)) {
             return buildTypeAnnotationDeclarationTypes(typeScope(annotation.id.name), parentName);
           }
-          return true;
+          return {};
         case 'ObjectTypeAnnotation':
           if (skipShapeProps) {
-            return true;
+            return {};
           }
           const shapeTypeDefinition = {
             type: 'shape',
@@ -494,10 +494,7 @@ module.exports = {
           };
           iterateProperties(annotation.properties, (childKey, childValue) => {
             const fullName = [parentName, childKey].join('.');
-            let types = buildTypeAnnotationDeclarationTypes(childValue, fullName);
-            if (types === true) {
-              types = {};
-            }
+            const types = buildTypeAnnotationDeclarationTypes(childValue, fullName);
             types.fullName = fullName;
             types.name = childKey;
             types.node = childValue;
@@ -512,7 +509,7 @@ module.exports = {
           for (let i = 0, j = annotation.types.length; i < j; i++) {
             const type = buildTypeAnnotationDeclarationTypes(annotation.types[i], parentName);
             // keep only complex type
-            if (type !== true) {
+            if (Object.keys(type).length > 0) {
               if (type.children === true) {
                 // every child is accepted for one type, abort type analysis
                 unionTypeDefinition.children = true;
@@ -523,16 +520,13 @@ module.exports = {
             unionTypeDefinition.children.push(type);
           }
           if (unionTypeDefinition.children.length === 0) {
-            // no complex type found, simply accept everything
-            return true;
+            // no complex type found
+            return {};
           }
           return unionTypeDefinition;
         case 'ArrayTypeAnnotation':
           const fullName = [parentName, '*'].join('.');
-          let child = buildTypeAnnotationDeclarationTypes(annotation.elementType, fullName);
-          if (child === true) {
-            child = {};
-          }
+          const child = buildTypeAnnotationDeclarationTypes(annotation.elementType, fullName);
           child.fullName = fullName;
           child.name = '__ANY_KEY__';
           child.node = annotation;
@@ -542,7 +536,7 @@ module.exports = {
           };
         default:
           // Unknown or accepts everything.
-          return true;
+          return {};
       }
     }
 
@@ -741,10 +735,7 @@ module.exports = {
           return;
         }
 
-        let types = buildTypeAnnotationDeclarationTypes(value, key);
-        if (types === true) {
-          types = {};
-        }
+        const types = buildTypeAnnotationDeclarationTypes(value, key);
         types.fullName = key;
         types.name = key;
         types.node = value;
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -725,6 +725,50 @@ module.exports = {
     }
 
     /**
+     * Marks all props found inside IntersectionTypeAnnotation as declared.
+     * Since InterSectionTypeAnnotations can be nested, this handles recursively. 
+     * 
+     * Modifies the declaredPropTypes object
+     * @param {ASTNode} propTypes 
+     * @param {Object} declaredPropTypes 
+     * @returns {Boolean} True if propTypes should be ignored (e.g. when a type can't be resolved, when it is imported)
+     */
+    function declarePropTypesForIntersectionTypeAnnotation(propTypes, declaredPropTypes) {
+      let ignorePropsValidation = false;
+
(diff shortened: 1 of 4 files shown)


lib/rules/no-unused-prop-types.js
```diff
@@ -725,6 +725,70 @@ module.exports = {
...
+          if (!typeNode) {
+            ignorePropsValidation = true;
+          } else if (typeNode.type === 'IntersectionTypeAnnotation') {
+            ignorePropsValidation = declarePropTypesForIntersectionTypeAnnotation(typeNode, declaredPropTypes);
+          } else {
+            ignorePropsValidation = declarePropTypesForObjectTypeAnnotation(typeNode, declaredPropTypes);
+          }
+        }
+
+        return ignorePropsValidation;
```
ljharb: i can't decide if this style, or early-returning each place this var is assigned, is better

lib/rules/no-unused-prop-types.js
```diff
@@ -725,6 +725,70 @@ module.exports = {
...
+     * Marks all props found inside IntersectionTypeAnnotation as declared.
+     * Since InterSectionTypeAnnotations can be nested, this handles recursively. 
+     * 
+     * Modifies the declaredPropTypes object
+     * @param {ASTNode} propTypes 
+     * @param {Object} declaredPropTypes 
+     * @returns {Boolean} True if propTypes should be ignored (e.g. when a type can't be resolved, when it is imported)
+     */
+    function declarePropTypesForIntersectionTypeAnnotation(propTypes, declaredPropTypes) {
+      return propTypes.types.reduce((ignorePropsValidation, annotation) => {
```
ljharb: since this is reducing to a boolean, could it use `some` or `every`, and that way it will bail early?

lib/rules/no-unused-prop-types.js
```diff
@@ -725,6 +725,70 @@ module.exports = {
...
+      let ignorePropsValidation = false;
+
+      iterateProperties(propTypes.properties, (key, value) => {
+        if (!value) {
+          ignorePropsValidation = true;
+          return;
+        }
+
+        let types = buildTypeAnnotationDeclarationTypes(value, key);
+        if (types === true) {
```
ljharb: instead of reusing `types` here, can we use multiple vars and use `const`?

lib/rules/no-unused-prop-types.js
jseminck: Honestly I'm in general not really happy about the fact that a function with a name `declarePropTypes***` returns a boolean at all. It's a bit confusing. That's why I liked this style a bit more, because it's very explicit to say what those methods return and what the boolean actually means.

Perhaps a `try/catch` may be even better?

lib/rules/no-unused-prop-types.js
jseminck: This is from the existing code so didn't touch it yet. But I think we can simply do: 

```js
const types = buildTypeAnnotationDeclarationTypes(value, key) || {};
declaredPropTypes.push({
  ...types,
  fullName: key,
  name: key,
  node: value
});
```

jseminck: Meh, another problem is that `buildTypeAnnotationDeclarationTypes` returns either a boolean or an object which isn't great either. I try to refactor it to always return an object, but it needs changes in more places than just here. 

But I do think it's a good refactor to do 😄

lib/rules/no-unused-prop-types.js
ljharb: We could rename the function instead :-)

jseminck:
I started some refactoring based on the comments, but now this currently valid scenario becomes invalid. But IMHO this should be invalid because `firstname` is indeed not declared.

WDYT?

```js
type Props = {name: Object;};
  props: Props;
  render () {
    return <div>Hello {this.props.name.firstname}</div>;
  }
}
```

jseminck:
Made some additional changes. Did a smoke test on the material-ui code which is currently breaking with rc0. I changed those functions to return early, but haven't changed the name as I couldn't come up with something.

@yannickcr please also have a look as this is required for 7.4.0.

lib/rules/prop-types.js
```diff
@@ -261,7 +262,7 @@ module.exports = {
           // If it's a computed property, we can't make any further analysis, but is valid
           return key === '__COMPUTED_PROP__';
         }
-        if (propType === true) {
+        if (typeof propType === 'object' && Object.keys(propType).length === 0) {
```
ljharb: What is the typeof guarding against here? This allows `null`, for example.

lib/rules/no-unused-prop-types.js
```diff
@@ -725,6 +713,60 @@ module.exports = {
...
+    /**
+     * Marks all props found inside IntersectionTypeAnnotation as declared.
+     * Since InterSectionTypeAnnotations can be nested, this handles recursively. 
+     * 
+     * Modifies the declaredPropTypes object
+     * @param {ASTNode} propTypes 
+     * @param {Object} declaredPropTypes 
+     * @returns {Boolean} True if propTypes should be ignored (e.g. when a type can't be resolved, when it is imported)
+     */
+    function declarePropTypesForIntersectionTypeAnnotation(propTypes, declaredPropTypes) {
```
ljharb: Can these two functions live somewhere they can be shared across the rules that currently duplicate them?

jseminck: Yes. I'm working on it 😄 The idea will be that the rule will be declared as:

```
create: propTypes.detect((context, components, utils) => {
...
})
```

And this will take care of ALL the shared code between `prop-types` and `no-unused-prop-types`. Perhaps also for the other rules.

lib/rules/prop-types.js
jseminck: `null` would be checked in the if statement before:

```
if (!propType) {
    // If it's a computed property, we can't make any further analysis, but is valid
    return key === '__COMPUTED_PROP__';
}
```

Otherwise, if I remove the `typeof` condition, one test starts failing. Turns out `propType` in this case is a `function()` (not sure why) and it satifies the condition `Object.keys(propType).length === 0`. Perhaps should use https://www.npmjs.com/package/lodash.isplainobject ?

```
function Hello(props) {
  return <div>{props.name.constructor.firstname}</div>
}
Hello.propTypes = {
  name: PropTypes.shape({
    firstname: PropTypes.object
  })
};:

AssertionError: Should have 1 error but had 0: []
```

ljharb: Gotcha; makes sense.

PR #1416, opened 2017-09-07
Add no-mutation-props

### LINKED ISSUE #1113: Rule idea: no-direct-mutation-props

joeybaker:
Fixes #1113

This is based on the work in https://github.com/yannickcr/eslint-plugin-react/pull/1145 but reports on many more cases.

Now warns on:

* array mutations
* `Object.assign`
* `Object.defineProperty`
* `delete`
* all of the above also works with array and object destructuring and
variable assignment


joeybaker:
Once this is in a good place, it's probably a good idea to use the same logic in the `no-direct-mutation-state` rule.

lib/rules/no-mutation-props.js
```diff
@@ -0,0 +1,193 @@
...
+            // if this is object destructuring `const {b} = this.props`
+            || (id.type === 'ObjectPattern' && id.properties.map(property => property.value))
+            // if this is array destructuring `const [b] = this.props.list`
+            || (id.type === 'ArrayPattern' && id.elements);
+
+          // remember that this node is a prop, `isFromProps` relies on this
+          assignments.forEach(assignment => {
+            components.add(assignment);
+            components.set(assignment, {
+              isFromProps: true
```
jseminck: I'm not sure how I feel about using the `components` as a storage for non-component nodes... Maybe we can use some local collection instead?

joeybaker: Fair enough, I was just trying to re-use the logic since what I need is all there. But, I'm happy to copy/paste what I need into here?

jseminck:
Can you also add support for `this.props.foo++`? There's a PR that adds this to the `no-direct-mutation-state` rule, so I think we'll want it here too 😄 

https://github.com/yannickcr/eslint-plugin-react/pull/1387

On the other hand, this rule supports other cases that `no-direct-mutation-state` supports, e.g. those array mutations. I'm not sure what would be a good way to share that code.

lib/rules/no-mutation-props.js
jseminck: Something like:

```
          assignments.forEach(assignment => {
            allAssignments.push(assignment.range.join(':'));
          });
```

And where it's checked:

```
        const currentIdentifier = allAssignments.indexOf(originalIdentifiers[originalIdentifiers.length - 1].range.join(':')) >= 0;
        return currentIdentifier;
```

This way, at the end of the file there will only be the actual components inside the list, and we skip looping over all those assignments. For all tests, list would only contain 1 value.

```
        Object.keys(list).forEach(key => {
          const component = list[key];

          reportMutations(component);
        });
```

What do you think?

joeybaker:
Yea, `++` and `--` are good calls. Can do.

I'd love to share code between this and `no-direct-mutation-state`. Perhaps the way to do that is to add `isPropsNode` and the new non-`components` set/get code to `utils`?

joeybaker:
@jseminck changes made. I'd be happy to look at sharing code between this rule and `no-direct-mutation-state`, but I suggest we do that in a separate PR that focuses on improving `no-direct-mutation-state`.

jseminck:
Yeah. I agree!

Looks good for me. It's a pleasure to read your code. 👍 Someone else should also still go over the code since I'm not an official maintainer though!

docs/rules/no-mutation-props.md
```diff
@@ -0,0 +1,63 @@
...
+  render: function() {
+    this.props.name = this.props.name.toUpperCase();
+    return <div>Hello {this.props.name}</div>;
+  }
+});
+
+var Hello = React.createClass({
+  render: function() {
+    const {list} = this.props;
+    list.push(2);
```
ljharb: i don't see how this can safely be considered a mutation; `list` could be anything with a "push" method, there's no way to know it's an array.

docs/rules/no-mutation-props.md
```diff
@@ -0,0 +1,63 @@
...
+var Hello = React.createClass({
+  render: function() {
+    delete this.props.foo;
+    return <div>{Object.keys(this.props).length} props</div>;
+  }
+});
+
+var Hello = React.createClass({
+  render: function() {
+    Object.assign(this.props.foo, {bar: 'baz'})
```
ljharb: If we're going to support builtins like `Object.assign` and `Object.defineProperty` (where's `Object.defineProperties` btw), we'd need to support a way to allow this to be extensible - in other words, to allow devs to specify a custom function name, and *which argument(s) are mutated* as well.

I don't think we should support functions in this way in the initial version of the rule.

docs/rules/no-mutation-props.md
```diff
@@ -0,0 +1,63 @@
...
+    return <div>{this.props.foo.bar}</div>;
+  }
+});
+
+var Hello = React.createClass({
+  render: function() {
+    Object.defineProperty(this.props, 'foo')
```
ljharb: please ensure that all examples properly use semicolons.

lib/rules/no-mutation-props.js
```diff
@@ -0,0 +1,222 @@
+/**
+ * @fileoverview Prevent mutation of this.props
+ * @author Ian Schmitz, Joey Baker
+ */
+'use strict';
+
+const Components = require('../util/Components');
+const ARRAY_MUTATIONS = ['push', 'pop', 'shift', 'unshift'];
```
ljharb: although i don't think we can safely check for any of these, you're also missing the other array mutator methods: `splice`, `reverse`, `sort`, `fill`, `copyWithin`.

lib/rules/no-mutation-props.js
```diff
@@ -0,0 +1,222 @@
...
+    const getId = node => node && node.range.join(':');
+
+    const propsNodes = {};
+    /**
+     * Adds a node to the list of nodes that come from this.props
+     * @param {Object} node The node to add
+     */
+    const addToPropsNodes = node => {
+      const id = getId(node);
+      if (propsNodes[id]) {
```
ljharb:
```js
if (!propsNodes[id]) {
  propsNodes[id] = node;
}
return propsNodes[id];
```

lib/rules/no-mutation-props.js
```diff
@@ -0,0 +1,222 @@
...
+      recommended: false
+    }
+  },
+
+  create: Components.detect((context, components, utils) => {
+    /**
+     * Gets a unique ID for a node
+     * @param {Object} node The node to get an ID for
+     */
+    const getId = node => node && node.range.join(':');
```
ljharb: what happens if the range only has one item, and that item is `toString`?

lib/rules/no-mutation-props.js
```diff
@@ -0,0 +1,222 @@
...
+
+    const propsNodes = {};
+    /**
+     * Adds a node to the list of nodes that come from this.props
+     * @param {Object} node The node to add
+     */
+    const addToPropsNodes = node => {
+      const id = getId(node);
```
ljharb: `getId` can return a falsy value - that would get coerced to a string here, and probably provide undesired results.

ljharb (changes requested):

lib/rules/no-mutation-props.js
joeybaker: I'm not sure I know what you mean? I didn't think it was possible for range to have only one item? This logic is copied from https://github.com/yannickcr/eslint-plugin-react/blob/f1e86b5ba32ed9359eb7823b98f7a1dcc52376e3/lib/util/Components.js#L43-L45

lib/rules/no-mutation-props.js
joeybaker: Sure? It would just try to find a falsey value key which would be `undefined`. I don't think there's any harm, but I agree it's safer and slightly faster to add a guard.

lib/rules/no-mutation-props.js
joeybaker: Make sense.

docs/rules/no-mutation-props.md
joeybaker: That's a great point. I'd hate to loose the functionality though. Any thoughts on how we could determine it's an array?

docs/rules/no-mutation-props.md
joeybaker: I'm not sure I follow? `Object.assign` is never safe to use when `this.props` is the first argument, no?

joeybaker: `Object.defineProperties` is a good catch! Added.

lib/rules/no-mutation-props.js
joeybaker: Added, but I'll hold of on adding tests until we resolve what to do here.

lib/rules/no-mutation-props.js
joeybaker: Done.

lib/rules/no-mutation-props.js
ljharb: Fair, but since it's being used as an object key, without any prefix, there's still the possibility that it could conflict with an Object.prototype key.

joeybaker: Sorry, I'm still not following you. Will the AST allow for `range` to be anything other than a tuple of integers?

joeybaker:
@ljharb I appreciate you taking a look. I'm sure you're busy, but I'd like to work with you to get this merged. I believe it's a nice feature. At the very least, it would help my team quite a bit.

docs/rules/no-mutation-props.md
joeybaker: Meanwhile, [on Gitter](https://gitter.im/eslint/eslint?at=59b1e7a8bac826f05490b79b)… 

@not-an-aardvark suggests that leave this rule as-is since these method names tend to imply mutation anyway (also, it would be pretty weird to have a `prop` that isn't a plain object). Further, people can disable custom use-cases with an inline `eslint-disable`.

If we want to be very safe, we could put this behind an option flag?

ljharb: I think putting it behind an option is a great compromise - what about two booleans, and two lists of 1) instance methods, and 2) static methods? That way the instance method list could default to `ARRAY_MUTATIONS`, and the static method list could default to `['Object.assign', 'Object.defineProperties', 'Object.defineProperty']` - and both booleans would default to `false`.

That way most people wouldn't touch the lists at all, they'd just enable the booleans - but those who wanted to update the lists could do so.

lib/rules/no-mutation-props.js
ljharb: Yes, I'm just not that familiar with what the AST could return. If `range` always contains > 1 item, then this is totally safe when `node` is truthy.

lib/rules/no-mutation-props.js
```diff
@@ -0,0 +1,227 @@
...
+     * @param {Object} node The node to examine
+     */
+    function isPropsNode (node) {
+      if (!node) {
+        return false;
+      }
+
+      let topObject = node;
+
+      // when looking at an object path, get the top-most object
```
ljharb: this might be helpful to move to a helper method, so that this function's code can be simpler.

lib/rules/no-mutation-props.js
```diff
@@ -0,0 +1,227 @@
...
+        return currentIdentifier;
+      }
+
+      return false;
+    }
+
+    function setPropMutation(mutation, node) {
+      const component = components.get(utils.getParentComponent());
+      const propMutations = component && component.propMutations
+        ? component.propMutations.concat([mutation])
```
ljharb: is `mutation` an array? If not, the wrapping brackets aren't needed.

lib/rules/no-mutation-props.js
```diff
@@ -0,0 +1,227 @@
...
+        const currentIdentifier = getPropsNode(originalIdentifiers[originalIdentifiers.length - 1]);
+        return currentIdentifier;
+      }
+
+      return false;
+    }
+
+    function setPropMutation(mutation, node) {
+      const component = components.get(utils.getParentComponent());
+      const propMutations = component && component.propMutations
```
ljharb: Can we make `component.propMutations` always start out an empty array, so we don't need this check?

lib/rules/no-mutation-props.js
joeybaker: Yea, the spec says `range` must be an array with 2 integers: https://eslint.org/docs/developer-guide/working-with-plugins#all-nodes

docs/rules/no-mutation-props.md
joeybaker: I'm with you on the array mutations, but I'm not sure I understand the danger of the static methods. Can you run that by me one more time? When is `Object.assign(this.props, {…})` ever a safe operation?

lib/rules/no-mutation-props.js
joeybaker: I hear you, but since this is only used in one place, I'm not sure the abstraction is worth it.

lib/rules/no-mutation-props.js
joeybaker: Could do, but that feels like unnecessary baggage across the rest of the rules if someone opts to not enable this rule. Thoughts?

lib/rules/no-mutation-props.js
joeybaker: No, mutation is not an array, it a node.

joeybaker:
@ljharb okay, I've added an option for `allowArrayMutation`, but I'd love more clarification on when the static `Object` methods would ever be okay.

lib/rules/no-mutation-props.js
ljharb: I don't imagine that an empty array will slow things down much, compared the the speedup and clarity of not having to ever check for truthiness.

ljharb:
I'm thinking a schema like this:
```json
{
  "enableInstanceMethods": true,
  "instanceMethods": ["push", "shift"],
  "enableStaticMethods": true,
  "staticMethod": ["Object.assign", "Object.defineProperty"]
}
```

joeybaker:
Interesting. A few questions:

1. I went with `enableInstanceMethods` over `allowArrayMutation`? "Instance methods" seem a little vague to me.
2. Why bother with the separate arrays and booleans? Seems like if you provide an array, the boolean value is superfluous.
3. I'm still not clear on the danger of the object static methods. Could you give me an example when `Object.assign(this.props, {})` would ever be safe?

ljharb:
The value in providing both a boolean and an array is that a shared config (like the airbnb config) can specify the array without having to enable it, and users can enable it without having to copy/paste the array.

`Object.assign` is always safe when an object literal is the first argument :-) Regardless, `Object.assign = function () { return 42; }` would be totally safe and mutation-free.

joeybaker:
Fair enough on the config option. Will change.

I really don't see the value in accommodating people who override the language builtins. Adding a config option will make the documentation a little more confusing, but your call. I'll go ahead and add it.

lib/rules/no-mutation-props.js
joeybaker: Done.

joeybaker:
@ljharb options added, and all feedback addressed!

docs/rules/no-mutation-props.md
```diff
@@ -0,0 +1,115 @@
...
+  "allowableArrayMutations": <array<string>>,
+  "allowObjectStatics": <boolean>,
+  "allowableObjectStatics": <array<string>>,
+}]
+...
+```
+
+### `allowArrayMutations`
+
+When `true` the rule ignores [array mutation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#Mutator_methods) of props.
```
ljharb: Boolean options should always default to `false`, so that absence and `false` are equivalent - this implies that anyone who hasn't passed an explicit `true` here, will get the warnings.

Regardless of what you want the default behavior to be, the booleans should be named such that they can default to `false`.

Separately, why do you think this behavior should be on by default, especially since there's caveats?

docs/rules/no-mutation-props.md
```diff
@@ -0,0 +1,115 @@
...
+this.props.list.push(1)
+```
+
+### `allowablArrayMutations`
+
+When `allowArrayMutation: true` the rule specifies specific [array mutations](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#Mutator_methods) of props to allow. Mutations not listed will remain as warnings.
+
+You must used this option in conjunction with `allowArrayMutation`. You should only use this option if you have a prop that contains an object with methods that conflict with array mutation methods.
+
+### `allowObjectStatics`
```
ljharb: I don't think this should be restricted to `Object` statics - there's `Reflect` statics that mutate too, and users might provide a single identifier (like `lodash.assign`) that they'd want to be warned on as well.

ljharb (changes requested):

docs/rules/no-mutation-props.md
joeybaker: Agreed with the first part. This does default to `false`, I can specify that if you think it needs to be clearer.

I like my tools to work out of the box. I really think that both of the cases you outlined are corner cases.

The chances of a `prop` that's an object, with methods, that also has methods named the same as array mutation methods are low. Frankly, doing so seems like bad practice (outside of immutable.js objects, but you shouldn't be calling the mutation methods in a component). 

Similarly, if you choose to override a language builtin like `Object.assign`, it's think it's reasonable for your linter to complain. As we've discussed, I prefer not to even have the object statics option – I think it encourages bad behavior.

The options are here to help you get around extremely unusual circumstances, and I struggle to see a case where they should ever be used.

docs/rules/no-mutation-props.md
joeybaker: `Reflect` is a good call! I forgot about that. I'll add that in.

Unfortunately, this isn't setup for arbitrary `lodash.assign` calls, but I can see that being a good feature. However, adding support for that is quite a bit of work. I suggest that we look at that in a issue and see if that's something the community wants/expects.

docs/rules/no-mutation-props.md
ljharb: Defaults need to be safe; the instance methods in particular are not safe by default.

If the chance is nonzero, it's not an acceptable amount.

I'd be fine defaulting the static methods to be enabled, because as you say, the only way that'd be a problem is if someone intentionally broke the environment.

docs/rules/no-mutation-props.md
ljharb: That was my intention with allowing the option - I should be able to put any `foo.bar` string in there and have it work. Specifically, I want it to work for `import assign from 'object.assign'; assign()`, and `import lodash from 'lodash'; lodash.assign()`.

joeybaker: I gotcha, the issue is that parsing the AST to determine that is more difficult. As I say, I'd love to support that, but it feels like something that can be done on the next iteration of this rule?

docs/rules/no-mutation-props.md
joeybaker: Hmm… I'd argue it's more important for defaults to solve the common case, even if there's a little bit of danger involved. If I were using this rule, I wouldn't expect array mutations to not be reported.

I appreciate that point of view, and I'm not sure what your preferred method of resolving issues like this are. Maybe we can ask another maintainer to weigh in?

joeybaker:
@ljharb Okay, `Reflect` mutations added. Where do we stand?

joeybaker:
@ljharb 👋

docs/rules/no-mutation-props.md
ljharb: Until other maintainers have weighed in, my vote remains that instance method mutations (which is a better name than "array method", and I'd like that changed also) are not safe to enable by default.

Anything less than 100% correct isn't a safe default.

docs/rules/no-mutation-props.md
```diff
@@ -0,0 +1,135 @@
...
+Object.defineProperties(this.props.foo, [{foo: {value: 'bar'}}])
+```
+
+### `allowableObjectStatics`
+
+When `allowObjectStatics: true` the rule specifies specific [`Object` methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/object#Methods_of_the_Object_constructor) of `Object` to allow. Mutation methods not listed will remain as warnings.
+
+You must used this option in conjunction with `allowObjectStatics`. You should only use this option if you've overridden the language built-in `Object`.
+
+### `allowReflectStatics`
```
ljharb: Instead of Object and Reflect statics, this really needs to just be "statics" - all of them should appear in the same list, and I should be able to define other static methods on other objects (like eslint's `no-restricted-properties` rule) and have it work.

ljharb (changes requested):

docs/rules/no-mutation-props.md
joeybaker: I understand and want the same thing, but it's significantly more complex to support arbitrary objects and their statics. It seems to me that that this API, while not ideal, is a good v1. We can add complexity down the road.

joeybaker:
I'm unsure of who else to ping, so @yannickcr @jseminck @jackyho112  what are your thoughts on https://github.com/yannickcr/eslint-plugin-react/pull/1416#discussion_r139571616

docs/rules/no-mutation-props.md
ljharb: eslint core handles it with `{ object: 'Object', property: 'assign', message: 'some custom message' }` - I'm not sure why we couldn't handle it the same way (as `no-restricted-properties`)

joeybaker: Sure! That would work for `Reflect` and `Object` (though I'm not sure it's a more clear API), but it wouldn't enable lodash, or any other non-built-in as far as I know. Unless there's something I'm missing?

ljharb: `{ property: 'lodash' }` is used in eslint core to refer to the bare identifier.

joeybaker: Oh, interesting. So, that doesn't actually guarantee that the object is imported from lodash. e.g. it could throw a false positive on something like `const lodash = {assign: noop}`, and wouldn't catch `import _ from 'lodash'`, nor `import {assign} from 'lodash'`. I can do that, but it feels a little gross. Thoughts?

joeybaker: (I get that you could set `_.assign` in the `statics` array, but as I say … it's a bit gross)

ljharb: Yes, but I don't think there's much of a way around that, since adding awareness of imports/requires would be a bit impractical.

joeybaker: Okay, fair enough. I'll add that! :)

docs/rules/no-mutation-props.md
```diff
@@ -87,49 +93,29 @@ The following patterns are considered okay and do not cause warnings:
...
 
-### `allowObjectStatics`
+### `disabledMethods`
 
-When `true` the rule ignores [`Object` methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/object#Methods_of_the_Object_constructor) that can mutation props.
+By default, this is set to: `['Object.assign', 'Object.defineProperty', 'Object.defineProperties', 'Reflect.defineProperty', 'Reflect.deleteProperty', 'Reflect.set', '_.fill', '_.reverse', '_.assign', '_.extend', '_.assignIn', '_.assignInWith', '_.extendWith', '_.assignWith', '_.defaults', '_.defaultsDeep', '_.merge', '_.mergeWith', '_.set', '_.setWith']`. This contains language built-ins that can mutate objects (see: [`Object` methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/object#Methods_of_the_Object_constructor) and [`Reflect` mutation methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Reflects/Reflect#Methods)) as well as methods on [lodash](https://lodash.com) that will mutate either arrays or objects. You can override this with your own set of object/method pairs.
```
ljharb: I think all the lodash/underscore methods should not be included by default; that's presuming that more people use those libs than do.

Tests should certainly ensure these would work, ofc :-)

ljharb (changes requested):

docs/rules/no-mutation-props.md
joeybaker: I figured you'd say that, so I held off on tests ;)

Okay, I'll move them out, and just have them in the docs for easy reference.

joeybaker:
@ljharb okay, options have been changed up and lodash methods removed. We now have just a `disabledMethods` option instead of 2 separate settings for each of `Object` and `Reflect`.

@yannickcr @jseminck @jackyho112 The githubs tell me that you're also maintainers on this package. I'd love to get your thoughts on disabling array mutations on `this.props` by default.

@ljharb makes the point that we can't do this with 100% safety. It's true, but I contend that the possibility of a `prop` that is not a plain object, that has methods named the same as array mutators, that in turn need to be called within a component is extremely limited. However, in the case it does occur, we've provide an option to disable that default.

We're at a bit of a deadlock, so any of you weighing in would be greatly appreciated. In general, disallowing `this.props` mutation would be really nice feature for my team (and I imagine many others) so I'd like to get this merged and released!

jseminck:
I'm just a contributor 😄 I'll try to go through the whole PR tomorrow and see if I can weigh in, but honestly from following this from the side I have to admit some of the technical discussions are a bit above my technical knowledge level 😞

jackyho112:
@joeybaker 

I am merely just a contributor as well. I am not sure if I will have time to review it this week, but I will try my best. 😄

joeybaker:
@jseminck @jackyho112 whoops sorry about that! Opinions are very welcome though!

ljharb:
@joeybaker it's been awhile, but if you're still interested in landing this, it'd be great if you rebased it and marked it as ready for review :-)

oteoe:
@joeybaker @ljharb 
any news on this?

PR #1437, opened 2017-09-17
Move valid test case to invalid for no-typos due to ESLint upgrade

jseminck:
With the latest release of ESLint, a case that was previously not supported is now supported. This currently fails tests on master and also for any PRs that are created.

We can't remove the work-around code yet though, because it would mean an upgrade of the ESLint peerDependency which would probably mean a major version bump?

We should remove that work-around code only when a new version is planned, right?

### DIFF
--- a/tests/lib/rules/no-typos.js
+++ b/tests/lib/rules/no-typos.js
@@ -239,15 +239,6 @@ ruleTester.run('no-typos', rule, {
     ].join('\n'),
     parserOptions: parserOptions
   }, {
-    // PropTypes declared on a component that is detected through JSDoc comments and is
-    // declared AFTER the PropTypes assignment does not work.
-    code: `
-      MyComponent.PROPTYPES = {}
-      /** @extends React.Component */
-      class MyComponent extends BaseComponent {}
-    `,
-    parserOptions: parserOptions
-  }, {
     // https://github.com/yannickcr/eslint-plugin-react/issues/1353
     code: `
       function test(b) {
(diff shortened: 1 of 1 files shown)


tests/lib/rules/no-typos.js
```diff
@@ -763,5 +754,17 @@ ruleTester.run('no-typos', rule, {
     }, {
       message: 'Typo in declared prop type: objectof'
     }]
+  }, {
+  // PropTypes declared on a component that is detected through JSDoc comments and is
+  // declared AFTER the PropTypes only works when ESLint version is >= 4.7.0
```
ljharb: We should support both eslint < v4.7, and >= v4.7, and this code should be made to fail as expected in both ranges.

jseminck: Hmm - not sure how we can make it fail with ESLint >= 4.7 😄 

It does feel a bit strange to really force a working case to be invalid though? Currently we do support both version ranges, but the plugin behaves differently in both... but yes, I agree that's kind of confusing as well.

I'll see what can be done.

ljharb: To clarify; I think that the code warning or not warning should behave identically in 4.6 and 4.7 - I don't really care which we pick, as long as it doesn't change.

If that's not possible, then that will mean 4.7 was a breaking change, and eslint core needs to revert something so that it again works as expected.

jseminck: We definitely cannot easily detect this case with ESLint 4.6. That's why we had to choose to ignore it until the ESLint issue fixed.

Indeed, 4.7 did some kind of internal refactor of ESLint core. If you're interested, this is the change: https://github.com/eslint/eslint/pull/9283

ljharb: in that case, we need to explicitly ignore it in eslint `^4.7`.

not-an-aardvark:
This change has been reverted in `eslint@4.7.2`, so your tests should be passing again. Sorry about the breakage.

ljharb:
Great! Sounds like this can be closed for the time being.

PR #1441, opened 2017-09-18
[New] Rule jsx-no-uppercase-data-aria-attributes

donabrams:
As per https://facebook.github.io/react/docs/dom-elements.html , `data-*` and `aria-*` attributes should be lowercase. This rule enforces that.

Justification to be added to recommended rules: Without this rule, `data-myProp` and `aria-describedBy` will both throw warnings. It's also pretty lightweight.

Small strange thing: `no-typos.js` tests were failing on master. I had to guess at the intention, but unsure how `master` was passing tests before.


jseminck:
There's a PR for the `no-typos` test, so you can actually ignore it.

jseminck:
Is this still valid with React 16? https://facebook.github.io/react/blog/2017/09/08/dom-attributes-in-react-16.html

It seems the data attribute is no longer ignored, but it does get lowercased: https://codesandbox.io/s/6nx6p1r0rn

screen shot 2017-09-18 at 23 19 07

lib/rules/jsx-no-uppercase-data-aria-attributes.js
```diff
@@ -0,0 +1,39 @@
...
+ * @author Don Abrams
+ */
+'use strict';
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+function isDataOrAriaAttributeWithUppercase(attr) {
+  return (attr.name.name.indexOf('data-') === 0 ||
```
jseminck: `.startsWith()` ?

donabrams:
This is definitely still useful to avoid warnings; however it's not as critical as it was in React < 16.
screen shot 2017-09-18 at 3 44 18 pm

ljharb (changes requested):
Please add more test cases that illustrate that this rule only applies to host components (HTML), and not to custom components.

Separately, we should probably add an option to define an array of custom component names that should be *treated* like they're HTML components (it would default to `[]`).

However, I don't think this rule belongs in eslint-plugin-react at all - are you familiar with https://github.com/evcohen/eslint-plugin-jsx-a11y/ ?

donabrams:
The `aria-*` restriction could go in `js-a11y`, but the `data-*` rules would be irrelevant there. This is also clearly a react/jsx constraint and not an accessibility helper which is why it's documented in the 2nd paragraph of https://facebook.github.io/react/docs/dom-elements.html and not just in https://facebook.github.io/react/docs/accessibility.html

Currently this rule does apply to custom AND host components. I could have restricted it to just custom components, but the performance would suffer and complexity of the rule would increase with little reward: I find it very unlikely that custom components would have an attribute key named `data-*` or `aria-*` except as pass-throughs to HTML components.

ljharb:
Unlikely isn't impossible, however.

Using semantic HTML attributes is still part of a11y, including data- attributes.

PR #1458, opened 2017-09-30
[jsx-curly-brace-presence] Bail out in warning JSX expression when some chars exist

### LINKED ISSUE #1449: jsx-curly-brace-presence removes braces that are required
```
<style type="text/css">{'.mt0 { margin-top: 0; }'}</style>
```
with the `7.4.0` release these braces are removed and then eslint chokes on the syntax.

### LINKED ISSUE #1479: Template string with special character are getting inappropriately autofix to quoted string
```<input placeholder={`line1\nline2`} ></input>``` gets autofix into 
`<input placeholder="line1\nline2" />` which treats the `\n` literally. 

Versions:
    "eslint-config-airbnb": "^16.0.0",
    "eslint-plugin-import": "^2.7.0",
    "eslint-plugin-jsx-a11y": "^6.0.2",
    "eslint-plugin-react": "^7.4.0"

### PR DESCRIPTION
**Pull request by jackyho112**
This is an attempt to fix https://github.com/yannickcr/eslint-plugin-react/issues/1449, so that the jsx-curly-brace-presence rule, when set to get rid of unnecessary curly braces(JSX expression), will bail out or deem a JSX expression necessary when a character that needs to be escaped in its JSX form is found in the expression. It also improves the fix in enforcing curly braces by escaping backslashes in its JSX form.

Gigantic thanks to @lydell for suggesting the implementation [plan](https://github.com/yannickcr/eslint-plugin-react/issues/1449#issuecomment-332750344) for this PR. I hope I didn't screw it up too much. 😄 

Let me know if there is any improvement you have in mind.

Fixes #1479. Fixes #1449.

### DIFF
--- a/lib/rules/jsx-curly-brace-presence.js
+++ b/lib/rules/jsx-curly-brace-presence.js
@@ -59,12 +59,37 @@ module.exports = {
       {props: ruleOptions, children: ruleOptions} :
       Object.assign({}, DEFAULT_CONFIG, ruleOptions);
 
-    function containsBackslashForEscaping(rawStringValue) {
+    function containsBackslash(rawStringValue) {
       return rawStringValue.includes('\\');
+    }
+
+    function containsHTMLEntity(rawStringValue) {
+      return rawStringValue.match(/&([A-Za-z\d]+);/g);
+    }
+
+    function containsDisallowedJSXTextChars(rawStringValue) {
+      return rawStringValue.match(/{|<|>|}/g);
+    }
+
+    function containsQuoteCharacters(value) {
+      return value.match(/'|"/g);
     }
 
     function escapeDoubleQuotes(rawStringValue) {
       return rawStringValue.replace(/\\"/g, '"').replace(/"/g, '\\"');
+    }
+
+    function escapeBackslashes(rawStringValue) {
+      return rawStringValue.replace(/\\/g, '\\\\');
+    }
+
+    function needToEscapeCharacterForJSX(raw, cooked) {
+      return (
+        containsBackslash(raw) ||
+        containsHTMLEntity(raw) ||
+        containsDisallowedJSXTextChars(raw) ||
+        containsQuoteCharacters(cooked)
+      );
     }
 
     /**
@@ -83,11 +108,10 @@ module.exports = {
 
           let textToReplace;
           if (parentType === 'JSXAttribute') {
-            textToReplace = `"${escapeDoubleQuotes(
-              expressionType === 'TemplateLiteral' ?
-                expression.quasis[0].value.raw :
-                expression.raw.substring(1, expression.raw.length - 1)
-            )}"`;
+            textToReplace = `"${expressionType === 'TemplateLiteral' ?
+              expression.quasis[0].value.raw :
+              expression.raw.substring(1, expression.raw.length - 1)
+            }"`;
           } else {
             textToReplace = expressionType === 'TemplateLiteral' ?
               expression.quasis[0].value.cooked : expression.value;
@@ -103,10 +127,16 @@ module.exports = {
         node: literalNode,
         message: 'Need to wrap this literal in a JSX expression.',
         fix: function(fixer) {
+          // Leave it to the author to fix as it can be fixed by either a
+          // real character or unicode
+          if (containsHTMLEntity(literalNode.raw)) {
+            return null;
+          }
+
           const expression = literalNode.parent.type === 'JSXAttribute' ?
-            `{"${escapeDoubleQuotes(
+            `{"${escapeDoubleQuotes(escapeBackslashes(
               literalNode.raw.substring(1, literalNode.raw.length - 1)
-            )}"}` :
+            ))}"}` :
             `{${JSON.stringify(literalNode.value)}}`;
 
           return fixer.replaceText(literalNode, expression);
@@ -114,23 +144,25 @@ module.exports = {
       });
     }
 
+    // Bail out if there is any character that needs to be escaped in JSX
+    // because escaping decreases readiblity and the original code may be more
+    // readible anyway or intentional for other specific reasons
     function lintUnnecessaryCurly(JSXExpressionNode) {
       const expression = JSXExpressionNode.expression;
       const expressionType = expression.type;
-      const parentType = JSXExpressionNode.parent.type;
 
       if (
         expressionType === 'Literal' &&
-          typeof expression.value === 'string' && (
-          parentType === 'JSXAttribute' ||
-          !containsBackslashForEscaping(expression.raw))
+          typeof expression.value === 'string' &&
+          !needToEscapeCharacterForJSX(expression.raw, expression.value)
       ) {
         reportUnnecessaryCurly(JSXExpressionNode);
       } else if (
         expressionType === 'TemplateLiteral' &&
-          expression.expressions.length === 0 && (
-          parentType === 'JSXAttribute' ||
-          !containsBackslashForEscaping(expression.quasis[0].value.raw))
+          expression.expressions.length === 0 &&
+          !needToEscapeCharacterForJSX(
+            expression.quasis[0].value.raw, expression.quasis[0].value.cooked
+          )
       ) {
         reportUnnecessaryCurly(JSXExpressionNode);
       }
@@ -170,17 +202,13 @@ module.exports = {
 
     return {
       JSXExpressionContainer: node => {
-        const parent = node.parent;
-
-        if (shouldCheckForUnnecessaryCurly(parent, userConfig)) {
+        if (shouldCheckForUnnecessaryCurly(node.parent, userConfig)) {
           lintUnnecessaryCurly(node);
         }
       },
 
       Literal: node => {
-        const parentType = node.parent.type;
-
-        if (shouldCheckForMissingCurly(parentType, userConfig)) {
+        if (shouldCheckForMissingCurly(node.parent.type, userConfig)) {
           reportMissingCurly(node);
         }
       }
--- a/lib/rules/jsx-curly-brace-presence.js
+++ b/lib/rules/jsx-curly-brace-presence.js
@@ -69,7 +69,7 @@ module.exports = {
     }
 
     function containsHTMLEntity(rawStringValue) {
-      return /&([A-Za-z\d#]+);/.test(rawStringValue);
+      return /&[A-Za-z\d#]+;/.test(rawStringValue);
     }
 
     function containsDisallowedJSXTextChars(rawStringValue) {
--- a/lib/rules/jsx-curly-brace-presence.js
+++ b/lib/rules/jsx-curly-brace-presence.js
@@ -1,6 +1,7 @@
 /**
  * @fileoverview Enforce curly braces or disallow unnecessary curly brace in JSX
  * @author Jacky Ho
+ * @author Simon Lydell
  */
 'use strict';
 
--- a/lib/rules/jsx-curly-brace-presence.js
+++ b/lib/rules/jsx-curly-brace-presence.js
@@ -60,7 +60,7 @@ module.exports = {
       Object.assign({}, DEFAULT_CONFIG, ruleOptions);
 
     function containsLineTerminators(rawStringValue) {
-      return /[\n\r\u2028\u2029]+/.test(rawStringValue);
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lib/rules/jsx-curly-brace-presence.js [hunk 1]
```diff
@@ -103,34 +127,42 @@ module.exports = {
         node: literalNode,
         message: 'Need to wrap this literal in a JSX expression.',
         fix: function(fixer) {
+          // Leave it to the author to fix as it can be fixed by either a
+          // real character or unicode
+          if (containsHTMLEntity(literalNode.raw)) {
```
jackyho112: We could use `String.fromCharCode` here to fix but based on your [comment](https://github.com/yannickcr/eslint-plugin-react/issues/1449#issuecomment-332750344), it may not be what the author has in mind.

lib/rules/jsx-curly-brace-presence.js
```diff
@@ -59,14 +59,39 @@ module.exports = {
...
+
     function escapeDoubleQuotes(rawStringValue) {
       return rawStringValue.replace(/\\"/g, '"').replace(/"/g, '\\"');
     }
 
+    function escapeBackslashes(rawStringValue) {
+      return rawStringValue.replace(/\\/g, '\\\\');
+    }
+
+    function needToEscapeCharacterForJSX(raw, cooked) {
```
jackyho112: Not sure if it is also necessary to check for patterns like `U+2029` because it doesn't seem like it is recognized in both JS and JSX. Correct me if I am wrong.

tests/lib/rules/jsx-curly-brace-presence.js
```diff
@@ -256,18 +254,6 @@ ruleTester.run('jsx-curly-brace-presence', rule, {
...
-    },
-    {
-      code: '<MyComponent prop={\'bar \"foo\" \'}>foo</MyComponent>',
-      output: '<MyComponent prop=\"bar \\\"foo\\\" \">foo</MyComponent>',
-      options: [{props: 'never'}],
-      errors: [{message: unnecessaryCurlyMessage}]
-    },
-    {
       code: '<MyComponent prop=\'bar\'>foo</MyComponent>',
       output: '<MyComponent prop={\"bar\"}>foo</MyComponent>',
```
lydell: `\"` is a useless escape here. Do you mean `"` or `\\"`? There are several other occurrences of this throughout this file; you might want to go through it from top to bottom correcting all the useless escapes you come across.

(It might even be worth using template literals for all/most of these tests to make them easier to read: No need to escape `'`s.)

tests/lib/rules/jsx-curly-brace-presence.js
```diff
@@ -338,30 +336,6 @@ ruleTester.run('jsx-curly-brace-presence', rule, {
       ]
     },
     {
-      code: '<App>{\'foo "bar"\'}</App>',
-      output: '<App>foo \"bar\"</App>',
```
lydell: This autofix could have been kept. `<App>foo "bar"</App>` is valid.

tests/lib/rules/jsx-curly-brace-presence.js
```diff
@@ -338,30 +336,6 @@ ruleTester.run('jsx-curly-brace-presence', rule, {
...
-      options: [{children: 'never'}],
-      errors: [{message: unnecessaryCurlyMessage}]
-    },
-    {
-      code: '<App>{\"foo \'bar\'\"}</App>',
-      output: '<App>foo \'bar\'</App>',
```
lydell: This autofix could also have been kept.

tests/lib/rules/jsx-curly-brace-presence.js
```diff
@@ -388,6 +362,18 @@ ruleTester.run('jsx-curly-brace-presence', rule, {
       output: '<App prop={\'foo\'} attr={\"bar\"} />',
       errors: [{message: missingCurlyMessage}],
       options: [{props: 'always'}]
+    },
+    {
+      code: '<App prop=\'foo &middot; bar\' />',
+      output: '<App prop=\'foo &middot; bar\' />',
```
lydell: I can't spot the difference between `code` and `output` here.

tests/lib/rules/jsx-curly-brace-presence.js
```diff
@@ -388,6 +362,18 @@ ruleTester.run('jsx-curly-brace-presence', rule, {
...
+      errors: [{message: missingCurlyMessage}],
+      options: [{props: 'always'}]
+    },
+    {
+      code: '<App>foo &middot; bar</App>',
+      output: '<App>foo &middot; bar</App>',
```
lydell: Not here either.

lib/rules/jsx-curly-brace-presence.js
```diff
@@ -103,34 +127,42 @@ module.exports = {
...
         fix: function(fixer) {
+          // Leave it to the author to fix as it can be fixed by either a
+          // real character or unicode
+          if (containsHTMLEntity(literalNode.raw)) {
+            return null;
+          }
+
           const expression = literalNode.parent.type === 'JSXAttribute' ?
-            `{"${escapeDoubleQuotes(
+            `{"${escapeDoubleQuotes(escapeBackslashes(
```
lydell: We also need to escape line terminators here.

lib/rules/jsx-curly-brace-presence.js
```diff
@@ -59,14 +59,39 @@ module.exports = {
...
+        containsBackslash(raw) ||
+        containsHTMLEntity(raw) ||
+        containsDisallowedJSXTextChars(raw) ||
+        containsQuoteCharacters(cooked)
+      );
+    }
+
     /**
      * Report and fix an unnecessary curly brace violation on a node
      * @param {ASTNode} node - The AST node with an unnecessary JSX expression
```
lydell: The `@param` on the next line does not seem to exist.

tests/lib/rules/jsx-curly-brace-presence.js
jackyho112: I remember I did this because of an Eslint rule. But I agree with you. I will try to use template literals on at least the tests in which I have to escape quote characters.

lib/rules/jsx-curly-brace-presence.js
jackyho112: I actually wanted to ask you about them. I know you mentioned them [here](https://github.com/yannickcr/eslint-plugin-react/issues/1449#issuecomment-332750344). 

> similarly for the line terminators: CR, LF, U+2028 and U+2029

I am not quite familiar with those. Do you type exactly those characters of any of the sets into the string to terminate a line? I tried having them in a JSX expression and it just displays those characters. Sorry about my lack of knowledge in the subject matter. 😀

jackyho112:
@lydell 

Can't thank you enough for the feedback! Will update the PR soon based on your suggestions!

tests/lib/rules/jsx-curly-brace-presence.js
lydell: IMO, it's fine to add `/* eslint-disable that-rule */` at the top of this file if needed. You should be allowed to break rules when you can motivate why. But I'm not a maintainer of this repo so I can't say for sure if that's OK. But it would be sad if it's not allowed to make it easier to succeed writing correct tests.

lib/rules/jsx-curly-brace-presence.js
lydell: Press the Enter key to type a LF (`\n`) (or most likely CRLF `\r\n` on Windows).

You can also run one of the following in the Chrome or Firefox dev tools console to copy the character to the clipboard:

- `copy('\n')`
- `copy('\r')`
- `copy('\u2028')`
- `copy('\u2029')`

Then you can simply paste the character.

But since the tests are written as JavaScript strings you should be able to use JavaScript escapes:

```js
input: 'propWithLineTerminators="\n\r\u2028\u2029"',
```

tests/lib/rules/jsx-curly-brace-presence.js
```diff
@@ -228,166 +234,218 @@ ruleTester.run('jsx-curly-brace-presence', rule, {
...
+        '<App prop={"    ',
+        '   a     ',
+        '     b      c',
+        '        d',
+        '"}>{"\\n  a\\n      b     c   \\n         d      \\n"}</App>'
+      ].join('\n'),
+      errors: [
+        {message: missingCurlyMessage}, {message: missingCurlyMessage}
+      ],
+      options: ['always']
```
lydell: Check out the [babel repl](http://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&code_lz=DwQQBADgTg9hC8AiAUGMBDVawCNtgGMt80ATZRAPi03z2yJO3OAHoRrkBuZZUSWAgDeAAxrFc-RkzDkRAX0qjxdKRPxz5bDkA&debug=false&circleciRepo=&evaluate=false&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=6.26.0). It strips some whitespace. We either have to replicate that, or bail out.

jackyho112: Thanks for pointing it out! If we were to bail out, I would check for multiple whitespaces?

lib/rules/jsx-curly-brace-presence.js
jackyho112: For the example you mentioned, the output would be `<div propWithLineTerminators="\\n\\r\\u2028\\u2029"/>`? Because these characters would be displayed as they are in a JSX attribute. 

In that case, wouldn't the parser have already handled that? 
screen shot 2017-10-01 at 12 44 28 pm

tests/lib/rules/jsx-curly-brace-presence.js
lydell: I _think_ we should bail out if there's a line terminator in there. Multiple spaces on a single line should be fine.

lib/rules/jsx-curly-brace-presence.js
lydell: Actually, I think we should bail out if we find any line terminators at all, because of Babel's whitespace stripping thing.

jackyho112:
@lydell 

I think I have covered most of the things you said. Let me know if you have any other improvement in mind. 😀

lib/rules/jsx-curly-brace-presence.js
```diff
@@ -59,6 +59,10 @@ module.exports = {
       {props: ruleOptions, children: ruleOptions} :
       Object.assign({}, DEFAULT_CONFIG, ruleOptions);
 
+    function containsLineTerminators(rawStringValue) {
+      return /[\n\r\u2028\u2029]+/.test(rawStringValue);
```
lydell: There's no need for the `+` here.

tests/lib/rules/jsx-curly-brace-presence.js
lydell: I don't remember – have we addressed or talked about this one?

jackyho112: @lydell 

Yeah, since it is valid, I added back the fix for cases like this and the tests.

lib/rules/jsx-curly-brace-presence.js
```diff
@@ -59,18 +60,45 @@ module.exports = {
...
+    function containsLineTerminators(rawStringValue) {
+      return /[\n\r\u2028\u2029]/.test(rawStringValue);
+    }
+
+    function containsBackslash(rawStringValue) {
       return rawStringValue.includes('\\');
     }
 
+    function containsHTMLEntity(rawStringValue) {
+      return /&([A-Za-z\d#]+);/.test(rawStringValue);
```
lydell: Final little nitpick: The parentheses in the regex are unnecessary. (Sorry, I can't help myself when I see a regex.)

jackyho112: It's all good. I am actually quite grateful for all your regex comments. I learned a lot from them! 😄

lydell:
Reviewers: Don't feel intimidated by the walls of comments me and @jackyho112 have exchanged! If you look a the implementation (one file), you'll see that it's pretty clear and concise. The rest of the changes are just docs and tests updates.

**TL;DR** on the whole issue: Some characters make it unclear\* if braces are unnecessary or not. If we find such characters, bail out.

<small>\* Because readability can suffer badly.</small>

PR #1459, opened 2017-10-02
[jsx-no-bind] Only warn for props and account for variable declaration

### LINKED ISSUE #1395: Accidental bypass for react/jsx-no-bind allowArrowFunctions
We've recently started to get rid of our usage of arrow functions in component props for similar reasons to that mentioned [here](https://medium.freecodecamp.org/why-arrow-functions-and-bind-in-reacts-render-are-problematic-f1c08b060e36). We went ahead and set: 
```JS
'react/jsx-no-bind': ['error', {
      allowArrowFunctions: false,
      allowBind: false,
}]
```
but realized we were still able to get away with using something like:
```JS
const someFunc = () => onClickHandler(userId)
<SomeComponent onClick={someFunc} />
```
So we didn't fix the problem, we just accidentally tricked it. Is this expected behavior or is this a bug?

### LINKED ISSUE #1417: react/jsx-no-bind false positive?
Example snippet:

```jsx
import React from 'react';

class FalsePositive extends React.PureComponent {
  static propTypes = {
    list: PropTypes.arrayOf(PropTypes.node).isRequired,
  };

  render() {
    return (
      <div>
        {this.props.list.map(this.wrap.bind(this, 'span'))}
      </div>
    );
  }

  wrap(Tag, item, i) {
    return <Tag key={i}>{item}</Tag>;
  }
}

export default FalsePositive;
```

`this.wrap.bind` triggers the rule, even though the bound function isn't the actual prop being passed, but is being used during the render to build the children, so it shouldn't break purity.

### LINKED ISSUE #1444: Jsx no bind rule
The [no bind rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md) flags of the following jsx code.

```javascript
renderButton(someParams){
  return(
    <ButtonComponent
      onClick={this.handleClick.bind(this, someParams)} // <- Error here
    >
      Submit
    </ButtonComponent>  
  )
}
```

How ever if i modify the same function

```javascript
renderButton(someParams){
  const clickHandler = this.handleClick.bind(this, someParams);
  return(
    <ButtonComponent
      onClick={clickHandler}
    >
      Submit
    </ButtonComponent>  
  )
}
```

The error is not thrown, how ever effectively the same thing of creating a new function when the ```renderButton``` method is called takes place, only in the second case a separate variable has been assigned. But in terms of memory allocation a new function is being created in both cases and though the rule will give a sign off it'll still not impact performance.

**Proposal**
Modify the scope of the rule to see if bind is called within that function scope..

I may be completely wrong in some assumptions but would like a discussion here , if possible.

jackyho112:
This will fix #1444, #1395 and #1417, so that the rule will only warn if there are violations in JSX attributes and will keep track of `const` variable declarations of violations. 

For examples,

```jsx
renderButton(someParams){
  const clickHandler = this.handleClick.bind(this, someParams);
  return(
    <ButtonComponent
      onClick={clickHandler}
    >
      Submit
    </ButtonComponent>  
  )
} 
```

will warn.

But

```jsx
 render() {
    return (
      <div>
        {this.props.list.map(this.wrap.bind(this, 'span'))}
      </div>
    );
  }
```

will not.

The new implementation is simply keeping track of each `const` variable declaration that references a violation in each block statement and reports them if they got passed into JSX props.

With the update, this rule does not account for any assignment expression and variable declarations that are not `const`. I feel that it is acceptable to leave them for future updates.

Please let me know if there is any idea for improvements. 😄 

Fixes #1444. Fixes #1395. Fixes #1417.

### DIFF
--- a/lib/rules/jsx-no-bind.js
+++ b/lib/rules/jsx-no-bind.js
@@ -1,7 +1,7 @@
 /**
  * @fileoverview Prevents usage of Function.prototype.bind and arrow functions
- *               in React component definition.
- * @author Daniel Lo Nigro <dan.cx>
+ *               in React component props.
+ * @author Daniel Lo Nigro <dan.cx>, Jacky Ho
  */
 'use strict';
 
@@ -12,10 +12,14 @@ const propName = require('jsx-ast-utils/propName');
 // Rule Definition
 // -----------------------------------------------------------------------------
 
+const bindViolationMessage = 'JSX props should not use .bind()';
+const arrowViolationMessage = 'JSX props should not use arrow functions';
+const BindExpressionViolationMessage = 'JSX props should not use ::';
+
 module.exports = {
   meta: {
     docs: {
-      description: 'Prevents usage of Function.prototype.bind and arrow functions in React component definition',
+      description: 'Prevents usage of Function.prototype.bind and arrow functions in React component props',
       category: 'Best Practices',
       recommended: false
     },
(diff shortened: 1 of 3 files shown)


lib/rules/jsx-no-bind.js
```diff
@@ -40,33 +45,97 @@ module.exports = {
...
         if (
-          !configuration.allowBind &&
-          (callee.type !== 'MemberExpression' || callee.property.name !== 'bind')
+          blockAncestors.length > 0 &&
+          variableViolationType &&
+          node.parent.kind === 'const' // only support const right now
         ) {
-          return;
-        }
-        const ancestors = context.getAncestors(callee).reverse();
```
jackyho112: This block of code reports any violation in a component's `render` method. This rule is supposed to only report violations in JSX props, so this is not needed.

tests/lib/rules/jsx-no-bind.js
```diff
@@ -115,6 +116,121 @@ ruleTester.run('jsx-no-bind', rule, {
...
+      code: [
+        'class Hello extends Component {',
+        '  render() {',
+        '    return (<div>{',
+        '      this.props.list.map(this.wrap.bind(this, "span"))',
+        '    }</div>);',
+        '  }',
+        '};'
+      ].join('\n'),
+      parser: 'babel-eslint'
```
ljharb: most of these test cases don't need `babel-eslint`; can we use the default parser wherever possible?

jackyho112:
@ljharb Forgot about them. Will fix. 😅

jackyho112:
@ljharb Thanks for the feedback! Now I will always think of async arrow functions in future contributions. Let me know if there is anything else I can improve on this PR. 😄

ljharb:
@jackyho112 I was mainly asking about ensuring that `<Foo prop={async () => {}} />` warns too :-)

jackyho112:
@ljharb 😄 , didn't know. I'm on it!

ljharb:
Let's also ensure that `function () {}`, `function * () {}`, and `async function () {}` all warn as well :-)

jackyho112:
@ljharb Good call. Gonna have to do it tomorrow though.

jackyho112:
Thanks for the approval!

windhost:
537b6fdc-f92b-4cdc-b55e-09271415c8f1
However, with react-navigation options, it can not pass the rule

ljharb:
@windhost can you file a separate issue for that? since `NavigationOptions` returns an object, it shouldn't be treated like a component.

windhost:
@ljharb 👌

windhost:
#1736

PR #1462, opened 2017-10-03
[destructuring-assignment] : new rule

### LINKED ISSUE #216: Rule proposal (for 0.14): require stateless function components use destructuring on the props parameter
It would warn on this:

``` jsx
    var SomeThing = props => (
        <Foo bar={props.fizz} baz={props.buzz} />
    );
```

Suggesting that this is a more expressive syntax:

``` jsx
    var SomeThing = ({fizz, buzz}) => (
        <Foo bar={fizz} baz={buzz} />
    );
```

An exception where props would be ok is if it's just being passed wholesale to an inner component via spread operator. This would be consider valid:

``` jsx
    var SomeThing = props => (
        <Wrapper>
            <Foo someDefault="blah" {...props} />
        </Wrapper>

    )
```

### LINKED ISSUE #278: Rule idea: Enforce using property destructructuring when properties are referred to N > 1 times
A common pattern that I often see (and like) is to use property destructors in render methods.

React Native:

``` jsx
render() {
  const {
    firstName,
    lastName,
    photo,
  } = this.props;

  return <View>
    <Text>{firstName}</Text>
    <Text>{lastName}</Text>
    <Image source={photo} />
  </View>;
}
```

### LINKED ISSUE #556: Suggestion: require or forbid prop object destructuring syntax for SFCs
Some people may prefer this syntax:

``` js
const MyComponent = ({ id, className }) => (
  <div id={id} className={className} />
);
```

while others may prefer this:

``` js
const MyComponent = props => (
  <div id={props.id} className={props.className} />
);
```

It would be nice to be able to enforce one or the other.

DianaSuvorova:
to enforce consistent usage of destructuring assignment closes #216, closes #556, closes #278

### DIFF
--- a/docs/rules/destructuring-assignment.md
+++ b/docs/rules/destructuring-assignment.md
@@ -1,7 +1,6 @@
 # Enforce consostent usage of destructuring assignment of props, state and context (react/destructuring-assignment)
 
-Rule can be set to either of `always`, `never`, `ignore` for SFC and to `always` or `ignore` for class components.
-
+Rule can be set to either of `always`, `never`, `ignore` for `class` and `SFC` components.
 ```js
 "react/destructuring-assignment": [<enabled>, { "SFC": "always", "class": "always"}]
 ```
@@ -33,19 +32,42 @@ const MyComponent = ({id}) => {
 ```
 
 ```js
+const MyComponent = (props, context) => {
+  const { id } = props;
+  return (<div id={id} />)
+};
+```
+
+```js
 const Foo = class extends React.PureComponent {
   render() {
-    const { foo } = this.props;
-    return <div>{foo}</div>;
+    const { title } = this.context;
+    return <div>{title}</div>;
   }
 };
 ```
 
-If rule option is set to `never` for SFC, the following pattern is considered warning:
+If rule option is set to `never`, the following patterns are considered warning:
 
 ```js
 const MyComponent = ({id}) => {
   return (<div id={id} />)
+};
+```
+
+```js
+const MyComponent = (props) => {
+  const { id } = props;
+  return (<div id={id} />)
+};
+```
+
+```js
+const Foo = class extends React.PureComponent {
+  render() {
+    const { title } = this.state;
+    return <div>{title}</div>;
+  }
 };
 ```
 
@@ -56,3 +78,10 @@ const MyComponent = (props) => {
   return (<div id={props.id} />)
 };
 ```
+
+```js
+const Foo = class extends React.PureComponent {
+  render() {
+    return <div>{this.state.title}</div>;
+  }
+};
--- a/docs/rules/destructuring-assignment.md
+++ b/docs/rules/destructuring-assignment.md
@@ -1,0 +1,58 @@
+# Enforce consostent usage of destructuring assignment of props, state and context (react/destructuring-assignment)
+
+Rule can be set to either of `always`, `never`, `ignore` for SFC and to `always` or `ignore` for class components.
+
+```js
+"react/destructuring-assignment": [<enabled>, { "SFC": "always", "class": "always"}]
+```
+
+## Rule Details
+
+By default rule is set to `always` enforce destructuring assignment. The following patterns are considered warnings:
+
+```js
+const MyComponent = (props) => {
+  return (<div id={props.id} />)
+};
+```
+
+```js
+const Foo = class extends React.PureComponent {
+  render() {
+    return <div>{this.context.foo}</div>;
+  }
+};
+```
+
+Below pattern is correct:
+
+```js
+const MyComponent = ({id}) => {
+  return (<div id={id} />)
+};
+```
+
+```js
+const Foo = class extends React.PureComponent {
+  render() {
+    const { foo } = this.props;
+    return <div>{foo}</div>;
+  }
+};
+```
+
+If rule option is set to `never` for SFC, the following pattern is considered warning:
+
+```js
+const MyComponent = ({id}) => {
+  return (<div id={id} />)
+};
+```
+
+and below pattern is correct:
+
+```js
+const MyComponent = (props) => {
+  return (<div id={props.id} />)
+};
+```
--- a/docs/rules/destructuring-assignment.md
+++ b/docs/rules/destructuring-assignment.md
@@ -1,4 +1,4 @@
-# Enforce consostent usage of destructuring assignment of props, state and context (react/destructuring-assignment)
+# Enforce consistent usage of destructuring assignment of props, state, and context (react/destructuring-assignment)
 
 Rule can be set to either of `always` or `never`;
 ```js
--- a/docs/rules/destructuring-assignment.md
+++ b/docs/rules/destructuring-assignment.md
@@ -1,8 +1,8 @@
 # Enforce consostent usage of destructuring assignment of props, state and context (react/destructuring-assignment)
 
-Rule can be set to either of `always`, `never`, `ignore` for `class` and `SFC` components.
+Rule can be set to either of `always` or `never`;
 ```js
-"react/destructuring-assignment": [<enabled>, { "SFC": "always", "class": "always"}]
+"react/destructuring-assignment": [<enabled>, 'always']
 ```
 
 ## Rule Details
@@ -47,7 +47,7 @@ const Foo = class extends React.PureComponent {
 };
 ```
 
-If rule option is set to `never`, the following patterns are considered warning:
+If rule is set to `never`, the following patterns are considered warning:
 
 ```js
 const MyComponent = ({id}) => {
--- a/lib/rules/destructuring-assignment.js
+++ b/lib/rules/destructuring-assignment.js
@@ -5,10 +5,7 @@
 
 const Components = require('../util/Components');
 
-const DEFAULT_OPTIONS = {
-  SFC: 'always',
-  class: 'always'
-};
+const DEFAULT_OPTION = 'always';
 
 module.exports = {
   meta: {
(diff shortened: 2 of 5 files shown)


docs/rules/destructuring-props-argument.md
```diff
@@ -0,0 +1,45 @@
...
+Rule can be set to either of `always`, `never`, `ignore` for each type of the component.
+Currently only SFC is implemented.
+
+```js
+"react/destructuring-props-argument": [<enabled>, { "SFC": "always"}]
+```
+
+## Rule Details
+
+By default rule is set to `always` enforce destructuring assignment. The following pattern is considered warning:
```
ljharb: presumably we'd want the default to be "always" on all component types; releasing the rule with only SFC support means we couldn't default class and createClass components to "always" without a major release

lib/rules/destructuring-props-argument.js
```diff
@@ -0,0 +1,89 @@
...
+
+      FunctionExpression: handleStatelessComponent,
+
+      MemberExpression: function(node) {
+        const relatedComponent = components.get(context.getScope(node).block);
+        const isStatelessFunctionUsage = node.object.name === 'props' && !isAssignmentToProp(node);
+        if (relatedComponent && isStatelessFunctionUsage && options.SFC === 'always') {
+          context.report({
+            node: node,
+            message: 'Should use destructuring props assignment in SFC argument'
```
ljharb: `s/Should/Must/g`?

docs/rules/destructuring-props-argument.md
DianaSuvorova: I am not sure about your major release concern. Would releasing only SFC support require a major release? What's the problem eith having to do a major release?

Irrespective of that I have only one use case for "other" component types as per issue #278 and I am not particularly keen about it.  Are you? Maybe it would make sense to have this rule for SFC only?

lib/rules/destructuring-props-argument.js
DianaSuvorova: hm, what do you mean?

ljharb: it's not that you *should* use it, it's that you *must* because of the rule

docs/rules/destructuring-props-argument.md
ljharb: releasing only SFC support would be a minor.

Later releasing class-based component supported, defaulted to "ignore", would also be a minor.

However, later releasing class-based component supported, defaulted to "always", would be a major.

There's not a *problem* per se about releasing a major release, but it's always best to minimize breaking changes.

My use case for non-SFC components is that I want *all* `this.props`, `this.state`, and `this.context` properties destructured at the top of every function that uses them (that also includes `context` in SFCs).

It would be great if one rule could handle all of these cases - if not, I'd expect this rule to only *ever* handle SFCs (props and context), and a new rule to be added for "inside non-SFC functions".

DianaSuvorova: Ok, well in this case I can  try to add the rest of the logic to this pr.
So both of the component types will enforce `context` and `props`. Class component will in addition enforce `state`. 
Config structure would be: `{ SFC: always/never/ignore,  class: always/never/ignore}` (I don't like name `class` is there a better term?)
default: `{ SFC: always, class: always }`

there is also a nextProps, prevProps which I would ignore for now.
Does this make sense?

ljharb: i think `class` is probably the best name, unfortunately.

I agree ignoring next/prev props is fine for now, in SFCs, it'd be `props` and `context`, and in class-based and createClass components, it'd be `this.props` and `this.state` and `this.context`.

DianaSuvorova:
@ljharb , FYI, this is ready for review.

lib/rules/destructuring-assignment.js
```diff
@@ -0,0 +1,133 @@
...
+      }
+    }
+
+    function handleSFCUsage(node) {
+      // props.aProp || context.aProp
+      const isPropUsed = (node.object.name === 'props' || node.object.name === 'context') && !isAssignmentToProp(node);
+      if (isPropUsed && options.SFC === 'always') {
+        context.report({
+          node: node,
+          message: `Should use destructuring ${node.object.name} assignment`
```
ljharb: i think the word "must" here is better than "should"

lib/rules/destructuring-assignment.js
```diff
@@ -0,0 +1,133 @@
...
+      MemberExpression: function(node) {
+        const SFCComponent = components.get(context.getScope(node).block);
+        const classComponent = utils.getParentComponent(node);
+        if (SFCComponent) {
+          handleSFCUsage(node);
+        }
+        if (classComponent) {
+          handleClassUsage(node, classComponent);
+        }
+      }
```
ljharb: can we also handle createClassComponents here?

DianaSuvorova:
@ljharb Ready for re-review

```class Foo extends React.Component {
  render() { return this.foo(); }
  foo() {
    return this.props.children; // should be warned when "always"
  }
}``` 

```var Hello = React.createClass({
      render: function() {
        return <Text>{this.props.foo}</Text>;
      }
    });```

Two above examples are added to test cases. 

Changed `should` to `must`. I am not a big fan of `must` but I trust your judgement.

**Comment by ljharb** [comment]
In specifications, "should" means the implementation can choose not to comply; "must" means they are forced to comply. If the rule is enabled, they're being forced to comply :-)

**Comment by DianaSuvorova** [comment]
I think that `never` setting doesn't really makes sense for class component. As it would flag any
`const { foo } = this.props`.
In SFC `never` is only enforced within the argument.

For example:
```
{
    code: `const MyComponent = ({ id, className }) => (
      <div id={id} className={className} />
    );`,
    options: [{SFC: 'never'}],
    errors: [
      {message: 'Must never use destructuring props assignment in SFC argument'}
    ]
  }
```
 
```
,{
    code: `const MyComponent = (props) => {
      const { id, className } = props;
      return <div id={id} className={className} />
    };`,
    options; [SFC: 'never']
  }
```

> I'm also wondering if class-based components and createClass components should have separate configurability?
I would do it separately if it is a popular request.

I added explicit always test cases (which match default ones). And few more tests for class components. Hope it covers everything.

**Comment by DianaSuvorova** [comment]
@ljharb , btw I am squashing any new commits I add. It may look the last commit was 8days ago but I just pushed the change.

**Comment by ljharb** [comment]
> I think that never setting doesn't really makes sense for class component. As it would flag any
const { foo } = this.props.

Yes, that's exactly what I'm expecting - if the user chooses "never", then they want `this.props.foo` to be used every time.

**Comment by DianaSuvorova** [comment]
@ljharb, this is ready for review

File: docs/rules/destructuring-assignment.md [hunk 5]
```diff
@@ -0,0 +1,87 @@
+# Enforce consostent usage of destructuring assignment of props, state and context (react/destructuring-assignment)
+
+Rule can be set to either of `always`, `never`, `ignore` for `class` and `SFC` components.
```
ljharb: this should probably mention `createClass` components as well

docs/rules/destructuring-assignment.md
```diff
@@ -0,0 +1,87 @@
...
+```js
+"react/destructuring-assignment": [<enabled>, { "SFC": "always", "class": "always"}]
```
ljharb: also here

lib/rules/destructuring-assignment.js
```diff
@@ -0,0 +1,154 @@
...
+
+const DEFAULT_OPTIONS = {
+  SFC: 'always',
+  class: 'always'
+};
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Enforce consostent usage of destructuring assignment of props, state and context',
```
ljharb: s/consostent/consistent

also, please use an Oxford comma after `state`

hendrikswan:
@DianaSuvorova, thanks for creating this rule - my team is looking forward to when this gets merged in :)

docs/rules/destructuring-assignment.md
DianaSuvorova: @ljharb, do you suggest `createClass` as additional option? So it would be [`SFC`, `createClass`, `class`]?  If so can it be done as a separate PR? Right now `class` option includes `createClass`.

ljharb: Yes, I think they should be separate options - i think it'd be fine to add it in a separate PR, but I don't think this should be merged with `class` handling both :-/

DianaSuvorova: hmmm, could you explain rationale behind that? To me this rule almost can be monolithic (cover all possible react components with a single setting). I can see the need for SFC vs Class components distinction, but not class vs createClass. What is the use case?

ljharb: For one, createClass components are often legacy, and legacy things need different rules. Mainly though, all three are actually distinct constructs - createClass components have different behavior than class-based components, in particular around auto-binding all the "instance" methods;

I think that it'd be fine to allow a string *or* an object option, where the string option is a single setting for all components - but the object option is intended to allow fine-grained control, so it's supposed to be *not* monolithic by design.

DianaSuvorova: I agree yeah, technically if we are going to list component types, we should list them all. But even though it is very little code to be added to support the 3rd `createClass` option, I am not very comfortable adding a special config option for a deprecated functionality. Additionally the  difference in behaviour between `class` and `createClass` component doesn't seem to be related to what this rule is checking (in both props are referenced with `this.props` right, or am I wrong?).  

So seems like the single string option controlling all components is an acceptable solution for both of us and is common pattern across the library. I will update my pr shortly.

DianaSuvorova: @ljharb I added an incremental change to switch to a single string option so it might be easier to review.

DianaSuvorova:
@ljharb anything else is missing for this PR?

hendrikswan:
@DianaSuvorova it seems like the changes were approved?

DianaSuvorova:
@hendrikswan, yeah, but I don't have permissions to merge to master.

hendrikswan:
On realized that after my comment.. I admire your perseverance!

lib/rules/destructuring-assignment.js
```diff
@@ -0,0 +1,137 @@
...
+          message: `Must use destructuring ${node.object.name} assignment`
+        });
+      }
+    }
+
+    function handleClassUsage(node) {
+      // this.props.Aprop || this.context.aProp || this.state.aState
+      const isPropUsed = (node.object.type === 'MemberExpression' && node.object.object.type === 'ThisExpression' &&
+        (node.object.property.name === 'props' || node.object.property.name === 'context' || node.object.property.name === 'state') &&
+        node.object.type === 'MemberExpression'
```
SimenB: `node.object.type === 'MemberExpression'` is asserted twice

ljharb: whoops, thanks - want to make a PR to clean it up? :-D

SimenB: Sure: #1502

PR #1467, opened 2017-10-07
[WIP] Update styling of files documenting rules

felicio:
Increase legibility of negative sentences in counter examples.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,14 @@
+# Contributing to ESLint-plugin-React
+
+## ESLint Rules
+
+### Markup
+
+Files documenting individual rules live in docs/rules folder. They're annotated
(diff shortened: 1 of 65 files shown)


felicio:
I'd definitely agree. At first it might seem as a tooth-picking request, but when I was creating this PR I thought about that the docs are not always read without stress and little things like _that_ could save someone from extra frustration.

The pattern here seems to be in 50+ markdown files where negative verbs (e.g. is/are not) within a paragraph perceived by and followed by a code block should be __bold__.

If no one jumps in with some clever `sed` or `awk` implementations, I could have a look at the rest myself.

felicio:
I've just `squash`ed and updated the branch with force. Shouldn't it trigger another review request? Nevertheless, I labeled this one as WIP, since it's missing the CONTRIBUTING.md.

PR #1470, opened 2017-10-10
Add instance-methods and instance-variables to sort-comp.

### LINKED ISSUE #599: sort-comp: Consider separating static fields & instance fields into separate groups

RDGthree:
Allow sorting of instance methods and instance variables optionally with the rules `instance-methods` and `instance-variables` respectively. Rules are optional, and are applied immediately before `everything-else` to avoid overriding other rules.

Adjusting `static-methods` and adding `static-variables` would have been breaking so I skipped that for now.

Fixes #599.


danieljvdm:
Hey guys, changes LGTM, let's get the ball rolling on merging this.

GerritSe:
Any updates on the merge?

michaelhogg:
FYI: discussion of this PR is continuing in #1774.

Issue #1469, opened 2017-10-10
'react/jsx-wrap-multilines' behaves strange


SidKwok:
The [rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-wrap-multilines.md) tells me that this is good:
```
var hello = (
    <div>
      <p>Hello</p>
    </div>
)
```

but this is also accepted:
```
var hello = (<div>
  <p>Hello</p>
</div>)
```

```
'react/jsx-wrap-multilines': 'error'
```

Did I miss something?

ljharb:
Nope. Both of your examples are allowed, but the docs only mention the first one, since it's generally preferred (and more readable) than the second.

SidKwok:
How can I enforce the first format?

ljharb:
It sounds like we'd need another possible value for the three syntaxes: instead of `true`/`false`, we'd want an enum of `ignore`/`parens`/`parens-separate-lines` (or some better name for the last option).

SidKwok:
That will be a good feature for this rule.

lukemwillis:
@ljharb It looks like these options are supported in master now, but in v7.4.0, I get errors. This is my rule:

    "react/jsx-wrap-multilines": ["error", { 
        "declaration": "parens-new-line", 
        "assignment": "parens-new-line", 
        "return": "parens-new-line", 
        "arrow": "parens-new-line", 
        "condition": "parens-new-line", 
        "logical": "parens-new-line", 
        "prop": "parens-new-line"
    }]

Which gives me the error:

> Error: .eslintrc:
	Configuration for rule "react/jsx-wrap-multilines" is invalid:
	Value "[object Object]" should NOT have additional properties.

If I remove the rules for `condition`, `logical`, and `prop`, then I get the error:

> Error: .eslintrc:
	Configuration for rule "react/jsx-wrap-multilines" is invalid:
	Value "parens-new-line" should be boolean.

Are the changes to this rule in v7.4.0? Or are they slated for a future release?

ljharb:
Slated for a future release.

dfsq:
Are there plans to make a release in near future? This rule is important.

vzaidman:
Also, I'd like to specify default behavior:
```
"react/jsx-wrap-multilines": ["error", "parens-new-line", {
    "arrow": "parens", 
}]
```

ljharb:
I believe this has been released.

mqklin:
[...]

PR #1475, opened 2017-10-14
Check parentheses with new lines for multiline JSX

jomasti:
This changes the options from a boolean to an enum with three possible
values: `ignore`, `parens`, and `parens-new-line`. Using `ignore` is like
the previous `false` while `parens` is like the previous `true`. The new
option (`parens-new-line`) will check that the parentheses around the
multiline JSX are on separate lines for better readability.

This PR attempts to address #1469.

### DIFF
--- a/tests/lib/rules/jsx-wrap-multilines.js
+++ b/tests/lib/rules/jsx-wrap-multilines.js
@@ -21,8 +21,11 @@ const parserOptions = {
 };
 
 // ------------------------------------------------------------------------------
-// Code Snippets
+// Constants/Code Snippets
 // ------------------------------------------------------------------------------
+
+const MISSING_PARENS = 'Missing parentheses around multilines JSX';
+const PARENS_NEW_LINES = 'Parentheses around JSX should be on separate lines';
 
 const RETURN_SINGLE_LINE = `
   var Hello = createReactClass({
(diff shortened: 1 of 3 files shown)


tests/lib/rules/jsx-wrap-multilines.js
```diff
@@ -147,96 +216,258 @@ ruleTester.run('jsx-wrap-multilines', rule, {
...
+      code: ARROW_PAREN
+    }, {
+      code: ARROW_SINGLE_LINE
+    }, {
       code: ARROW_PAREN,
-      options: []
+      options: [{arrow: true}]
     }, {
       code: ARROW_SINGLE_LINE,
-      options: []
```
ljharb: should we not keep this test against "the default"?

tests/lib/rules/jsx-wrap-multilines.js
```diff
@@ -147,96 +216,258 @@ ruleTester.run('jsx-wrap-multilines', rule, {
...
+    }, {
       code: ASSIGNMENT_NO_PAREN,
       options: [{assignment: false}]
     }, {
+      code: ARROW_PAREN
+    }, {
+      code: ARROW_SINGLE_LINE
+    }, {
       code: ARROW_PAREN,
-      options: []
```
ljharb: should we not keep this test against "the default"?

jomasti: The two tests above these are using the defaults.

lencioni:
Looks like this needs a rebase

PR #1478, opened 2017-10-14
Support propWrapperFunctions for boolean-prop-naming

### LINKED ISSUE #1452: `boolean-prop-naming` triggers "Cannot read property 'properties' of undefined"
Hi,

I think this is similar to: https://github.com/yannickcr/eslint-plugin-react/issues/1369

I’m running 7.4.0 and still getting this error:

**package.json:**
`"eslint-plugin-react": "^7.4.0"`

`eslint` is 4.7.2

**.eslintrc.js:**
```
"react/boolean-prop-naming": [
    "error",
    { "rule": "^(is|has)[A-Z]([A-Za-z0-9]?)+" }
],
```

**Error:**
```
Cannot read property 'properties' of undefined
TypeError: Cannot read property 'properties' of undefined
    at EventEmitter.MemberExpression (/Users/user/repos/myrepo/node_modules/eslint-plugin-react/lib/rules/boolean-prop-naming.js:183:45)
    at emitOne (events.js:120:20)
    at EventEmitter.emit (events.js:210:7)
    at NodeEventGenerator.applySelector (/Users/user/repos/myrepo/node_modules/eslint/lib/util/node-event-generator.js:265:26)
    at NodeEventGenerator.applySelectors (/Users/user/repos/myrepo/node_modules/eslint/lib/util/node-event-generator.js:294:22)
    at NodeEventGenerator.enterNode (/Users/user/repos/myrepo/node_modules/eslint/lib/util/node-event-generator.js:308:14)
    at CodePathAnalyzer.enterNode (/Users/user/repos/myrepo/node_modules/eslint/lib/code-path-analysis/code-path-analyzer.js:606:23)
    at Traverser.enter (/Users/user/repos/myrepo/node_modules/eslint/lib/linter.js:962:32)
    at Traverser.__execute (/Users/user/repos/myrepo/node_modules/estraverse/estraverse.js:397:31)
    at Traverser.traverse (/Users/user/repos/myrepo/node_modules/estraverse/estraverse.js:501:28)
```

**The issue is stemming from this code:**
```
Card.propTypes = merge({}, Card.propTypes, {
    showScore: propTypes.bool
});
```

jomasti:
This adds support for checking prop types  in `boolean-prop-naming` when using the `propWrapperFunctions` setting.

Closes #1452.

### DIFF
--- a/lib/rules/boolean-prop-naming.js
+++ b/lib/rules/boolean-prop-naming.js
@@ -139,6 +139,13 @@ module.exports = {
       });
     }
 
+    function checkPropWrapperArguments(node, args) {
+      if (!node || !Array.isArray(args)) {
+        return;
+      }
+      args.filter(arg => arg.type === 'ObjectExpression').forEach(object => validatePropNaming(node, object.properties));
+    }
+
     // --------------------------------------------------------------------------
     // Public
     // --------------------------------------------------------------------------
@@ -147,6 +154,9 @@ module.exports = {
       ClassProperty: function(node) {
         if (!rule || !propsUtil.isPropTypesDeclaration(node)) {
           return;
+        }
+        if (node.value && node.value.type === 'CallExpression' && propWrapperFunctions.has(sourceCode.getText(node.value.callee))) {
+          checkPropWrapperArguments(node, node.value.arguments);
         }
         if (node.value && node.value.properties) {
           validatePropNaming(node, node.value.properties);
@@ -166,7 +176,7 @@ module.exports = {
         }
         const right = node.parent.right;
         if (right.type === 'CallExpression' && propWrapperFunctions.has(sourceCode.getText(right.callee))) {
-          right.arguments.filter(arg => arg.type === 'ObjectExpression').forEach(object => validatePropNaming(component.node, object.properties));
+          checkPropWrapperArguments(component.node, right.arguments);
           return;
         }
         validatePropNaming(component.node, node.parent.right.properties);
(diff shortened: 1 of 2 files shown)


louisscruz:
Any progress on this?

jomasti:
I think this is good based on the previous feedback, but @ljharb will have to review again.

jseminck:
@ljharb last I remembered ESLint supports node 4 which does not support object spread. I've had the same problems in the past. Has it changed?

ljharb:
@jseminck I’m not asking about the rule implementation, I’m asking about users in their own codebase - also, object spread is trivially transpileable with Babel to Object.assign, or to any other package name via transform-replace-object-assign.

jomasti:
@ljharb Now that I think about it, this change should just use the current `propWrapperFunctions` setting instead of a new one since we would want to support the other functions that could be used like `forbidExtraProps` and whatnot.

ljharb:
@jomasti it could; but `propWrapperFunctions` is meant to convey "this function returns an object", which "exact" prop functions also do.

However, this rule deals only with "exact prop wrapper functions" - `Object.freeze`, for example, is a valid propWrapperFunction but not a valid exact prop wrapper function.

jomasti:
@ljharb You may be thinking about my other PR for `prefer-exact-props`, but for this one (`boolean-prop-naming`), any function that returns an object should be supported, right? That would be beyond the current implementation of adding a new property (and probably unnecessary currently) (`propObjectAssignFunctions`).

Also, avoiding adding this should make it easy to change the schema of `propWrapperFunctions` to support your suggested schema in #1547 without having to remove this new setting.

ljharb:
@jomasti oh lol i got confused, yes, i was thinking about the other one.

Yes, this rule should just use propWrapperFunctions, good call :-D

Issue #1476, opened 2017-10-14
react/sort-prop-types and sort order of shape values


jackwilsdon:
It would be nice to be able to require an alphabetical sort order of `PropTypes.shape` values, maybe as an option for `react/sort-prop-types`?

shepmaster:
Appears to be closed by  #1481.

PR #1481, opened 2017-10-16
Add `sortShapeProp` option to sort-prop-types rule

jomasti:
This allows the rule to enforce the sorting rules in a shape prop object.

I think it makes more sense to happen by default, but I expect it would be better to wait until the next major version for that. This PR should allow the rule to satisfy the request in #1476.

### DIFF
--- a/tests/lib/rules/sort-prop-types.js
+++ b/tests/lib/rules/sort-prop-types.js
@@ -344,6 +344,27 @@ ruleTester.run('sort-prop-types', rule, {
         }),
       };
     `
+  }, {
+    code: `
+      class Component extends React.Component {
+        render() {
+          return <div />;
+        }
+      }
+      Component.propTypes = {
+        a: PropTypes.any,
+        b: PropTypes.any,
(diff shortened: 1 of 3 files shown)


tests/lib/rules/sort-prop-types.js
```diff
@@ -325,8 +325,47 @@ ruleTester.run('sort-prop-types', rule, {
...
+          b: PropTypes.bool,
+        }),
+      };
+    `,
+    options: [{
+      sortShapeProp: true
+    }]
   }],
-
+  // Invalid code, should not be validated
```
jseminck: What's the reason for this comment?

jomasti: I think I saw the comment as an intended marking of where the invalid test cases began and that it was in the wrong place. Looking back over it, the comment was actually in the right spot. I'll rebase to put this back.

ljharb:
Please rebase this one more time; tests should pass now :-)

PR #1482, opened 2017-10-16
Fix `boolean-prop-naming` with Object.spread syntax

track0x1:
There is an issue where using Object.spread syntax with this rule causes Eslint to break as it attempts to look for `node.value.property` in a spread object. We avoid validating spread nodes by checking for the `ExperimentalSpreadProperty` type.


track0x1:
I pushed changes to address your comment, but I'm not seeing the new commit show up here. 🤔

track0x1:
I'm going to close this PR and see if the new changes will show.

ljharb:
Please don't open new PRs; there's always a way to fix the first one if you don't close it.

PR #1483, opened 2017-10-16
Add jsx-sort-default-props rule

b0gok:
Related to https://github.com/yannickcr/eslint-plugin-react/issues/281

- [x] Add support for
```js
var First = createReactClass({
  getDefaultProps: function() {
    return {
      barRequired: "barRequired",
      onBar: "onBar"
    };
  }
});
```
- [x] Add more tests
- [x] Remove useless code

### DIFF
--- a/docs/rules/jsx-sort-default-props.md
+++ b/docs/rules/jsx-sort-default-props.md
@@ -49,7 +49,7 @@ Component.defaultProps = {
 };
 ```
 
-The following patterns are considered okay and do not cause warnings:
+The following patterns are considered okay and do **not** cause warnings:
 
 ```jsx
 var Component = createReactClass({
--- a/docs/rules/jsx-sort-default-props.md
+++ b/docs/rules/jsx-sort-default-props.md
@@ -65,6 +65,25 @@ StatelessComponentWithSpreadInPropTypes.defaultProps = {
   a: "a",
   ...defaults,
 };
+
+export default class ClassWithSpreadInPropTypes extends BaseClass {
+  static propTypes = {
+    a: PropTypes.string,
+    b: PropTypes.string,
+    c: PropTypes.string,
+    d: PropTypes.string,
+    e: PropTypes.string,
+    f: PropTypes.string
+  }
+  static defaultProps = {
+    b: "b",
+    a: "a",
+    ...c.defaultProps,
+    f: "f",
+    e: "e",
+    ...d.defaultProps
+  }
+}
 ```
 
 The following patterns are considered okay and do **not** cause warnings:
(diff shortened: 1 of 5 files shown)


docs/rules/jsx-sort-default-props.md
```diff
@@ -0,0 +1,97 @@
+# Enforce propTypes declarations alphabetical sorting (react/jsx-sort-default-props)
+
+Some developers prefer to sort propTypes declarations alphabetically to be able to find necessary declaration easier at the later time. Others feel that it adds complexity and becomes burden to maintain.
```
ljharb: let's make sure all `propTypes` and `defaultProps` mentions are surrounded by backticks.

Also, "to be able to find necessary declarations easier at a later time" and "becomes a burden to maintain"

docs/rules/jsx-sort-default-props.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Rule Details
+
+This rule checks all components and verifies that all defaultProps declarations are sorted alphabetically. A spread attribute resets the verification. The default configuration of the rule is case-sensitive.
```
ljharb: Can you elaborate about the spread attribute?

I would expect that sorts do not cross a spread boundary, but `d, a, ...obj, c, b` would be sorted into `a, d, ...obj, b, c`

fsmaia:
Any advances? I have some spare time, may I help?

b0gok:
@fsmaia sorry, haven't free time for a while. So help is welcome :)

ljharb:
@b0gok the ideal is for you to add @fsmaia to your fork, so they can update this PR directly - if not, then @fsmaia, please post here a link to your commits on your own fork, and I'll add them for you. Please do *not* open up a duplicate PR.

b0gok:
@ljharb ok.
@fsmaia I add you as collaborator in my fork.

docs/rules/jsx-sort-default-props.md
fsmaia: Done! Should I fix these typos in `sort-prop-types` and `jsx-sort-props` documentation too? Their introduction are identical.

fsmaia:
Added some SFC examples in the documentation and tests. Had to change the rule source, 'cause unnecessary/commented code was useful right in SFC scenario.

fsmaia:
@ljharb: Are we ready to merge? Any further work or change needed? Thanks

docs/rules/jsx-sort-default-props.md
```diff
@@ -0,0 +1,111 @@
...
+
+const Component = (props) => (...);
+Component.defaultProps = {
+  z: "z",
+  y: "y",
+  a: "a"
+};
+```
+
+The following patterns are considered okay and do not cause warnings:
```
ljharb: Is the wording of this sentence consistent with the other rules?

ljharb:
Also, please rebase and force push this so as to remove merge commits.

b0gok:
> Can we add some good and bad examples in the docs that use spreads?

@ljharb you meen something like this?
```js
const defaults = {
  foo: "foo"
};
const types = {
  foo: PropTypes.string,
  bar: PropTypes.string
};

function MyStatelessComponent({ foo, bar }) {
  return <div>{foo}{bar}</div>;
}

MyStatelessComponent.propTypes = types;
  ...defaults,
  bar: "bar"
};
```

ljharb:
Yes, just like that.

b0gok:
Added more tests with spreads and examples.

EvHaus:
Just curious, why not update the existing `jsx-sort-props` rule to sort defaultProps instead of create a separate rule? Is there a situations where you'd want proptypes sorted one way, but defaults sorted a different way?

PR #1485, opened 2017-10-17
Fix `boolean-prop-naming` with Object.spread syntax

track0x1:
There is an issue where using Object.spread syntax with this rule causes Eslint to break as it attempts to look for `node.value.property` in a spread object. We avoid validating spread nodes by checking for the `ExperimentalSpreadProperty` type.

Reopened #1482 (github wasn't showing new commit) @ljharb


track0x1:
@ljharb Done!

PR #1497, opened 2017-10-25
New rule: jsx-one-expression-per-line

TSMMark:
From issue https://github.com/yannickcr/eslint-plugin-react/issues/1491.

Tested against a small react-native app with 50-100 components and it works great.


jseminck:
There are only tests for components. Should it also warn on `<div><span>Hello World</span></div>` ?

TSMMark:
> I gather that you're choosing an indent of zero, so that the indent rule can correct it properly?

That's right. 

> There are only tests for components. Should it also warn on `<div><span>Hello World</span></div>` ?

Not sure why that would make a difference, but I added a case for ya.

tests/lib/rules/jsx-one-element-per-line.js
```diff
@@ -101,6 +101,20 @@ ruleTester.run('jsx-max-elements-per-line', rule, {
     parserOptions: parserOptions
   }, {
     code: [
+      '<div>',
+      '  <span />foo<input />',
+      '</div>'
+    ].join('\n'),
+    output: [
+      '<div>',
+      '  <span />foo',
```
ljharb: If the lack of whitespace before "foo" is important, then it's just as important after "foo" - meaning, this example can't be autofixed.

However,
```jsx
<div>
  <span /> foo <input />
</div>
```
definitely should become:
```jsx
<div>
  <span />
  foo
  <input />
</div>
```

TSMMark: @ljharb I believe white space in jsx is differnt from HTML and in your example the two jsx snippets are actually different. Foo in the first snippet will be rendered with a whitspace on either side, while foo in the second snippet will have no white space. 

However even if I'm correct the rule still does need to be patched. Currently the autofix does change the white space under certain conditions which is obviously unacceptable.

TSMMark: Using curlies and transforming into a string literal with spacing on either side seems the most accurate. Is this acceptible?

```jsx
<div>
  <span />
  {' foo '}
  <input />
</div>
```

TSMMark:
Hey @ljharb I spec'd out what I think is the best approach for handling white space. Basically, it comes down to *not* handling it and letting the user fix each case on their own. ptal if you could: 799dfa68accaaa50b38f01d176d96ad83c898864

If y'all agree I'll implement this behavior in the next couple days

tests/lib/rules/jsx-one-element-per-line.js
ljharb: Sure, that'd be great!

As long as the autofix is not changing the resulting whitespace that's rendered, we're good.

ljharb:
@TSMMark i think we'd need to either add an option, or have this be the default, so that the things that aren't autofixable *aren't* warned.

TSMMark:
For the record, I am attempting to handle all autofix cases but it is proving rather difficult.

TSMMark:
@ljharb Well, I had to throw everything out and start over, but now we can handle all cases! 😌 

How's it look?

tests/lib/rules/jsx-one-expression-per-line.js
```diff
@@ -0,0 +1,839 @@
...
+    parserOptions: parserOptions
+  }, {
+    code: [
+      '<div>',
+      '  foo {"bar"}',
+      '</div>'
+    ].join('\n'),
+    output: [
+      '<div>',
+      '  foo ',
```
ljharb: this can cause issues due to rules that strip trailing spaces; i think this needs to remain as `foo {"bar"}`, untouched.

tests/lib/rules/jsx-one-expression-per-line.js
```diff
@@ -0,0 +1,839 @@
...
+    parserOptions: parserOptions
+  }, {
+    code: [
+      '<div>',
+      '  {"foo"} { I18n.t(\'baz\') }',
+      '</div>'
+    ].join('\n'),
+    output: [
+      '<div>',
+      '  {"foo"} ',
```
ljharb: this trailing space should be removed, if the space on the next line is being explicitly preserved?

tests/lib/rules/jsx-one-expression-per-line.js
```diff
@@ -0,0 +1,839 @@
...
+      {message: '`{ I18n.t(\'baz\') }` must be placed on a new line'}
+    ],
+    parserOptions: parserOptions
+  }, {
+    code: [
+      '<Text style={styles.foo}>{ bar } <Text/> { I18n.t(\'baz\') }</Text>'
+    ].join('\n'),
+    output: [
+      '<Text style={styles.foo}>',
+      '{ bar } ',
```
ljharb: also here?

tests/lib/rules/jsx-one-expression-per-line.js
TSMMark: `foo ` and `{"bar"}` are 2 different expressions and I don't want them on the same line.

tests/lib/rules/jsx-one-expression-per-line.js
TSMMark: React parser already ignores trailing spaces, and `no-trailing-spaces` will chop them off anyway... but let me see if I can get rid of them

tests/lib/rules/jsx-one-expression-per-line.js
ljharb: right - but the jsx parser doesn't ignore the space in between `foo {"bar" }` - so to make it the same, with an autofix, you'd need:
```jsx
foo
{' '}
{'bar'}
```

tests/lib/rules/jsx-one-expression-per-line.js
ljharb: i agree that no-trailing-spaces could do the cleanup, but it seems easy enough to `.trim()` any newly autofixed lines :-)

tests/lib/rules/jsx-one-expression-per-line.js
TSMMark: Yes, if you look on the next line 117 you will see `{\' \'}`

TSMMark: It results in the same output after JSX parsing. I am following the guidelines:

> Since all rules are run again after the initial round of fixes is applied, it’s not necessary for a rule to check whether the code style of a fix will cause errors to be reported by another rule.

https://eslint.org/docs/developer-guide/working-with-rules#applying-fixes

The trailing space in `foo ` is inconsequential. Its only impact is offending `no-trailing-spaces`, which, as stated in the guide, is not our concern here.

tests/lib/rules/jsx-one-expression-per-line.js
TSMMark: See latest commit, best I can do right now.

lib/rules/jsx-one-expression-per-line.js
```diff
@@ -0,0 +1,181 @@
...
+
+        const childrenGroupedByLine = {};
+        const fixDetailsByNode = {};
+
+        children.forEach(child => {
+          let countNewLinesBeforeContent = 0;
+          let countNewLinesAfterContent = 0;
+
+          if (child.type === 'Literal') {
+            if (child.value.match(/^\s*$/)) {
```
lencioni: This should probably use `regex.test(str)` instead of match (here and elsewhere).

lib/rules/jsx-one-expression-per-line.js
```diff
@@ -0,0 +1,181 @@
...
+            childrenGroupedByLine[startLine].push(child);
+            if (!childrenGroupedByLine[endLine]) {
+              childrenGroupedByLine[endLine] = [];
+            }
+            childrenGroupedByLine[endLine].push(child);
+          }
+        });
+
+        Object.keys(childrenGroupedByLine).forEach(line => {
+          line = parseInt(line, 10);
```
lencioni: Instead of reassigning the argument here, can we use a new variable?

PR #1507, opened 2017-10-29
Fix no-unused-prop-types setState updater 

### LINKED ISSUE #1506: Invalid no-unused-prop-types catch
Hi,

Try this code:
```
import React from 'react';

class Test extends React.Component {
    constructor(props) {
        super(props);
        this.state = {
            foo: false,
        };
    }

    onFoo() {
        this.setState((prevState, props) => {
            props.doSomething();
            return {
                foo: true,
            };
        });
    }

    render() {
        return (
            <div>{this.state.foo}</div>
        );
    }
}

Test.propTypes = {
    doSomething: PropTypes.func.isRequired,
};

export default Test;
```
You will get an error on line `doSomething: PropTypes.func.isRequired,` despite that `doSomething` is used inside the async version of `setState()`. Yep, I know, it's a strange usage, but it should be handled imho ;)

Thanks!

petersendidit:
Handle when props are used in the setState update callback

Fix #1506

### DIFF
--- a/tests/lib/rules/no-unused-prop-types.js
+++ b/tests/lib/rules/no-unused-prop-types.js
@@ -829,10 +829,10 @@ ruleTester.run('no-unused-prop-types', rule, {
       type PropsA = { a: string }
       type PropsB = { b: string }
       type Props = PropsA & PropsB;
-      
+
       class MyComponent extends React.Component {
         props: Props;
-        
+
         render() {
           return <div>{this.props.a} - {this.props.b}</div>
         }
@@ -848,7 +848,7 @@ ruleTester.run('no-unused-prop-types', rule, {
 
         class Bar extends React.Component {
           props: Props & PropsC;
-          
+
           render() {
             return <div>{this.props.foo} - {this.props.bar} - {this.props.zap}</div>
           }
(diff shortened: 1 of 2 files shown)


tests/lib/rules/no-unused-prop-types.js
```diff
@@ -2111,6 +2111,24 @@ ruleTester.run('no-unused-prop-types', rule, {
       parser: 'babel-eslint',
       options: [{skipShapeProps: false}]
     }, {
+      // issue #1506
+      code: [
+        'class MyComponent extends React.Component {',
+        '  onFoo() {',
+        '    this.setState((prevState, props) => {',
+        '      props.doSomething();',
```
ljharb: does this work independent of what "props" is called?

jseminck:
Does this case work with the prop-types rule? If not: the code is really similar and it should be possible to apply the same logic there.

tests/lib/rules/no-unused-prop-types.js
petersendidit: Not yet, working a better version that will handle more cases. This code doesn't ] handle destructuring or anything like that.

ljharb: I think it's important that it handles that from the get-go.

petersendidit:
Should handle this case much better now. Added a test case for when the second argument to the setState updater is not named `props` and is not destructured. Let me know if there are other test cases that should be added.

jseminck:
What about the prop-types rule? Does it also need this functionality?

ljharb:
Yes, I think it does.

Issue #1435, opened 2017-09-14
Rule Suggestion: Disallow this (or this.props) in stateless components


benjick:
Would be super helpful when refactoring from classes to SFC

ljharb:
Done in #1509.

PR #1509, opened 2017-10-31
Add new rule for preventing `this` in SFCs

jomasti:
This rule will report errors when a stateless functional component is
attempting to use `this`.

This attempts to address the request in #1435.

### DIFF
--- a/docs/rules/no-this-in-sfc.md
+++ b/docs/rules/no-this-in-sfc.md
@@ -1,0 +1,61 @@
+# Prevent `this` from being used in stateless functional components (react/no-this-in-sfc)
+
+When using a stateless functional component (SFC), props/context aren't accessed in the same way as a class component or the `create-react-class` format. Both props and context are passed as separate arguments to the component instead. Also, as the name suggests, a stateless component does not have state on `this.state`.
+
+Attempting to access properties on `this` can be a potential error if someone is unaware of the differences when writing a SFC or missed when converting a class component to a SFC.
+
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```jsx
+function Foo(props) {
+  return (
+    <div>{this.props.bar}</div>
+  );
+}
+```
+
+```jsx
+function Foo(props, context) {
+  return (
+    <div>
+      {this.context.foo ? this.props.bar : ''}
+    </div>
+  );
+}
+```
+
+```jsx
+function Foo(props) {
+  if (this.state.loading) {
+    return <Loader />;
+  }
+  return (
+    <div>
+      {this.props.bar}
+    </div>
+  );
+}
+```
+
+The following patterns are **not** considered warnings:
+
+```jsx
+function Foo(props) {
+  return (
+    <div>{props.bar}</div>
+  );
+}
+```
+
+```jsx
+function Foo(props, context) {
+  return (
+    <div>
+      {context.foo ? props.bar : ''}
+    </div>
+  );
+}
+```
--- a/docs/rules/no-this-in-sfc.md
+++ b/docs/rules/no-this-in-sfc.md
@@ -18,10 +18,31 @@ function Foo(props) {
 ```
 
 ```jsx
+function Foo(props) {
+  const { bar } = this.props;
+  return (
+    <div>{bar}</div>
+  );
+}
+```
+
+```jsx
 function Foo(props, context) {
   return (
     <div>
       {this.context.foo ? this.props.bar : ''}
+    </div>
+  );
+}
+```
+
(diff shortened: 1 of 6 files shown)


jomasti:
Okay, I can update the rule to report any usage of this. You mean all usages of `this` beyond just the typical React usages?

ljharb:
Yes; an SFC has no reason to use `this` at all

jomasti:
Updated

tests/lib/rules/no-this-in-sfc.js
```diff
@@ -0,0 +1,106 @@
...
+    experimentalObjectRestSpread: true,
+    jsx: true
+  }
+};
+
+const ruleTester = new RuleTester({parserOptions});
+ruleTester.run('no-this-in-sfc', rule, {
+  valid: [{
+    code: `
+    function Foo(props) {
```
jseminck: Sorry, this is really a nit :) don't these template literals need an extra indentation?

```
code: `
  mystring...
`,
```

lib/util/Components.js
```diff
@@ -314,6 +314,13 @@ function componentRule(rule, context) {
           break;
         case 'ArrowFunctionExpression':
           property = 'body';
+          if (node[property] && node[property].type === 'BlockStatement') {
+            node = utils.findReturnStatement(node);
+            if (!node) {
+              return false;
+            }
+            property = 'argument';
+          }
```
jseminck: Is this code really needed?

`no-typos` rule also detects SFCs, it does it so ( code is at: https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/no-typos.js#L137 )

```
      MemberExpression: function(node) {
        const relatedComponent = utils.getRelatedComponent(node);
        if (
          relatedComponent && utils.isReturningJSX(relatedComponent.node)
        ) {
          // Should be an SFC at this point
        }
```

On the other hand, perhaps this is a better approach, I don't know too much about the internals there...

jomasti: Well, when I was calling `getParentStatelessComponent` before the change, it was returning any parent function, which is not what I wanted. So, I updated the function to check that the function is also returning JSX. When I did that with the previous version of `isReturningJSX`, it would fail on an `ArrowFunctionExpression` with a body. This caused several tests to fail. After I updated this case to check for the body, it resolved the failures. There may be multiple rules that do that sort of check, but I would prefer the logic to be in one place.

tests/lib/rules/no-this-in-sfc.js
jomasti: I think the need is up to personal preference. I can change it if that is the standard preference.

ljharb: imo the indentation of test cases in a non-indentation rule isn't that important, as long as it's readable.

docs/rules/no-this-in-sfc.md
```diff
@@ -0,0 +1,61 @@
...
+  }
+  return (
+    <div>
+      {this.props.bar}
+    </div>
+  );
+}
+```
+
+The following patterns are **not** considered warnings:
```
ljharb: let's also add non-warning patterns (and test cases) for destructured props/context

lib/rules/no-this-in-sfc.js
```diff
@@ -0,0 +1,42 @@
...
+    docs: {
+      description: 'Report "this" being used in stateless components',
+      category: 'Possible Errors',
+      recommended: false
+    },
+    schema: []
+  },
+
+  create: Components.detect((context, components, utils) => ({
+    MemberExpression: function(node) {
```
ljharb: nit: `MemberExpression(node) {`

lib/util/Components.js
```diff
@@ -442,7 +449,7 @@ function componentRule(rule, context) {
           return null;
         }
         // Return the node if it is a function that is not a class method and is not inside a JSX Element
-        if (isFunction && !isMethod && !isJSXExpressionContainer) {
+        if (isFunction && !isMethod && !isJSXExpressionContainer && utils.isReturningJSX(node)) {
```
ljharb: does `isReturningJSX` return true when something returns null?

Also, in React 16, components can return strings or arrays, so I'm not sure this detection will be reliable moving forward.

"React 16 fragments" will probably be something we have to solve in a separate PR anyways tho, so no need to address that now (i filed #1510)

jseminck:
Is there anything stopping from getting this merged? Just ran into an issue with this, using `this.props` in SFC by accident that caused a bug. Would be cool if linter catches it.

PR #1518, opened 2017-11-06
Handle no-props-unused-props in custom validators

### LINKED ISSUE #1068: unused prop warning Destructuring assignment in parameters list
This code will result in a unused prop warning for the `validate` prop even if its used inside the value validation.

```
MyComponent extends Component {
  static propTypes = {
    validate: PropTypes.bool,
    options: PropTypes.array,
    value: ({options, value, validate}) => {
      if (!validate) return; 
      if (options.indexOf(value) < 0)
        throw new Errow(`Value should be one of ${options.join(', ')}, passed: ${value}`);
    }
  }

  render() {
    return <ul>
      {this.props.options.map(option => 
        <li className={this.props.value == option && "active"}>{option}<li>
      )}
    </ul>
  }
}
```

If I change the value validation to the following, the warning disappears.

```
    ...
    value: (props) => {
      const {options, value, validate} = props;
      if (!validate) return; 
      if (options.indexOf(value) < 0)
        throw new Errow(`Value should be one of ${options.join(', ')}, passed: ${value}`);
    }
```

petersendidit:
When a custom prop validator is used mark props used inside the
validator as used.

Fixes #1068


tests/lib/rules/no-unused-prop-types.js
```diff
@@ -2208,6 +2208,54 @@ ruleTester.run('no-unused-prop-types', rule, {
...
+      `,
+      parser: 'babel-eslint'
+    }, {
+      // Issue #1068
+      code: `
+      class MyComponent extends Component {
+        static propTypes = {
+          validate: PropTypes.bool,
+          options: PropTypes.array,
+          value: function ({options, value, validate}) {
```
ljharb: can you also add a test case for `value({options, value, validate}) {`?

petersendidit: 👍 Added the test case.

ljharb:
Will merge this tomorrow if no objections.

PR #1522, opened 2017-11-08
Account for defaultProps in prefer-stateless rule

### LINKED ISSUE #1521: static defaultProps breaks prefer-stateless-function
The following is **not** caught by `prefer-stateless-function`:

```jsx
import React from 'react';
import PropTypes from 'prop_types';

export default class Thing extends React.Component {
  static defaultProps = {
     foo: ''
  };

  static propTypes = {
    foo: PropTypes.string
  };

  render() {
  }
}
```

However, by removing `static defaultProps`, the error is reported:
```jsx
import React from 'react';
import PropTypes from 'prop_types';

export default class Thing extends React.Component {
  static propTypes = {
    foo: PropTypes.string
  };

  render() {
  }
}
```

This is happening in 7.4.0.

jomasti:
Previously, a class component that could be converted to a stateless
functional component would not be reported as such if the component had
`defaultProps` defined as a class property.

Closes #1521

### DIFF
--- a/docs/rules/prefer-stateless-function.md
+++ b/docs/rules/prefer-stateless-function.md
@@ -6,7 +6,7 @@ Stateless functional components are simpler than class based components and will
 
 This rule will check your class based React components for
 
-* methods/properties other than `displayName`, `propTypes`, `render` and useless constructor (same detection as ESLint [no-useless-constructor rule](http://eslint.org/docs/rules/no-useless-constructor))
+* methods/properties other than `displayName`, `propTypes`, `contextTypes`, `defaultProps`, `render` and useless constructor (same detection as ESLint [no-useless-constructor rule](http://eslint.org/docs/rules/no-useless-constructor))
 * instance property other than `this.props` and `this.context`
 * extension of `React.PureComponent` (if the `ignorePureComponents` flag is true)
 * presence of `ref` attribute in JSX
--- a/tests/lib/rules/prefer-stateless-function.js
+++ b/tests/lib/rules/prefer-stateless-function.js
@@ -547,7 +547,54 @@ ruleTester.run('prefer-stateless-function', rule, {
       errors: [{
         message: 'Component should be written as a pure function'
       }]
-
+    }, {
+      code: `
+        class Foo extends React.Component {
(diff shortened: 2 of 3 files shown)


evandavis:
My hero

docs/rules/prefer-stateless-function.md
```diff
@@ -6,7 +6,7 @@ Stateless functional components are simpler than class based components and will
 
 This rule will check your class based React components for
 
-* methods/properties other than `displayName`, `propTypes`, `render` and useless constructor (same detection as ESLint [no-useless-constructor rule](http://eslint.org/docs/rules/no-useless-constructor))
+* methods/properties other than `displayName`, `propTypes`, `contextTypes`, `defaultProps`, `render` and useless constructor (same detection as ESLint [no-useless-constructor rule](http://eslint.org/docs/rules/no-useless-constructor))
```
ljharb: awesome, i was going to ask about contextTypes :-)

tests/lib/rules/prefer-stateless-function.js
```diff
@@ -496,6 +496,39 @@ ruleTester.run('prefer-stateless-function', rule, {
...
+        }
+      `,
+      parser: 'babel-eslint',
+      errors: [{
+        message: 'Component should be written as a pure function'
+      }]
+    }, {
+      code: `
+        class Foo extends React.Component {
+          static get defaultProps() {
```
ljharb: Let's also add a test case with `Foo.defaultProps = {}`, that doesn't use the babel-eslint parser (also this test case shouldn't need it either)

docs/rules/prefer-stateless-function.md
petersendidit: Doesn't seem to be any tests for contextTypes.

ljharb: If you felt like adding them here, that'd be super awesome :-D

jomasti: Oops. I thought I saw them, but it was `childContextTypes`. I added tests for `contextTypes` to this.

PR #1525, opened 2017-11-09
Add a rule enforcing explicit "type" attributes for buttons

Hypnosphi:
By default, `button` elements have `type="submit"`, which usually isn't what you actually want. This rule enforces the button type to be explicit

See also: https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/186#issuecomment-283727911


Hypnosphi:
I'm open to renaming it BTW, maybe `jsx-explicit-button-type` would be more... explicit =D

ljharb:
As for the name, `button-has-type` might work?

Hypnosphi:
> Can this perhaps not just apply to jsx, and also apply to React.createElement?

Are there some examples of rules that check both JSX and `createElement`?

petersendidit:
@Hypnosphi https://github.com/yannickcr/eslint-plugin-react/blob/d672588cc39334e215ff5993f859f159fc977d57/tests/lib/rules/no-children-prop.js#L40-L50

ljharb:
I'll give this a day, and then merge if there's no objections.

PR #1526, opened 2017-11-09
Code clean up

jomasti:
In the previous times I have made changes, I have noticed some duplication in code and wanted to deduplicate it. I have started that in this PR, and I wanted to open it up for some feedback. Is this desired? Should I go further?

### DIFF
--- a/lib/util/props.js
+++ b/lib/util/props.js
@@ -1,0 +1,46 @@
+/**
+ * @fileoverview Utility functions for props
+ */
+'use strict';
+
+const astUtil = require('./ast');
+
+/**
+ * Checks if the Identifier node passed in looks like a propTypes declaration.
+ * @param {ASTNode} node The node to check. Must be an Identifier node.
+ * @returns {Boolean} `true` if the node is a propTypes declaration, `false` if not
+ */
+function isPropTypesDeclaration(node) {
+  if (node && node.type === 'ClassProperty') {
+    // Flow support
+    if (node.typeAnnotation && node.key.name === 'props') {
+      return true;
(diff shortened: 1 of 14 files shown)


lib/util/props.js
```diff
@@ -0,0 +1,46 @@
...
+ */
+function isDefaultPropsDeclaration(node) {
+  const propName = astUtil.getPropertyName(node);
+  return (propName === 'defaultProps' || propName === 'getDefaultProps');
+}
+
+/**
+ * Checks if the PropTypes MemberExpression node passed in declares a required propType.
+ * @param   {ASTNode} propTypeExpression node to check. Must be a `PropTypes` MemberExpression.
+ * @returns {Boolean}                    `true` if this PropType is required, `false` if not.
```
ljharb: i think we can remove all these extra spaces; it's not aligned anyways

lib/util/props.js
```diff
@@ -0,0 +1,46 @@
...
+      return true;
+    }
+  }
+  return astUtil.getPropertyName(node) === 'propTypes';
+}
+
+/**
+ * Checks if the Identifier node passed in looks like a defaultProps declaration.
+ * @param   {ASTNode}  node The node to check. Must be an Identifier node.
+ * @returns {Boolean}       `true` if the node is a defaultProps declaration, `false` if not
```
ljharb: also here

jseminck:
This is really great. There's a lot of duplicate code between prop-types and no-unused-prop-types rules that should be extracted out. I think this is a good first step. 👍

PR #1527, opened 2017-11-09
Fix bad report by jsx-no-comment-textnodes

### LINKED ISSUE #1517: jsx-no-comment-textnodes rejecting html-entity "&#x2F;&#x2F;"
The rule `jsx-no-comment-textnodes` is warning when using html-entities for slashes
(`//` -> `&#x2F;&#x2F;`)

Example:
```html
<pre>&#x2F;&#x2F; TODO: Put instructions about usage here</pre>
```
This should not warn, since it is not valid comment.

jomasti:
The rule was reporting usage of two slashes via html entities at the
beginning of a literal between JSX elements. Since that is not a valid
comment, there should be no error reported by this rule.

Closes #1517

### DIFF
--- a/lib/rules/jsx-no-comment-textnodes.js
+++ b/lib/rules/jsx-no-comment-textnodes.js
@@ -34,7 +34,10 @@ module.exports = {
 
     return {
       Literal: function(node) {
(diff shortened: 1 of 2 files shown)


ljharb:
Thanks!

PR #1532, opened 2017-11-11
Add forbidDefaultForRequired option

### LINKED ISSUE #1524: Add option to `require-default-props` to forbid default props for required props
The [`require-default-props`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-default-props.md) rule allows required props to also have a default prop. It would be nice to add an option to warn against this behavior. 

Option:
`forbidDefaultForRequired` (`boolean`) - Setting this to `true` will forbid required props from also having a default prop declaration. Default `false`.

```js
schema: [{
 type: 'object',
      properties: {
        forbidDefaultForRequired: {
          type: 'boolean'
        }
      },
      additionalProperties: false,
}]
```
```js
function MyStatelessComponent({ foo, bar }) {
  return <div>{foo}{bar}</div>;
}

MyStatelessComponent.propTypes = {
  foo: PropTypes.string.isRequired,
};

  foo: 'something',
};
```

```js
  render() {
    return (
      <h1>Hello, {this.props.foo}</h1>
    );
  }

  static propTypes = {
    foo: PropTypes.string,
  };

  static defaultProps = {
    foo: "foo",
  };
}
```

The following patterns are **not** considered warnings:
```js
function MyStatelessComponent({ foo, bar }) {
  return <div>{foo}{bar}</div>;
}

MyStatelessComponent.propTypes = {
  foo: PropTypes.string,
};

  foo: 'something',
};
```
```js
  render() {
    return (
      <h1>Hello, {this.props.foo}</h1>
    );
  }

  static propTypes = {
    foo: PropTypes.string,
  };

  static defaultProps = {
    foo: "foo",
  };
}
```

jomasti:
New option for require-default-props rule that forbids defining default
props for props that are required.

This doesn't handle spread properties since the work needed should also work for the default state of the rule. If that were to happen, it could cause reports for previously valid code, and that seems more suitable for a major version bump. Therefore, I think that should be in a separate PR.

Closes #1524

### DIFF
--- a/docs/rules/require-default-props.md
+++ b/docs/rules/require-default-props.md
@@ -103,6 +103,19 @@ Greeting.defaultProps = {
 ```
 
 ```jsx
+type Props = {
+  foo: string,
+  bar?: string
+};
+
+function MyStatelessComponent(props: Props) {
+  return <div>Hello {props.foo} {props.bar}</div>;
+}
+```
+
+The following patterns are **not** considered warnings:
+
+```jsx
 class Greeting extends React.Component {
   render() {
     return (
@@ -120,19 +133,6 @@ class Greeting extends React.Component {
   };
 }
 ```
-
-```jsx
-type Props = {
-  foo: string,
-  bar?: string
-};
-
-function MyStatelessComponent(props: Props) {
-  return <div>Hello {props.foo} {props.bar}</div>;
-}
-```
-
-The following patterns are **not** considered warnings:
 
 ```jsx
 function MyStatelessComponent({ foo, bar }) {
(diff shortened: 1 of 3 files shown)


docs/rules/require-default-props.md
```diff
@@ -184,6 +184,91 @@ NotAComponent.propTypes = {
...
 
+## Rule Options
+
+```js
+...
+"react/require-default-props": [<enabled>, { forbidDefaultForRequired: <boolean> }]
+...
+```
+
+* `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
```
ljharb: This is basic eslint documentation; individual rule docs don't tend to explain this.

ljharb:
cc @sharmilajesupaul; a review would be great :-)

tests/lib/rules/require-default-props.js
```diff
@@ -1896,6 +1910,102 @@ ruleTester.run('require-default-props', rule, {
...
+        '    };',
+        '  }',
+        '  render() {',
+        '    return <div>Hello {this.props.foo}</div>;',
+        '  }',
+        '}'
+      ].join('\n'),
+      options: [{forbidDefaultForRequired: true}],
+      errors: [{
+        message: 'propType "foo" is required and should not have a defaultProp declaration.'
```
sharmilajesupaul: Will this also warn on the case where a spread property has an object literal referenced in the same file? 
eg: 
```js
const defaults = {
  foo: ''
}

  static propTypes = {
    foo: PropTypes.string.isRequired
  }

  static defaultProps = { ...defaults } // <---

  render () {
    return <div>Hello {this.props.foo}</div>;
  }
}
```

sharmilajesupaul: nvm read your pr discription

docs/rules/require-default-props.md
jomasti: I just modified this block from another rule. Should I remove it?

jomasti:
I created #1535 for tracking the spread property enhancement.

docs/rules/require-default-props.md
ljharb: Not a blocker then, altho i don’t think any rules should have this.

ljharb:
Will merge this tomorrow if no objections.

PR #1534, opened 2017-11-13
Fix docs for button-has-type

jzaefferer:
Looks like a copy&pasta error.

Ref #1525


jzaefferer:
PS: A release with the new rule would be neat

PR #1536, opened 2017-11-14
Extract duplicate code between prop-types and no-unused-prop-types #1

jseminck:
Follow-up for: https://github.com/yannickcr/eslint-plugin-react/pull/1526

Basically in `prop-types` and `no-unused-prop-types` there is quite a bit of duplicate code. I tried to do a HUGE refactoring to remove a lot of it but it turned out... well... not good 😄 So maybe smaller PRs is a better idea.

Basically made the "are we in a lifecycle method" detection similar between `prop-types` and `no-unused-prop-types`. Both rules were doing it differently but the results were the same. There was also duplicated methods like `inComponentWillReceiveProps`.

There should be some new cases detected now, e.g. in `prop-types` rule: `nextProps` should be considered not only for `componentWillReceiveProps`, but also for `shouldComponentUpdate`.

 On that note, I noticed that `prevProps` is not supported for `prop-types` rule, with a lifecycle method of `componentDidUpdate` for example. There's some further refactoring needed to align the two rules more to get that properly working which I'd like to tackle in a PR after this.

### DIFF
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -84,7 +84,7 @@ module.exports = {
 
     /**
      * Check if we are in a lifecycle method
-     * @return {boolean} true if we are in a class constructor, false if not
+     * @return {boolean} true if we are in a lifecycle method, false if not
      **/
     function inLifeCycleMethod() {
       let scope = context.getScope();
@@ -239,40 +239,6 @@ module.exports = {
         component &&
         !component.ignorePropsValidation
       );
-    }
-
-    /**
-     * Returns true if the given node is a React Component lifecycle method
-     * @param {ASTNode} node The AST node being checked.
-     * @return {Boolean} True if the node is a lifecycle method
-     */
-    function isNodeALifeCycleMethod(node) {
-      const nodeKeyName = (node.key || {}).name;
-      return (
-        node.kind === 'constructor' ||
-        nodeKeyName === 'componentWillReceiveProps' ||
-        nodeKeyName === 'shouldComponentUpdate' ||
-        nodeKeyName === 'componentWillUpdate' ||
-        nodeKeyName === 'componentDidUpdate'
-      );
-    }
-
-    /**
-     * Returns true if the given node is inside a React Component lifecycle
-     * method.
-     * @param {ASTNode} node The AST node being checked.
-     * @return {Boolean} True if the node is inside a lifecycle method
-     */
-    function isInLifeCycleMethod(node) {
-      if (node.type === 'MethodDefinition' && isNodeALifeCycleMethod(node)) {
-        return true;
-      }
-
-      if (node.parent) {
-        return isInLifeCycleMethod(node.parent);
-      }
-
-      return false;
     }
 
     /**
@@ -670,7 +636,7 @@ module.exports = {
             // let {firstname} = props
             const genericDestructuring = isPropAttributeName(node) && (
               utils.getParentStatelessComponent() ||
-              isInLifeCycleMethod(node)
+              inLifeCycleMethod()
             );
 
             if (thisDestructuring) {
@@ -992,7 +958,7 @@ module.exports = {
         // let {firstname} = props
         const statelessDestructuring = destructuring && isPropAttributeName(node) && (
           utils.getParentStatelessComponent() ||
-          isInLifeCycleMethod(node)
+          inLifeCycleMethod()
         );
 
         if (!thisDestructuring && !statelessDestructuring) {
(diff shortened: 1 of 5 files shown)


lib/rules/no-unused-prop-types.js
```diff
@@ -1058,7 +989,7 @@ module.exports = {
       ObjectPattern: function(node) {
         // If the object pattern is a destructured props object in a lifecycle
         // method -- mark it for used props.
-        if (isNodeALifeCycleMethod(node.parent.parent)) {
+        if (astUtil.inLifeCycleMethod(context)) {
```
ljharb: is this necessarily equivalent? `context.getScope()` won't be the same as `node.parent.parent`, necessarily.

jseminck:
Thanks for feedback. I'm at a conference for the next 2 days, but I will make "is in constructor" separately and deeper investigate the differences `isNodeALifeCycleMethod(node.parent.parent)` vs `astUtil.inLifeCycleMethod(context)`, which I do believe achieve the same thing, but technically they are doing something different.

lib/rules/no-unused-prop-types.js
```diff
@@ -668,10 +602,12 @@ module.exports = {
...
             );
             // let {firstname} = props
-            const genericDestructuring = isPropAttributeName(node) && (
-              utils.getParentStatelessComponent() ||
-              isInLifeCycleMethod(node)
-            );
+            const genericDestructuring =
+              isPropAttributeName(node) && (
+                utils.getParentStatelessComponent() ||
+                astUtil.inConstructor(context) ||
```
ljharb: which test case exposes this missing inConstructor check?

lib/rules/no-unused-prop-types.js
```diff
@@ -992,7 +928,8 @@ module.exports = {
         // let {firstname} = props
         const statelessDestructuring = destructuring && isPropAttributeName(node) && (
           utils.getParentStatelessComponent() ||
-          isInLifeCycleMethod(node)
+          astUtil.inConstructor(context) ||
```
ljharb: also here, which test case exposes the need for this check?

lib/rules/no-unused-prop-types.js
```diff
@@ -1058,7 +995,7 @@ module.exports = {
       ObjectPattern: function(node) {
         // If the object pattern is a destructured props object in a lifecycle
         // method -- mark it for used props.
-        if (isNodeALifeCycleMethod(node.parent.parent)) {
+        if (astUtil.inConstructor(context) || astUtil.inLifeCycleMethod(context)) {
```
ljharb: also here

lib/rules/prop-types.js
```diff
@@ -967,7 +942,7 @@ module.exports = {
         const directDestructuring =
           destructuring &&
           PROPS_REGEX.test(node.init.name) &&
-          (utils.getParentStatelessComponent() || inConstructor() || inComponentWillReceiveProps())
```
ljharb: also, which test cases warrant this change from just checking willReceiveProps, to checking all the lifecycle methods?

lib/rules/no-unused-prop-types.js
jseminck: Note that the old `isInLifeCycleMethod` function already was checking whether we are in constructor: `node.kind === 'constructor'`. Now it is split into `inConstructor` and `inLifecycleMethod` so the code is the same between this rule and `prop-types`.

In this case, the following two tests will fail when removing this line:

* https://github.com/yannickcr/eslint-plugin-react/blob/master/tests/lib/rules/no-unused-prop-types.js#L2037
* https://github.com/yannickcr/eslint-plugin-react/blob/master/tests/lib/rules/no-unused-prop-types.js#L1213

lib/rules/no-unused-prop-types.js
jseminck: Same as above: Note that the old isInLifeCycleMethod function already was checking whether we are in constructor: node.kind === 'constructor'. Now it is split into inConstructor and inLifecycleMethod so the code is the same between this rule and prop-types.

However, this case seems to be untested. 😞 In master we can remove this whole block and the tests still pass.

```
&& ( utils.getParentStatelessComponent() || isInLifeCycleMethod(node))
```

lib/rules/no-unused-prop-types.js
jseminck: Same as above.

Fails when removing inConstructor:
* https://github.com/yannickcr/eslint-plugin-react/blob/master/tests/lib/rules/no-unused-prop-types.js#L1228

When removing the `inLifecycleMethod` then 5 tests fail (2 for `componentWillReceiveProps`, 1 for `shouldComponentUpdate` and 2 for `componentDidUpdate`)

I suppose that we can also test https://reactjs.org/docs/react-component.html#componentwillupdate which also has `nextProps` as the first argument.

lib/rules/prop-types.js
jseminck: This is missing tests! Indeed we will now check more cases than just `componentWillReceiveProps`

ljharb:
@jseminck @alexzherdev is this PR still useful, now that #1911 and friends have landed?

alexzherdev:
It seems like what this PR contains has been achieved in spirit: there is now a single place that defines `inConstructor` and `inLifecycleMethod`, which is `lib/util/usedPropTypes.js`.

ljharb:
Are there test cases here we could salvage?

PR #1538, opened 2017-11-15
change allowTransparency to allowtransparency

dawidvdh:
Changed `allowTransparency`to `allowtransparency` to account for the following react error:

Warning: React does not recognize the `allowTransparency` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `allowtransparency` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

react - `16.1.1`
react-dom - `16.1.1`

### DIFF
--- a/lib/rules/no-unknown-property.js
+++ b/lib/rules/no-unknown-property.js
@@ -6,6 +6,7 @@
 'use strict';
 
 const docsUrl = require('../util/docsUrl');
+const versionUtil = require('../util/version');
 
 // ------------------------------------------------------------------------------
 // Constants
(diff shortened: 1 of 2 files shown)


lib/rules/no-unknown-property.js
```diff
@@ -108,7 +108,7 @@ const SVGDOM_ATTRIBUTE_NAMES = {
 
 const DOM_PROPERTY_NAMES = [
   // Standard
-  'acceptCharset', 'accessKey', 'allowFullScreen', 'allowTransparency', 'autoComplete', 'autoFocus', 'autoPlay',
+  'acceptCharset', 'accessKey', 'allowFullScreen', 'allowtransparency', 'autoComplete', 'autoFocus', 'autoPlay',
```
ljharb: Did react remove support for this property? I’m reasonably sure it was supported at the time this list was created. What version dropped support for it?

The lowercase one is just a generic custom property name; there’s no need to include it in the list at all.

jomasti: Looks like it was removed in facebook/react#10823, which was released in 16.1.0. It must have been missed when making the release notes.

lib/rules/no-unknown-property.js
```diff
@@ -108,7 +108,7 @@ const SVGDOM_ATTRIBUTE_NAMES = {
 
 const DOM_PROPERTY_NAMES = [
   // Standard
-  'acceptCharset', 'accessKey', 'allowFullScreen', 'allowTransparency', 'autoComplete', 'autoFocus', 'autoPlay',
```
ljharb: In that case, it should be present unless the React version declared in the settings is 16.1+.

gaearon:
Can you completely remove it? That's what React did. This attribute only works for IE8 and lower. React stopped supporting them more than a year ago.

ljharb:
We should be supporting what *is*, not what the documentation claims things are. It seems like the attribute arrives in the HTML in every browser prior to React 16.1; whether it has functionality or not isn't particularly relevant.

ljharb:
I rebased this and added the version logic.

PR #1539, opened 2017-11-16
Fix alignment bug in jsx-indent (and extract duplicate code)

### LINKED ISSUE #1246: Alignment bug?
Hey, I think this should be picked up but its not:

```
  return (
    <P>
      <Clickable inNewTab to={guideUrls[productType]}>
        <Icon className={styles.icon} name="document" size="medium" />Product guide (PDF)
    </Clickable>
    </P>
  );
```
I figure `</Clickable>` should have six spaces behind it.

jseminck:
Fixed this bug in the jsx-indent rule: Fixes #1246

Along the way, discovered that a very similar function was defined 3 different times with 3 different implementations in 3 different rules. Using a more advanced version of this function fixed the bug, and then decided to use the same function in all three rules thanks to the new `astUtil` helper file that we have.

### DIFF
--- a/lib/util/ast.js
+++ b/lib/util/ast.js
@@ -4,6 +4,30 @@
 'use strict';
 
 /**
+ * Find a return statment in the current node
+ *
+ * @param {ASTNode} ASTnode The AST node being checked
+ */
+function findReturnStatement(node) {
+  if (
+    (!node.value || !node.value.body || !node.value.body.body) &&
+    (!node.body || !node.body.body)
+  ) {
+    return false;
+  }
+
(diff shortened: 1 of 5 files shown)


lib/util/ast.js
```diff
@@ -4,6 +4,30 @@
 'use strict';
 
 /**
+ * Find a return statment in the current node
+ *
+ * @param {ASTNode} ASTnode The AST node being checked
+ */
+function findReturnStatement(node) {
```
jseminck: I re-ordered the functions alphabetically.

PR #1547, opened 2017-11-19
[New] add `prefer-exact-props` rule

### LINKED ISSUE #1455: Add a rule prefer-exact-props
With react and flow, it's very common to type props like this:

```javascript
type Props = {
  prop1: string,
  prop2: number
};
```

However this makes it possible to pass other props as well, as by default objects in Flow are not exact. So this can hide an issue where the developer removed a prop but the caller still use it:
```javascript
<Component prop1='foo' prop2={42} prop3='bar' />
```
Or made a typo (this example is possible is `prop2` is optional):
```javascript
<Component prop1='foo' porp2={42} />
```

To fix this with flow, the solution is to use exact objects:
```javascript
type Props = {|
  prop1: string,
  prop2: number
|};
```

It would be useful to have an eslint rule to force this pattern.

jomasti:
This rule checks if non-exact objects are used for defining prop types when using Flow along with using normal objects for prop types instead of something like `prop-types-exact` to enforce only passing the exact props to a component.

The exact prop wrapper functions for non-Flow usage are configurable by adding the `exact` property to an object setting in the `propWrapperFunctions` config option.

Closes #1455


docs/rules/prefer-exact-props.md
```diff
@@ -0,0 +1,128 @@
...
+  type Props = {|
+    foo: string
+  |}
+  function Component(props: Props) {
+    return <div />;
+  }
+```
+
+```jsx
+  import exact from 'prop-types-exact';
```
ljharb: This is a great default, but there's also `import { forbidExtraProps } from 'airbnb-prop-types'`, and I'm sure others might have a similar package. Can this be made configurable?

jomasti:
Do you think it should be a global setting instead of just an option for the rule? I'm unsure of the other use cases for it being a global setting.

ljharb:
@jomasti if it were a global setting, then other rules could access it to "unwrap" propTypes for their own uses.

jomasti:
@ljharb Wouldn't these particular "exact props" functions still be in `propWrapperFunctions` in order to facilitate that?

ljharb:
@jomasti right; the idea is that i'd be able to *move* `forbidExtraProps` etc out of `propWrapperFunctions` and into this new setting.

ljharb:
I guess an alternative schema for the existing setting could be: 
```json
"propWrapperFunctions": [
  "Object",
  { "property": "freeze", "object": "Object" },
  { "property": "forbidExtraProps", "exact": true }
]
```

maxammann:
What exactly is missing before this can be merged? Can we maybe split this task and add support for libs like airbnb/prop-types in an other task?

ljharb:
@maxammann that support already exists; see the "propWrapperFunctions" setting in the readme.

This PR needs to be updated per the above comments.

jomasti:
After doing some digging, it looks like the failures on ESLint 3 are due to versions 7.22+ of babel-eslint using `eslint-scope` over `escope` if available. This combination of the two results is weird errors for TypeAlias and maybe others. I'm not sure what I should do to rectify this.

ljharb:
We could modify the dep range, and for eslint 3 in Travis.yml manually downgrade it?

jomasti:
> We could modify the dep range, and for eslint 3 in Travis.yml manually downgrade it?

Would you prefer for the `before_script` to be extracted to a separate script with this extra logic?

jomasti:
Downgrading to a working version of `babel-eslint` breaks other tests, so it doesn't seem viable.

jomasti:
Yes, using `babel-eslint@7.2.1` with `eslint@3` breaks tests that use object spread in type aliases and the short syntax for fragments.

ljharb:
So what is it exactly about this pr that causes this breakage?

jomasti:
It's looking for the TypeAlias variable in the scope. This works fine on `eslint` 4 and above because both it and `babel-eslint` use `eslint-scope`.

lib/util/propWrapper.js
```diff
@@ -18,7 +13,41 @@ function isPropWrapperFunction(context, name) {
...
+  return new Set(exactPropWrappers);
+}
+
+function isExactPropWrapperFunction(context, name) {
+  const exactPropWrappers = getExactPropWrapperFunctions(context);
+  return searchPropWrapperFunctions(name, exactPropWrappers);
+}
+
+function formatPropWrapperFunctions(propWrapperFunctions) {
+  return Array.from(propWrapperFunctions).map(func => {
```
ljharb:
```suggestion
  return Array.from(propWrapperFunctions, func => {
```

lib/util/propWrapper.js
```diff
@@ -18,7 +13,41 @@ function isPropWrapperFunction(context, name) {
...
+    if (func.object && func.property) {
+      return `'${func.object}.${func.property}'`;
+    } else if (func.property) {
```
ljharb:
```suggestion
    }
    if (func.property) {
```

EvHaus:
This rule seems to be eerily similar to `no-unused-prop-types`. Can you help me understand what the major difference is? We might even want to add a note to the README to explain why you might use this over `no-unused-prop-types`.

alexzherdev:
I could be wrong, but my understanding of this is `forbidExtraProps` warns at runtime when a component's consumer passes in more than the component needs; whereas `no-unused-prop-types` warns if the component itself is not using some of the propTypes it declares. So it's two separate things.

FelixButzbach:
What happened to this PR? Is there any other rule that can/should be used?

I would love to enforce the usage of exact from prop-types-exact in our project as it seams to have only advantages and no disadvantages...

PR #1549, opened 2017-11-19
Add link to "button-has-type" rule docs in the main README

jseminck:
Added the link. Partially resolves: https://github.com/yannickcr/eslint-plugin-react/issues/1546
 
I was now wondering though, should this rule not have the `jsx-` prefix? As it's dealing with JSX specific linting only.

Unfortunately that would be a breaking change, since it is already released without this prefix right now.


ljharb:
No, because it shouldn't be dealing only with jsx - it should be able to catch React.createElement calls as well.

Issue #1546, opened 2017-11-19
Missing documentation


nwoltman:
There is no documentation for the `jsx-one-expression-per-line` rule (#1497).

Also, the readme is missing a link to the documentation for the `button-has-type` rule (#1525).

jseminck:
Thanks for reporting this! I added the link to `button-has-type` rule.

@TSMMark since you wrote the code for the `jsx-one-expression-per-line` rule, would you be willing to add the documentation as well?

TSMMark:
Sure thing. Anything I need to do other than the following?

 - add a file in docs/rules/
 - update readme with a link

jseminck:
That's it 👍

TSMMark:
See #1552

PR #1552, opened 2017-11-19
[Docs] jsx-one-expression-per-line

TSMMark:
related issue #1546


docs/rules/jsx-one-expression-per-line.md
```diff
@@ -0,0 +1,104 @@
+# One JSX Element Per Line (react/jsx-indent)
```
jomasti: Wrong rule name

ljharb: lol whoops, good catch. i'll fix.

TSMMark: My bad!

PR #1562, opened 2017-11-22
Add a forbid-dom-props rule

davazp:
A new rule forbid-dom-props was added. It is based on
forbid-component-props but it only forbid property on DOM Node
elements.

This can be useful to enforce some good practices, as not using 'id'
attributes or inline styles.

### DIFF
--- a/lib/rules/forbid-dom-props.js
+++ b/lib/rules/forbid-dom-props.js
@@ -28,11 +28,13 @@ module.exports = {
         forbid: {
           type: 'array',
           items: {
-            type: 'string'
-          }
+            type: 'string',
+            minLength: 1
+          },
+          uniqueItems: true
         }
       },
-      additionalProperties: true
+      additionalProperties: false
     }]
   },
 
--- a/lib/rules/forbid-dom-props.js
+++ b/lib/rules/forbid-dom-props.js
@@ -1,0 +1,68 @@
+/**
+ * @fileoverview Forbid certain props on DOM Nodes
+ * @author David Vázquez
+ */
+'use strict';
+
+// ------------------------------------------------------------------------------
+// Constants
+// ------------------------------------------------------------------------------
+
+const DEFAULTS = [];
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Forbid certain props on DOM Nodes',
+      category: 'Best Practices',
+      recommended: false
+    },
+
+    schema: [{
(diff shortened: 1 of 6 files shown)


davazp:
I used `dom` as they keyword for this rule as I didn't find any other rule acting only on DOM Nodes, and element refers to both Component and DOM Nodes.  Hopefully it is a good name!

I can imagine a more flexible set of option where you can forbid some properties based on the specific DOM Node, but I preferred to keep it similar to  `forbid-component-props` for now. I guess it can be extended if somebody finds that useful.

lib/rules/forbid-dom-props.js
```diff
@@ -0,0 +1,68 @@
...
+      type: 'object',
+      properties: {
+        forbid: {
+          type: 'array',
+          items: {
+            type: 'string'
+          }
+        }
+      },
+      additionalProperties: true
```
ljharb: this should be set to `false`

lib/rules/forbid-dom-props.js
```diff
@@ -0,0 +1,68 @@
...
+      description: 'Forbid certain props on DOM Nodes',
+      category: 'Best Practices',
+      recommended: false
+    },
+
+    schema: [{
+      type: 'object',
+      properties: {
+        forbid: {
+          type: 'array',
```
ljharb: this array should be forced to be unique

lib/rules/forbid-dom-props.js
```diff
@@ -0,0 +1,68 @@
...
+          items: {
+            type: 'string'
```
ljharb: these items should have a minLength of 1

davazp:
@ljharb There were tests for the 3 cases already I think?  I have anyway added extra SFC and class-based  *invalid* tests, that there was just one.

All code was derived from `forbid-component-props`, including the schema.  Would you like me to change also the schema of `forbid-component-props` to follow your requested changes as well?  Or keep it for backward compatibility?

Thanks,
David.

davazp:
I have also added Related section to the components to link to each other, following the example from `default-props-match-prop-types.md`.

ljharb:
Changing the schema of the existing rule would be a breaking change; leave that the same, and let's update this one.

davazp:
I updated the PR with the refined schema 👍

lib/rules/forbid-dom-props.js
ljharb: sorry, to clarify: 0 items in the array is OK, but an empty string should not be.

davazp:
I misread your original comment. I fixed the schema now.

PR #1563, opened 2017-11-22
Fix crash when using Unions in flow propTypes

justinanastos:
This will fix the crash when using Union flowtypes for the following rules:

- `prop-types`
- `no-unused-prop-types`

The library already escapes early when facing a `typeNode` it doesn't recognize; this will escape early when we can't even figure out the `typeNode` because `UnionTypeAnnotation`s don't have an `annotation.id`.

This will not actually resolve the issue of using unions; it will simply prevent eslint from crashing.

Fixes issue https://github.com/yannickcr/eslint-plugin-react/issues/1468

### DIFF
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -780,6 +780,10 @@ module.exports = {
           return declarePropTypesForObjectTypeAnnotation(annotation, declaredPropTypes);
         }
 
+        if (annotation.type === 'UnionTypeAnnotation') {
+          return true;
+        }
+
         const typeNode = typeScope(annotation.id.name);
 
         if (!typeNode) {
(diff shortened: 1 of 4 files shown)


jseminck:
This seems to work, but it would be nice to also have some test with some imported type that is used with union, like the main one reported from the issue. WDYT?

```
import type { FieldProps } from "redux-form"
type Props = {
  name: string,
} & FieldProps;

  props: Props;
  render() {
    const {name} = this.props;
    return name;
  }
}
```

ljharb:
@jseminck that's a great idea :-) more tests is always better!

justinanastos:
Absolutely agree! I’ll send an MR with another test with an import in the morning.

ljharb:
Thanks, a PR would be great (not sure what an MR is)

justinanastos:
Sorry @ljharb , been spending too much time using GitLab (MR vs PR).

justinanastos:
I created a new PR with additional tests (https://github.com/yannickcr/eslint-plugin-react/pull/1570). @jseminck I want to point out that this does _not_ actually apply the rules with the `imports` or the union propTypes in the previous PR; this just prevents eslint from crashing.

PR #1565, opened 2017-11-22
Fix two bugs with jsx-curly-spacing

lukeapage:
The first bug is that comments in children get deleted when the rule auto fixes

```
<a>
{ /*test*/ }
</a>
```
==>
```
<a>
{}
</a>
```
The second bug is that if the penultimate token spans multiple lines, like it is for example a template literal, then it reports an error when none should.
```
<pre>
{`
text
`}
</pre>
```
rule options and more info is in the test cases

### DIFF
--- a/lib/rules/jsx-curly-spacing.js
+++ b/lib/rules/jsx-curly-spacing.js
@@ -148,7 +148,33 @@ module.exports = {
      * @returns {boolean} Whether or not there is a newline between the tokens.
      */
     function isMultiline(left, right) {
-      return left.loc.start.line !== right.loc.start.line;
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -17,6 +17,7 @@ build
 node_modules
 npm-debug.log
 sftp-config.json
+.idea
```
ljharb: this should be reverted; individuals' IDEs' clutter shouldn't pollute upstream repos. You can add this to your global gitignore tho!

lukeapage: Ok fine. I just saw some other editors in there. will revert.

PR #1568, opened 2017-11-23
Update README.md

briandipalma:
Removed ESLint 1.x config information as it's probably used by very, very few people. Restructured config section to focus on quick set-up with defaults over more time consuming individual rule selection. No need to specify the plugin configuration as it works without it so it seems to be noise?

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,38 @@ $ npm install eslint-plugin-react --save-dev
 
 # Configuration
 
-Add `plugins` section and specify ESLint-plugin-React as a plugin.
+Use [our preset](#recommended) to get reasonable defaults:
+
+```json
+  "extends": [
+    "eslint:recommended",
+    "plugin:react/recommended"
+  ]
+```
+
+You should also specify settings that will be shared across all the plugin rules.
+
+```json5
+{
+  "settings": {
+    "react": {
+      "createClass": "createReactClass", // Regex for Component Factory to use,
+                                         // default to "createReactClass"
+      "pragma": "React",  // Pragma to use, default to "React"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,17 +21,39 @@ $ npm install eslint-plugin-react --save-dev
...
 # Configuration
 
-Add `plugins` section and specify ESLint-plugin-React as a plugin.
+Enable JSX support.
+
+With ESLint 2+
 
 ```json
 {
-  "plugins": [
```
ljharb: this is definitely necessary, and will not work without it being present somewhere (note that if using a shared config, like the airbnb config, it might be declaring this for you)

briandipalma: It's not necessary when using the recommended rule config (it defines itself). I'll edit to make that clearer.

briandipalma:
OK, added back in the plugins configuration, I see how it work now, thanks.

briandipalma:
I want to make more tweaks, we are repeating configuration.

briandipalma:
Actually no this is fine, keeping separate recommend/all blocks seems OK.

README.md
ljharb: gotcha, thanks.

README.md
```diff
@@ -80,6 +65,22 @@ Finally, enable all of the rules that you would like to use.  Use [our preset](#
   }
 ```
 
+You can also specify settings that will be shared across all the plugin rules.
```
ljharb: We should probably put this higher up, and change "can" to "should" - the react version in particular is pretty important.

briandipalma:
Moved and amended.

PR #1569, opened 2017-11-24
Not allow constructor in React components

zoomchik:
In most cases, we don't need a constructor in React Components instead we can use a static propTypes property to pass props in a component. This rule does not allow you to use a constructor in components.

### DIFF
--- a/docs/rules/no-constructor-in-component.md
+++ b/docs/rules/no-constructor-in-component.md
@@ -1,0 +1,52 @@
+# Do not allow constructor in React components (react/no-constructor-in-component)
+
+In most cases, we don't need a constructor in React Components instead we can use a static propTypes property to pass props in a component. This rule does not allow you to use a constructor in components.
+
+## Rule Details
+
+Will enforce not to use a constructor for React Components.
+
+The following patterns are considered bad:
+
+```jsx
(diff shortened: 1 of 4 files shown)


docs/rules/no-constructor-in-component.md
```diff
@@ -0,0 +1,24 @@
...
+In most cases, we don't need a constructor in React Components instead we can use a static propTypes property to pass props in a component. This rule does not allow you to use a constructor in components.
+
+## Rule Details
+
+Will enforce not to use a constructor for React Components.
+
+The following patterns are considered bad:
+```jsx
+  class OneComponent extends Component {
+    constructor(props) {
```
ljharb: @there's already an eslint core rule to cover this: `no-useless-constructor`.

Once class properties are stage 4, and actually in the language, there will doubtless be an eslint rule to `prefer-class-properties`, which would then reduce the constructor down to being useless, which that core rule would then eliminate.

In other words, I don't think there's any benefit to having a React-specific rule for an already-half-solved problem that applies to the entire language.

(Additionally, you *should* have a constructor if you're doing anything in it that can't be done with class properties - and there are definitely use cases for that)

PR #1573, opened 2017-11-28
Fix values of JSX-context configuration

MikeTaylor:
The document previously specified `"parens"` and `"ignore"` as the values in the optional configuration structure, but in face `true` and `false` are used.

### DIFF
--- a/docs/rules/jsx-wrap-multilines.md
+++ b/docs/rules/jsx-wrap-multilines.md
@@ -10,13 +10,13 @@ This rule optionally takes a second parameter in the form of an object, containi
 
 ```json
 {
-  "declaration": "parens",
-  "assignment": "parens",
-  "return": "parens",
-  "arrow": "parens",
-  "condition": "ignore",
-  "logical": "ignore",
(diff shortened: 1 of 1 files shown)


jomasti:
Are you referencing the README from `master` when using a version of the plugin before 7.5.0? The values given in the README are correct and reflect the defaults in the rule. I just verified that the configuration works as-is using v7.5.1.

MikeTaylor:
Yes, the version from `master`.

I see that the ESLin plugin we're using imports rules from eslint-plugin-react ^7.3.0 -- and that's the circumstances in which I found that I needed `true` and `false` values. So you're saying that you changed from `true`/`false` to `"parens"`/`"ignore"` in v7.5.0? If so, sorry to have wasted your time.

MikeTaylor:
Ah yes, I see that is the case: https://github.com/yannickcr/eslint-plugin-react/commit/4a719e8b03355ff5be26c7a211b645641430f71a

So I think you should simply close this PR without merging. Thanks for checking it.

PR #1576, opened 2017-11-30
Potential autofix improvement for jsx-wrap-mulitlines

sharmilajesupaul:
This change removes the newline that comes before opening parentheses, if `"parens-new-line"` is specified as an option and the multiline JSX starts on a new line.

i.e. when `"parens-new-line"` is specified,  instead of fixing these cases: 
```jsx
  <div prop={
    <div>
      <p>Hello</p>
    </div>
  }>
    <p>Hello</p>
  </div>

// &

  <div>
    {foo &&
      <div>
        <p>Hello World</p>
      </div>
    }
  </div>
```
like this: 
```jsx
  <div prop={
(
    <div>
      <p>Hello</p>
    </div>
)
  }>

// &

  <div>
    {foo &&
(
      <div>
        <p>Hello World</p>
      </div>
)
}
  </div>
```
this change will inline the opening parentheses: 

```jsx
  <div prop={(
    <div>
      <p>Hello</p>
    </div>
)}>

// &

  <div>
    {foo && (
      <div>
        <p>Hello World</p>
      </div>
)}
  </div>
```

~I did not change the position of closing parens in this PR, I'm not sure if there is another rule that might cover that? if not, I can also try and collapse the bottom parens to be more consistent.~

cc. @ljharb 

**Update**: this change will now inject the closing paren before the token after the jsx body, so the autofix will not leave the closing paren on its own line.

### DIFF
--- a/tests/lib/rules/jsx-wrap-multilines.js
+++ b/tests/lib/rules/jsx-wrap-multilines.js
@@ -249,6 +249,16 @@ const LOGICAL_NO_PAREN = `
   </div>
 `;
 
+const LOGICAL_PAREN_NEW_LINE_AUTOFIX = `
+  <div>
+    {foo && (
+<div>
+        <p>Hello World</p>
+      </div>
+)}
+  </div>
+`;
+
 const LOGICAL_PAREN_NEW_LINE = `
   <div>
     {foo && (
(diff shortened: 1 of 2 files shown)


tests/lib/rules/jsx-wrap-multilines.js
```diff
@@ -291,6 +300,15 @@ const ATTR_PAREN_NEW_LINE = `
   </div>
 `;
 
+const ATTR_PAREN_NEW_LINE_AUTOFIX = `
+  <div prop={(\n<div>
+      <p>Hello</p>
+    </div>\n)
```
ljharb: i think this one needs to be `)}>` on the next line

sharmilajesupaul:
@ljharb added change to cover closing paren, ptal thanks!

PR #1578, opened 2017-12-01
Export defaultConfig from sort-comp rule for programmatic use

Andarist:


ljharb:
This seems good; what’s the use case?

Andarist:
Just programmatic manipulation of this `defaultConfig` - so it's easy to insert i.e. `type-annotations` in the desired place without actually hardcoding the original default config, which also might (although probably won't) change over time.

I just hate hardcoding :P

PR #1580, opened 2017-12-02
New rule: Report usages of class methods besides render returning JSX

### LINKED ISSUE #578: Rule Proposal: functions returning JSX should be refactored into React components
Hey @yannickcr thanks for being awesome and making this great plugin, super valuable 😎.

I have noticed, that while working with a team of React devs - some devs will write functions that return JSX, but aren't components. This usually makes things difficult to refactor later on, as we need to split out all of these "jsx fragment functions" into real React components. 

For example, we often see things like this slip into our codebase:

```js
export default class ArticleList extends React.Component {
  
  renderImage(src, alt, caption) {
    return (
      <figure>
      <img src={src} />
      { caption ? <figcaption>{caption}>/figcaption> : null }
      </figure>
    );
  }

  renderArticle(articles) {
    return articles.map((article) => (
      <a href={item.url}>{this.renderImage(item.src, item.alt, item.caption)}</a>
    ));
  }

  render() {
    return (
      <div class="article-list">
        {this.renderItems(this.props.articles)}
      </div>
    );
  }  
}
```

The problem with the above code, other than from a purity standpoint, is that it becomes a bit of a rats nest at any level of complexity, plus it is difficult to refactor when you need to. Ideally this code would be written as such:

```js
function ArticleImage({ src, alt, caption }) {
  return (
    <figure>
    <img src={src} />
    { caption ? <figcaption>{caption}>/figcaption> : null }
    </figure>
  );
}

function ArticleItem({ url, src, alt, caption }) {
  return (
    <a href={url}>
      <ArticleImage src={src} alt={alt} caption={caption}/>
    </a>
  );
}

export default function ArticleList({ articles }) {
  return (
    <div class="article-list">
      {articles.map((article) => <ArticleItem {...article} />)}
    </div>
  );
}
```

The above code is much easier to reason about - each function that returns JSX has a well defined function signature (that of a react component) and each piece can be easily pulled out into its own component if the time comes.

### Proposal 

All of the above is just to demonstrate the proposal for the following rule:

> react/no-jsx-outside-component: If a function returns JSX, it should be a React component (class or function).

The rule would check all functions in a file:

 - if the function returned JSX
 - or if the function returned React.createElement calls
 - or if the function returned map/reduce calls which themselves return JSX
 - the function is not a React.Component#render method
 - or the function does not follow the React stateless function signature (`fn(props = {})`)

Then the rule would warn that the function should be a React (stateless or class) component.

---

Let me know what your thoughts are on this. If you're happy with the above proposal I'm happy to put the work into a PR for this.

jomasti:
I saw #578 and thought that the requested rule was useful. I personally don't like components that define extra render* methods that return JSX. It is a way to clean up the main render method, but the better choice is to move the logic to separate components. This rule will try to enforce that.

I wasn't sure about the rule name, so I'm open to a better name if it's offered.

Closes #578

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -303,6 +303,16 @@ function componentRule(rule, context) {
       return calledOnReact;
     },
 
+    isReturningArrayMap(node) {
+      return (
+        node.type === 'ReturnStatement' &&
+        node.argument &&
+        node.argument.type === 'CallExpression' &&
+        node.argument.callee.property &&
+        node.argument.callee.property.name === 'map'
+      );
+    },
+
     getReturnPropertyAndNode(ASTnode) {
       let property;
       let node = ASTnode;
@@ -320,6 +330,9 @@ function componentRule(rule, context) {
         default:
           node = utils.findReturnStatement(node);
           property = 'argument';
+      }
+      if (this.isReturningArrayMap(node)) {
+        return this.getReturnPropertyAndNode(node.argument.arguments[0]);
       }
       return {
         node: node,
--- a/tests/lib/rules/prop-types.js
+++ b/tests/lib/rules/prop-types.js
@@ -1275,15 +1275,6 @@ ruleTester.run('prop-types', rule, {
       ].join('\n')
     }, {
       code: [
-        'function Hello({names}) {',
-        '  return names.map((name) => {',
-        '    return <div>{name}</div>;',
-        '  });',
-        '}'
-      ].join('\n'),
-      parser: 'babel-eslint'
-    }, {
-      code: [
         'type Target = { target: EventTarget }',
         'class MyComponent extends Component {',
         '  static propTypes = {',
@@ -3531,6 +3522,18 @@ ruleTester.run('prop-types', rule, {
       `,
       errors: [{
         message: '\'fooBar\' is missing in props validation'
+      }],
+      parser: 'babel-eslint'
+    }, {
+      code: [
+        'function Hello({names}) {',
+        '  return names.map((name) => {',
+        '    return <div>{name}</div>;',
+        '  });',
+        '}'
+      ].join('\n'),
+      errors: [{
+        message: '\'names\' is missing in props validation'
       }],
       parser: 'babel-eslint'
     }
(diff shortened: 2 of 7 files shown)


lib/util/Components.js
```diff
@@ -303,6 +303,16 @@ function componentRule(rule, context) {
       return calledOnReact;
     },
 
+    isReturningArrayMap(node) {
```
ljharb: i'm concerned about this approach; there's other things that have a `.map` method. How can we know this is an array?

tests/lib/rules/prop-types.js
```diff
@@ -3533,6 +3524,18 @@ ruleTester.run('prop-types', rule, {
...
+    }, {
+      code: [
+        'function Hello({names}) {',
+        '  return names.map((name) => {',
+        '    return <div>{name}</div>;',
+        '  });',
+        '}'
+      ].join('\n'),
+      errors: [{
+        message: '\'names\' is missing in props validation'
```
ljharb: this is a good fix; but this should be able to go in a separate PR

ljharb (changes requested):

tests/lib/rules/prop-types.js
jomasti: It's a by-product of checking `map`. Considering this request and your concern above, should I remove the `map` support for now with it being added later as an enhancement?

ljharb: In the case of the `prop-types` rule, `names` isn't inside a `map`, so it should be caught regardless.

jomasti: It doesn't have to do with the JSX being returned in `map`. The reason it's being caught now is because `isReturningJSX` wasn't considering functions with return statements that are calling `Array#map`. So, in order to have this test case have an error, it required the same change that was made to check for returning `Array#map` in a component class method.

ljharb: Right - I'm saying that the bug fix for `prop-types` can be to consider these return statements, even if it ignores the code inside the `.map` callback.

jomasti: Okay, so in order to still support the `Array#map` support for this rule, you want me to find a solution that makes it work while also not resolving the issue with that particular test case?

ljharb: I'm thinking that the `prop-types` issue can be solved without having to guess that `.map` is "array map".

Then separately, this rule may be able to avoid making the same guess.

jseminck:
I like this rule also. 

For reference, I have a PR that adds similar functionality to the existing `prefer-stateless-function` rule, but there were a few cases that I didn't manage to catch: https://github.com/yannickcr/eslint-plugin-react/pull/1313

PR #1588, opened 2017-12-07
Allow for Boolean Prop Naming Custom Message

louisscruz:
The default message for boolean-prop-naming does not make it clear why the rule is being flagged. I'd like my organization to be able to immediately know what the issue is without them having to parse the Regex.

This PR introduces functionality such that the following configuration:

```js
{
    'react/boolean-prop-naming': [
      'error',
      {
        rule: '^(is|has)[A-Z]([A-Za-z0-9]?)+',
        message: 'Boolean prop names must begin with either \'is\' or \'has\'.'
      }
    ]
}
```

displays the custom message of `Boolean prop names must begin with either 'is' or 'has'.`.

### DIFF
--- a/lib/rules/boolean-prop-naming.js
+++ b/lib/rules/boolean-prop-naming.js
@@ -35,6 +35,10 @@ module.exports = {
           default: '^(is|has)[A-Z]([A-Za-z0-9]?)+',
           minLength: 1,
           type: 'string'
+        },
+        message: {
+          minLength: 1,
+          type: 'string'
         }
       },
       type: 'object'
@@ -131,7 +135,7 @@ module.exports = {
         const propName = getPropName(propNode);
         context.report({
           node: propNode,
-          message: `Prop name (${propName}) doesn't match rule (${config.rule})`,
+          message: config.message || `Prop name (${propName}) doesn't match rule (${config.rule})`,
           data: {
             component: propName
           }
--- a/lib/rules/boolean-prop-naming.js
+++ b/lib/rules/boolean-prop-naming.js
@@ -135,9 +135,11 @@ module.exports = {
         const propName = getPropName(propNode);
         context.report({
           node: propNode,
-          message: config.message || `Prop name (${propName}) doesn't match rule (${config.rule})`,
+          message: config.message || 'Prop name ({{ propName }}) doesn\'t match rule ({{ pattern }})',
           data: {
-            component: propName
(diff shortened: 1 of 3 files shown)


lib/rules/boolean-prop-naming.js
```diff
@@ -35,6 +35,10 @@ module.exports = {
           default: '^(is|has)[A-Z]([A-Za-z0-9]?)+',
           minLength: 1,
           type: 'string'
+        },
+        message: {
+          default: '',
```
ljharb: I don't think eslint pays attention to `default` here, fwiw.

Also, could we add `minLength: 1`? nobody should need to configure an empty string

rapheld:
@ljharb Did you mean to request changes on this PR. From your comment, it looks like you meant to approve.

ljharb:
@rapheld if you click on "show outdated" you'll see the change i requested.

lib/rules/boolean-prop-naming.js
```diff
@@ -131,7 +135,7 @@ module.exports = {
         const propName = getPropName(propNode);
         context.report({
           node: propNode,
-          message: `Prop name (${propName}) doesn't match rule (${config.rule})`,
+          message: config.message || `Prop name (${propName}) doesn't match rule (${config.rule})`,
```
ljharb: I think if we're making this configurable, then instead of using a template literal to interpolate `propName` and `config.rule`, we should use eslint's built-in string templating - that way the configurable message can use placeholders for those two things.

rapheld:
I had tried that, but for whatever reason, it wasn’t displaying a comment. Worked this time though.

lib/rules/boolean-prop-naming.js
louisscruz: Adding now.

louisscruz:
@ljharb String templating added!

louisscruz:
@ljharb Glad to see this merged! This is my first contribution to this project, and I'm trying to determine how I should bring this into my own project. What's the timeline for publication of this enhancement?

ljharb:
@louisscruz once merged, you're welcome to delete your branch whenever you like, or keep it forever :-) For this project, only @yannickcr can do a release, so it'll be released when they get a chance to do so (I can't speak to their timeline).

louisscruz:
@yannickcr Any plans for publication?

PR #1591, opened 2017-12-10
Ignore jsx-key if inside React.Children.toArray()

### LINKED ISSUE #1574: (react/jsx-key) still warns when wrapping with React.Children.toArray
The (react/jsx-key) rule seems to still warn for instances wrapped with the `React.Children.toArray` utility. 

This seems to be a bug since React no longer warns about missing keys when wrapping and considers it valid (the toArray utility provides keys to each item automatically)

```javascript

// React and lint rule both warn
{
  things.map(item => <Thing {...item} />)
}

// lint rule warns but React considers this to be valid
{
  React.Children.toArray(
    things.map(item => <Thing {...item} />)
  )
}
```

### Suggested Solution:

The rule accounts for instances wrapped with `React.Children.toArray` correctly and does not warn.

**_Edit:_**
I incorrectly described this as the no-array-index (as key) rule from this eslint plugin, but it is actually: `[eslint] Missing "key" prop for element in iterator (react/jsx-key)`. Updated the body of the issue plus the title to reflect the correct rule.

silvenon:
`jsx-key` rule should always succeed if we're inside `React.Children.toArray()` because omitting `key` there doesn't cause a React warning.

Fixes #1574.

### DIFF
--- a/tests/lib/rules/jsx-key.js
+++ b/tests/lib/rules/jsx-key.js
@@ -176,6 +176,33 @@ ruleTester.run('jsx-key', rule, {
       `,
       features: ['types', 'no-babel-old'],
     },
+    { code: 'React.Children.toArray([1, 2 ,3].map(x => <App />));' },
+    {
+      code: `
+        import { Children } from "react";
+        Children.toArray([1, 2 ,3].map(x => <App />));
+      `,
+    },
+    {
+      // TODO: uncomment the commented lines below
+      code: `
+        import Act from 'react';
+        import { Children as ReactChildren } from 'react';
+        
+        const { Children } = Act;
+        const { toArray } = Children;
+        
+        Act.Children.toArray([1, 2 ,3].map(x => <App />));
+        Act.Children.toArray(Array.from([1, 2 ,3], x => <App />));
+        Children.toArray([1, 2 ,3].map(x => <App />));
+        Children.toArray(Array.from([1, 2 ,3], x => <App />));
+        // ReactChildren.toArray([1, 2 ,3].map(x => <App />));
+        // ReactChildren.toArray(Array.from([1, 2 ,3], x => <App />));
+        // toArray([1, 2 ,3].map(x => <App />));
+        // toArray(Array.from([1, 2 ,3], x => <App />));
+      `,
+      settings,
+    },
   ]),
   invalid: parsers.all([
     {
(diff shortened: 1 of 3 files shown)


tests/lib/rules/jsx-key.js
```diff
@@ -37,7 +37,15 @@ ruleTester.run('jsx-key', rule, {
...
     {code: 'var App = () => <div />;'},
     {code: '[1, 2, 3].map(function(x) { return; });'},
-    {code: 'foo(() => <div />);'}
+    {code: 'foo(() => <div />);'},
+    // #1574
+    {code: 'React.Children.toArray([1, 2 ,3].map(x => <App />));'},
+    {
+      code: `
+      import { Children } from "react";
+      Children.toArray([1, 2 ,3].map(x => <App />));
```
ljharb: Let's also include a test case that has `const { toArray } = Children`, and another that has something called `toArray` and `Children` that aren't imported from React.

silvenon: Good idea, my changes wouldn't catch those cases. I'm also worried about other likely cases, this is a test case that I'm currently trying to get working as valid:

```js
import React from 'react';
import { Children as ReactChildren } from 'react';

const { Children } = React;
const { toArray } = Children;

React.Children.toArray([1, 2 ,3].map(x => <App />));
ReactChildren.toArray([1, 2 ,3].map(x => <App />));
Children.toArray([1, 2 ,3].map(x => <App />));
toArray([1, 2 ,3].map(x => <App />));
```

I need to get used to hunting down variable/import declarations. 😄

silvenon: One more question: should CommonJS also work?

```js
const { Children } = require('react');
```

silvenon: One more question, should CommonJS also work?

```js
const { Children } = require('react');
```

ljharb: Yes, definitely.

ljharb: ping @silvenon, are you interested in completing this PR?

silvenon: I am, but I don't have the skills for it yet, sorry 😥

ljharb:
I rebased this, and with virtually no modification, the tests (which still fail on master) seem to pass now on this PR!

ljharb:
I'm going to go ahead and merge as-is, since it now respects the React pragma, and this is a long-awaited fix.

ljharb (approved):

silvenon:
Nice, thank you for finishing this! ❤️

PR #1610, opened 2017-12-15
Exclude references to this.state in setState callback

pfhayes:


pfhayes:
Addresses https://github.com/yannickcr/eslint-plugin-react/issues/1597

pfhayes:
Any updates on this? This bug is still present in 7.6.1

SlaUA:
The bug is still present in 7.7.0

pfhayes:
@slaua can you supply an example that reproduces the issue? I have not witnessed this

SlaUA:
Looks like I faced with cache...

Aequitas737:
Still present in 7.10.0 when the callback calls a function which references `this.state`.

`myFunc = () => {
   this.setState({
            ...prop,
        }, () => this.doStuff());
    }

    doStuff= () => {
        console.log(this.state.asdf);
    }
`

Can work around this by just passing this.state into doStuff: `this.doStuff(this.state)`

ljharb:
@Aequitas737 altho you shouldn't put arrow functions in class fields, that seems like it should work. Can you file a new issue?

PR #1614, opened 2017-12-16
 Allow using this.state within cb in this.setState(updater, cb)

### LINKED ISSUE #1604: False positive no-access-state-in-setstate
Rule `no-access-state-in-setstate` is false positive if `this.state` is used in callback (`setState`'s second argument).

Example:
```js
  handleChange = (value) => {
    this.setState(prevState => someLogicToMutateState(prevState, value), () => {
      this.props.onChange(this.state); // accessing this.state here is save but it's reported
    });
  };
```
[React documentation](https://reactjs.org/docs/react-component.html#setstate) says: 

> …use `componentDidUpdate` or a `setState` callback (`setState(updater, callback)`), either of which are guaranteed to fire after the update has been applied.

jaaberg:
Fixes #1604. 

It should be allowed to use this.state within cb in this.setState(updater, cb).

### DIFF
--- a/lib/rules/no-access-state-in-setstate.js
+++ b/lib/rules/no-access-state-in-setstate.js
@@ -79,6 +79,13 @@ module.exports = {
         ) {
           let current = node;
           while (current.type !== 'Program') {
+            // Do not trigger if this.state is used within a callback of this.setState (#1604)
+            if (
+              isSetStateCall(current.parent) &&
+              current.parent.arguments[1] === current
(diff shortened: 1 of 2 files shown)


jaaberg:
Wops, seems like I was late to the party. Consider #1610 instead of this one.

ljharb:
Thanks! Indeed tho, closing as a duplicate of #1610 :-)

PR #1616, opened 2017-12-17
Find string literals as a part of BinaryExpressions in jsx-no-literals

### LINKED ISSUE #1511: react/jsx-no-literals rule shows warning with template strings
Hello,

i'm using eslint@4.10.0 and eslint-plugin-react@7.4.0. The linter shows a warning if a join a string with "template strings", but not if I join it with the + operator:

```
<div className={'test' + name} />
// No warning
```

```
<div className={`test ${name}`} />
// Strings not allowed in JSX files (react/jsx-no-literals)
```

Is this a known issue?

Best regards

jaaberg:
Fixes #1511

### DIFF
--- a/lib/rules/jsx-no-literals.js
+++ b/lib/rules/jsx-no-literals.js
@@ -42,20 +42,24 @@ module.exports = {
       });
     }
 
-    function getValidation(node) {
+    function getParentIgnoringBinaryExpressions(node) {
       let current = node;
       while (current.parent.type === 'BinaryExpression') {
         current = current.parent;
       }
+      return current.parent;
+    }
+
+    function getValidation(node) {
+      const parent = getParentIgnoringBinaryExpressions(node);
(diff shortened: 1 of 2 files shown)


jaaberg:
There exists tests for ``<Foo bar={`Test`} />`` and  ``<Foo>{\`Test\`}</Foo>``, but we could add one for ``<Foo bar={`Test ${baz}`} />`` to be explicit(?)

ljharb:
@jaaberg that'd be great; also `{`foo` + 'bar'}`, `{`foo` + `bar`}` and `{'foo' + `bar`}`?

jaaberg:
@ljharb Great, the tests actually uncovered another bug. The rule should find all strings/literals, so ``{`foo` + 'bar'}`` should actually give two errors, one for `foo` and another one for `bar`. 

Added tests in f27ebc2 and fixed in 28581df.

PR #1617, opened 2017-12-17
jsx-no-target-blank to check for forms with target="_blank"

### LINKED ISSUE #1143: Add or modify rule for no target="_blank" to include form elements
```
      <form
        method="POST"
        action={url}
        target="_blank"
      >
          <formStuff />
      </form>
```

I could be wrong but isn't a form post open to the same vulnerabilities as a <a> with target="_blank".  The current implementation doesn't flag this.

jaaberg:
This should fix #1143.

I am not sure if this should be an option or not (as suggested in the issue)(?).

### DIFF
--- a/lib/rules/jsx-no-target-blank.js
+++ b/lib/rules/jsx-no-target-blank.js
@@ -121,6 +121,14 @@ module.exports = {
         },
         warnOnSpreadAttributes: {
           type: 'boolean'
+        },
+        links: {
+          type: 'boolean',
+          default: true
+        },
+        forms: {
+          type: 'boolean',
+          default: false
         }
       },
       additionalProperties: false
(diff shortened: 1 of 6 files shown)


jaaberg:
Added option `preventInForms` to opt-in to validating target-blank in forms with tests and docs.

ljharb:
@jaaberg what about a schema like this:
```json5
{
  "forms": <boolean>, // default false
  "links": <boolean>, // default true
}
```

That way it's an easy change to flip the default from false to true in the next semver-major, and the schema is also extensible if we want to add other categories in the future?

jaaberg:
@ljharb Thanks for feedback :) Updated the PR.

lib/rules/jsx-no-target-blank.js
```diff
@@ -38,19 +41,40 @@ module.exports = {
...
     return {
       JSXAttribute: function(node) {
-        if (node.parent.name.name !== 'a') {
+        const config = Object.assign({
+          links: true,
+          forms: false
+        }, context.options[0]);
+
+        if (
+          (config.links ? node.parent.name.name !== 'a' : true) &&
```
ljharb: `!config.links || node.parent.name.name !== 'a'`?

jaaberg: Yeah I was thinking about it, do you think it is more readable?

ljharb: hm, i'm not really sure. i'm not a fan of ternaries that select booleans, but as-is might be more readable in that case.

ljharb:
Rebased, and added a global `formComponents` setting to go with the `linkComponents` setting that was added after this PR was created.

ljharb:
@jaaberg i think this can land if it's rebased on top of the newly added `allowReferrer` and `warnOnSpreadAttributes` options.

ljharb (approved):

PR #1628, opened 2017-12-27
Teach no-unused-prop-types about parameter aliases

akx:
This PR extends `no-unused-prop-types` to understand `const {...} = props;` destructuring even in methods that are not lifecycle methods, _iff_ the `props` name refers to a function parameter whose default value refers to the same field in the instance. (Now that's a complex sentence...)

Point being, it's fairly common to refactor code called from a lifecycle method into reusable bits, but when not being called from a lifecycle method (that might have the `nextProps` to pass in), one wants to use the component's current props, like so:

```javascript
  componentDidMount() {
    this.helperMethod();
  }
  componentWillReceiveProps(nextProps) {
    this.helperMethod(nextProps);
  }
  helperMethod(props = this.props) {
    const { message } = props;
    // ....
  }
}
MyComponent.propTypes = { message: PropTypes.string };
```

Previously, this would have failed (`message` was considered a defined prop that was never used), since the plugin did (and still doesn't) do code path analysis to realize that `helperMethod` is called from a lifecycle method.

Instead of full-blown code path analysis, we just assume that the pattern `x = this.x` (where `x` is one of the "well-known" props names `props`, `nextProps` and `prevProps`) in function arguments is this usage.

This is my first ever foray into writing an Eslint plugin or manipulating a Babylon AST, so I'm all ears for advice in case I'm doing something extra silly. :)

### DIFF
--- a/tests/lib/rules/no-unused-prop-types.js
+++ b/tests/lib/rules/no-unused-prop-types.js
@@ -2480,6 +2480,22 @@ ruleTester.run('no-unused-prop-types', rule, {
         }
         MyComponent.propTypes = { * other() {} };
       `
+    },
+    {
+      code: `
+        class MyComponent extends React.Component {
+          componentDidMount() {
+            this.helperMethod();
+          }
+          componentWillReceiveProps(nextProps) {
+            this.helperMethod(nextProps);
+          }
+          helperMethod(props = this.props) {
+            const { message } = props;
+          }
+        }
+        MyComponent.propTypes = { message: PropTypes.string };
+      `
     }
   ],
 
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -276,6 +276,20 @@ module.exports = {
     }
 
     /**
+     * Figure out whether the source of this destructuring node has a default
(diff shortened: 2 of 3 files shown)


tests/lib/rules/no-unused-prop-types.js
```diff
@@ -2480,6 +2480,22 @@ ruleTester.run('no-unused-prop-types', rule, {
...
+    {
+      code: `
+        class MyComponent extends React.Component {
+          componentDidMount() {
+            this.helperMethod();
+          }
+          componentWillReceiveProps(nextProps) {
+            this.helperMethod(nextProps);
+          }
+          helperMethod(props = this.props) {
```
ljharb: This should work. Passing around the entire props object is a massive antipattern anyways, but in this case, there’s no guarantee that nobody isnt calling it at runtime with “not a props object”.

The correct and only solution here is to destructured the props first, and pass individual props into the methods.

akx:
Closing for inactivity and since I use TypeScript these days anyway. :)

ljharb:
Note that propTypes are much more expressive than TS/Flow types, so you should be using them along with TS.

PR #1635, opened 2018-01-08
Add URL to rule documentation to the metadata

Arcanemagus:
ESLint v4.15.0 added an official location for rules to store a URL to their documentation in the rule metadata in eslint/eslint#9788. This adds the URL to all the existing rules so anything consuming them can
know where their documentation is without having to resort to external packages to guess.

### DIFF
--- a/lib/util/docsUrl.js
+++ b/lib/util/docsUrl.js
@@ -1,0 +1,7 @@
+'use strict';
+
+function docsUrl(ruleName) {
+  return `https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules/${ruleName}.md`;
+}
+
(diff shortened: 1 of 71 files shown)


lib/util/docsUrl.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+function docsUrl(ruleName) {
+  return `https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules/${ruleName}.md`;
+}
+
+module.exports = {
+  docsUrl
```
ljharb: this should just be `module.exports = docsUrl`

Arcanemagus: Fixed.

PR #1646, opened 2018-01-14
`no-unused-prop-types`: mark all props as used when spreading props

### LINKED ISSUE #1601: react/no-unused-prop-types no working when use `{...this.state}`
HI ALL,
I found a problem with `react/no-unused-prop-types` rule.
When I use `{...this.state}` syntax in React component, the rule can't find the unused prop.
I don't know is this a bug with eslint-plugin-react? 
Please Help. Thanks.

### Component

```js
import logo from './logo.svg';
import './App.css';

class App extends Component {
  static propTypes = {
    bar: PropTypes.string.isRequired,
  };

  state = {
    style: { color: 'red' },
  };

  render() {
    return (
      <div className="App">
        <header className="App-header">
          <img src={logo} className="App-logo" alt="logo" />
          // when I add `{...this.state}`, the rule is no working
          <h1 className="App-title" {...this.state} >Welcome to React</h1>
        </header>
        <p className="App-intro">
          To get started, edit <code>src/App.js</code> and save to reload.
        </p>
      </div>
    );
  }
}

export default App;
```

### .eslintrc

```json
{
  "extends": "react-app",
  "rules": {
    "react/no-unused-prop-types": [
      "error",
      {
        "customValidators": [],
        "skipShapeProps": true
      }
    ]
  }
}
```
### versions

* eslint: 4.10.0
* eslint-plugin-react: 7.4.0

### Test github repos

https://github.com/zhaozhiming/eslint-plugin-react-no-unused-prop-types

zhaozhiming:
Hi, I try to fix #1601. Please review. Thanks.

### DIFF
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -1041,10 +1041,19 @@ module.exports = {
       },
 
       JSXSpreadAttribute: function(node) {
-        const component = components.get(utils.getParentComponent());
-        components.set(component ? component.node : node, {
-          ignorePropsValidation: true
-        });
+        // includes `{...props}`, `{...others}` etc.
+        const isIdentifier = node.argument.type === 'Identifier';
+        // includes `{...this.props}`
+        const isThisProps =
+          node.argument.type === 'MemberExpression' &&
+          node.argument.object.name === 'this' &&
+          node.argument.property.name === 'props';
+        if (isIdentifier || isThisProps) {
+          const component = components.get(utils.getParentComponent());
+          components.set(component ? component.node : node, {
+            ignorePropsValidation: true
+          });
+        }
       },
 
       MethodDefinition: function(node) {
--- a/tests/lib/rules/no-unused-prop-types.js
+++ b/tests/lib/rules/no-unused-prop-types.js
@@ -4027,7 +4027,28 @@ ruleTester.run('no-unused-prop-types', rule, {
       errors: [{
         message: '\'lastname\' PropType is defined but prop is never used'
       }]
+    }, {
+      // `no-unused-prop-types` with `{...this.state}` - [Issue #1601]
+      code: `
+        class MyComponent extends React.Component {
+          static propTypes = {
+            foo: PropTypes.string,
+          }
+          constructor(props, context) {
+            super(props, context)
+            this.state = { className: 'bar' }
+          }
+          render() {
+            return <div {...this.state}>Hello</div>;
+          }
+        }
+      `,
+      parser: 'babel-eslint',
+      errors: [{
+        message: '\'foo\' PropType is defined but prop is never used'
+      }]
     }
+
 
     /* , {
       // Enable this when the following issue is fixed


lib/rules/no-unused-prop-types.js
```diff
@@ -1041,10 +1041,19 @@ module.exports = {
       },
 
       JSXSpreadAttribute: function(node) {
-        const component = components.get(utils.getParentComponent());
-        components.set(component ? component.node : node, {
-          ignorePropsValidation: true
-        });
+        // includes `{...props}`, `{...others}` etc.
+        const isIdentifier = node.argument.type === 'Identifier';
```
ljharb: Does this mean that if i do `<div {...someRandomThing} />` it will cause props not to be marked as unused, as well?

zhaozhiming: Yes, it's not good. But I found a test case use `{...other}` should props be marked as used.  
This test case and `<div {... someRandomThing} />` seems have conflict.

https://github.com/yannickcr/eslint-plugin-react/blob/8237551309e2beaaa751d3d3fcc8bdb15e0f9a1b/tests/lib/rules/no-unused-prop-types.js#L756-L763

ljharb: That should be marked as used because `other` resolves to the props object. If it *doesn't* resolve to that, it should not be marked as used.

zhaozhiming: I debug into the method `JSXSpreadAttribute`.  
Both the 2 test case as below the parameter `node` object is the same.  
It seems difficult to verify the `node` object does resolve to the props object or not.
Could you give me some suggestions? Thanks. @ljharb 

```js
// props should be marked as used
render() {
  var {...other} = this.props;
  render (<div {...other} />); // `node` object is `{...other}`

// props should be marked as unused
render() {
  const other = [{ className: 'foo' }];
  render (<div {...other} />);  // `node` object is `{...other}` too
```

ljharb: That's a good question, and I'm not sure. You may need to set up visitors for a bunch of other node types, and keep track of their names and what name they point to?

zhaozhiming: My thought is: 
1. find all nodes relate with the jsx spread attribute name (which node name is equal with the jsx spread attribute name, such as `other`) in the component.
2. check every related node whether resolves to the props object or not.
3. if yes, set the `ignorePropsValidation` to true and break loop.

But I'm not familiar with AST. I think it's difficult for me to handle this problem 😂.  
Maybe I need someone help.

zhaozhiming: Perhaps we could make the rule a bit looser.

The case as below should be marked as used, others are not: 
* `{...this.props}`
* `{...props}` (not matter what `props` is)

ljharb: that's probably a better place to leave it then wrongly marking them as used :-)

PR #1650, opened 2018-01-20
add missing meta to jsx-tag-spacing

flyerhzm:
just add some missing meta


lib/rules/jsx-tag-spacing.js
```diff
@@ -221,7 +221,11 @@ function validateBeforeClosing(context, node, option) {
 
 module.exports = {
   meta: {
+    deprecated: true,
```
ljharb: why is this marked deprecated?

flyerhzm:
@ljharb sorry, I read the wrong doc. It's updated now.

Issue #1607, opened 2017-12-14
`no-typos`: false positive


ljharb:
Here's the code:
```jsx
import React from 'react';
import PropTypes from 'prop-types';
import { forbidExtraProps } from 'airbnb-prop-types';

export const CustomShape = PropTypes.shape({});

export default function withWrapper(Component) {
  class Wrapper extends React.Component {
  }

  Wrapper.propTypes = forbidExtraProps({
    ...Component.propTypes,
    phrases: CustomShape.isRequired,
  });

  Wrapper.childContextTypes = {
    ...Component.contextTypes,
    phrases: CustomShape.isRequired,
  };

  Wrapper.displayName = `withWrapper(${Component.displayName || Component.name})`;

  return Wrapper;
}
```

The error:
```sh
Typo in declared prop type: isRequired  react/no-typos
```

This shouldn't error, both because it's not a typo, and because `PhraseShape` isn't a component.

Issue #1389, opened 2017-08-22
no-typos rule doesn't handle custom proptypes classes


modosc:
with the `7.2.1` -> `7.3.0` upgrade this rule started breaking on this code:
```
import * as MyPropTypes from 'lib/my-prop-types'

const Foo = ({ bar } = {} ) => <div className={bar} />

Foo.propTypes = {
  bar: MyPropTypes.bar.isRequired
}
```
with
```
  29:25  error  Typo in declared prop type: bar  react/no-typos
```
interestingly enough if i change to this:
```
Foo.propTypes = {
  bar: MyPropTypes.string.isRequired
}
```

then the rule passes, even though `MyPropTypes` doesn't export a `string` prop.

jseminck:
Just for information, indeed we added logic to validate the specific PropTypes. Since `string` is a key that is exported from the official `prop-types` package, it doesn't fail. `bar` is not a prop type defined in the official `prop-types` package, so it fails.

Not sure how we should handle this case yet. Just explaining what changes and why the error is introduced. 😄

ljharb:
This seems like a bug in no-typos - we should only be looking for typos on `React.PropTypes` or on an import/require of `prop-types`.

jseminck:
I'm trying to work on this... I guess we need to support the following cases? Did I miss anything

```js
import PropTypes from "prop-types"
import { PropTypes } from "react";
import { PropTypes as ValidPropTypes } from "react"
import React from "react" + React.PropTypes
```

ljharb:
@jseminck also `require('prop-types')` in any form.

martin-svk:
It also fail when importing specific prop types directly:

```
MyComponent.propTypes = {
    numberProp: number.isRequired,
    funcProp: func.isRequired
};
```

Resulting in `react/no-typos` Error.

oieduardorabelo:
same happens when using [`react-intl`](https://github.com/yahoo/react-intl) inject hoc:

```js
import { injectIntl, intlShape } from "react-intl";

function Example() { //code }

Example.propTypes = {
  // ESLint error: 'Typo in declared prop type: isRequired (react/no-typos)'
  intl: intlShape.isRequired
};

export default injectIntl(Example)
```

![screen shot 2017-10-10 at 9 56 00 am](https://user-images.githubusercontent.com/829902/31358227-3ea76112-ada1-11e7-9fae-b9357da61283.png)

[...]

PR #1652, opened 2018-01-20
support for custom propTypes with isRequired in no-typos

lfades:
fixes: #1607 and helps in #1389.

if will add support for the following cases:
```js
{
  a: string.isRequired,
  b: customProp.isRequired,
  // added support for isRequired in call expressions. No validation is done inside
  c: PropTypes.shape({
    d: string
  }).isRequired
  e: customShape({
    f: number
  }).isRequired
}
```

for a custom prop type like `string.isrequired` the error message will be: `Typo in declared prop type: isrequired` because currently it doesn't know that it's a prop type cualifier

for call expressions like `customShape({}).isrequired` the error message will be `Typo in prop type chain qualifier: isrequired` and it will work as long as it's not something like `customShape().number.isRequired`

### DIFF
--- a/tests/lib/rules/no-typos.js
+++ b/tests/lib/rules/no-typos.js
@@ -378,6 +378,16 @@ ruleTester.run('no-typos', rule, {
        }).isRequired
      }
    `,
+    parserOptions: parserOptions
+  }, {
+    code: `class Component extends React.Component {};
+     Component.propTypes = {
+       b: string.isRequired,
+       c: PropTypes.shape({
+         d: number.isRequired,
+       }).isRequired
+     }
+   `,
     parser: 'babel-eslint',
     parserOptions: parserOptions
   }],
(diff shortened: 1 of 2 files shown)


tests/lib/rules/no-typos.js
```diff
@@ -369,6 +369,17 @@ ruleTester.run('no-typos', rule, {
...
+  }, {
+    code: `class Component extends React.Component {};
+     Component.propTypes = {
+       b: string.isRequired,
+       c: PropTypes.shape({
+         d: number.isRequired,
+       }).isRequired
+     }
+   `,
+    parser: 'babel-eslint',
```
ljharb: please add another test like this that uses the default parser

tests/lib/rules/no-typos.js
```diff
@@ -795,5 +806,29 @@ ruleTester.run('no-typos', rule, {
     }, {
       message: 'Typo in declared prop type: objectof'
     }]
+  }, {
+    code: `class Component extends React.Component {};
+     Component.propTypes = {
+       b: string.isrequired
+     }
+   `,
+    parser: 'babel-eslint',
```
ljharb: same with these 2 tests; please duplicate them so there's 1 for babel-eslint and 1 for the default parser.

PR #1655, opened 2018-01-23
Add allowInPropTypes option to forbid-foreign-prop-types (#1647)

### LINKED ISSUE #1647: forbid-foreign-prop-types should allow another component's propTypes in a propTypes definition
Currently this results in a warning but shouldn't:

```
Button.propTypes = {
  ...Touchable.propTypes,
  color: PropTypes.string
}
```

The error message could also be improved to explain why using another component's propTypes is bad. For example: "Using propTypes from another component is not safe because they may be removed in production builds".

Comments via @gaearon on this issue: https://github.com/facebookincubator/create-react-app/pull/3818

iansu:
Add an option to allow using another components propTypes inside a propTypes declaration.

Closes #1647

### DIFF
--- a/lib/rules/forbid-foreign-prop-types.js
+++ b/lib/rules/forbid-foreign-prop-types.js
@@ -6,15 +6,6 @@
 
 const docsUrl = require('../util/docsUrl');
 
-// ------------------------------------------------------------------------------
-// Constants
-// ------------------------------------------------------------------------------
-
-
-// ------------------------------------------------------------------------------
-// Rule Definition
-// ------------------------------------------------------------------------------
-
 module.exports = {
   meta: {
     docs: {
@@ -22,10 +13,25 @@ module.exports = {
       category: 'Best Practices',
       recommended: false,
       url: docsUrl('forbid-foreign-prop-types')
-    }
+    },
+
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          allowInPropTypes: {
+            type: 'boolean'
+          }
+        },
+        additionalProperties: false
+      }
+    ]
   },
 
   create: function(context) {
+    const config = context.options[0] || {};
+    const allowInPropTypes = config.allowInPropTypes || false;
+
     // --------------------------------------------------------------------------
     // Helpers
     // --------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


lib/rules/forbid-foreign-prop-types.js
```diff
@@ -34,26 +48,58 @@ module.exports = {
       return node.parent.type === 'AssignmentExpression' && node.parent.left === node;
     }
 
+    function isInPropTypesAssignment(node) {
+      if (!allowInPropTypes) {
+        return false;
+      }
+
+      if (node.parent.parent.parent.parent.type === 'AssignmentExpression') {
```
ljharb: are we sure that the node will always have 4 parents?

lib/rules/forbid-foreign-prop-types.js
```diff
@@ -34,26 +48,58 @@ module.exports = {
...
+        const assignmentExpression = node.parent.parent.parent.parent;
```
ljharb: might as well create this var above the conditional, so it can be reused in there?

lib/rules/forbid-foreign-prop-types.js
ljharb: Also, rather than checking for AssignmentExpressions, is there a way we could create a visitor for AssignmentExpression instead?

lib/rules/forbid-foreign-prop-types.js
```diff
@@ -34,26 +48,58 @@ module.exports = {
...
+            node.property.type === 'Literal' &&
+            node.property.value === 'propTypes' &&
+            !isLeftSideOfAssignment(node) &&
+            !isInPropTypesAssignment(node))
+        ) {
           context.report({
             node: node.property,
-            message: 'Using another component\'s propTypes is forbidden'
+            message:
+              'Using propTypes from another component is not safe because they may be removed in production builds'
```
ljharb: this is a super weird line break; can this all go on one line? (line length limits aren't important)

lib/rules/forbid-foreign-prop-types.js
```diff
@@ -34,26 +48,58 @@ module.exports = {
...
           });
         }
       },
 
       ObjectPattern: function(node) {
-        const propTypesNode = node.properties.find(property => property.type === 'Property' && property.key.name === 'propTypes');
+        const propTypesNode = node.properties.find(
+          property => property.type === 'Property' && property.key.name === 'propTypes'
```
ljharb: similarly, please revert this back to one line

lib/rules/forbid-foreign-prop-types.js
```diff
@@ -34,26 +48,58 @@ module.exports = {
...
+        );
 
         if (propTypesNode) {
           context.report({
             node: propTypesNode,
-            message: 'Using another component\'s propTypes is forbidden'
+            message:
+              'Using propTypes from another component is not safe because they may be removed in production builds'
```
ljharb: also here, this should be one line.

lib/rules/forbid-foreign-prop-types.js
iansu: No, I'm not sure the node will always have 4 parents. I'm sure there's a better way to handle this but I don't know what it is. Is there a method I can use to search the tree for a parent that's an `AssignmentExpression`? 

Creating a visitor for `AssignmentExpression` might be better but I'm not sure how to coordinate between the two visitors. If we find an `AssignmentExpression` where `x.propTypes` is on the left and `allowInPropTypes` is true then we need to ignore the rest of that block of code but the `MemberExpression` visitor will still be called for expressions inside that block of code. I'm not sure how to solve that.

lib/rules/forbid-foreign-prop-types.js
iansu: This would have been done by Prettier. I'll revert it.

lib/rules/forbid-foreign-prop-types.js
ljharb: Good question.

Maybe we should make a shared utility function `isInAssignmentExpression` that walks upwards until it finds an `AssignmentExpression`?

iansu: On the one hand I'm glad I didn't miss something obvious but on the other hand I wish there was an easy solution to this problem. 😀 

I'll try adding a function that traverses the tree upwards until it reaches an `AssignmentExpression`. What are some other node types that should make it stop? I'm thinking `Program` and `FunctionExpression` at the very least. Anything else?

ljharb: Program, certainly. Not so sure about FunctionExpression.

iansu:
I've improved the traversal up the tree in search of a parent `AssignmentExpression`. I also changed some of the method names to (hopefully) make the code more readable.

lib/rules/forbid-foreign-prop-types.js
```diff
@@ -34,15 +48,56 @@ module.exports = {
...
+    function findParentAssignmentExpression(node) {
+      let parent = node.parent;
+
+      while (parent && parent.type !== 'Program') {
+        if (parent.type === 'AssignmentExpression') {
+          return parent;
+        } else if (parent.parent) {
+          parent = parent.parent;
+        } else {
+          return false;
```
ljharb: i think we can collapse the else if and the else together, and just do `parent = parent.parent`, and let the while condition catch it? (also, maybe `null` would make more sense than `false`)

iansu:
Any idea when this will be released? I want to get it into create-react-app.

ljharb:
v7.7.0 is released.

PR #1659, opened 2018-01-26
Fixed crossorigin naming convention and missing tag support.

### LINKED ISSUE #1656: no-unknown-property does not allow crossorigin on link tags
In the latest release, there was a [fix for the crossorigin property](https://github.com/yannickcr/eslint-plugin-react/pull/1643) (related to https://github.com/yannickcr/eslint-plugin-react/issues/1642). The updated rule only allows the `crossorigin` property on `script`, `img` and `video` tags. This means an error is now thrown for the `crossorigin` property on `link` tags.

You can see [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link) that `crossorigin` is an accepted attribute of `link` tags. I would therefore expect the linter to allow them. Something like this should pass:

```
<link href="path/to/file.ext" crossorigin="anonymous" />
```

Unless there's something I'm missing, happy to make a PR to fix this 🌈

jzDev:
Fixed crossorigin naming convention to match React's supported HTML attributes.
https://reactjs.org/docs/dom-elements.html

Fixes #1656. Per https://github.com/yannickcr/eslint-plugin-react/issues/1642#issuecomment-360881470

### DIFF
--- a/tests/lib/rules/no-unknown-property.js
+++ b/tests/lib/rules/no-unknown-property.js
@@ -82,6 +82,6 @@ ruleTester.run('no-unknown-property', rule, {
     errors: [{message: 'Unknown property \'clip-path\' found, use \'clipPath\' instead'}]
   }, {
     code: '<div crossOrigin />',
-    errors: [{message: 'Invalid property \'crossOrigin\' found on tag \'div\', but it is only allowed on: script, img, video'}]
+    errors: [{message: 'Invalid property \'crossOrigin\' found on tag \'div\', but it is only allowed on: script, img, video, link'}]
   }]
 });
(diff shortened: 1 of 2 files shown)


tests/lib/rules/no-unknown-property.js
```diff
@@ -80,14 +80,8 @@ ruleTester.run('no-unknown-property', rule, {
     code: '<rect clip-path="bar" />;',
     output: '<rect clipPath="bar" />;',
     errors: [{message: 'Unknown property \'clip-path\' found, use \'clipPath\' instead'}]
-  }, {
-    code: '<script crossOrigin />',
```
ljharb: let's replace this with `crossorigin`, so that it tests that this is an invalid property?

Issue #1663, opened 2018-01-28
react/prop-types doesn


realityking:

manjula91:
Hey anyone working on the issue so far?

ljharb:
I’m confused; does React even work with this?

At any rate, React themselves provides a codemod that can convert all your non-mixin-using createClasses to class-based components; is there a reason not to use that?

abenitesDC:
@ljharb I believe what you are pointing is a complete separate discussion from the actual problem reported. There are many teams that for various reasons are still using React without ES6 support. Unfortunately, for big projects, migrating code is not always as straight-forward as running a codemod.

And to answer your first question, this indeed works in React. (check: https://reactjs.org/docs/react-without-es6.html)

ljharb:
@abenitesDC fair. Those linked docs don't mention PropTypes, but I verified manually that indeed, it works in either form.

We should indeed modify the `prop-types` rule to detect static PropTypes with createClass components as well.

realityking:
Sorry for not getting back to this earlier. React indeed supports mixing these two.

We're unlikely to switch to ES6 classes until Node.js natively supports class fields. I'd however like to eliminate the amount of changes needed when we do, thus this usage.

PR #1671, opened 2018-02-01
Allow custom PropType classes with no-typos rule

### LINKED ISSUE #1389: no-typos rule doesn't handle custom proptypes classes

brettdh:
This is simply a rebase of @jseminck's branch, plus the removal of the
lingering failing test.

I removed the failing test because I believe it's a very small corner case
that should not block the resolution of this issue, which seems to affect
a lot of people and limits our ability to keep our components' proptypes DRY.

If it's determined that this corner case *should* be handled, I suggest creating
a new issue to track just that case.

Closes #1389.

### DIFF
--- a/lib/rules/no-typos.js
+++ b/lib/rules/no-typos.js
@@ -85,7 +85,6 @@ module.exports = {
         node.properties.forEach(prop => checkValidProp(prop.value));
       }
     }
-    /* eslint-enable no-use-before-define */
 
     function reportErrorIfClassPropertyCasingTypo(node, propertyName) {
       if (propertyName === 'propTypes' || propertyName === 'contextTypes' || propertyName === 'childContextTypes') {
--- a/lib/rules/no-typos.js
+++ b/lib/rules/no-typos.js
@@ -77,7 +77,9 @@ module.exports = {
       } else if (callee.type === 'MemberExpression' && callee.property.name === 'oneOfType') {
         const args = node.arguments[0];
         if (args && args.type === 'ArrayExpression') {
-          args.elements.forEach(el => checkValidProp(el));
+          args.elements.forEach(el => {
+            checkValidProp(el);
+          });
         }
       }
     }
--- a/lib/rules/no-typos.js
+++ b/lib/rules/no-typos.js
@@ -37,6 +37,7 @@ module.exports = {
 
   create: Components.detect((context, components, utils) => {
     let propTypesPackageName;
+    let reactPackageName;
 
     function checkValidPropTypeQualfier(node) {
       if (node.name !== 'isRequired') {
@@ -58,7 +59,7 @@ module.exports = {
 
     /* eslint-disable no-use-before-define */
     function checkValidProp(node) {
-      if (!propTypesPackageName || !node) {
+      if ((!propTypesPackageName && !reactPackageName) || !node) {
         return;
       }
 
@@ -132,12 +133,14 @@ module.exports = {
       ImportDeclaration: function(node) {
         if (node.source && node.source.value === 'prop-types') { // import PropType from "prop-types"
           propTypesPackageName = node.specifiers[0].local.name;
-        }
-      },
+        } else if (node.source && node.source.value === 'react') { // import { PropTypes } from "react"
+          reactPackageName = node.specifiers[0].local.name;
 
-      CallExpression: function(node) {
-        if (node.callee.name === 'require' && node.arguments[0].value === 'prop-types') {
-          propTypesPackageName = node.parent.id.name;
+          propTypesPackageName = node.specifiers.length > 1 && (
+            node.specifiers
+              .filter(specifier => specifier.imported && specifier.imported.name === 'PropTypes')
+              .map(specifier => specifier.local.name)
+          );
         }
       },
 
--- a/lib/rules/no-typos.js
+++ b/lib/rules/no-typos.js
@@ -36,6 +36,8 @@ module.exports = {
   },
 
   create: Components.detect((context, components, utils) => {
+    let propTypesPackageName;
+
     function checkValidPropTypeQualfier(node) {
       if (node.name !== 'isRequired') {
         context.report({
@@ -56,14 +58,28 @@ module.exports = {
 
     /* eslint-disable no-use-before-define */
     function checkValidProp(node) {
-      if (node && node.type === 'MemberExpression' && node.object.type === 'MemberExpression') {
+      if (!propTypesPackageName || !node) {
+        return;
+      }
+
+      if (
+        node.type === 'MemberExpression' &&
+        node.object.type === 'MemberExpression' &&
+        node.object.object.name === propTypesPackageName
+      ) { // PropTypes.myProp.isRequired
         checkValidPropType(node.object.property);
         checkValidPropTypeQualfier(node.property);
-      } else if (node && node.type === 'MemberExpression' && node.object.type === 'Identifier' && node.property.name !== 'isRequired') {
+      } else if (
+        node.type === 'MemberExpression' &&
+        node.object.type === 'Identifier' &&
+        node.object.name === propTypesPackageName &&
+        node.property.name !== 'isRequired'
+      ) { // PropTypes.myProp
         checkValidPropType(node.property);
-      } else if (node && (
-        node.type === 'MemberExpression' && node.object.type === 'CallExpression' || node.type === 'CallExpression'
-      )) {
+      } else if (
+        (node.type === 'MemberExpression' && node.object.type === 'CallExpression') ||
+        node.type === 'CallExpression'
+      ) { // Shapes
         if (node.type === 'MemberExpression') {
           checkValidPropTypeQualfier(node.property);
           node = node.object;
@@ -113,6 +129,18 @@ module.exports = {
     }
 
     return {
+      ImportDeclaration: function(node) {
+        if (node.source && node.source.value === 'prop-types') { // import PropType from "prop-types"
+          propTypesPackageName = node.specifiers[0].local.name;
+        }
+      },
+
+      CallExpression: function(node) {
+        if (node.callee.name === 'require' && node.arguments[0].value === 'prop-types') {
+          propTypesPackageName = node.parent.id.name;
+        }
+      },
+
       ClassProperty: function(node) {
         if (!node.static || !utils.isES6Component(node.parent.parent)) {
           return;
--- a/lib/rules/no-typos.js
+++ b/lib/rules/no-typos.js
@@ -36,8 +36,8 @@ module.exports = {
   },
 
   create: Components.detect((context, components, utils) => {
-    let propTypesPackageName;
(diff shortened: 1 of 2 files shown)


brettdh:
...well, I *thought* I'd rebased on master. Fixing.

brettdh:
Ok, rebased again, conflicts resolved, tests fixed up to handle changes on master. PTAL

ljharb:
In the future, please do not open up duplicate PRs; this permanently pollutes the git graph.

Posting a link to your branch on the original PR allows collaborators to bring it in to the original PR branch for you.

At this point, please leave this PR open; I'll force push to both of them so their git refs can match.

ljharb:
Thanks, will review soon.

brettdh:
Added them. They fail. :confused: Looking into it.

brettdh:
Ok, new tests (and fixes) pushed. Apologies if it's a touch sloppy; I just now got this to work and wanted to get it out before I unplug for the day. PTAL, thanks for the feedback!

brettdh:
@ljharb How crucial is that? There are no pre-existing `createClass` tests for the `no-typos` rule;  adding that coverage strikes me as out of scope here.

Also, there's an implicit test matrix in the `no-typos` tests, along at least these dimensions:
- Method of importing `React`/`PropTypes`
- Method of declaring component
- Method of declaring `propTypes` object
I wouldn't feel confident about `createClass` coverage without doing the refactoring to create parameterized tests on these dimensions, which also strikes me as out of scope here.

I created a couple quick-and-dirty `createClass` versions of other `no-typos` tests, and [they fail when run against `master`](https://gist.github.com/brettdh/bec038b45a8c2ab5bdf34defd45b48c1). So, I'm inclined to think that this is another instance of an issue like #1663, where `createClass` just isn't supported currently. (I didn't see any other issue related to `createClass` and `no-typos`, so I'm guessing it's not something a lot of people are encountering.)

ljharb:
I think it's crucial in that no createClass-related bugs have been reported for it, so even though there's no tests, any change to the rule might break it - and tests ensure that it's not broken.

DRY is generally a bad idea in tests; it'd be better to copy-paste all the needed tests.

**However** - if, in fact, this rule never worked on createClass components, which your tests seem to suggest, would you be willing to add commented-out createClass tests in this PR, and file a new PR after this one is merged that uncomments the tests? (From that point I can try to get the rule passing, if you don't have the time)

brettdh:
> DRY is generally a bad idea in tests; it'd be better to copy-paste all the needed tests.

Strongly disagree on "generally" and definitely in this particular case. The repetition makes it hard to know which are "all the needed tests" - it's hard to see which tests test which behavior, since there is clearly a cross product of behaviors under test. It would be better IMO if that was extracted and made explicit. Thankfully, we don't have to agree on that to make progress with this PR. :)

> However - if, in fact, this rule never worked on createClass components, which your tests seem to suggest, would you be willing to add commented-out createClass tests in this PR, and file a new PR after this one is merged that uncomments the tests? (From that point I can try to get the rule passing, if you don't have the time)

Sure, sounds good. I only have a couple such tests, but it'll be a start, anyway. Pushing shortly.

lib/rules/prefer-stateless-function.js
```diff
@@ -298,6 +302,20 @@ module.exports = {
...
+      // Report non-"render" methods that return JSX
+      MethodDefinition: function(node) {
+        if (
+          utils.isES6Component(node.parent.parent) &&
+          node.key.name !== 'render' &&
+          utils.isReturningJSX(node.value)
+        ) {
+          context.report({
+            node: node,
+            message: 'Method returning JSX should be written as stateless function component'
```
ljharb: This should probably be pluralized; `Methods returning JSX should be written as stateless functional components`

lib/rules/prefer-stateless-function.js
```diff
@@ -298,6 +302,20 @@ module.exports = {
         markThisAsUsed(node);
       },
 
+      // Report non-"render" methods that return JSX
+      MethodDefinition: function(node) {
```
ljharb: what about computed methods? like `['foo']() { return <div /> }` or `[Symbol()]() { return <div />; }`?

brettdh: TBH I'm not sure why this commit was in the original branch that I started from here. I'd be fine just removing that commit from this PR, but I'm not sure what would be correct in regards to your comment here.

ljharb: ah yes, good call. This PR is about the no-typos rule; but it has prefer-stateless-function changes. Let's remove all the parts that aren't about no-typos :-)

brettdh:
Removed `prefer-stateless-function` commits and rebased on master.

lib/rules/no-typos.js
```diff
@@ -54,38 +57,69 @@ module.exports = {
...
     /* eslint-disable no-use-before-define */
+
+    function checkValidCallExpression(node) {
+      const callee = node.callee;
+      if (callee.type === 'MemberExpression' && callee.property.name === 'shape') {
+        checkValidPropObject(node.arguments[0]);
+      } else if (callee.type === 'MemberExpression' && callee.property.name === 'oneOfType') {
+        const args = node.arguments[0];
+        if (args && args.type === 'ArrayExpression') {
+          args.elements.forEach(el => checkValidProp(el));
```
ljharb: since forEach ignores its return value, this should be an explicit-return arrow function instead.

lib/rules/no-typos.js
```diff
@@ -54,38 +57,69 @@ module.exports = {
       }
     }
 
+    function isPropTypesPackage(node) {
+      // Note: we really do want == with the package names here,
+      // since we need value equality, not identity - and
```
ljharb: i'm confused about this - every string that has the same contents is the same string by identity, iow `'abc' === 'abc'`. Why do we need `==` here?

brettdh: I thought that too, but when I was testing, it seemed that only every string *literal* with the same contents is the same string by identity. I found that was not the case when comparing a literal to a value read from a file.

But maybe there was another issue that I didn't fully understand; I made [this jsfiddle](https://jsfiddle.net/3yu2op32/5/) to test this, and the results indicate no difference between `===` and `==`.

I can dig into this more tomorrow, probably.

ljharb: Perhaps there's invisible characters in the strings? When you run into the difference, try comparing `.length` (altho then `==` would fail too). All strings in JS that have the same contents are `===`.

Specifically, `==` is "`===` after type coercion", so if they're both `typeof "string"`, `==` and `===` are identical, by spec.

brettdh: WOW ok silly mistake. `propTypesPackageName` was an array containing one string, which is indeed `==` to that string value itself, but not `===`.

Thanks for not letting that slide. 😳 Pushing fix shortly.

lib/rules/no-typos.js
brettdh: You're saying you'd prefer `args.elements.forEach(el => { checkValidProp(el); });`? It seems strictly noisier to me, but I don't feel strongly ¯\\\_(ツ)_/¯

brettdh:
Thanks for your feedback. However, I'm concerned that we're getting solidly into scope creep territory again.

> Can you add test cases for require as well? we should be able to handle top-level static requires along with import.

There are no such tests currently; it's not part of this PR to add that coverage. Can we make a separate issue for that?

> Separately, can we make sure that every test using babel-eslint has a duplicate test that uses the default parser?

It would be similarly out of scope to ask this for every `no-typos` test. I assume you mean every new test that this PR adds? Also, this would be a perfect use case for parameterization; it seems quite noisy to copy the entire tests' content just to remove the `parser` line, and glancing through the tests as they are now, it's quite hard to pick out which tests use which parser. But I feel hesitant to start going down that road, due to the opposition to it you expressed above, and the uncertainty I've expressed about getting it right.

Also, are all tests valid with both parsers? I don't really understand the difference between babel-eslint and the "default parser", so I'm not sure I'm the one to make that change anyway, even if it were in scope here.

lib/rules/no-typos.js
ljharb: yep, exactly; implicit return should only be used where the return value matters.

ljharb:
> I assume you mean every new test that this PR adds?

Yes, certainly only the added tests :-) if the others are missing tests for both the default parser and babel-eslint, that's something we'll have to fix separately.

"Noisy" isn't really a concern in tests; copy-pasting in tests is also fine. The value in testing with both parsers is that each parser might generate different node names, making the rule behave differently (even though it's unlikely). Including both tests prevents any regressions on that in the future as we update eslint and/or babel-eslint.

brettdh:
Added copies of this PR's new tests with the default parser.

brettdh:
Anything else needed on this one? I'd love to get it merged soon.

lib/rules/no-typos.js
```diff
@@ -114,6 +145,20 @@ module.exports = {
...
+        if (node.source && node.source.value === 'prop-types') { // import PropType from "prop-types"
+          propTypesPackageName = node.specifiers[0].local.name;
+        } else if (node.source && node.source.value === 'react') { // import { PropTypes } from "react"
+          reactPackageName = node.specifiers[0].local.name;
+
+          propTypesPackageName = node.specifiers.length >= 1 ? (
+            node.specifiers
+              .filter(specifier => specifier.imported && specifier.imported.name === 'PropTypes')
+              .map(specifier => specifier.local.name)
+          )[0] : null;
```
ljharb: should this use `.find` instead of `.filter`?

brettdh: Sure; it makes this code clearer. Updated.

jseminck:
Thanks for dealing with this @brettdh - I also hope to see this merged and released soon 👍

lucaslim:
Any idea when is this going to be released? Or is that a temporary workaround for it until then?

brettdh:
@lucaslim looks like there's an RC now: https://github.com/yannickcr/eslint-plugin-react/releases/tag/v7.8.0-rc.0

yannickcr:
@lucaslim As @brettdh noticed, I published a pre-release a few hours ago.
You can test it using `npm i eslint-plugin-react@next -D`.
If no major issues are found the final release will be most likely published tomorrow or the day after.

lucaslim:
@yannickcr That's awesome! Thanks :)

PR #1675, opened 2018-02-03
Fixed "PropType is defined but prop is never used  react/no-unused-prop-types"

yuri-sakharov:
We got `PropType is defined but prop is never used  react/no-unused-prop-types` when used **createReactClass**.
Added fix, tests and removed 2 duplicated tests.

### DIFF
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -266,7 +266,7 @@ module.exports = {
      * @return {Boolean} True if the node is inside a lifecycle method
      */
     function isInLifeCycleMethod(node) {
-      if (node.type === 'MethodDefinition' && isNodeALifeCycleMethod(node)) {
+      if ((node.type === 'MethodDefinition' || node.type === 'Property') && isNodeALifeCycleMethod(node)) {
         return true;
       }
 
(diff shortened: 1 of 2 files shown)


yuri-sakharov:
Yes, of course.

yuri-sakharov:
Added requested tests.

yuri-sakharov:
@ljharb what about this case:
```class Hello extends Component {
          static propTypes = {
            something: PropTypes.bool
          }
          componentWillUpdate (nextProps, nextState) {
            const {something} = nextProps;
          }
        }```
does it should throw error or not?

**Comment by ljharb** [comment]
That syntax requires babel-eslint; but we should have a test with the default parser that covers `Hello.propTypes = { … }` instead.

**Comment by yuri-sakharov** [comment]
@ljharb  is it the same 
```
code: [
        'var Hello = createReactClass({',
        '  propTypes: {',
        '    something: PropTypes.bool,',
        '  },',
        '  componentWillReceiveProps (nextProps) {',
        '    const {something} = nextProps;',
        '    doSomething(something);',
        '  }',
        '})'
      ].join('\n'),
      parser: 'babel-eslint'
```
and 
```
code: [
        'var Hello = createReactClass({',
        '  propTypes: {',
        '    something: PropTypes.bool,',
        '  },',
        '  componentWillReceiveProps (nextProps) {',
        '    const {something} = nextProps;',
        '    doSomething(something);',
        '  }',
        '})'
      ].join('\n')
```
?
will all use parser: 'babel-eslint'?

ljharb:
yep, those all look the same :-)

PR #1681, opened 2018-02-07
Update no-unused-prop-types rule for new React class component lifecycles

bvaughn:
The upcoming React 16.3 release will make a couple of changes to the class component lifecycles: `componentWillMount`, `componentWillReceiveProps`, and `componentWillUpdate` are being aliased to `UNSAFE_componentWillMount`, `UNSAFE_componentWillReceiveProps`, and `UNSAFE_componentWillUpdate` and a new static lifecycle, `getDerivedStateFromProps`, is being added.

For background information, see [RFC 6](https://github.com/reactjs/rfcs/blob/master/text/0006-static-lifecycle-methods.md).

When running codemods at Facebook to rename these lifecycles, I noticed that it caused a couple of false positives with the `react/no-unused-prop-types` lint rule since that rule was not yet aware of the lifecycle changes. I _think_ this PR corrects that. I added a couple of sanity check tests (which failed before my changes) to verify this.

### DIFF
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -22,15 +22,8 @@ const docsUrl = require('../util/docsUrl');
 const DIRECT_PROPS_REGEX = /^props\s*(\.|\[)/;
 const DIRECT_NEXT_PROPS_REGEX = /^nextProps\s*(\.|\[)/;
 const DIRECT_PREV_PROPS_REGEX = /^prevProps\s*(\.|\[)/;
-const LIFE_CYCLE_METHODS = [
-  'componentDidUpdate',
-  'componentWillReceiveProps',
-  'componentWillUpdate',
-  'getDerivedStateFromProps',
-  'shouldComponentUpdate',
-  'UNSAFE_componentWillReceiveProps',
-  'UNSAFE_componentWillUpdate'
-];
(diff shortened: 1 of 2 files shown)


lib/rules/no-unused-prop-types.js
```diff
@@ -22,7 +22,15 @@ const docsUrl = require('../util/docsUrl');
 const DIRECT_PROPS_REGEX = /^props\s*(\.|\[)/;
 const DIRECT_NEXT_PROPS_REGEX = /^nextProps\s*(\.|\[)/;
 const DIRECT_PREV_PROPS_REGEX = /^prevProps\s*(\.|\[)/;
-const LIFE_CYCLE_METHODS = ['componentWillReceiveProps', 'shouldComponentUpdate', 'componentWillUpdate', 'componentDidUpdate'];
+const LIFE_CYCLE_METHODS = [
+  'componentDidUpdate',
+  'componentWillReceiveProps',
+  'componentWillUpdate',
+  'getDerivedStateFromProps',
```
ljharb: these new methods should only be checked if the React version pragma from settings is >= 16.3.

bvaughn: Ah, yes. Thanks for pointing that out.

dhansiddh:
Can we get this PR merged ?

rgabs:
I checked the PR and it works. Could you please merge this PR?

PR #1688, opened 2018-02-10
Adding the option to provide a custom message to jsx-no-bind

03k64:
Changes
- added the ability to provide a custom message to the `jsx-no-bind` rule
- added tests around the rule to ensure the custom message _always_ overrides the default
- added documentation illustrating the usage of the rule

Context
- in my current role we are trying to educate developers about best practice and performance concerns
- custom messages on linter rules allow us to provide information about _why_ they should follow these practices (or conversely, why they should avoid patterns detected by linter rules) as an expansion on what the default messages provide (ie. the pattern they should avoid)

If it will help this get merged I can amend the request to enable custom messages to be set for each of the four configuration options. Let me know if you would rather I pursued this option.

### DIFF
--- a/docs/rules/jsx-no-bind.md
+++ b/docs/rules/jsx-no-bind.md
@@ -9,11 +9,13 @@ The following patterns are considered warnings:
 ```jsx
 <div onClick={this._handleClick.bind(this)}></div>
 ```
+
 ```jsx
 <div onClick={() => console.log('Hello!')}></div>
 ```
 
 The following patterns are **not** considered warnings:
+
 ```jsx
 <div onClick={this._handleClick}></div>
 ```
@@ -26,6 +28,7 @@ The following patterns are **not** considered warnings:
   "allowArrowFunctions": <boolean> || false,
   "allowFunctions": <boolean> || false,
   "allowBind": <boolean> || false
+  "customMessage": <string> || undefined
 }]
 ```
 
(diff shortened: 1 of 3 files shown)


ljharb:
Is there a reason you think that the default shouldn't just be improved?

darArch:
The default could certainly be improved to address _why_ `jsx-no-bind` is a pattern best avoided in addition to advising against its use.

I still think a custom message is useful, it allows the message to be customised in order to formatted in a manner best suited for the tooling being used. For example, a lot of our developers use a combination of Visual Studio Code and plugins which display a pop-up message containing the linter warning when hovering over a linter violation in code.

ljharb:
In general I agree that custom messages are useful; however for this case, I'm still not sure what custom message anyone could provide that wouldn't be better as the unconfigurable default.

darArch:
I agree that expanding the default message here will likely satisfy 99% of users insofar as it would provide context as to why the lint rule in question is necessary. To be clear, I absolutely agree that, provided the default messaging is expanded, the content of the messaging does not need to be tweaked per user.

I do however think the custom message facility allows the _format_ of the message to be tweaked to best utilise the chosen development environment. For instance, the Visual Studio Code plugin providing lint rule tooltips I mentioned has a maximum width of 64 characters. For non-users of Visual Studio Code, inserting line breaks/wrapping the message at this length does not necessarily make sense.

docs/rules/jsx-no-bind.md
```diff
@@ -62,6 +65,26 @@ When `true` the following is **not** considered a warning:
...
 ```
 
+### `customMessage`
+
+When `customMessage` is set, it will override the default error message for each rule violation,
+for example, with the following config:
+
+```js
+"react/jsx-no-bind": [true, {
+  "customMessage": "A bind call in a JSX prop will cause performance issues related to garbage collection"
```
ljharb: Let's just call this "message", but let's also improve the default such that as few people as possible would need/want to customize it.

docs/rules/jsx-no-bind.md
```diff
@@ -62,6 +65,26 @@ When `true` the following is **not** considered a warning:
...
+}]
+```
+
+And the below code:
+
+```jsx
+<div onCilck={this._handleClick.bind(this)} />
```
ljharb: onCilck → onClick

ljharb:
Unfortunately, the fork appears to be deleted, so this can't land.

Issue #1686, opened 2018-02-09
jsx-handler-names: false positive when handler name begins with number


jsphstls:
![image](https://user-images.githubusercontent.com/4481784/36049520-4164c64c-0da0-11e8-8205-e7f459aff9b9.png)

Fail:
```javascript
onClick={this.handle123LogoClick}
```

Pass
```javascript
onClick={this.handleLogo123Click}
```

123Logo is a fictional example.

jsphstls:
Pull request: https://github.com/yannickcr/eslint-plugin-react/pull/1689

Moong0122:
May I try to solve this bug?

ljharb:
@Moong0122 any issue with a "help wanted" label that doesn't have a linked PR is up for grabs. This one has #1689, though (which I'll take another look at) so i don't think any additional help is needed on this issue.

PR #1689, opened 2018-02-10
Issue #1686 - jsx-handler-names: false positive when handler name begins with number

jsphstls:
Updated unit test and EVENT_HANDLER_REGEX regex.

### DIFF
--- a/tests/lib/rules/jsx-handler-names.js
+++ b/tests/lib/rules/jsx-handler-names.js
@@ -31,6 +31,9 @@ ruleTester.run('jsx-handler-names', rule, {
   valid: [{
     code: '<TestComponent onChange={this.handleChange} />'
   }, {
+    // TODO: make this an invalid test
+    code: '<TestComponent onChange={this.handle123Change} />'
+  }, {
     code: '<TestComponent onChange={this.props.onChange} />'
   },
   {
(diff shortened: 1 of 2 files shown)


jsphstls:
I adjusted the regex so that a capital letter is still required at some point after the word "handle" and added more test cases to demonstrate that minimum requirement.

tests/lib/rules/jsx-handler-names.js
```diff
@@ -69,6 +73,18 @@ ruleTester.run('jsx-handler-names', rule, {
...
+    errors: [{message: 'Handler function for onChange prop key must begin with \'handle\''}]
+  }, {
+    code: '<TestComponent onChange={this.handle2} />',
+    errors: [{message: 'Handler function for onChange prop key must begin with \'handle\''}]
+  }, {
+    code: '<TestComponent onChange={this.handl3Change} />',
+    errors: [{message: 'Handler function for onChange prop key must begin with \'handle\''}]
+  }, {
+    code: '<TestComponent onChange={this.handle4change} />',
+    errors: [{message: 'Handler function for onChange prop key must begin with \'handle\''}]
```
ljharb: hm, we should probably change the error message to add " and be in camelCase", otherwise this one's pretty confusing

jsphstls: I updated the error message to be more specific and removed the handleC test case that does pass, but should not be considered valid.

ljharb: if the test wrongly passes, we should move it to the "invalid" section, even if it's commented out for now.

PR #1690, opened 2018-02-12
Rule prop-types includes nextProps checking in shouldComponentUpdate

### LINKED ISSUE #1684: Rule prop-types should include nextProps checking in shouldComponentUpdate
Consider this example:
``` javascript
class Hello extends Component {
  static propTypes = {
    bar: PropTypes.func
  }
    return nextProps.foo !== null
  }
  render() {
    return <div bar={this.props.bar} />;
  }
}
```

This is a case where we want to check to make sure that `foo` is properly defined in `propTypes`. I think this case should be included in the `prop-type` rule. Any thoughts?

Please refer to my comment here: https://github.com/yannickcr/eslint-plugin-react/issues/1636#issuecomment-364195848

I've made the changes in a forked repo and seems to work just fine. If we wanted to get this in, I'd need some help with the test cases but here is my fork: https://github.com/mindspacepdx/eslint-plugin-react/tree/am-should-component-update

amerryma:
Resolves #1684 

I've added the `shouldComponentUpdate` to the list of functions to include with this rule, following how `componentWillReceiveProps` was implemented.

The only thing outstanding are the tests, I've only included a single test but it may need to be put in a different spot.

### DIFF
--- a/lib/rules/prop-types.js
+++ b/lib/rules/prop-types.js
@@ -122,6 +122,24 @@ module.exports = {
     }
 
     /**
+     * Check if we are in a class constructor
+     * @return {boolean} true if we are in a class constructor, false if not
+     */
+    function inShouldComponentUpdate() {
+      let scope = context.getScope();
+      while (scope) {
+        if (
+          scope.block && scope.block.parent &&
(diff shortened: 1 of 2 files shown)


amerryma:
@ljharb Thanks for your review! I've made the changes requested.

PR #1699, opened 2018-02-20
[Fix] `react/jsx-no-bind` Add props to example

ronanmathew:
https://reactjs.org/blog/2015/01/27/react-v0.13.0-beta-1.html#autobinding

### DIFF
--- a/docs/rules/jsx-no-bind.md
+++ b/docs/rules/jsx-no-bind.md
@@ -121,8 +121,8 @@ Unfortunately [React ES6 classes](https://facebook.github.io/react/blog/2015/01/
 
 ```jsx
 class Foo extends React.Component {
-  constructor() {
-    super();
+  constructor(...args) {
+    super(...args);
     this._onClick = this._onClick.bind(this);
   }
   render() {


docs/rules/jsx-no-bind.md
```diff
@@ -121,8 +121,8 @@ Unfortunately [React ES6 classes](https://facebook.github.io/react/blog/2015/01/
 
 ```jsx
-  constructor() {
-    super();
+  constructor(props) {
```
ljharb: This isn't actually needed, and to be fully correct, it'd be `props, context`. It's a fine change tho.

ronanmathew: While not strictly necessary it allows use of `this.props` in the constructor and is suggested best practice by the React docs:

"Class components should always call the base constructor with props."
- https://reactjs.org/docs/state-and-lifecycle.html#adding-local-state-to-a-class

ljharb: In that case, what about `constructor(...args) { super(...args);` ? that way people copying the example will always forward all args.

ronanmathew: Comments addressed @ljharb. Anything else or are you happy to merge?

PR #1701, opened 2018-02-23
Handle isRequired for boolean-prop-naming

louisscruz:
# The Problem

There was a bug in boolean-prop-naming where the following would trip the rule:

```js
class Something extends React.Component {
  static propTypes = {
    something: PropTypes.bool
  };
  // ...
}
```

but the following would not:

```js
class Something extends React.Component {
  static propTypes = {
    something: PropTypes.bool.isRequired
  };
  // ...
}
```

# The Solution

I added an extra conditional to `getPropKey` to handle this scenario.

### DIFF
--- a/lib/rules/boolean-prop-naming.js
+++ b/lib/rules/boolean-prop-naming.js
@@ -62,7 +62,8 @@ module.exports = {
      * propTypes: {
      *   full: React.PropTypes.bool,
      *   short: PropTypes.bool,
-     *   direct: bool
+     *   direct: bool,
+     *   required: PropTypes.bool.isRequired
      * }
      * @param {Object} node The node we're getting the name of
      */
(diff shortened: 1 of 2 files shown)


lib/rules/boolean-prop-naming.js
```diff
@@ -73,6 +74,10 @@ module.exports = {
         return null;
       }
       if (node.value.property) {
+        const name = node.value.property.name;
+        if (name === 'isRequired') {
+          return node.value.object.property.name;
+        }
         return node.value.property.name;
```
ljharb: this can be `return name;` now?

louisscruz:
@ljharb Absolutely. On it.

louisscruz:
@ljharb Actually, just gave it a shot, and it seems that any test that uses class properties needs that parser. If I remove it, I get `A fatal parsing error occurred: Parsing error: Unexpected token =` (even for other tests in this file).

rapheld:
Would love to see this get in

ljharb:
@louisscruz indeed; the code examples using class properties would need to be tweaked so as not to use them.

louisscruz:
@ljharb Oh, I see what you're getting at. Will add those.

louisscruz:
@yannickcr Any chance you could publish this so I can get access to this bug fix by default?

PR #1702, opened 2018-02-24
[Fix] `boolean-prop-naming`: add missing default config

### LINKED ISSUE #1540: boolean-prop-naming does not check Flow props
The `boolean-prop-naming` does not complain about boolean props defined in Flow which do not match the rule pattern.

### LINKED ISSUE #1551: boolean-prop-naming is missing default config
It seems that the `boolean-prop-naming` rule does not enforce any default config (especially for `rule` as the docs mention [here](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/boolean-prop-naming.md#rule))

Sample code:

```js

class ReactClassComponent extends React.Component {
	static propTypes = {
		error: PropTypes.bool,
		hasError: PropTypes.bool,
		isOptional: PropTypes.bool,
		optional: PropTypes.bool,
	};

	render() {
		return null;
	}
}
```

Sample eslint configuration:

```js
	plugins: ['react'],
	rules: {
		'react/boolean-prop-naming': ['error'],
	},
```

Running `eslint .` gives no errors.

However switching the rule config to `['error', { rule: '^(is|has)[A-Z]([A-Za-z0-9]?)+' }]` will throw linting error for the sample code above.

ljharb:
This might be semver-major, since the rule has never had any default behavior previously.

Fixes #1551. Probably fixes #1540.

### DIFF
--- a/lib/rules/boolean-prop-naming.js
+++ b/lib/rules/boolean-prop-naming.js
@@ -13,6 +13,8 @@ const docsUrl = require('../util/docsUrl');
 // Rule Definition
 // ------------------------------------------------------------------------------
 
+const ruleDefault = '^(is|has)[A-Z]([A-Za-z0-9]?)+';
+
 module.exports = {
   meta: {
     docs: {
@@ -34,7 +36,7 @@ module.exports = {
           uniqueItems: true
         },
         rule: {
-          default: '^(is|has)[A-Z]([A-Za-z0-9]?)+',
+          default: ruleDefault,
           minLength: 1,
           type: 'string'
         },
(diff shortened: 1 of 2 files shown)


EvHaus (approved):
LGTM. I agree that semver should be considered as we don't want users to complain about the unexpected change.

ljharb:
In that case we can leave this unmerged for awhile until we’re ready for a breaking change.

louisscruz (approved):

lencioni (approved):

yannickcr (approved):

PR #1720, opened 2018-03-12
Fix failures of jsx-max-depth on empty nodes

j-martyn:
On some my code (didn't check what code exactly) I got this:

```
Cannot read property 'type' of null
TypeError: Cannot read property 'type' of null
    at find (/path/to/package/node_modules/eslint-plugin-react/lib/rules/jsx-max-depth.js:93:27)
    at findJSXElement (/path/to/package/node_modules/eslint-plugin-react/lib/rules/jsx-max-depth.js:102:49)
    at JSXExpressionContainer (/path/to/package/node_modules/eslint-plugin-react/lib/rules/jsx-max-depth.js:137:25)
```

So, for now, I just have fixed it. No tests have been changed as I didn't investigate it.

### DIFF
--- a/lib/rules/jsx-max-depth.js
+++ b/lib/rules/jsx-max-depth.js
@@ -38,7 +38,7 @@ module.exports = {
     const maxDepth = has(option, 'max') ? option.max : DEFAULT_DEPTH;
 
     function isJSXElement(node) {
-      return node.type === 'JSXElement';
+      return node && node.type === 'JSXElement';
     }
 
     function isExpression(node) {
@@ -90,7 +90,8 @@ module.exports = {
 
             return isJSXElement(writeExpr)
               && writeExpr
-              || writeExpr.type === 'Identifier'
+              || writeExpr 
+              && writeExpr.type === 'Identifier'
               && findJSXElement(variables, writeExpr.name);
           }
         }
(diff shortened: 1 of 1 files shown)


lib/rules/jsx-max-depth.js
```diff
@@ -90,7 +90,8 @@ module.exports = {
 
             return isJSXElement(writeExpr)
               && writeExpr
-              || writeExpr.type === 'Identifier'
+              || writeExpr
```
ljharb: probably this whole chain should be rewritten for clarity; at the least, let's use parens so as to not mix && and || :-)

ljharb (changes requested):
Thanks for the fix!

It's pretty important to find the code this broke on, so that we can add a regression test.

ferhatelmas:
@j-martyn @ljharb Any ETA ? I see this crash too.

FYI, it happens on files with [jsx-control-statements](https://github.com/AlexGilleran/jsx-control-statements).

ljharb:
@ferhatelmas if you could provide a failing test case (on a branch in your fork), and link it here, i'll add it to the PR and we can get this in.

ljharb:
cc @alexzherdev if you're interested in taking a look at this one, i'm happy to add any commits you like to this PR branch :-)

alexzherdev:
I'll see if I can reproduce this 👍

ljharb:
ping @alexzherdev, any progress?

Issue #1722, opened 2018-03-13
'react/jsx-no-bind' doesn't recognise ternary conditions


gwenaellarmet:
```jsx
render () {
  return (
     <MyComponent
       onValidate={this.isAdmin ? this.onValidateAdmin.bind(this) : this.onValidate.bind(this)}
       onRequestClose={this.onRequestClose.bind(this)}
       {...props}
    >
  )
}
```

In the code above, the rules 'react/jsx-no-bind' is marking onRequestClose as a warning but not onValidate, although it should in my opinion.

I'll open a PR fixing this bug.

gwenaellarmet:
PR #1723 fix this bug, need review and merge if accepted

gwenaellarmet:
Bug was fixed :+1:

PR #1723, opened 2018-03-13
handle ternary operation for jsx-no-bind

gwenaellarmet:
fix for issue #1722

### DIFF
--- a/lib/rules/jsx-no-bind.js
+++ b/lib/rules/jsx-no-bind.js
@@ -82,6 +82,10 @@ module.exports = {
       ) {
         return 'bindCall';
       } else if (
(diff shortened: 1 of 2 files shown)


lib/rules/jsx-no-bind.js
```diff
@@ -81,6 +81,12 @@ module.exports = {
         node.callee.property.name === 'bind'
       ) {
         return 'bindCall';
+      } else if (
+        nodeType === 'ConditionalExpression'
```
ljharb: I think that this if condition can fit on one line :-)

gwenaellarmet: I try to match the code style of the function, I can change it if needed

PR #1729, opened 2018-03-19
Add JSX prop alignment based on the first prop & between-prop space consistency

### LINKED ISSUE #398: jsx-indent-props for aligned indentation
It would be nice if this rule would allow to set an option to lint all the props declarations one below the other just like the next snippet shows.

``` jsx
<div styleName='content'>
  <TableOfContents content={props.model.tableOfContents} />
    <ContentView content={props.model.contents}
                 editing={props.model.editing}
                 actions={props.actions}
                 bind={this.props.bind} />
</div>
```

ThiefMaster:
This adds support for the JSX prop style suggested in #398 which is the default of the various JetBrains IDEs.
I haven't updated the docs yet since writing docs isn't fun so I'd rather know first that the PR has a chance of getting accepted before doing that.

Fixes #398.

### DIFF
--- a/lib/rules/jsx-indent-props.js
+++ b/lib/rules/jsx-indent-props.js
@@ -47,7 +47,7 @@ module.exports = {
 
     schema: [{
       oneOf: [{
-        enum: ['tab', 'aligned']
+        enum: ['tab', 'first']
       }, {
         type: 'integer'
       }]
@@ -64,8 +64,8 @@ module.exports = {
     const sourceCode = context.getSourceCode();
 
     if (context.options.length) {
-      if (context.options[0] === 'aligned') {
-        indentSize = 'aligned';
+      if (context.options[0] === 'first') {
+        indentSize = 'first';
         indentType = 'space';
       } else if (context.options[0] === 'tab') {
         indentSize = 1;
@@ -167,24 +167,9 @@ module.exports = {
           return;
         }
         let propIndent;
-        if (indentSize === 'aligned') {
+        if (indentSize === 'first') {
           const firstPropNode = node.attributes[0];
-          // '<' and any space following it
-          const beforeNameLength = node.name.range[0] - node.range[0];
-          // the name itself
-          const nameLength = node.name.range[1] - node.name.range[0];
-          // any space between the name and the first attribute
-          const afterNameLength = Math.max(1, firstPropNode.loc.start.column - node.name.loc.end.column);
-          if (afterNameLength !== 1) {
-            // this is not easily fixable, so we just report another error for it.
-            // it should be unlikely enough to be triggered anyway
-            context.report({
-              node: node,
-              message: 'Found too much whitespace between tag name and first attribute.'
-            });
-          }
-          // use the indentation of the element relative to the beginning of the line
-          propIndent = node.loc.start.column + beforeNameLength + nameLength + afterNameLength;
+          propIndent = firstPropNode.loc.start.column;
         } else {
           const elementIndent = getNodeIndent(node);
           propIndent = elementIndent + indentSize;
--- a/lib/rules/jsx-indent-props.js
+++ b/lib/rules/jsx-indent-props.js
@@ -47,7 +47,7 @@ module.exports = {
 
     schema: [{
       oneOf: [{
-        enum: ['tab']
+        enum: ['tab', 'aligned']
       }, {
         type: 'integer'
       }]
@@ -64,7 +64,10 @@ module.exports = {
     const sourceCode = context.getSourceCode();
 
     if (context.options.length) {
-      if (context.options[0] === 'tab') {
+      if (context.options[0] === 'aligned') {
+        indentSize = 'aligned';
+        indentType = 'space';
+      } else if (context.options[0] === 'tab') {
         indentSize = 1;
         indentType = 'tab';
       } else if (typeof context.options[0] === 'number') {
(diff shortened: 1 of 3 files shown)


lib/rules/jsx-indent-props.js
```diff
@@ -160,8 +163,33 @@ module.exports = {
...
+        if (indentSize === 'aligned') {
+          const firstPropNode = node.attributes[0];
+          // '<' and any space following it
+          const beforeNameLength = node.name.range[0] - node.range[0];
+          // the name itself
+          const nameLength = node.name.range[1] - node.name.range[0];
+          // any space between the name and the first attribute
+          const afterNameLength = Math.max(1, firstPropNode.loc.start.column - node.name.loc.end.column);
+          if (afterNameLength !== 1) {
+            // this is not easily fixable, so we just report another error for it.
```
ThiefMaster: I went for the lazy option here - the default fixing logic in `report()` cannot handle the case where the property isn't the first thing in the line and `checkNodesIndent` would need further changes as well.

If you think having an automated fixer for this is important I can have another look though.

lib/rules/jsx-indent-props.js
```diff
@@ -47,7 +47,7 @@ module.exports = {
 
     schema: [{
       oneOf: [{
-        enum: ['tab']
+        enum: ['tab', 'aligned']
```
ljharb: aligned with what? opening tag, closing tag, first used prop name, tag name, etc?

lib/rules/jsx-indent-props.js
ljharb: Yes, I think it's very important to have an automated fixer for as many things as possible.

lib/rules/jsx-indent-props.js
ThiefMaster: Any better suggestion? Naming is hard :p

My first idea was `tag-aligned` but `jsx-closing-bracket-location` usess that term with a different meaning, and just `tag` is too similar to `tab` IMO.

lib/rules/jsx-indent-props.js
ThiefMaster: OK I'll have another look at this

lib/rules/jsx-indent-props.js
ljharb: Seems like it's aligned with the first prop?

What happens with:
```jsx
<C a b c
   d
     e
       f
/>
```
Which is properly aligned, d, e, or f?

ThiefMaster: yep, `d`. So yeah, `first-prop` sounds actually a good name.

Just thinking.. should fixing double-space between tag and prop (the stuff from the other comment below) even be part of this rule or a completely new rule? In that case I'd look into creating such a rule and just remove the `afterNameLength !== 1` error here.

ljharb: I think that this rule should only cover prop indentation. I'm not sure what your specific question is tho, could you provide a code example?

ThiefMaster: `<Foo  bar>` --> `<Foo bar>`

So this should then be handled by a separate rule (or is there already one which does that?)

klimashkin:
Awesome!
Similar to eslint indent rule
```javascript
indent: [2, 2, {
  'FunctionDeclaration': {
    'parameters': 'first',
    'body': 1,
  },
}]
```

we can call it first
`'react/jsx-indent-props': [2, 'first']`

ThiefMaster:
PR updated. Also removed a bunch of unused code from the rule in a separate commit (probably copy&paste from somewhere else).

lib/rules/jsx-indent-props.js
ljharb: Yes, that's not prop indentation - indentation is only "start of line + whitespace + thing"

ThiefMaster:
PR updated with the rule to deal with excessive spacing between tag/prop or props in the same line

ThiefMaster:
Anything missing to get this merged? Would be nice if we could start using it (by installing the plugin from this git repo until it's part of a release, without having to maintain our own branch containing this and my other PR).

lib/rules/jsx-indent-props.js
```diff
@@ -78,62 +81,41 @@ module.exports = {
      * @param {ASTNode} node Node violating the indent rule
      * @param {Number} needed Expected indentation character count
      * @param {Number} gotten Indentation character count in the actual node/code
-     * @param {Object=} loc Error line and column location
      */
-    function report(node, needed, gotten, loc) {
```
ljharb: is the `loc` no longer required because eslint gets it off the node?

docs/rules/jsx-props-no-multi-spaces.md
```diff
@@ -0,0 +1,29 @@
+# Disallow multiple spaces between inline JSX props
```
ljharb: i'm not sure what this rule is, but please remove it from this PR :-)

ThiefMaster: It's the one I mentioned here: https://github.com/yannickcr/eslint-plugin-react/pull/1729#discussion_r175829190

I can move it to a separate PR.

lib/rules/jsx-indent-props.js
ThiefMaster: It was never provided when calling the function - probably just a copy&paste leftover from somewhere else.

docs/rules/jsx-props-no-multi-spaces.md
ljharb: ahh, ok. yes, thanks, a separate PR for it would be great.

ThiefMaster:
PR updated

ThiefMaster:
Any chance to get it merged so it makes it into the next version?

klimashkin:
Released in `7.9.0`.
But seems like there is a conflict with eslint `indent` rule:

image

image

@ThiefMaster Is there a workaround for this? Probably i'm missing something in config
```javascript
    'react/jsx-indent-props': [2, 'first'],
    'indent': [2, 2, {
      'SwitchCase': 1,
      'VariableDeclarator': {'var': 2, 'let': 2, 'const': 3},
      'outerIIFEBody': 1,
      'MemberExpression': 1,
      'FunctionDeclaration': {
        'parameters': 'first', 
        'body': 1, 
      },
      'FunctionExpression': {
        'parameters': 'first',
        'body': 1,
      },
      'CallExpression': {
        'arguments': 1,
      },
      'ArrayExpression': 1,
      'ObjectExpression': 1,
      'ImportDeclaration': 1,
      'flatTernaryExpressions': true,
      'ignoredNodes': [],
      'ignoreComments': false,
    }],
```

ThiefMaster:
You have to exclude the JSX nodes from the indent rules:

      ignoredNodes:
        - JSXAttribute
        - JSXSpreadAttribute

klimashkin:
@ThiefMaster Thank you for quick response, that works! 
Probably that should be added to [the rule page](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-indent-props.md)

Issue #1724, opened 2018-03-13
react/require-default-props not support type params of React Component


Miziak:
When we define Component prop types like this:
```
type SelectProps = {
  isOpen?: boolean
}
class Select extends PureComponent<SelectProps, SelectState> {
  static defaultProps = {
    // isOpen: false,
  }
  ...
}
```
We don't get error like this:
> 17:3  error  propType "isOpen" is not required, but has no corresponding defaultProp declaration  react/require-default-props

To get this error we need to define component like this:
```
type SelectProps = {
  isOpen?: boolean
}
class Select extends PureComponent {
  props: SelectProps
  static defaultProps = {
    // isOpen: false,
  }
  ...
}
```

eslint v: `^4.17.0`
eslint-plugin-react v: `^7.6.1`

ljharb:
Fixed by #1733.

PR #1733, opened 2018-03-20
Issue #1724 support for Flow generic PropTypes declaration in require-default-props rule

Miziak:
Hi it's my first PR to this repo so if something is wrong please comment me.

### DIFF
--- a/lib/rules/require-default-props.js
+++ b/lib/rules/require-default-props.js
@@ -330,7 +330,7 @@ module.exports = {
     }
 
     function hasPropTypesAsGeneric(node) {
-      return node.parent && node.parent.type === 'ClassDeclaration';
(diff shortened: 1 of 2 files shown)


lib/rules/require-default-props.js
```diff
@@ -300,6 +300,50 @@ module.exports = {
...
+        return {
+          name: name,
+          isRequired: !property.optional,
+          node: property
+        };
+      });
+    }
+
+    function hasPropTypesAsGeneric(node) {
+      return node.parent && node.parent.type === 'ClassDeclaration';
```
ljharb: let's do `node && node.parent` here just to be sure it won't crash :-)

Miziak:
Ok, I added suggested undefined check on node.

PR #1735, opened 2018-03-22
forbid-component-props: add component whitelist

### LINKED ISSUE #1732: forbid-component-props should have a component whitelist
For custom components it makes sense to have the rule, but when using third-party components violating the rule (`ReactModal` using `className` is probably one of the more prominent examples) there isn't much you can do about it...

Having to add `eslint-disable react/forbid-component-props` comments isn't a very pretty solution.

ThiefMaster:
Docs are still missing ;)

fixes #1732

### DIFF
--- a/lib/rules/forbid-component-props.js
+++ b/lib/rules/forbid-component-props.js
@@ -57,8 +57,8 @@ module.exports = {
   create: function(context) {
     const configuration = context.options[0] || {};
     const forbid = new Map((configuration.forbid || DEFAULTS).map(value => {
-      const propName = (typeof value === 'string') ? value : value.propName;
-      const whitelist = (typeof value === 'string') ? [] : (value.allowedFor || []);
+      const propName = typeof value === 'string' ? value : value.propName;
+      const whitelist = typeof value === 'string' ? [] : (value.allowedFor || []);
       return [propName, whitelist];
     }));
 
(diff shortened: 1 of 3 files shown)


lib/rules/forbid-component-props.js
```diff
@@ -31,20 +31,40 @@ module.exports = {
...
-            type: 'string'
+            oneOf: [{
+              type: 'string'
+            }, {
+              type: 'object',
+              properties: {
+                propName: {
+                  type: 'string'
+                },
+                allowedFor: {
```
ljharb: let's make sure this list is unique

lib/rules/forbid-component-props.js
```diff
@@ -31,20 +31,40 @@ module.exports = {
...
+      }
     }]
   },
 
   create: function(context) {
-    function isForbidden(prop) {
-      const configuration = context.options[0] || {};
+    const configuration = context.options[0] || {};
+    const forbid = new Map((configuration.forbid || DEFAULTS).map(value => {
+      const propName = (typeof value === 'string') ? value : value.propName;
```
ljharb: let's remove the redundant parens on the ternary conditionals here and below

lib/rules/forbid-component-props.js
ThiefMaster: done; didn't know json schema has an option to require uniqueness

ThiefMaster:
Docs added, changes done.

ThiefMaster:
Anything missing to get this merged? Would be nice if we could start using it (by installing the plugin from this git repo until it's part of a release, without having to maintain our own branch containing this and my other PR).

Issue #1709, opened 2018-03-01
destructuring-assignment for different component types


HenriBeck:
I would like to propose an option for allowing different behaviors for a functional component and a class component for the destructuring-assignment rule.

ljharb:
Can you elaborate, with code examples, and explain why you’d want it to be different?

HenriBeck:
Personally, I like to destructure my props in a functional component, for the simple reason of passing additional extra props down to a specific child. Though I tend to not do that for class components.
Just having the option would be nice.
Of course, it would still allow for just a string as the first configuration. This would just add a few extra configuration options.

New rule configuration example:
```js
module.exports = {
  rules: {
    'react/destructuring-assignment': ['error', {
      functional: 'always',
      class: 'never',
    }],
  },
};
```

The correct code for the configuration example: 
```js
function FunctionalComponent({ children }) {
  return children;
}

class Component extends React.Component {
  render() {
    return this.props.children;
  }
}
```


Incorrect code for the configuration example: 
```js
class Component extends React.Component {
  render() {
    const { children } = this.props;

    return children;
  }
}
```

ljharb:
We'd need to also account for createClass components; but if we had all three, and all three defaulted to `true` (or, we had an "ignore" object with all three defaulting to "false", which is probably better) that could work.

HenriBeck:
I would personally prefer rather allowing an object for more detailed configuration rather than turning the rule off for certain component types.
```js
// Still support configuration for all three component types
module.exports = {
  rules: {
    'react/destructuring-assignment': ['error',  'always'],
  },
};

// Individual configuration for the component types
module.exports = {
  rules: {
    'react/destructuring-assignment': ['error', {
      functional: 'always',
      class: 'never',
      createClass: 'always',
    }],
  },
};

// Or allow for specific component overrides
module.exports = {
  rules: {
    'react/destructuring-assignment': ['error', 'always', {
      class: 'never',
    }],
  },
};
```

ljharb:
Using always/never seems most appropriate here. Want to make a PR?

HenriBeck:
Sure, I will start working on it next week. Which configuration should I implement?

ljharb:
The always/never one you suggested, with "class", "createClass", and "SFC" as options, would be great.

HenriBeck:
Also, I discovered this bug:
In both cases when destructuring is set to always it's not reporting an error.

```js
function Baz({ ...rest }) {
[...]

PR #1740, opened 2018-03-23
Partial rewrite of the destructuring rule

HenriBeck:
Pull Request for #1709 

This introduces an object as an option for specifying different behavior for different component types.

I also added a check if we are in a `render` method for class components (don't know if this is wanted but the docs only showed examples with `render` methods).

What is still missing:
- [x] Checking for destructuring props inside the functional component body
- [x] Update tests
- [x] Update Docs
- [ ] Add an option for allowing the props inside class properties (Future PR)
- [ ] Add an option for also validating class methods (Future PR)

This also fixes a few bugs:

Fixes when the `props` aren't named `props`:
```js
function Baz(properties) {
  return (<div>{properties.children}</div>); // Didn' report here before
}
```

### DIFF
--- a/lib/rules/destructuring-assignment.js
+++ b/lib/rules/destructuring-assignment.js
@@ -15,6 +15,14 @@ const SCHEMA = {
   ]
 };
 
+function isAssignmentToProp(node) {
+  return (
+    node.parent &&
+    node.parent.type === 'AssignmentExpression' &&
+    node.parent.left === node
+  );
+}
+
 /**
  * Check if the node is in a render method of the component.
  *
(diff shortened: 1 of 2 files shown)


HenriBeck:
@ljharb could you take a look?

ljharb:
@HenriBeck i know it's been awhile; but did you have to delete the fork? This still seems useful, and I'd prefer not to have to have a different PR.

HenriBeck:
@ljharb yeah, I just cleaned up my forks and forgot that I still had a PR open here.
I can recreate it when I have time, you can still see the changes so it shouldn't be too big of an effort.

HenriBeck:
I also have the changes on my laptop.

PR #1742, opened 2018-03-26
[Docs] Re-sort all rules alphabetically

ybiquitous:
To make it easier to find rules.

It will be appreciated if you review it. 😊

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,8 @@ before_script:
   - 'if [ -n "${ESLINT-}" ]; then npm install --no-save "eslint@${ESLINT}" ; fi'
 script:
   - 'if [ -n "${PRETEST-}" ]; then npm run pretest ; fi'
-  - 'if [ -z "${PRETEST-}" ] && [ -n "${TEST-}" ]; then npm run unit-test ; fi'
+  - 'if [ -n "${TEST-}" ]; then npm run unit-test ; fi'
+  - 'if [ -n "${README-}" ]; then npm run generate-list-of-rules:check ; fi'
 env:
   global:
     - TEST=true
@@ -29,6 +30,8 @@ matrix:
   include:
     - node_js: 'lts/*'
       env: PRETEST=true
+    - node_js: 'lts/*'
+      env: README=true
   exclude:
     - node_js: '4'
       env: ESLINT=5
--- a/README.md
+++ b/README.md
@@ -97,94 +97,100 @@ Enable the rules that you would like to use.
 
 # List of supported rules
 
+<!-- AUTO-GENERATED-CONTENT:START (BASIC_RULES) -->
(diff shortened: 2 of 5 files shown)


ybiquitous:
Thanks for your review!

> is there any way we could add a CI check that prevents these from being mis-sorted?

I think it is easiest to use regular expressions, so what do you think?

For example:

```js
const pattern = /# List of supported rules([\s\S]+?)#/;
const source = '<README.md contents...>';
const ruleListText = source.match(pattern)[1];
const ruleList = ruleListText.trim().split('*');
// check sorting...
```

(I think using the markdown parser is too heavy for this task...)

ljharb:
I'm not sure using an X parser is ever too heavy a task to parse X :-)

ljharb:
An alternative would be to make an `npm run-script` that *generates* this list automatically, and then the CI check would simply be "run the check, and fail if there's a diff".

ybiquitous:
Sorry for late reply 🙏 

> An alternative would be to make an npm run-script that generates this list automatically, and then the CI check would simply be "run the check, and fail if there's a diff".

Thanks, I will try it!
But diffs in this PR becomes too large, so can I do this with another PR?

ljharb:
@ybiquitous i'm not concerned with a large diff since it can be in a different commit; is it OK if we do it in this PR?

ybiquitous:
> is it OK if we do it in this PR?

Yes! No problem! 🙏

ybiquitous:
This is so old. I close.

ljharb:
That’s unfortunate; old is irrelevant. If you restore the branch and reopen the PR, and add the requested CI check, this can land.

ybiquitous:
@ljharb Thanks for your comment. I retried via PR #2541.

ljharb:
There's no "restore branch" button?

ybiquitous:
I know the restore button, but I thought the branch name `readme-resort-rules` of this PR does not express the new PR #2541.
So, I opened as another PR.

Sorry for confusing you... 🙇

ljharb:
I'd still prefer to consolidate PR refs - the branch name doesn't stick around but the PR ref does - so it'd be great if you could reopen this, and I'll be happy to manually keep this one in sync with your new one.

ybiquitous:
OK, I will sync both. 👍

ybiquitous:
I've cherry-picked from #2541.

ybiquitous:
@ljharb Thanks for your suggestion. [markdown-magic](https://github.com/DavidWells/markdown-magic) can reduce the script and works very well! 

If we would publish and share the transform script, it might be better to publish as a [markdown-magic plugin](https://github.com/DavidWells/markdown-magic#adding-custom-transforms).

See also https://github.com/yannickcr/eslint-plugin-react/pull/2541#discussion_r365995925

ybiquitous:
@ljharb Could you please review this again?

PR #1750, opened 2018-03-29
Adjusted no-deprecated rule for React 16.3.0

sergei-startsev:
Added warnings for deprecated lifecycle methods:
* componentWillMount,
* componentWillReceiveProps,
* componentWillUpdate.

See details [React 16.3.0](https://reactjs.org/blog/2018/03/29/react-v-16-3.html#component-lifecycle-changes).

### DIFF
--- a/lib/rules/no-deprecated.js
+++ b/lib/rules/no-deprecated.js
@@ -2,14 +2,17 @@
  * @fileoverview Prevent usage of deprecated methods
  * @author Yannick Croissant
  * @author Scott Feeney
+ * @author Sergei Startsev
  */
 'use strict';
 
 const has = require('has');
 
+const Components = require('../util/Components');
+const astUtil = require('../util/ast');
+const docsUrl = require('../util/docsUrl');
 const pragmaUtil = require('../util/pragma');
 const versionUtil = require('../util/version');
-const docsUrl = require('../util/docsUrl');
 
 // ------------------------------------------------------------------------------
 // Constants
@@ -37,7 +40,7 @@ module.exports = {
     schema: []
   },
 
-  create: function(context) {
+  create: Components.detect((context, components, utils) => {
     const sourceCode = context.getSourceCode();
     const pragma = pragmaUtil.getFromContext(context);
 
@@ -73,6 +76,10 @@ module.exports = {
       deprecated[`${pragma}.PropTypes`] = ['15.5.0', 'the npm module prop-types'];
       // 15.6.0
       deprecated[`${pragma}.DOM`] = ['15.6.0', 'the npm module react-dom-factories'];
+      // 16.3.0
+      deprecated.componentWillMount = ['16.3.0'];
+      deprecated.componentWillReceiveProps = ['16.3.0'];
+      deprecated.componentWillUpdate = ['16.3.0'];
       return deprecated;
     }
 
@@ -119,6 +126,27 @@ module.exports = {
       return moduleName;
     }
 
+    /**
+     * Returns life cycle methods if available
+     * @param {ASTNode} node The AST node being checked.
+     * @returns {Array} The array of methods.
+     */
+    function getLifeCycleMethods(node) {
+      const properties = astUtil.getComponentProperties(node);
+      return properties.map(property => astUtil.getPropertyName(property));
+    }
+
+    /**
+     * Checks life cycle methods
+     * @param {ASTNode} node The AST node being checked.
+     */
+    function checkLifeCycleMethods(node) {
+      if (utils.isES5Component(node) || utils.isES6Component(node)) {
+        const methods = getLifeCycleMethods(node);
+        methods.forEach(method => checkDeprecation(node, method));
+      }
+    }
+
     // --------------------------------------------------------------------------
     // Public
     // --------------------------------------------------------------------------
@@ -160,8 +188,11 @@ module.exports = {
         node.id.properties.forEach(property => {
           checkDeprecation(node, `${reactModuleName || pragma}.${property.key.name}`);
         });
-      }
+      },
 
+      ClassDeclaration: checkLifeCycleMethods,
+      ClassExpression: checkLifeCycleMethods,
+      ObjectExpression: checkLifeCycleMethods
     };
-  }
+  })
 };
--- a/lib/rules/no-deprecated.js
+++ b/lib/rules/no-deprecated.js
@@ -23,7 +23,7 @@ const MODULES = {
   'react-addons-perf': ['ReactPerf', 'Perf']
 };
 
-const DEPRECATED_MESSAGE = '{{oldMethod}} is deprecated since React {{version}}{{newMethod}}';
+const DEPRECATED_MESSAGE = '{{oldMethod}} is deprecated since React {{version}}{{newMethod}}{{refs}}';
 
 // ------------------------------------------------------------------------------
 // Rule Definition
@@ -77,9 +77,21 @@ module.exports = {
       // 15.6.0
       deprecated[`${pragma}.DOM`] = ['15.6.0', 'the npm module react-dom-factories'];
       // 16.3.0
-      deprecated.componentWillMount = ['16.3.0'];
-      deprecated.componentWillReceiveProps = ['16.3.0'];
-      deprecated.componentWillUpdate = ['16.3.0'];
+      deprecated.componentWillMount = [
+        '16.3.0',
+        'constructor',
+        'https://reactjs.org/docs/react-component.html#unsafe_componentwillmount'
+      ];
+      deprecated.componentWillReceiveProps = [
+        '16.3.0',
+        'getDerivedStateFromProps',
+        'https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops'
+      ];
+      deprecated.componentWillUpdate = [
+        '16.3.0',
+        'getDerivedStateFromProps',
+        'https://reactjs.org/docs/react-component.html#unsafe_componentwillupdate'
+      ];
       return deprecated;
     }
 
@@ -98,13 +110,17 @@ module.exports = {
         return;
       }
       const deprecated = getDeprecated();
+      const version = deprecated[method][0];
+      const newMethod = deprecated[method][1];
+      const refs = deprecated[method][2];
       context.report({
         node: node,
         message: DEPRECATED_MESSAGE,
         data: {
           oldMethod: method,
-          version: deprecated[method][0],
-          newMethod: deprecated[method][1] ? `, use ${deprecated[method][1]} instead` : ''
+          version,
+          newMethod: newMethod ? `, use ${newMethod} instead` : '',
+          refs: refs ? `, see ${refs}` : ''
         }
       });
     }
@@ -173,10 +189,10 @@ module.exports = {
       VariableDeclarator: function(node) {
         const reactModuleName = getReactModuleName(node);
         const isRequire = node.init && node.init.callee && node.init.callee.name === 'require';
-        const isReactRequire =
-          node.init && node.init.arguments &&
-          node.init.arguments.length && typeof MODULES[node.init.arguments[0].value] !== 'undefined'
-        ;
+        const isReactRequire = node.init
+          && node.init.arguments
+          && node.init.arguments.length
+          && typeof MODULES[node.init.arguments[0].value] !== 'undefined';
         const isDestructuring = node.id && node.id.type === 'ObjectPattern';
 
         if (
--- a/lib/rules/no-deprecated.js
+++ b/lib/rules/no-deprecated.js
@@ -79,17 +79,17 @@ module.exports = {
       // 16.3.0
       deprecated.componentWillMount = [
         '16.3.0',
-        'constructor',
+        'UNSAFE_componentWillMount',
         'https://reactjs.org/docs/react-component.html#unsafe_componentwillmount'
       ];
       deprecated.componentWillReceiveProps = [
         '16.3.0',
-        'getDerivedStateFromProps',
+        'UNSAFE_componentWillReceiveProps',
         'https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops'
       ];
       deprecated.componentWillUpdate = [
         '16.3.0',
-        'getDerivedStateFromProps',
+        'UNSAFE_componentWillUpdate',
         'https://reactjs.org/docs/react-component.html#unsafe_componentwillupdate'
       ];
       return deprecated;
--- a/tests/lib/rules/no-deprecated.js
+++ b/tests/lib/rules/no-deprecated.js
@@ -24,11 +24,10 @@ const parserOptions = {
 
 require('babel-eslint');
 
-function errorMessage(oldMethod, version, newMethod) {
-  if (!newMethod) {
-    newMethod = '';
-  }
-  return `${oldMethod} is deprecated since React ${version}${newMethod}`;
+function errorMessage(oldMethod, version, newMethod, refs) {
+  newMethod = newMethod ? `, use ${newMethod} instead` : '';
+  refs = refs ? `, see ${refs}` : '';
+  return `${oldMethod} is deprecated since React ${version}${newMethod}${refs}`;
 }
 
 // ------------------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


lib/rules/no-deprecated.js
```diff
@@ -73,6 +76,10 @@ module.exports = {
       deprecated[`${pragma}.PropTypes`] = ['15.5.0', 'the npm module prop-types'];
       // 15.6.0
       deprecated[`${pragma}.DOM`] = ['15.6.0', 'the npm module react-dom-factories'];
+      // 16.3.0
+      deprecated.componentWillMount = ['16.3.0'];
```
ljharb: can we supply messages in all these three, that ideally suggest replacements and link to the react docs on the deprecation?

lib/rules/no-deprecated.js
```diff
@@ -37,7 +40,7 @@ module.exports = {
     schema: []
   },
 
-  create: function(context) {
+  create: Components.detect((context, components, utils) => {
```
ljharb: good change

tests/lib/rules/no-deprecated.js
```diff
@@ -39,151 +47,316 @@ ruleTester.run('no-deprecated', rule, {
...
-  }]
+      parser: 'babel-eslint',
+      errors: [{
+        message: 'ReactPerf.printDOM is deprecated since React 15.0.0, use ReactPerf.printOperations instead'
+      }]
+    },
+    {
+      code: 'React.DOM.div',
+      errors: [{
+        message: 'React.DOM is deprecated since React 15.6.0, use the npm module react-dom-factories instead'
```
ljharb: can all these tests be migrated to use the `errorMessage` helper as well?

sergei-startsev: sure thing, see the updated tests

lib/rules/no-deprecated.js
sergei-startsev: Could you take a look at the new message pattern?

lib/rules/no-deprecated.js
```diff
@@ -73,6 +76,22 @@ module.exports = {
       deprecated[`${pragma}.PropTypes`] = ['15.5.0', 'the npm module prop-types'];
       // 15.6.0
       deprecated[`${pragma}.DOM`] = ['15.6.0', 'the npm module react-dom-factories'];
+      // 16.3.0
+      deprecated.componentWillMount = [
+        '16.3.0',
+        'constructor',
```
ljharb: hmm, i'm actually not comfortable recommending the constructor here - side-effecting things should never go in a constructor, and there's still some use cases for `componentWillMount` that aren't covered by "constructor" or "getDerivedStateFromProps".

I wonder if maybe we should also add an "ignore" array of method names; ie so I can configure the rule to ignore `componentWillMount` but warn on the rest.

asapach: How about recommending `componentDidMount()` (or [`UNSAFE_componentWillMount()`](https://reactjs.org/docs/react-component.html#unsafe_componentwillmount) as a temporary measure)?

ljharb: I guess the unsafe one; but it'd still probably be a good idea to add an ignore list.

sergei-startsev: Agree, in terms of consistency it would be better to deprecated them in favor of `UNSAFE_` methods, it should allow to avoid confusion. @ljharb, does it sound reasonable to leave warnings with `UNSAFE_` methods as new methods for now and deprecate them in 17.0.0?

sergei-startsev: Regarding the ignore list for that rule, I vote for adding a dedicated issue for further discussion. I could take a look at it.

sergei-startsev:
@ljharb, is there anything else that we have to discuss here?

lib/rules/no-deprecated.js
ljharb: A separate issue for the "UNSAFE_" methods would be ideal.

lib/rules/no-deprecated.js
```diff
@@ -142,13 +186,13 @@ module.exports = {
...
 
-      VariableDeclarator: function(node) {
+      VariableDeclarator: function (node) {
         const reactModuleName = getReactModuleName(node);
         const isRequire = node.init && node.init.callee && node.init.callee.name === 'require';
         const isReactRequire =
           node.init && node.init.arguments &&
           node.init.arguments.length && typeof MODULES[node.init.arguments[0].value] !== 'undefined'
-        ;
+          ;
```
ljharb: this line should be de-indented

lib/rules/no-deprecated.js
```diff
@@ -119,17 +142,38 @@ module.exports = {
...
+    }
+
     // --------------------------------------------------------------------------
     // Public
     // --------------------------------------------------------------------------
 
     return {
 
-      MemberExpression: function(node) {
+      MemberExpression: function (node) {
```
ljharb: please revert all unrelated formatting changes in this PR

lib/rules/no-deprecated.js
```diff
@@ -79,17 +79,17 @@ module.exports = {
...
       deprecated.componentWillMount = [
         '16.3.0',
-        'constructor',
+        'UNSAFE_componentWillMount',
         'https://reactjs.org/docs/react-component.html#unsafe_componentwillmount'
       ];
       deprecated.componentWillReceiveProps = [
         '16.3.0',
-        'getDerivedStateFromProps',
+        'UNSAFE_componentWillReceiveProps',
```
sahrens: Why not suggest getDerivedStateFromProps? And why not lint against UNSAFE_ lifecycle methods?

ljharb: Because that's only one of the use cases for `componentWillReceiveProps`, so it's not always the right recommendation.

sahrens: What's the point of linting against `componentWillReceiveProps` if we're just going to recommend `UNSAFE_componentWillReceiveProps` anyway? Both are equally bad, no? They do the exact same thing, just one has a scary name.

ljharb: So that a human can look at it and make a human judgement about which to use.

sergei-startsev: @sahrens, legacy lifecycles (including `componentWillReceiveProps`) will continue to work until version 17, however you will see deprecation warnings for them. In version 17, it will still be possible to use them, but they will be aliased with an “UNSAFE_”, see [details](https://reactjs.org/blog/2018/03/29/react-v-16-3.html#component-lifecycle-changes).

sahrens: To clarify, you don't think we should lint against using any UNSAFE_ methods, even if we don't give a specific recommendation of what to use instead? Seems like linting against them earlier rather than later would help folks track down and remove usage before they start firing a bunch of deprecation warnings.

sahrens: Perhaps it would make more sense to make a new "strict-mode" lint rule? Any plans in that direction?

ljharb: I think that `no-deprecated` can and will warn against them, but not until *every* use case of them has a replacement (this is not currently the case).

sergei-startsev: There're no deprecation warnings regarding using `UNSAFE_` methods in strict mode, there is unsafe lifecycle methods warning: 
unsafe lifecycle methods warning

I'm not sure regarding `strict-mode` rule, I vote for adding `no-unsafe` rule that will be more specific. In the same way that it was done for `no-string-refs` rule -- using string ref API also cause warnings in strict mode. @ljharb, I could take a look at it if you find it reasonable.

ljharb: Modes are not a conceptual pattern I'd want to encourage, "strict" or otherwise.

I don't think a "no-unsafe" rule makes sense when for users of React < 16, it is perfectly safe, and for users of React >= 17, the methods won't work at all.

The purpose of these deprecations from React is to encourage people to migrate to replacements as early as possible, so the breaking change in v17 that removes the old ones will be as painless as possible. When React 17 comes out, we'll certainly lint against invalid lifecycle methods, including newly-invalid ones - but I'm not sure how much value we'd get out of warnings in React 16 outside of `no-deprecated`.

sergei-startsev: >I don't think a "no-unsafe" rule makes sense when for users of React < 16, it is perfectly safe, and for users of React >= 17, the methods won't work at all.

It's not true, even in version 17, it will still be possible to use UNSAFE_ methods, see [Component Lifecycle Changes](https://reactjs.org/blog/2018/03/29/react-v-16-3.html#component-lifecycle-changes).

ljharb: In that case, a separate "no-unsafe" rule, that only focuses on UNSAFE_ methods, seems reasonable.

gaearon:
Technically React didn't yet "deprecate" these methods. It *will* in a later version, but saying they're deprecated in 16.3.0 is not correct. For now they're just "legacy". React doesn't warn about them yet.

The warning also regrettably doesn't link to the codemod. So people seeing this will try to edit by hand because they have no idea the codemod exists.

ljharb:
Let’s update the warning text at least.

Should we maybe move these methods behind an option then?

gaearon:
While I appreciate the good intent behind this change, I think it was a mistake and has caused some damage. To be clear, it's out mistake that we haven't engaged with this plugin's maintainers to ensure it matches our strategy. Let's try not to repeat this mistake.

It's a big problem that people still see this rule and think those methods are already "deprecated". It makes our messaging inconsistent and, worse, spams people's consoles. **It makes updating React seem like a difficult thing to do.** A lot of people think ESLint messages "are" React messages, and just pick a popular config that includes this rule.

**In React, we prefer runtime warnings *first* so that we can effectively coalesce them. We put great care into our runtime warnings, and include links to migration instructions and codemods.** This rule has no link to migration instructions, so people assume they need to rename methods by hand. Its [documentation](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-deprecated.md) includes no upgrade instructions whatsoever.

Again, it's our fault we didn't look into this before, but it's a huge cost to the community who aren't necessarily educated about codemods. This reflected badly on React ecosystem.

In the future, I'd like to propose that:

* PRs that add deprecation warnings aren't merged without a look from somebody on the React team
* Deprecation lint warnings aren't introduced until we actually *deprecate* a feature in React (which hasn't quite happened yet).
* Deprecation lint warning messages must include links to codemods and migration instructions.
* Deprecation rule documentation page must include links to codemods and migration instructions.
* Deprecation lint rule is not added to "recommended" preset until the next React major. In other words, we want people to see React runtime warnings first, and update at their own pace. They can they *opt into* checking deprecations with a linter. But we want React runtime warnings to be the primary mechanism, and the linter being an opt-in way to avoid regressing. It should never be the case that somebody picks a popular or "recommended" preset, and that preset contains warnings we haven't even had a chance to surface at runtime in React.

Hope this makes sense! Again, I'm thankful for everyone's effort on this — but we can do better to educate the community and avoid a bad perception of complex upgrade paths. It pains me to think of all the beginners who had their terminals spammed with hundreds of warnings and no accessible explanation about e.g. codemods.

gaearon:
I suggest the following plan (I can do it myself):

* Change `no-deprecations` doc page to include a list of deprecations (as separate subheaders), and links to recommended migration path instructions for each of them.
* Link each method warning to the relevant section of `no-deprecations` rule.
* Ensure `recommended` preset doesn't yet include these particular deprecations (since they're not even deprecated inside React yet).

sergei-startsev:
@gaearon I'm thinking about linting these legacy lefecycle methods with a separate rule, e.g. `no-legacy`. Thoughts?

kof:
@sergei-startsev probably better to call them specifically, e.g. no-class-component etc. Those things are not officially legacy yet.

sergei-startsev:
Just curious why is it beneficial to see React runtime warnings first? I'm not sure if it's expected to see leaked warnings in production without mechanism to catch them in build time.

sergei-startsev:
@kof take a look at [Dan's comment](https://github.com/yannickcr/eslint-plugin-react/pull/1750#issuecomment-425975934), they're legacy for now.

sergei-startsev:
@ljharb it seems to be released in [7.8.0](https://github.com/yannickcr/eslint-plugin-react/releases/tag/v7.8.0).

ljharb:
ahhh, you're right, i misread the merge date. i'll update my comment.

sergei-startsev:
@ljharb, @gaearon what are our next steps regarding `legacy` methods? Would like to move it to a dedicated rule, adjust the existing? I can pick up the task if u specify the direction.

gaearon:
>I'm thinking about linting these legacy lefecycle methods with a separate rule, e.g. no-legacy. Thoughts?

Another option would be for `no-deprecated` to take React version as an argument. These methods could be tagged as something like `16.999` until we've decided which release deprecates them. People who want to opt in can specify `17` if they want to see the messages.

Additionally, there should probably be a `strict-mode` rule which replicates the warnings currently emitted by [`<StrictMode>`](https://reactjs.org/docs/strict-mode.html) and similarly take a version as argument.

>I'm not sure if it's expected to see leaked warnings in production without mechanism to catch them in build time.

Not sure what you mean about seeing them in production. React warnings are DEV-only. You'd see them during development. Their only purpose is to warn about removals in future majors — so their presence also doesn't mean your production code is buggy. Unlike many other rules.

>Just curious why is it beneficial to see React runtime warnings first? 

A few reasons.

The first is that we coalesce runtime warnings where possible. We can show a single warning for many components. We try to avoid showing the same warning twice (if we do, it's usually a bug).

The second is that you only see warnings for code that actually runs while you develop. This may seem like it's bad. (Don't we want to find all issues?) But the all-or-nothing approach often discourages people from doing the migration work. If you see all warnings at once (instead of your little corner of the product) you're less incentivized to fix them. So we consider limiting your initial exposure to just the parts you interact with to be a feature, not a bug.

Of course once you fixed them (or at least some of them) you want to avoid regressing. That's where the rule comes it. I think *lint rules are helpful for preventing regressions, but aren't the best instrument to get notified about the issues for the first time*. At least with the open source tools today that just show you all lint warnings at once.

>Sounds good! How should I best ping you all?

GH notifications in other repos are too noisy but filing a React issue or pinging in Twitter DM are good ways.

>That sounds reasonable, but even a runtime warning is imo a deprecation in React.

We don't have runtime warnings for these methods in React outside of Strict Mode.

>I'm not sure I agree that runtime warnings should be the primary mechanism; runtime warnings are incredibly easy to miss, suppressed in production, and the linter is a far more effective way (in a large organization, at least) to get attention on these warnings.

I agree with that for *new code*. The thing I'm worried about is upgrade path. You bump a preset everybody uses, and suddenly it says you have 500 violations. Even if you're on an older React minor that doesn't consider this feature deprecated yet. I think that's pretty confusing.

sergei-startsev:
> Another option would be for no-deprecated to take React version as an argument. These methods could be tagged as something like 16.999 until we've decided which release deprecates them. People who want to opt in can specify 17 if they want to see the messages.

I'm concerned about version 17 in lint configuration, it can confuse... As alternative I'd propose to move it to the existing [no-unsafe](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unsafe.md) rule since `componentWillMount`, `componentWillReceiveProps`, `componentWillUpdate` also cause _Unsafe lifecycle methods_ warnings in strict mode. Does it sound reasonable to you?

>If you see all warnings at once (instead of your little corner of the product) you're less incentivized to fix them. 

Sometimes it's also useful to see the entire scope. If you decide to turn on a new lint rule, you usually expect to see warnings/errors. `no-deprecated` is a bad example here since it's included into `recommended` preset, however it should work fine with `no-unsafe`.

gaearon:
>As alternative I'd propose to move it to the existing no-unsafe rule since componentWillMount, componentWillReceiveProps, componentWillUpdate also cause Unsafe lifecycle methods warnings in strict mode.

👍

sergei-startsev:
🤝 if you don't mind I'll request a code review for PR

gaearon:
Note: writing the doc page is still on my todo list but I'm going on a vacation soon so maybe I'll come back to this after.

ljharb:
I think it'd be a bit weird to have a second place to configure a React version - at the very least, it sounds like there's a bunch of deprecations in the rule that are activating at React versions that are lower than they should be - a PR to fix that (even if it hoists the warnings to 16.999) would be appreciated.

PR #1755, opened 2018-04-01
Add jsx-props-no-multi-spaces rule

ThiefMaster:
This rule enforced consistent spacing of JSX tag attributes and is especially useful together with `jsx-indent-props` in "first" mode (#1729) to actually require consistency of where the first attribute begins.


ThiefMaster:
Any chance to get it merged so it makes it into the next version?

Issue #1767, opened 2018-04-12
Update sort-comp for React 16.3+


webOS101:
The `sort-comp` rule will need to be updated to include the new lifecycle methods (while also keeping the older ones, optionally?).  These include:

```
getDerivedStateFromProps
getSnapshotBeforeUpdate
UNSAFE_componentWillMount
UNSAFE_componentWillReceiveProps
UNSAFE_componentWillUpdate
```

otakustay:
Also, `componentDidCatch` is not listed in current `sort-comp` rule

joe-denea:
I made an attempt to add this feature with #1771 

Would love to hear any feedback/suggestions

ljharb:
Closed in #1771.

PR #1771, opened 2018-04-18
Add support for React 16.3 lifecycle methods in sort-comp

joe-denea:
Relevant issue: #1767 

The new lifecycle methods added in React 16.3 need to be added to the `sort-comp` rule for compatibility.

Also the rule needs to be updated to allow static method positions to be overridden so that `getDerivedStateFromProps` can be placed in the ordering that exists in [React documentation](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops), which appears to be how the other methods are currently ordered by default (aside from the render method).

Would love to hear any feedback/suggested changes.

I detailed potential issues etc. in my commit notes here:
- Add React 16.3 lifecycle methods to default config
- Allow static method position to be overriden by specification
  - Note: this update means that a static method of the same name as a non-static method in the sort-comp will be considered an error if it is not placed in the position corresponding to its name
- Add `instance-variables` to the lifecycle after the `constructor` and before `getDefaultProps`
  - Note: This update is not required, but is a nice to have since the default order has to be updated anyway, and current behavior forces all instance variables to come after `componentWillUnmount` by default, which is a strange ordering.
  - Note: The super constructor runs prior to instance variables being instantiated, so `this.props` can be referenced within an instance variable definition. However, instance variables are defined prior to the current class' constructor executing. Since it is hard to represent this flow I thought it would be better to place instance variables below the constructor so as to make it obvious that `this.props` is defined, since defining an instance variable then using it in the constructor seems like an unnecessary use case.
- Add a test case for a React 16.3 class
- Update documentation to reflect changes

### DIFF
--- a/docs/rules/sort-comp.md
+++ b/docs/rules/sort-comp.md
@@ -1,6 +1,6 @@
 # Enforce component methods order (react/sort-comp)
 
-When creating React components it is more convenient to always follow the same organisation for method order to help you easily find lifecyle methods, event handlers, etc.
+When creating React components it is more convenient to always follow the same organisation for method order to help you easily find lifecycle methods, event handlers, etc.
 
 **Fixable:** This rule is automatically fixable using the [`sort-comp` transform](https://github.com/reactjs/react-codemod/blob/master/transforms/sort-comp.js) in [react-codemod](https://www.npmjs.com/package/react-codemod).
 
@@ -9,7 +9,7 @@ When creating React components it is more convenient to always follow the same o
 The default configuration ensures that the following order must be followed:
 
   1. static methods and properties
-  2. lifecycle methods: `displayName`, `propTypes`, `contextTypes`, `childContextTypes`, `mixins`, `statics`,`defaultProps`, `constructor`, `getDefaultProps`, `getInitialState`, `state`, `getChildContext`, `componentWillMount`, `componentDidMount`, `componentWillReceiveProps`, `shouldComponentUpdate`, `componentWillUpdate`, `componentDidUpdate`, `componentWillUnmount` (in this order).
+  2. lifecycle methods: `displayName`, `propTypes`, `contextTypes`, `childContextTypes`, `mixins`, `statics`, `defaultProps`, `constructor`, `getDefaultProps`, `state`, `getInitialState`, `getChildContext`, `getDerivedStateFromProps`, `componentWillMount`, `UNSAFE_componentWillMount`, `componentDidMount`, `componentWillReceiveProps`, `UNSAFE_componentWillReceiveProps`, `shouldComponentUpdate`, `componentWillUpdate`, `UNSAFE_componentWillUpdate`, `getSnapshotBeforeUpdate`, `componentDidUpdate`, `componentDidCatch`, `componentWillUnmount` (in this order).
   3. custom methods
   4. `render` method
 
@@ -70,15 +70,21 @@ The default configuration is:
       'defaultProps',
       'constructor',
       'getDefaultProps',
-      'getInitialState',
       'state',
+      'getInitialState',
       'getChildContext',
+      'getDerivedStateFromProps',
       'componentWillMount',
+      'UNSAFE_componentWillMount',
       'componentDidMount',
       'componentWillReceiveProps',
+      'UNSAFE_componentWillReceiveProps',
       'shouldComponentUpdate',
       'componentWillUpdate',
+      'UNSAFE_componentWillUpdate',
+      'getSnapshotBeforeUpdate',
       'componentDidUpdate',
+      'componentDidCatch',
       'componentWillUnmount'
     ]
   }
@@ -86,14 +92,14 @@ The default configuration is:
 ```
 
 * `static-methods` is a special keyword that refers to static class methods.
-* `lifecycle` is referring to the `lifecycle` group defined in `groups`.
-* `everything-else` is a special group that match all the methods that do not match any of the other groups.
-* `render` is referring to the `render` method.
-* `type-annotations`. This group is not specified by default, but can be used to enforce flow annotations positioning.
-* `getters` This group is not specified by default, but can be used to enforce class getters positioning.
-* `setters` This group is not specified by default, but can be used to enforce class setters positioning.
-* `instance-variables` This group is not specified by default, but can be used to enforce all other instance variables positioning.
-* `instance-methods` This group is not specified by default, but can be used to enforce all other instance methods positioning.
+* `lifecycle` refers to the `lifecycle` group defined in `groups`.
+* `everything-else` is a special group that matches all of the methods that do not match any of the other groups.
+* `render` refers to the `render` method.
+* `type-annotations`. This group is not specified by default, but can be used to enforce flow annotations' positioning.
+* `getters` This group is not specified by default, but can be used to enforce class getters' positioning.
+* `setters` This group is not specified by default, but can be used to enforce class setters' positioning.
+* `instance-variables` This group is not specified by default, but can be used to enforce all other instance variables' positioning.
+* `instance-methods` This group is not specified by default, but can be used to enforce all other instance methods' positioning.
 
 You can override this configuration to match your needs.
 
@@ -216,7 +222,6 @@ class Hello extends React.Component<any, Props, void> {
   }
 }
 ```
-
 
 ## When Not To Use It
 
--- a/tests/lib/rules/sort-comp.js
+++ b/tests/lib/rules/sort-comp.js
@@ -92,6 +92,77 @@ ruleTester.run('sort-comp', rule, {
       ]
     }]
   }, {
+    // Must validate a full React 16.3 createReactClass class
+    code: [
+      'var Hello = createReactClass({',
+      '  displayName : \'\',',
+      '  propTypes: {},',
+      '  contextTypes: {},',
+      '  childContextTypes: {},',
+      '  mixins: [],',
+      '  statics: {},',
+      '  getDefaultProps: function() {},',
+      '  getInitialState: function() {},',
+      '  getChildContext: function() {},',
+      '  UNSAFE_componentWillMount: function() {},',
+      '  componentDidMount: function() {},',
+      '  UNSAFE_componentWillReceiveProps: function() {},',
+      '  shouldComponentUpdate: function() {},',
+      '  UNSAFE_componentWillUpdate: function() {},',
+      '  getSnapshotBeforeUpdate: function() {},',
+      '  componentDidUpdate: function() {},',
+      '  componentDidCatch: function() {},',
+      '  componentWillUnmount: function() {},',
+      '  render: function() {',
+      '    return <div>Hello</div>;',
+      '  }',
+      '});'
+    ].join('\n')
+  }, {
+    // Must validate React 16.3 lifecycle methods with the default parser
+    code: [
+      'class Hello extends React.Component {',
+      '  constructor() {}',
+      '  static getDerivedStateFromProps() {}',
+      '  UNSAFE_componentWillMount() {}',
+      '  componentDidMount() {}',
+      '  UNSAFE_componentWillReceiveProps() {}',
+      '  shouldComponentUpdate() {}',
+      '  UNSAFE_componentWillUpdate() {}',
+      '  getSnapshotBeforeUpdate() {}',
+      '  componentDidUpdate() {}',
+      '  componentDidCatch() {}',
+      '  componentWillUnmount() {}',
+      '  testInstanceMethod() {}',
+      '  render() { return (<div>Hello</div>); }',
+      '}'
+    ].join('\n')
+  }, {
+    // Must validate a full React 16.3 ES6 class
+    code: [
+      'class Hello extends React.Component {',
+      '  static displayName = \'\'',
+      '  static propTypes = {}',
+      '  static defaultProps = {}',
+      '  constructor() {}',
+      '  state = {}',
+      '  static getDerivedStateFromProps = () => {}',
+      '  UNSAFE_componentWillMount = () => {}',
+      '  componentDidMount = () => {}',
+      '  UNSAFE_componentWillReceiveProps = () => {}',
+      '  shouldComponentUpdate = () => {}',
+      '  UNSAFE_componentWillUpdate = () => {}',
+      '  getSnapshotBeforeUpdate = () => {}',
+      '  componentDidUpdate = () => {}',
+      '  componentDidCatch = () => {}',
+      '  componentWillUnmount = () => {}',
+      '  testArrowMethod = () => {}',
+      '  testInstanceMethod() {}',
+      '  render = () => (<div>Hello</div>)',
+      '}'
+    ].join('\n'),
+    parser: 'babel-eslint'
+  }, {
     // Must allow us to create a RegExp-based group
     code: [
       'class Hello extends React.Component {',
--- a/lib/rules/sort-comp.js
+++ b/lib/rules/sort-comp.js
@@ -32,12 +32,18 @@ const defaultConfig = {
       'state',
       'getInitialState',
       'getChildContext',
+      'getDerivedStateFromProps',
       'componentWillMount',
+      'UNSAFE_componentWillMount',
       'componentDidMount',
       'componentWillReceiveProps',
+      'UNSAFE_componentWillReceiveProps',
       'shouldComponentUpdate',
       'componentWillUpdate',
+      'UNSAFE_componentWillUpdate',
+      'getSnapshotBeforeUpdate',
       'componentDidUpdate',
+      'componentDidCatch',
       'componentWillUnmount'
     ]
   }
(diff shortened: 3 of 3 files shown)


docs/rules/sort-comp.md
```diff
@@ -9,7 +9,7 @@ When creating React components it is more convenient to always follow the same o
 The default configuration ensures that the following order must be followed:
 
   1. static methods and properties
-  2. lifecycle methods: `displayName`, `propTypes`, `contextTypes`, `childContextTypes`, `mixins`, `statics`,`defaultProps`, `constructor`, `getDefaultProps`, `getInitialState`, `state`, `getChildContext`, `componentWillMount`, `componentDidMount`, `componentWillReceiveProps`, `shouldComponentUpdate`, `componentWillUpdate`, `componentDidUpdate`, `componentWillUnmount` (in this order).
+  2. lifecycle methods: `displayName`, `propTypes`, `contextTypes`, `childContextTypes`, `mixins`, `statics`, `defaultProps`, `constructor`, `instance-variables`, `getDefaultProps`, `state`, `getInitialState`, `getChildContext`, `getDerivedStateFromProps`, `componentWillMount`, `UNSAFE_componentWillMount`, `componentDidMount`, `componentWillReceiveProps`, `UNSAFE_componentWillReceiveProps`, `shouldComponentUpdate`, `componentWillUpdate`, `UNSAFE_componentWillUpdate`, `getSnapshotBeforeUpdate`, `componentDidUpdate`, `componentWillUnmount`, `componentDidCatch` (in this order).
```
otakustay: I think `componentDidCatch` should be placed before `componentWillUnmount` since it is a method when component is "live"

joe-denea: Where exactly would you recommend putting it? Do you think it would make more sense directly after `componentDidMount` or directly before `componentWillUnmount`?

I think part of the reason it was separated from the lifecycle and placed at the bottom in the docs is that it doesn't fall within the update cycle like the other methods between `componentDidMount` and `componentWillUnmount`, but I would agree that it would be strange to write it after `componentWillUnmount`.

webOS101:
I think I would prefer to see the `instance-variables` removed from this PR so it could be debated further.  This will cause all previously passing source code to fail.  Keep in mind that many methods are defined as instance variables and should not appear before the lifecycle methods.  Further, I don't see any additional tests validating the change to `instance-variables`.

docs/rules/sort-comp.md
webOS101: fwiw, you can throw in the `componentWillUnmount` of a child and it will be caught by `componentDidCatch`.

joe-denea:
@webOS101 The nomenclature is a bit strange on `instance-variables` as they actually explicitly exclude `instance-methods` even if `instance-methods` is not a specified key. `instance-variables` only refers to non-static class property instantiations. Essentially the current behavior if you use class properties is for `sort-comp` to error out unless you define all your class properties after `componentWillUnmount`, which is extremely strange, considering that they are essentially equivalent to `this.myClassProp = 'my value'` in the constructor. I will add an instance variable to the test case I added. Let me know if you still think I should separate it though.

docs/rules/sort-comp.md
joe-denea: I'm not sure, but you may even be able to throw an error it would catch from an unmounted child component, assuming you have an unhandled timeout somewhere. Putting `componentDidCatch` at the end does make it clear that it catches errors from the preceding lifecycle hooks, but then again it doesn't actually catch errors within the component it was defined inside of. I just went with the docs on this one originally since the existing ordering followed them, but I don't know whether they put it at the bottom to separate it documentation-wise or because they would actually suggest writing it at the bottom, since all of the examples I have seen have been error boundary components with no other lifecycle methods.

yenshih: It will be better to put `getDerivedStateFromProps` right before `static`, since it is a static lifecycle.

otakustay: `componentDidCatch` is a method with behavior very close to `componentDidUpdate`, I'd prefer it to be placed right after `componentDidUpdate`

joe-denea: @yenshih That is not the ordering that the React documentation provides, as `getDerivedStateFromProps` is a lifecycle method, unlike other static methods and should be in a relevant position. Also `statics` only refers to the `React.createClass` statics object, and `static-methods` is actually where static methods on an ES6 class would be sorted to currently, which is before `statics` and actually lies at the very top of the component in the current default ordering.

@otakustay I don't really see the similarity there, `componentDidCatch` can be called at any time based on child component errors, whereas `componentDidUpdate` is called after every update to the component and is part of the update lifecycle. Also `componentDidCatch` is not something that is supposed to be triggered during the normal flow of your app, and it is explicitly stated in the documentation that it should not be used for control flow, which is what the lifecycle represents. I believe that that is the reason why React documentation has `componentDidCatch` at the bottom, after the normal lifecycle methods.

At this point I think it would be best to just stick with the React documentation's ordering since that seems to be what the ordering here has been based on in the past for the default configuration. If that ordering is not someone's preference that person can choose to override the default configuration to suit his or her needs.

otakustay: @joe-denea 

> getDerivedStateFromProps is invoked after a component is instantiated as well as when it receives new props

Since it is a replacement for `willMount` and `willReceiveProps`, it can be confused when being placed at the position of either `willMount` or `willReceiveProps`

I propose we order them like:

```
static displayName = '';

static defaultProps = {
};

... statics

static getDerivedStateFromProps() {
}

state = {
};

... instance fields
```

In this way we can easily have an overview about the structure of our `state` and how our state is synced with our `props`, the entire private state is clear before any life cycle occured

---

For `componentDidCatch`, I don't have a very strong insist, place it after `willUnmount` is OK

otakustay: Currently we may create initial state from props like:

```
constructor(props) {
  super(props);
  this.state = {
    foo: 1,
    bar: props.value
  };
}
```

Do we recommend, after `getDerivedStateFromProps` is available, to change it to:

```
state = {
  foo: 1,
  bar: null
};

static getDerivedStateFromProps(nextProps) {
  return {bar: nextProps.value};
}
```

In case we recommend "splitting self controlled and derived state", I'd prefer to place `getDerivedStateFromProps` after `constructor` so they have a better collabration

joe-denea: Since the super constructor is run prior to instance variable initialization, you can reference `this.props` in the state definition and write what you have above as:

```JavaScript
state = {
  foo: 1,
  bar: this.props.value,
}
```

The code you wrote above with `getDerivedStateFromProps` is not equivalent to the code with `constructor` as it would update the state of `bar` when the `value` prop was updated.

`getDerivedStateFromProps` also receives the previous state as an argument, so it receives the initial state definition as an argument when the initial mount is going to be performed. It would therefore be backwards to the actual ordering of execution to place `getDerivedStateFromProps` prior to the initialization of `state`. That ordering would also be strange as defining instance variables and state as an instance variable is essentially shorthand for what is often done within a constructor in React, so you are proposing that `getDerivedStateFromProps`, a lifecycle method that is triggered on all component mounts and prop updates, be placed between an analogue for the constructor's functionality and the constructor itself.

otakustay: What I mean is that we should recommend the separation of "self controlled state" and those "derived from props" even they can be written in the form of constructor or class property, this may be a best practice:

```js
getDerivedStateFromProps(nextProps, prevState) {
  if (prevState.prevousPropValue === nextProps.value) {
    return {
      bar: value,
      // reset other states
    };
  }
}
```

In most case, we should not have different logics between initial and update state compute, one derived state property should always be synced to props

otakustay: My consideration is we should choose to place `getDerivedStateFromProps` either:

1. right after `constructor` if we think it should be responsible to compute the initial state
2. at the place where `componentWillReceiveProps` was if we recommend to use it only on prop changes (not initial instantiation)

webOS101:
I certainly understand the reasoning and agree that you're probably right about where it belongs.  I'm just concerned about migrating an entire code base to the new rule to take advantage of the new lifecycle methods.  Also, if we had to revert one or the other parts of this PR, they would be tied together and not easy to separate.

docs/rules/sort-comp.md
joe-denea: To be clear I think there is also a use case for the initial state declaration as a set of defaults, like the following

```JavaScript
state = {
  localeString: '0',
}

static getDerivedStateFromProps = (nextProps, prevState) => ({
  localeString: (!Math.isNaN(Number(nextProps.number)) && nextProps.number.toLocaleString())
    || prevState.localeString,
})
```

since this allows us to easily say that the component will convert the prop `number` to a locale string if it is a number and cache the value in state, and fall back to displaying the default of `'0'` or the last valid `number` that we received. In this case even if we wanted to avoid recalculating the value if the prop hadn't changed by caching the unmodified `number` prop in state for comparisons, we would have a sensible default of `0` for it since we wouldn't need to recalculate the default `localeString` if it were `0` initially.

If there is a sensible value for the initial state that isn't derived from props and the code reads clearly I would prefer to avoid setting the initial state based off of props in general, since it is an antipattern as it does not respect prop changes after mounting. I just would not leave out required state values from the initial state declaration as it serves to annotate the expected state fields in a component.

`getDerivedStateFromProps` remains a lifecycle method that is used for updates as well as mounting, and that is fired after the state initialization on the initial mount. If we were to place the state declaration below `getDerivedStateFromProps` in the code above it would appear as though `localeString` would come through as `undefined` in the initial `prevState` based on the ordering, which is not the case.

joe-denea:
I don't believe that there is an auto fix in place for this rule. However, I would argue that the current sorting behavior for `instance-variables` is broken, and that it would be worse to perpetuate the broken pattern in a code base than to migrate over. I see your point that it could be separated out but I would consider the current behavior a bug and if someone wanted to keep the old behavior due to their code base having been written to that standard that person could copy the default ordering and remove `instance-variables` from it in their `eslintrc`. This option seems better to me than potentially exposing new users to broken behavior. Also the addition of `instance-variables` was extremely simple and could be removed very easily if necessary. Mostly I just didn't want to create a potentially broken commit. Just let me know if you want me to pull it out. Thanks for the feedback!

docs/rules/sort-comp.md
ljharb: Put didCatch directly before WillUnmount.

docs/rules/sort-comp.md
```diff
@@ -92,7 +99,7 @@ The default configuration is:
 * `type-annotations`. This group is not specified by default, but can be used to enforce flow annotations positioning.
 * `getters` This group is not specified by default, but can be used to enforce class getters positioning.
 * `setters` This group is not specified by default, but can be used to enforce class setters positioning.
-* `instance-variables` This group is not specified by default, but can be used to enforce all other instance variables positioning.
+* `instance-variables` This group can be used to enforce all other instance variables' positioning.
```
ljharb: specifying this by default makes this a breaking change; please revert that piece.

tests/lib/rules/sort-comp.js
```diff
@@ -91,6 +91,31 @@ ruleTester.run('sort-comp', rule, {
...
+      '  UNSAFE_componentWillUpdate = () => {}',
+      '  getSnapshotBeforeUpdate = () => {}',
+      '  componentDidUpdate = () => {}',
+      '  componentWillUnmount = () => {}',
+      '  componentDidCatch = () => {}',
+      '  myInstanceMethod = () => {}',
+      '  render = () => (<div>Hello</div>)',
+      '}'
+    ].join('\n'),
+    parser: 'babel-eslint'
```
ljharb: Please add a test case that parses with the default parser and also covers all the new lifecycle methods.

joe-denea: As far as I know I don't think I can test `getDerivedStateFromProps` with `React.createClass`. Is there a way I could specify a static method within the class body? I can write one for the other new methods though.

joe-denea:
@ljharb Technically the current behavior catches both arrow functions and regular methods as `instance-methods` interchangeably, and, as I mentioned above, excludes arrow functions from `instance-variables`. I can definitely add a prototype method version as well though. I can also make all of the lifecycle methods prototype methods but technically they can use arrow function syntax and still work so it might be good to test both cases.

tests/lib/rules/sort-comp.js
ljharb: `class Foo {} Foo.getDerivedStateFromProps = …` should work fine outside the class body

ljharb:
Arrow functions are not instance methods though :-/

tests/lib/rules/sort-comp.js
joe-denea: Ya I meant that writing it that way wouldn't test anything with this rule since it wouldn't be sorted within the component body. I checked earlier and if I remember correctly it doesn't get caught by the rule in that case (as one would expect) or in the case where it is placed within `statics` (which I am not sure is for declaring static methods anyway [since I don't have much experience with the `React.createClass` style of React]). I can just leave it out though, or preferably add it after the component definition like you mentioned to make sure that doesn't get caught as an error in future versions.

ljharb: ah true. you could do `static getDerivedStateFromProps() { }` since it's a method tho?

joe-denea: Oh, for some reason I didn't realize the default parser based tests had support for basic ES6 classes.

joe-denea:
I saw that in the error but I wasn't sure if I could add `'use-strict'` anywhere in order to fix the issue. I just attempted a fix by adding `'use-strict'` before the test using an ES6 class on the default parser, but adding it there didn't change anything. I can try removing that test as well to see if the issue is actually related to it.

joe-denea:
Ok it wasn't related to the test, and it in fact seems to be occurring on `npm install` inside of `npm-cli.js`.
```
$ nvm install 4
Downloading and installing node v4.9.1...
Downloading https://nodejs.org/dist/v4.9.1/node-v4.9.1-linux-x64.tar.xz...
Computing checksum with sha256sum
Checksums matched!
Now using node v4.9.1 (npm v2.15.11)
$ node --version
v4.9.1
$ npm --version
2.15.11
$ nvm --version
0.33.9
before_install
$ nvm install-latest-npm
Attempting to upgrade to the latest working version of npm...
* `npm` v2.x needs to first jump to the latest v2 to be able to upgrade further
/home/travis/.nvm/versions/node/v4.9.1/bin/npm -> /home/travis/.nvm/versions/node/v4.9.1/lib/node_modules/npm/bin/npm-cli.js
npm@2.15.12 /home/travis/.nvm/versions/node/v4.9.1/lib/node_modules/npm
* Installing latest `npm`; if this does not work on your node version, please report a bug!
/home/travis/.nvm/versions/node/v4.9.1/bin/npx -> /home/travis/.nvm/versions/node/v4.9.1/lib/node_modules/npm/bin/npx-cli.js
/home/travis/.nvm/versions/node/v4.9.1/bin/npm -> /home/travis/.nvm/versions/node/v4.9.1/lib/node_modules/npm/bin/npm-cli.js
npm@6.0.0 /home/travis/.nvm/versions/node/v4.9.1/lib/node_modules/npm
* npm upgraded to: v
$ npm install 
/home/travis/.nvm/versions/node/v4.9.1/lib/node_modules/npm/bin/npm-cli.js:79
      let notifier = require('update-notifier')({pkg})
```
The above shows up in the logs on the Travis CI build, which makes it seem like NPM has pushed an update that has broken compatibility with Node 4. It seems like `npm` was upgraded to `2.15.12` here but it is hard tell since the "upgraded to" display is broken and has just printed `v`.

joe-denea:
https://github.com/npm/npm/commit/834b46ff48ade4ab4e557566c10e83199d8778c6#diff-fff9baaea61bc82d9999fab7b965272c
This seems to be the commit that introduced the issue.

docs/rules/sort-comp.md
```diff
@@ -70,30 +70,36 @@ The default configuration is:
       'defaultProps',
       'constructor',
       'getDefaultProps',
-      'getInitialState',
```
ljharb: this could technically be a breaking change; is there a reason to swap the order of state and getInitialState?

ljharb:
@joe-denea it's now fixed in nvm itself, on travis - I've reran the errored builds.

docs/rules/sort-comp.md
joe-denea: I changed the order in the documentation to match the actual order in the code, I didn't actually change the order in the code. Also the two are mutually exclusive in actual usage so testing them together would be strange, and I wouldn't be sure how to even write that test considering that `state` only exists as an ES6 `Class` property and `getInitialState` only exists as a `createReactClass` method.

joe-denea:
@ljharb Looks like AppVeyor is still using `npm@6.0.0` and getting the error.

ljharb:
Ah, good call. I'll try to fix that too.

docs/rules/sort-comp.md
ljharb: Makes perfect sense :-) I didn't notice this was just the docs.

yannickcr:
@hamlim yes, it should be released in the next few days.

browne0:
Just curious this was 16 days ago, when will the next release be coming out? 😄

yannickcr:
@browne0 This feature was was released in `7.8.0` 17 days ago 😉

PR #1784, opened 2018-05-11
Target blank dynamic link should fail

kenearley:
Made from the suggestion in @ksdmitrieva's [issue](https://github.com/yannickcr/eslint-plugin-react/issues/1737), I believe this is the right approach even though you would get false positives since it cannot check for relative paths in dynamic links. However, I think that would be rare since it's not best practice to use `target='_blank'` on internal links anyway.

### DIFF
--- a/docs/rules/jsx-no-target-blank.md
+++ b/docs/rules/jsx-no-target-blank.md
@@ -4,14 +4,28 @@ When creating a JSX element that has an `a` tag, it is often desired to have
 the link open in a new tab using the `target='_blank'` attribute. Using this
 attribute unaccompanied by `rel='noreferrer noopener'`, however, is a severe
 security vulnerability ([see here for more details](https://mathiasbynens.github.io/rel-noopener))
-This rules requires that you accompany all `target='_blank'` attributes with `rel='noreferrer noopener'`.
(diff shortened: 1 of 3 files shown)


ljharb:
I think that since it can generate false positives, it should be behind an option.

kenearley:
Makes sense to me. I’ll update the pull request.

kenearley:
@ljharb I've added the option and set "always" as the default since it seems the safer state.

One thing I noticed is that you still would not get the error if you assigned the href using spread operator to assign props. Should that be covered in another option?

ljharb:
I think if you can expand the error so it applies to an inline spread-"href", that would be generally useful?

ljharb:
Yes, that's generally true - but spreading props is an antipattern anyways, so it also might be OK to allow the checks to be bypassed in that case.

ksdmitrieva:
Thanks, @kenearley, for moving this along. The whole idea of the vulnerability is when untrusted link is added without "rel=noopener". When the link is hard-coded, it's most likely trusted. Otherwise, why would you add untrusted links to your application. Untrusted links will most likely come from user input. Anyway, I'm glad this is added.
Interesting edge case with the spread operator. If it's an antipattern, I wonder if there is a rule that highlights that usage.

PR #1792, opened 2018-05-12
Fix crash in boolean-prop-naming when encountering a required shape p…

### LINKED ISSUE #1791: react/boolean-prop-naming error with isRequired on a shape
When upgrading to 7.8 from 7.7.0 eslint trigger the following error:

```bash
Cannot read property 'name' of undefined
TypeError: Cannot read property 'name' of undefined
    at getPropKey (/usr/app/node_modules/eslint-plugin-react/lib/rules/boolean-prop-naming.js:79:45)
    at forEach.prop (/usr/app/node_modules/eslint-plugin-react/lib/rules/boolean-prop-naming.js:114:25)
    at validatePropNaming (/usr/app/node_modules/eslint-plugin-react/lib/rules/boolean-prop-naming.js:113:25)
    at Object.ClassProperty (/usr/app/node_modules/eslint-plugin-react/lib/rules/boolean-prop-naming.js:175:11)
    at updatedRuleInstructions.(anonymous function) (/usr/app/node_modules/eslint-plugin-react/lib/util/Components.js:698:75)
    at listeners.(anonymous function).forEach.listener (/usr/app/node_modules/eslint/lib/util/safe-emitter.js:47:58)
    at Object.emit (/usr/app/node_modules/eslint/lib/util/safe-emitter.js:47:38)
    at NodeEventGenerator.applySelector (/usr/app/node_modules/eslint/lib/util/node-event-generator.js:251:26)
    at NodeEventGenerator.applySelectors (/usr/app/node_modules/eslint/lib/util/node-event-generator.js:280:22)
    at NodeEventGenerator.enterNode (/usr/app/node_modules/eslint/lib/util/node-event-generator.js:294:14)
    at CodePathAnalyzer.enterNode (/usr/app/node_modules/eslint/lib/code-path-analysis/code-path-analyzer.js:608:23)
    at Traverser.enter [as _enter] (/usr/app/node_modules/eslint/lib/linter.js:865:28)
    at Traverser._traverse (/usr/app/node_modules/eslint/lib/util/traverser.js:132:14)
    at Traverser._traverse (/usr/app/node_modules/eslint/lib/util/traverser.js:144:34)
```

I was able to reproduce with this simple code:
```js
import PropTypes from 'prop-types'
import React from 'react'


  static propTypes = {
    requiredObject: PropTypes.shape({}).isRequired,
  }

  render() {
    return null
  }
}
```

I believe this is due to https://github.com/yannickcr/eslint-plugin-react/commit/f980c3808460a2b983a00ce320bb5d6bd52760ae introducing a check for simple `PropTypes.bool.isRequired` but not handling the case where `isRequired` is added after a function.

pcorpet:
…rop type. Fixes #1791.

### DIFF
--- a/lib/rules/boolean-prop-naming.js
+++ b/lib/rules/boolean-prop-naming.js
@@ -76,7 +76,10 @@ module.exports = {
       if (node.value.property) {
         const name = node.value.property.name;
         if (name === 'isRequired') {
-          return node.value.object.property.name;
+          if (node.value.object && node.value.object.property) {
+            return node.value.object.property.name;
+          }
+          return null;
         }
         return name;
       }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
+## [Unreleased]
+### Fixed
+* Fix crash in [`boolean-prop-naming`] when encountering a required shape prop type ([#1791][] @pcorpet)
+
 ## [7.8.1] - 2018-05-12
 ### Fixed
 * Fix crash in [`no-deprecated`][] when encountering a class constructor ([#1785][] @taddei)
--- a/tests/lib/rules/boolean-prop-naming.js
+++ b/tests/lib/rules/boolean-prop-naming.js
@@ -348,6 +348,17 @@ ruleTester.run('boolean-prop-naming', rule, {
         hasValue: PropTypes.bool.isRequired
       }
     `
+  }, {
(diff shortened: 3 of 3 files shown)


lib/rules/boolean-prop-naming.js
```diff
@@ -76,7 +76,10 @@ module.exports = {
       if (node.value.property) {
         const name = node.value.property.name;
         if (name === 'isRequired') {
-          return node.value.object.property.name;
+          if (node.value.object.property) {
```
ljharb: what if `node.value.object` is falsy?

pcorpet: I couldn't find a case where node.value has a property but no object. But I added a guard anyways.

yannickcr:
Merged, thanks !

PR #1794, opened 2018-05-14
Add a when not to use case to forbid-foreign-prop-types

lencioni:


docs/rules/forbid-foreign-prop-types.md
```diff
@@ -28,3 +28,5 @@ import SomeComponent, {propTypes as someComponentPropTypes} from './SomeComponen
 ## When not to use
 
 This rule aims to make a certain production optimization, removing prop types, less prone to error. This rule may not be relevant to you if you do not wish to make use of this optimization.
+
+If you are writing a higher-order component that hoists the wrapped component's propTypes, you might want to disable this rule.
```
ljharb: Should we add something about doing it safely, so the HOC code path works both with and without propTypes present?

lencioni: Yes, that might be a good idea. That can be added in a separate PR though. :D

PR #1795, opened 2018-05-14
sort-comp: enforcing static lifecycle methods order

lynxtaa:
Fixes bug: https://github.com/yannickcr/eslint-plugin-react/issues/1793

Example of **incorrect** code with `"react/sort-comp": "error"`:
```jsx
class Test extends React.Component {
	static getDerivedStateFromProps() {
		return null
	}

	constructor(props) {
		super(props)
		this.state = { test: 'test' }
	}

	render() {
		return this.state.test
	}
}
```

Example of **correct** code with `"react/sort-comp": "error"`:
```jsx
class Test extends React.Component {
	constructor(props) {
		super(props)
		this.state = { test: 'test' }
	}

	static getDerivedStateFromProps() {
		return null
	}

	render() {
		return this.state.test
	}
}
```

### DIFF
--- a/lib/rules/sort-comp.js
+++ b/lib/rules/sort-comp.js
@@ -172,7 +172,7 @@ module.exports = {
         }
       }
 
-      if (method.static) {
+      if (method.static && methodsOrder.indexOf(method.name) === -1) {
         const staticIndex = methodsOrder.indexOf('static-methods');
         if (staticIndex >= 0) {
           indexes.push(staticIndex);
--- a/lib/rules/sort-comp.js
+++ b/lib/rules/sort-comp.js
@@ -172,7 +172,7 @@ module.exports = {
         }
       }
 
-      if (method.static && methodsOrder.indexOf(method.name) === -1) {
+      if (method.static && method.name !== 'getDerivedStateFromProps') {
         const staticIndex = methodsOrder.indexOf('static-methods');
         if (staticIndex >= 0) {
           indexes.push(staticIndex);
--- a/lib/rules/sort-comp.js
+++ b/lib/rules/sort-comp.js
@@ -172,7 +172,7 @@ module.exports = {
         }
       }
 
-      if (method.static && method.name !== 'getDerivedStateFromProps') {
+      if (indexes.length === 0 && method.static) {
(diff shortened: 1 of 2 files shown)


lib/rules/sort-comp.js
```diff
@@ -172,7 +172,7 @@ module.exports = {
         }
       }
 
-      if (method.static) {
+      if (method.static && method.name !== 'getDerivedStateFromProps') {
```
joe-denea: We might want to just add `indexes.length === 0` or something like that here so all named methods would be caught here rather than just `getDerivedStateFromProps` for future extensibility.

lynxtaa: Thanks, checking for `indexes.length === 0` will do the job. Fixed

tests/lib/rules/sort-comp.js
```diff
@@ -504,6 +504,16 @@ ruleTester.run('sort-comp', rule, {
...
     errors: [{message: 'render should be placed after displayName'}]
+  }, {
+    // Must validate static lifecycle methods
+    code: [
+      'class Hello extends React.Component {',
+      '  static getDerivedStateFromProps() {}',
+      '  constructor() {}',
+      '}'
+    ].join('\n'),
+    parser: 'babel-eslint',
```
ljharb: the default parser should be sufficient here; please either add a duplicate test (one using default, one using babel-eslint) or, since this is all ES6 syntax, just have the one using the default parser.

ThiefMaster:
Does this really make sense? Usually people group static methods together, and put them before the constructor since they are not related to a specific instance.

ThiefMaster:
Also it's impossible to remove something from the lifecycle list without hardcoding the whole list in my config (and thus having to maintain it whenever react adds new lifecycle methods)...

edmorley:
This appears to have caused #1821. Any ideas as to how to fix that case?

lynxtaa:
Static method declarations should be on top but it's not the case for `getDerivedStateFromProps`.

It was done that way because `getDerivedStateFromProps` is a lifecycle method (according to current rule description https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md#rule-details).

PR #1796, opened 2018-05-14
guard against accidental anonymous react imports

jsg2021:
There was a file in an old code base that improperly imported react anonymously...and had been unnoticed until we added this no-typos rule and eslint crashed. :)


jsg2021:
Surething.

PR #1825, opened 2018-06-14
Allow LHS in destructuring-assignment

### LINKED ISSUE #1728: Rule react/destructruring-assignement in a constructor
The rule `react/destructruring-assignement` should not throw in a component constructor.

image

alexzherdev:
Resolves #1728

Not sure if this needs to also check that assignment is in the constructor? Also, would the following do anything useful / should this be allowed?
```js
constructor(props, context) {
  this.context.foo = 'blah';
}
```

### DIFF
--- a/tests/lib/rules/destructuring-assignment.js
+++ b/tests/lib/rules/destructuring-assignment.js
@@ -133,8 +133,7 @@ ruleTester.run('destructuring-assignment', rule, {
         this.state.foo = 'bar';
       }
     };`,
-    options: ['always'],
-    parser: 'babel-eslint'
+    options: ['always']
   }],
 
   invalid: [{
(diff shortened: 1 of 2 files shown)


tests/lib/rules/destructuring-assignment.js
```diff
@@ -126,7 +126,17 @@ ruleTester.run('destructuring-assignment', rule, {
...
     parser: 'babel-eslint'
+  }, {
+    code: `const Foo = class extends React.PureComponent {
+      constructor() {
+        this.state = {};
+        this.state.foo = 'bar';
+      }
+    };`,
+    options: ['always'],
+    parser: 'babel-eslint'
```
ljharb: this test case doesn't require babel-eslint, let's remove this line

PR #1827, opened 2018-06-14
Fix static propTypes handling in no-typos

### LINKED ISSUE #1677: no-typos doesn't detect invalid PropTypes in static class fields
I've observed this with versions 7.5.1 & 7.6.1. The [docs](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-typos.md) indicate that the following should cause the following warning: `Typo in declared prop type: typo` however it doesn't.

```js
  static propTypes = {
    a: PropTypes.typo
  }
}
```

However, when you don't use static class fields, it produces the expected warning:

```js
class MyComponent extends React.Component {}
MyComponent.propTypes = {
  a: PropTypes.typo
};
```

alexzherdev:
Resolves #1677

### DIFF
--- a/lib/rules/no-typos.js
+++ b/lib/rules/no-typos.js
@@ -126,8 +126,7 @@ module.exports = {
 
     function reportErrorIfClassPropertyCasingTypo(node, propertyName) {
       if (propertyName === 'propTypes' || propertyName === 'contextTypes' || propertyName === 'childContextTypes') {
-        const propsNode = node && node.parent && node.parent.type === 'AssignmentExpression' && node.parent.right;
-        checkValidPropObject(propsNode);
+        checkValidPropObject(node);
       }
       STATIC_CLASS_PROPERTIES.forEach(CLASS_PROP => {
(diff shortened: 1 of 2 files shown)


ljharb:
(if those test cases already exist - because the bug is about class fields - then my bad)

ljharb:
That’s be great - generally it’s nice to have every test case repeated, one for each parser.

lib/rules/no-typos.js
```diff
@@ -45,7 +45,7 @@ module.exports = {
     let propTypesPackageName = null;
     let reactPackageName = null;
 
-    function checkValidPropTypeQualfier(node) {
+    function checkValidPropTypeQualifier(node) {
```
alexzherdev: Someone put typos in `no-typos` so I can fix typos while I fix `no-typos`.

EvHaus: 0ca36be035e2c8c79d6af4543c7a6990

PR #1828, opened 2018-06-15
Support createClass in no-typos

### LINKED ISSUE #1721: no-typos: Fix failing createClass test
Per #1671 (https://github.com/yannickcr/eslint-plugin-react/pull/1671/files#diff-a36b384ace30edac8eacb00c170eed54R1277) and #1389, there's a commented out no-typos test that should be uncommented and fixed.

alexzherdev:
Resolves #1721 

Also added some more tests for good measure.
Not sure about the exact message for an ES5 component, but mentioning `static class property` didn't make sense.

### DIFF
--- a/lib/rules/no-typos.js
+++ b/lib/rules/no-typos.js
@@ -124,15 +124,18 @@ module.exports = {
       }
     }
 
-    function reportErrorIfClassPropertyCasingTypo(node, propertyName) {
+    function reportErrorIfPropertyCasingTypo(node, propertyName, isClassProperty) {
       if (propertyName === 'propTypes' || propertyName === 'contextTypes' || propertyName === 'childContextTypes') {
         checkValidPropObject(node);
       }
       STATIC_CLASS_PROPERTIES.forEach(CLASS_PROP => {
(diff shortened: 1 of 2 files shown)


lib/rules/no-typos.js
```diff
@@ -124,15 +124,18 @@ module.exports = {
...
 
-    function reportErrorIfClassPropertyCasingTypo(node, propertyName) {
+    function reportErrorIfPropertyCasingTypo(node, propertyName, isClassProperty) {
       if (propertyName === 'propTypes' || propertyName === 'contextTypes' || propertyName === 'childContextTypes') {
         checkValidPropObject(node);
       }
       STATIC_CLASS_PROPERTIES.forEach(CLASS_PROP => {
         if (propertyName && CLASS_PROP.toLowerCase() === propertyName.toLowerCase() && CLASS_PROP !== propertyName) {
+          const message = isClassProperty
+            ? 'Typo in static class property declaration'
```
ljharb: i think it's fine to have different messages, but "static class property" is a conceptual term prior to the feature landing in the language, so i think that'd be reasonable.

Issue #1830, opened 2018-06-17
Add `no-unsafe` rule


sergei-startsev:
Certain legacy lifecycle methods are unsafe for use in async React applications and cause warnings in [_strict mode_](https://reactjs.org/docs/strict-mode.html#identifying-unsafe-lifecycles):
- `UNSAFE_componentWillMount`
- `UNSAFE_componentWillReceiveProps`
- `UNSAFE_componentWillUpdate`.

These also happen to be the lifecycles that cause the most [confusion within the React community](https://reactjs.org/blog/2018/03/29/react-v-16-3.html#component-lifecycle-changes). At the same time there're no deprecation warnings regarding using `UNSAFE_` methods and [it will still be possible to use them in React 17](https://reactjs.org/blog/2018/03/29/react-v-16-3.html#component-lifecycle-changes).
 
The rule prevents usage of of `UNSAFE_` lifecycle methods. See discussion details https://github.com/yannickcr/eslint-plugin-react/pull/1750#discussion_r194915657.

sergei-startsev:
The rule has been merged.

PR #1831, opened 2018-06-17
Added `no-unsafe` rule

sergei-startsev:
Added  `no-unsafe` rule to address the corresponding issue #1830. See discussion details https://github.com/yannickcr/eslint-plugin-react/pull/1750#discussion_r194915657.

### DIFF
--- a/index.js
+++ b/index.js
@@ -64,6 +64,7 @@ const allRules = {
   'no-typos': require('./lib/rules/no-typos'),
   'no-unescaped-entities': require('./lib/rules/no-unescaped-entities'),
   'no-unknown-property': require('./lib/rules/no-unknown-property'),
+  'no-unsafe': require('./lib/rules/no-unsafe'),
   'no-unused-prop-types': require('./lib/rules/no-unused-prop-types'),
   'no-unused-state': require('./lib/rules/no-unused-state'),
   'no-will-update-set-state': require('./lib/rules/no-will-update-set-state'),
@@ -139,6 +140,7 @@ module.exports = {
         'react/no-string-refs': 2,
         'react/no-unescaped-entities': 2,
         'react/no-unknown-property': 2,
+        'react/no-unsafe': 2,
         'react/prop-types': 2,
         'react/react-in-jsx-scope': 2,
         'react/require-render-return': 2
--- a/index.js
+++ b/index.js
@@ -140,7 +140,7 @@ module.exports = {
         'react/no-string-refs': 2,
         'react/no-unescaped-entities': 2,
         'react/no-unknown-property': 2,
-        'react/no-unsafe': 2,
+        'react/no-unsafe': 0,
         'react/prop-types': 2,
         'react/react-in-jsx-scope': 2,
         'react/require-render-return': 2
--- a/lib/rules/no-unsafe.js
+++ b/lib/rules/no-unsafe.js
@@ -26,6 +26,11 @@ module.exports = {
   },
 
   create: Components.detect((context, components, utils) => {
+    const isApplicable = versionUtil.testReactVersion(context, '16.3.0');
+    if (!isApplicable) {
+      return {};
+    }
+
     /**
      * Returns a list of unsafe methods
      * @returns {Array} A list of unsafe methods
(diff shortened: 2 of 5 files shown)


lib/rules/no-unsafe.js
```diff
@@ -0,0 +1,93 @@
...
+     * @param {string} method Life cycle method
+     */
+    function checkUnsafe(node, method) {
+      if (!isUnsafe(method)) {
+        return;
+      }
+
+      context.report({
+        node: node,
+        message: `Do not use ${method}`
```
ljharb: the message should explain why, and link to more info

tests/lib/rules/no-unsafe.js
```diff
@@ -0,0 +1,107 @@
...
+      code: `
+      class Foo extends React.Component {
+        UNSAFE_componentWillMount() {}
+        UNSAFE_componentWillReceiveProps() {}
+        UNSAFE_componentWillUpdate() {}
+      }
+    `,
+      errors: [
+        {
+          message: 'Do not use UNSAFE_componentWillMount'
```
ljharb: these tests should include position information, so that we can test that part as well

lib/rules/no-unsafe.js
sergei-startsev: Added details

tests/lib/rules/no-unsafe.js
sergei-startsev: Added a few more tests and position details for invalid cases

sergei-startsev:
The rule is applicable for 16.3.0+ only now, good point.

tests/lib/rules/no-unsafe.js
sergei-startsev: It seems that added position details causes issues in eslint 3, see `TEST=true ESLINT=3` failed travis builds.

lib/rules/no-unsafe.js
```diff
@@ -44,7 +45,8 @@ module.exports = {
      */
     function isUnsafe(method) {
       const unsafeMethods = getUnsafeMethods();
-      return unsafeMethods.indexOf(method) !== -1;
+      const isApplicable = versionUtil.testReactVersion(context, '16.3.0');
```
ljharb: This seems like a check we could do earlier, and prevent even returning any visitors in the first place?

sergei-startsev: Added early termination for the rule.

tests/lib/rules/no-unsafe.js
sergei-startsev: Adjusted tests to support ESLint 3

index.js
```diff
@@ -139,6 +140,7 @@ module.exports = {
         'react/no-string-refs': 2,
         'react/no-unescaped-entities': 2,
         'react/no-unknown-property': 2,
+        'react/no-unsafe': 2,
```
ljharb: enabling new rules in an exported config is a semver-major change; please set this to 0 for now.

lib/rules/no-unsafe.js
```diff
@@ -0,0 +1,99 @@
...
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Prevent usage of UNSAFE_ methods',
+      category: 'Best Practices',
+      recommended: true,
```
ljharb: ditto

sergei-startsev:
Excluded the rule from the recommended for now. Would we like to track it somehow to turn on it in the next major release?

EvHaus:
Why not built this logic into the existing **[no-deprecated](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-deprecated.md)** rule as opposed to creating a new dedicated rule?

ljharb:
I don't consider them deprecated yet, because there's not a replacement for all of their use cases.

sergei-startsev:
@EvHaus `UNSAFE_` methods aren't deprecated, so `no-deprecated` rule isn't the right place for that. See discussion details #1750 (comment).

PR #1843, opened 2018-06-23
Add support for eslint 5

papandreou:
Looks like it just works.

The peer dep change sees to it that this plugin can be used in projects that use eslint 5 and npm 2 (the default in node 4) without `npm install` failing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ env:
   global:
     - TEST=true
   matrix:
+    - ESLINT=5
     - ESLINT=4
     - ESLINT=3
 after_success:
@@ -28,8 +29,12 @@ matrix:
   include:
     - node_js: 'lts/*'
       env: PRETEST=true
+  exclude:
+    - node_js: '5'
+      env: ESLINT=5
+    - node_js: '4'
+      env: ESLINT=5
   allow_failures:
     - node_js: '9'
     - node_js: '7'
     - node_js: '5'
-    - env: TEST=true ESLINT=next
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,6 +19,7 @@ matrix:
     - nodejs_version: "9"
     - nodejs_version: "7"
     - nodejs_version: "5"
+    - nodejs_version: "4"
 
 platform:
   - x86
--- a/package.json
+++ b/package.json
@@ -32,12 +32,12 @@
   "devDependencies": {
     "babel-eslint": "^8.2.3",
     "coveralls": "^3.0.1",
-    "eslint": "^4.19.1",
+    "eslint": "^3.0.0 || ^4.0.0 || ^5.0.0",
     "istanbul": "^0.4.5",
     "mocha": "^5.2.0"
   },
   "peerDependencies": {
-    "eslint": "^3.0.0 || ^4.0.0"
+    "eslint": "^3.0.0 || ^4.0.0 || ^5.0.0"
   },
   "engines": {
     "node": ">=4"


papandreou:
The test failure also happens on master and is due to babel-eslint@8.2.4. Downgrading to 8.2.3 fixes it.

`git bisect` says:

```
e63962ddd340e3a412ee33d9f0ed443a88d3975f is the first bad commit
commit e63962ddd340e3a412ee33d9f0ed443a88d3975f
Author: Cristian Pallarés <cristian@pallares.io>
Date:   Wed Jun 6 00:18:13 2018 +0200

    refactor: rename babylon to @babel/parser

:040000 040000 c6b299854dde266cc8960551ab1ee334fd6f5e93 d08c0f67d7b0d9b795dbb8a8702ca787c3318182 M	lib
```

Perhaps the easiest fix is to just drop node 4 support now that everyone else in the eslint ecosystem is doing it.

existentialism:
@papandreou just released 8.2.5, which reverts the babel 7-b51 bump.

papandreou:
@ljharb, thanks for the quick response!

> Please also add testing for eslint 4 vs 5 here: https://github.com/yannickcr/eslint-plugin-react/blob/master/.travis.yml

Okay, do you mean like this? https://github.com/yannickcr/eslint-plugin-react/pull/1843/commits/d71706868923a4ff1a0bb772aca90f18530e7cc2

... I guess the exact configuration depends on your answer to the below question :)

> Also, please make the dev dep range for eslint identical to the peer dep range.

I tried that initially, but it makes the build fail in node 4: https://travis-ci.org/yannickcr/eslint-plugin-react/jobs/395825658#L856

Are you ready to part with node 4 support -- seems like everyone else is :)

.travis.yml
```diff
@@ -23,6 +23,14 @@ matrix:
...
     - node_js: 'lts/*'
       env: PRETEST=true
+    - node_js: '9'
+      env: TEST=true ESLINT=5
+    - node_js: '8'
+      env: TEST=true ESLINT=5
+    - node_js: '7'
+      env: TEST=true ESLINT=5
+    - node_js: '6'
+      env: TEST=true ESLINT=5
```
ljharb: sorry for being unclear :-) now we’re not testing eslint 4. Can you change al the “next”s to “4”s?

ljharb:
As for node 4, no, we should not drop support for it (nobody should). I’ll look into the failing tests.

.travis.yml
papandreou: Sure, fixed in d0bebd7da5a1805e57be54106c9b575e37a637da :)

papandreou:
@ljharb, thanks a lot!

I agree, but eslint 5 dropped node 4 support: https://eslint.org/docs/5.0.0/user-guide/migrating-to-5.0.0#drop-node-4

Would be great for this project to maintain backwards compatibility via older eslint versions, of course.

.travis.yml
```diff
@@ -47,4 +55,4 @@ matrix:
     - node_js: '9'
     - node_js: '7'
     - node_js: '5'
-    - env: TEST=true ESLINT=next
+    - env: TEST=true ESLINT=4
```
ljharb: let’s remove this line

ljharb:
ahhh well true, in that case for eslint 5 only, we should definitely not be testing on node 4.

.travis.yml
papandreou: Whatever you say, fixed in 61372b88cec82c453d65091068b3d37afe3d4d25 :)

papandreou:
> ahhh well true, in that case for eslint 5 only, we should definitely not be testing on node 4.

Sounds like the right way to handle it.
It's probably faster for you to set that up? :)

ljharb:
Sounds good, I’ll update this PR later today.

papandreou:
@ljharb, thanks for landing this! Would it be possible to release a new version to the npm registry?

ljharb:
I’m hoping to do that tonight or tomorrow.

ljharb:
Released in v7.10.0

PR #1845, opened 2018-06-24
Account for UNSAFE_ method in no-will-update-set-state

### LINKED ISSUE #1844: no-will-update-set-state not reporting for UNSAFE_ version
the rule no-will-update-set-state

stops reporting problems if you rename to the UNSAFE_ version of no will update set state

alexzherdev:
Resolves #1844

### DIFF
--- a/lib/util/makeNoMethodSetStateRule.js
+++ b/lib/util/makeNoMethodSetStateRule.js
@@ -10,7 +10,7 @@ const docsUrl = require('./docsUrl');
 // Rule Definition
 // ------------------------------------------------------------------------------
 
-function makeNoMethodSetStateRule(methodName) {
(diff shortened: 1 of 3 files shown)


lib/util/makeNoMethodSetStateRule.js
```diff
@@ -28,6 +28,15 @@ function makeNoMethodSetStateRule(methodName) {
     create: function(context) {
       const mode = context.options[0] || 'allow-in-func';
 
+      function nameMatches(name) {
+        let matches = name === methodName;
```
ljharb: This seems like it could return immediately if true, cleaning up the rest of the function

alexzherdev:
Ah, thanks!

PR #1854, opened 2018-06-26
For deprecated lifecycle methods, report identifier AST node instead of the class node

jsnajdr:
When the `no-deprecated` rule finds a deprecated lifecycle method, it reports the whole class declaration as the AST node for the error. Doesn't look good in editor: makes the whole class declaration red and obscures all other ESLint errors inside the class.

screen shot 2018-06-26 at 11 53 22

This patch changes the reported AST node to the method name identifier. The errors are much better targeted at the infringing code and look much better when shown in editor UI.

screen shot 2018-06-26 at 11 29 01

Added also test coverage to check for the reported AST node type.

### DIFF
--- a/lib/rules/no-deprecated.js
+++ b/lib/rules/no-deprecated.js
@@ -106,19 +106,19 @@ module.exports = {
       );
     }
 
-    function checkDeprecation(node, method) {
-      if (!isDeprecated(method)) {
+    function checkDeprecation(node, methodName, methodNode) {
+      if (!isDeprecated(methodName)) {
         return;
       }
       const deprecated = getDeprecated();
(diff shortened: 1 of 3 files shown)


jsnajdr:
@ljharb Thanks for reviewing this patch! I added a commit with line/column/endLine/endColumn checks.

ljharb:
@jsnajdr looks like eslint 3 is failing on the `endLine` parts?

jsnajdr:
Oops, ESLint 3 apparently didn't support the `endLine` and `endColumn` fields yet. I removed them from the tests, leaving only `line` and `column`.

PR #1857, opened 2018-06-27
Output a warning if React version is missing in settings

### LINKED ISSUE #1789: React 16 lifecycle errors appearing in a React 15 project
I'm getting `react/no-deprecated componentWillReceiveProps` errors in a React 15 project 😢 

- Build log: https://travis-ci.org/swagger-api/swagger-ui/builds/378016142#L506
- [`react@15.6.2`](https://github.com/swagger-api/swagger-ui/blob/master/package.json#L63), [`eslint-plugin-react@^7.1.0`](https://github.com/swagger-api/swagger-ui/blob/master/package.json#L118)

alexzherdev:
Resolves #1789

This message will appear when running some of the tests. Not sure if that should be avoided.

### DIFF
--- a/lib/util/log.js
+++ b/lib/util/log.js
@@ -1,0 +1,14 @@
+'use strict';
+
+/**
+ * Logs out a message if there is no format option set.
+ * @param {String} message - Message to log.
+ */
+function log(message) {
+  if (!/\=-(f|-format)=/.test(process.argv.join('='))) {
+    // eslint-disable-next-line no-console
+    console.log(message);
+  }
+}
+
+module.exports = log;
--- a/lib/rules/jsx-space-before-closing.js
+++ b/lib/rules/jsx-space-before-closing.js
@@ -7,6 +7,7 @@
 
 const getTokenBeforeClosingBracket = require('../util/getTokenBeforeClosingBracket');
 const docsUrl = require('../util/docsUrl');
+const log = require('../util/log');
 
 let isWarnedForDeprecation = false;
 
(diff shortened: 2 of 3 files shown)


lib/util/log.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+/**
+ * Logs out a message if there is no format option set.
+ * @param {String} message - Message to log.
+ */
+function log(message) {
+  if (!/\=-(f|-format)=/.test(process.argv.join('='))) {
+    /* eslint-disable no-console */
```
ljharb: i think we can use `// eslint-disable-next-line no-console` here

kud:
Another idea is to add "latest" as react version. :D

ljharb:
@kud i could see adding a separate option for "detect" (using the "version" as a fallback), which did `require('react').version` - want to open an issue or PR that adds that?

ljharb:
(Note that it'd have to use `process.cwd()`, and the `resolve` package, to get the proper require path)

alexzherdev:
I think I got the "detect" option to work. I found something I didn't expect though: when extending airbnb's config (as opposed to the recommended config from the plugin), there is a version there in `context.settings` even if you don't set it in `.eslintrc`. I believe it comes from here https://github.com/airbnb/javascript/blob/02b4eea3452b0ae6da666b01f9da23f26148f5b6/packages/eslint-config-airbnb/rules/react.js#L464
Is this intended? I found this surprising and non-obvious 🤔

ljharb:
Yes, absolutely it's intended.

alexzherdev:
Oh. Is this just the version used at Airbnb?
@ljharb you mentioned elsewhere you would make not specifying the version throw. But then (correct me if I'm wrong) if most of plugin's users use it via extending `airbnb`, they would never observe that because of this implicit (for them) setting.

ljharb:
Yes, that is true. However, with that behavior, the airbnb config would probably remove that setting.

With "detect", it'd probably use that, though.

PR #1858, opened 2018-06-27
sort-comp Allow methods to belong to any matching group.

### LINKED ISSUE #1808: [sort-comp] static getter should be treated as static method over common getter
Configuration (airbnb):

```js
{
  rules: {
    'react/sort-comp': ['error', {
      order: [
        'static-methods',
        'instance-variables',
        'lifecycle',
        '/^on.+$/',
        'getters',
        'setters',
        '/^(get|set)(?!(InitialState$|DefaultProps$|ChildContext$)).+$/',
        'instance-methods',
        'everything-else',
        'rendering',
      ],
  }
}
```

Here is a simple example: 

```jsx
import React from 'react';
import cx from 'classnames';

export class Icon extends React.Component {
  static propTypes = {
    name: PropTypes.string,
  };

  static get defaultProps() { // LINT ERROR
    return {
      name: 'defaultIcon',
    };
  }

    return this.props.name === nextProps.name;
  }

  render() {
    return (
      <i // eslint-disable-line jsx-a11y/no-static-element-interactions,jsx-a11y/click-events-have-key-events
        className={cx('s', `s-${this.props.name}`)}
      />
    );
  }
}
```

It causes an error on `static get defaultProps()` and error message is `getter functions should be placed after shouldComponentUpdate [react/sort-comp]`.

### LINKED ISSUE #1821: sort-comp wrong order with RegExp rules and static methods
I believe this issues comes from #1795 

The Airbnb rules for react include `sort-comp` with regex matching.  [source here](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/rules/react.js#L262)

If you have a static method that matches any of the 3 RegExp values (e.g. `renderCell`) then [this will push to indexes](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/sort-comp.js#L161), and [the following check for `method.static`](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/sort-comp.js#L175) will not happen because of the new conditional on empty `indexes`.

This should be valid according to the Airbnb Rules.

```js
class Row extends React.Component {
  static getDelta(data) {
    ...
  }

  static onTimeout() {
    ...
  }

  static renderCell(data) {
    ...
  }

  constructor(props) {
    ...
  }

  render() {
    ...
  }
}
```

### LINKED ISSUE #1838: sort-comp reorders static propTypes
Hello!

I recently started receiving errors about my `static propTypes` being out of order. It looks like with the grouping logic, all `static` variables must be in alphabetical order. This causes my prop types to be forced down.

I've gotten around this by adding the following regex to my config which will push displayName, propTypes, and defaultProps to the top while allowing all other static methods to be below them.

```JS
"react/sort-comp": [
      "error",
      {
        order: [
          'static-methods',
          'lifecycle',
          'everything-else',
          'render'
        ],
        groups: {
          'static-methods': [
            'displayName',
            'propTypes',
            'defaultProps',
            '/^((?!propTypes|defaultProps).)*$/',
          ],
          lifecycle: [
            'displayName',
            'propTypes',
            'contextTypes',
            'childContextTypes',
            'mixins',
            'statics',
            'defaultProps',
            'constructor',
            'getDefaultProps',
            'state',
            'getInitialState',
            'getChildContext',
            'getDerivedStateFromProps',
            'componentWillMount',
            'UNSAFE_componentWillMount',
            'componentDidMount',
            'componentWillReceiveProps',
            'UNSAFE_componentWillReceiveProps',
            'shouldComponentUpdate',
            'componentWillUpdate',
            'UNSAFE_componentWillUpdate',
            'getSnapshotBeforeUpdate',
            'componentDidUpdate',
            'componentDidCatch',
            'componentWillUnmount'
          ]
        }
      }
    ]
```

example error:
```
  13:3  error  propTypes should be placed after determineDropdownValue  react/sort-comp
``

### PR DESCRIPTION
**Pull request by nosilleg**
Fixes #1821. Fixes #1838. Fixes #1808. Also fixes a no-issue bug (see bottom in **bold**).

The history of the rule implies that it was intended that methods matching multiple groups should be able to be placed in any of those groups.  See [initial commit](https://github.com/yannickcr/eslint-plugin-react/commit/8b93659ded2ec0fc2278d68b700338625bbcfda7#diff-ced56472004c220d7b811bb48e0bacccR146) of the rule.

It seems over time that restrictions have come into the code that prevent this.   I didn't find any discussions about this change in behavior.  This change has lead to multiple bugs because the config isn't refined enough to accommodate individual preferences. 

This PR solves those bugs by reintroducing the ability for methods to be grouped with any matching group.

For example:

Static methods that match a RegExp can be placed with `static-methods` or with the RegExp group.
Static lifecycle methods can be placed with `static-methods` or with lifecycle methods.
Static getters can be placed with `static-methods` or `getters`.

This allows teams with different preferences to put the method where they think is appropriate, but it means that individuals within a team can still place methods in locations that the team doesn't like and this will need to be a manual check on code review.

This seems like the best compromise until team preferences can be fully supported by the rules options.

About the implementation:

The methodOrder is now processed in array order.  This was originally done because I took the approach that the methodOrder should determine stricter behavior.  While I no longer believe that to be the solution, I believe this code layout is more readable than previously, and there are fewer loops now than previously.  (Previously not all loops were executed, but that would have had to change in this PR.)

I moved the test descriptions into the `code` blocks, so that when there is a test error you actually know which test has failed.  (I didn't see this information otherwise.)

I added two additional tests to cover existing functionality that needed to be verified with the changes in this PR.

The first is that the existing code allows explicitly named functions in the group list.  I do not see this in the documentation and there were no tests for it.  A decision can be made if this needs to be added to the documentation, or if it's a bug, but the test is there to make sure the behavior isn't accidentally changed.

The second is that method names that match groups names were being forced into those groups.  For example, a method named `setters` would be forced into the `setters` group.  **I consider this a bug, and have corrected it in this PR.**  However, I'm willing to revert to existing behavior if it's deemed appropriate.

### DIFF
--- a/lib/rules/sort-comp.js
+++ b/lib/rules/sort-comp.js
@@ -8,6 +8,7 @@ const has = require('has');
 const util = require('util');
 
 const Components = require('../util/Components');
+const arrayIncludes = require('array-includes');
 const astUtil = require('../util/ast');
 const docsUrl = require('../util/docsUrl');
 
@@ -158,7 +159,7 @@ module.exports = {
           if (method.instanceMethod) {
             methodGroupIndexes.push(groupIndex);
           }
-        } else if ([
+        } else if (arrayIncludes([
           'displayName',
           'propTypes',
           'contextTypes',
@@ -185,7 +186,7 @@ module.exports = {
           'componentDidCatch',
           'componentWillUnmount',
           'render'
-        ].includes(currentGroup)) {
+        ], currentGroup)) {
           if (currentGroup === method.name) {
             methodGroupIndexes.push(groupIndex);
           }
--- a/lib/rules/sort-comp.js
+++ b/lib/rules/sort-comp.js
@@ -134,86 +134,71 @@ module.exports = {
       const methodGroupIndexes = [];
 
       methodsOrder.forEach((currentGroup, groupIndex) => {
-        switch (currentGroup) {
-          case 'getters': {
-            if (method.getter) {
-              methodGroupIndexes.push(groupIndex);
-            }
-            break;
+        if (currentGroup === 'getters') {
+          if (method.getter) {
+            methodGroupIndexes.push(groupIndex);
           }
-          case 'setters': {
-            if (method.setter) {
-              methodGroupIndexes.push(groupIndex);
-            }
-            break;
+        } else if (currentGroup === 'setters') {
+          if (method.setter) {
+            methodGroupIndexes.push(groupIndex);
           }
-          case 'type-annotations': {
-            if (method.typeAnnotation) {
-              methodGroupIndexes.push(groupIndex);
-            }
-            break;
+        } else if (currentGroup === 'type-annotations') {
+          if (method.typeAnnotation) {
+            methodGroupIndexes.push(groupIndex);
           }
-          case 'static-methods': {
-            if (method.static) {
-              methodGroupIndexes.push(groupIndex);
-            }
-            break;
+        } else if (currentGroup === 'static-methods') {
+          if (method.static) {
+            methodGroupIndexes.push(groupIndex);
           }
-          case 'instance-variables': {
-            if (method.instanceVariable) {
-              methodGroupIndexes.push(groupIndex);
-            }
-            break;
+        } else if (currentGroup === 'instance-variables') {
+          if (method.instanceVariable) {
+            methodGroupIndexes.push(groupIndex);
           }
-          case 'instance-methods': {
-            if (method.instanceMethod) {
-              methodGroupIndexes.push(groupIndex);
-            }
-            break;
+        } else if (currentGroup === 'instance-methods') {
+          if (method.instanceMethod) {
+            methodGroupIndexes.push(groupIndex);
           }
-          case 'displayName':
-          case 'propTypes':
-          case 'contextTypes':
-          case 'childContextTypes':
-          case 'mixins':
-          case 'statics':
-          case 'defaultProps':
-          case 'constructor':
-          case 'getDefaultProps':
-          case 'state':
-          case 'getInitialState':
-          case 'getChildContext':
-          case 'getDerivedStateFromProps':
-          case 'componentWillMount':
-          case 'UNSAFE_componentWillMount':
-          case 'componentDidMount':
-          case 'componentWillReceiveProps':
-          case 'UNSAFE_componentWillReceiveProps':
-          case 'shouldComponentUpdate':
-          case 'componentWillUpdate':
-          case 'UNSAFE_componentWillUpdate':
-          case 'getSnapshotBeforeUpdate':
-          case 'componentDidUpdate':
-          case 'componentDidCatch':
-          case 'componentWillUnmount':
-          case 'render': {
-            if (currentGroup === method.name) {
-              methodGroupIndexes.push(groupIndex);
-            }
-            break;
+        } else if ([
+          'displayName',
+          'propTypes',
+          'contextTypes',
+          'childContextTypes',
+          'mixins',
+          'statics',
+          'defaultProps',
+          'constructor',
+          'getDefaultProps',
+          'state',
+          'getInitialState',
+          'getChildContext',
+          'getDerivedStateFromProps',
+          'componentWillMount',
+          'UNSAFE_componentWillMount',
+          'componentDidMount',
+          'componentWillReceiveProps',
+          'UNSAFE_componentWillReceiveProps',
+          'shouldComponentUpdate',
+          'componentWillUpdate',
+          'UNSAFE_componentWillUpdate',
+          'getSnapshotBeforeUpdate',
+          'componentDidUpdate',
+          'componentDidCatch',
+          'componentWillUnmount',
+          'render'
+        ].includes(currentGroup)) {
+          if (currentGroup === method.name) {
+            methodGroupIndexes.push(groupIndex);
           }
-          default: {
-            // Is the group a regex?
-            const isRegExp = currentGroup.match(regExpRegExp);
-            if (isRegExp) {
-              const isMatching = new RegExp(isRegExp[1], isRegExp[2]).test(method.name);
-              if (isMatching) {
-                methodGroupIndexes.push(groupIndex);
-              }
-            } else if (currentGroup === method.name) {
+        } else {
+          // Is the group a regex?
+          const isRegExp = currentGroup.match(regExpRegExp);
+          if (isRegExp) {
+            const isMatching = new RegExp(isRegExp[1], isRegExp[2]).test(method.name);
+            if (isMatching) {
               methodGroupIndexes.push(groupIndex);
             }
-            break;
+          } else if (currentGroup === method.name) {
+            methodGroupIndexes.push(groupIndex);
           }
         }
       });
--- a/lib/rules/sort-comp.js
+++ b/lib/rules/sort-comp.js
@@ -133,9 +133,7 @@ module.exports = {
     function getRefPropIndexes(method) {
       const methodGroupIndexes = [];
 
-      for (let groupIndex = 0; groupIndex < methodsOrder.length; groupIndex++) {
-        const currentGroup = methodsOrder[groupIndex];
-
+      methodsOrder.forEach((currentGroup, groupIndex) => {
         switch (currentGroup) {
           case 'getters': {
             if (method.getter) {
@@ -218,7 +216,7 @@ module.exports = {
             break;
           }
         }
-      }
+      });
 
       // No matching pattern, return 'everything-else' index
       if (methodGroupIndexes.length === 0) {
--- a/lib/rules/sort-comp.js
+++ b/lib/rules/sort-comp.js
@@ -131,84 +131,108 @@ module.exports = {
      * @returns {Array} The matching patterns indexes. Return [Infinity] if there is no match.
      */
     function getRefPropIndexes(method) {
-      let isRegExp;
-      let matching;
-      let i;
-      let j;
-      const indexes = [];
+      const methodGroupIndexes = [];
 
-      if (method.getter) {
-        const getterIndex = methodsOrder.indexOf('getters');
-        if (getterIndex >= 0) {
-          indexes.push(getterIndex);
-        }
-      }
+      for (let groupIndex = 0; groupIndex < methodsOrder.length; groupIndex++) {
+        const currentGroup = methodsOrder[groupIndex];
 
-      if (method.setter) {
-        const setterIndex = methodsOrder.indexOf('setters');
-        if (setterIndex >= 0) {
-          indexes.push(setterIndex);
-        }
-      }
-
-      if (method.typeAnnotation) {
-        const annotationIndex = methodsOrder.indexOf('type-annotations');
-        if (annotationIndex >= 0) {
-          indexes.push(annotationIndex);
-        }
-      }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lib/rules/sort-comp.js [hunk 1]
```diff
@@ -131,84 +131,108 @@ module.exports = {
...
-      }
-
-      if (indexes.length === 0) {
-        for (i = 0, j = methodsOrder.length; i < j; i++) {
-          isRegExp = methodsOrder[i].match(regExpRegExp);
-          if (isRegExp) {
-            matching = new RegExp(isRegExp[1], isRegExp[2]).test(method.name);
-          } else {
-            matching = methodsOrder[i] === method.name;
+        switch (currentGroup) {
```
ljharb: does this have to be a switch? it's so much more verbose than if/elses would be.

lib/rules/sort-comp.js
```diff
@@ -131,84 +131,108 @@ module.exports = {
...
-        }
-      }
-
-      if (method.setter) {
-        const setterIndex = methodsOrder.indexOf('setters');
-        if (setterIndex >= 0) {
-          indexes.push(setterIndex);
-        }
-      }
+      for (let groupIndex = 0; groupIndex < methodsOrder.length; groupIndex++) {
```
ljharb: could this be a `.map` instead of a for loop?

lib/rules/sort-comp.js
```diff
@@ -131,84 +131,108 @@ module.exports = {
...
+            }
+            break;
+          }
+          case 'instance-methods': {
+            if (method.instanceMethod) {
+              methodGroupIndexes.push(groupIndex);
+            }
+            break;
+          }
+          case 'displayName':
```
ljharb: this seems like a list of lifecycle methods that should be a shared constant array, available in multiple rules

lib/rules/sort-comp.js
nosilleg: It doesn't have to be a switch.  

Generally `switch` is considered to be more readable and more performant than a chain of `if`/`else`es.  For me, the potential conciseness of the `if`/`else` chain wouldn't help with readability or maintainability; which would be my main concern.

I don't know the performance characteristics for this in JavaScript, but [this article](https://www.oreilly.com/library/view/high-performance-javascript/9781449382308/ch04s02.html) would indicate that it's true for JavaScript as well.

The fallthrough block of lifecycle methods is the most verbose part of this.  That could be changed to be a part of the `default` block, which would allow for the shared array of lifecycle method names (mentioned in another comment) with an `.includes(currentGroup)`.  I would say that this is less "pure" than the code as it stands, but not enough to block the suggestion.

Of course, you may just want it to be all `if`/`else`.

Let me know which option you want to go with.

lib/rules/sort-comp.js
nosilleg: I will update this to be a `forEach` (since we don't care about the resulting array that a `map` creates).

I had considered making it a `forEach`, but reading the code I had assumed that `for`s were preferable since there's a lot of places that `forEach` could be used but is not.

lib/rules/sort-comp.js
nosilleg: I think this is a sane suggestion, but...

It only works if you want to move to one of the two options above where we remove the lifecycle switch block (or `switch` completely).  

And there is code that can make use of the shared array.  (After a quick and rudimentary search I'm not seeing any other candidate code, but it may exist.)

lib/rules/sort-comp.js
ljharb: I definitely disagree; I consider switch to almost always be less readable than an alternative, and certainly less maintainable. It'd be great to go with all if/else.

nosilleg: I will update to be `if`/`else` chain.

I would be remiss if I didn't point out that your opinion goes against the popular opinion.  e.g. [why switch case and not if else if](https://stackoverflow.com/questions/1028437/why-switch-case-and-not-if-else-if)  So while this change may be more readable and maintainable for you, it potentially wont be for the masses.

nosilleg: Done.

ljharb: In JavaScript, in fact, the popular opinion is that switch is terrible and should be avoided. A random Stack Overflow post that isn't JavaScript-specific isn't particularly relevant to JavaScript.

> Done.

Thanks!

lib/rules/sort-comp.js
```diff
@@ -131,84 +131,91 @@ module.exports = {
...
+          'statics',
+          'defaultProps',
+          'constructor',
+          'getDefaultProps',
+          'state',
+          'getInitialState',
+          'getChildContext',
+          'getDerivedStateFromProps',
+          'componentWillMount',
+          'UNSAFE_componentWillMount',
```
ljharb: I think the UNSAFE_ methods should only be included in this array when the React version in settings is 16.3+ - same with gDSFP, gSBU, and cDC (for 16.0+)

lib/rules/sort-comp.js
nosilleg: Hi Jordan. It's a shame we have to belabour this point.

I didn't provide JavaScript specific discussions because you didn't specifically mention JavaScript in your opinion. Taking that into consideration is it fair to read your original statement as:

> I definitely disagree; I consider switch to almost always be less readable than an alternative, and certainly less maintainable _for JavaScript_. _I agree that it is more readable, maintainable, and performant for other languages. This is the overwhelming shared opinion within the JavaScript community, despite the similarities in syntax between JavaScript and other languages, and the similar performance implications with other languages._

If so, can you please provide references as I'd like to be following community best practices.

lib/rules/sort-comp.js
nosilleg: I disagree with this.  In my opinion this makes sense in the config code, but not in the rule implementation itself.

If someone has their config set up so that it includes the `UNSAFE_` methods and they have the `UNSAFE_` methods in their component then eslint should enforce that config no matter which version of React they are using.

lib/rules/sort-comp.js
ljharb: Since we're talking about a JS project, the context "in JavaScript" should be assumed. I offer no opinion here on use of `switch` in languages irrelevant to this repository (namely, everything that's not JavaScript).

Google, as well as common style guides (jslint, for example, since the 90s; airbnb's; etc), can all be consulted for information on that.
 - https://javascriptweblog.wordpress.com/2010/03/08/caseagainstswitch/
 - https://ericleads.wordpress.com/2012/12/11/switch-case-considered-harmful/
 - https://hackernoon.com/rethinking-javascript-eliminate-the-switch-statement-for-better-code-5c81c044716d
 - https://medium.com/tandemly/whats-wrong-with-the-switch-statement-in-javascript-c560e8ea3c0b
 - https://codeburst.io/alternative-to-javascripts-switch-statement-with-a-functional-twist-3f572787ba1c
etc.

lib/rules/sort-comp.js
ljharb: If someone is using React 15, React 16 lifecycle methods should not be sorted in the lifecycle group - because for them, they aren't lifecycle methods. Similarly, if someone is using React 17, where `componentWillMount` theoretically does not exist, cWM should not be sorted in the lifecycle group - because for them, it's not a lifecycle method.

lib/rules/sort-comp.js
nosilleg: You mention styleguides, but only link to one person opinion pieces.

The styleguides themselves make no claims that `switch` should be avoided and instead encourage sane syntax as per every other language that has `switch` statements.

The opinion pieces that you link to make the main complaints that switch doesn't follow their preferred programming paradigm. I.e. it isn't OO enough (very appropriate for a 2010 opinion piece.), or not functional enough (the new hotness). JavaScript is not a single paradigm language.

You say that JSLint has discourage the use of `switch` since the 90s, but considering it was created in 2002, and [Crockford himself doesn't advocate avoiding `switch`](https://crockford.com/javascript/code.html) I'm not sure what to make of your claims.

Others stating that premature deoptimisation is preferable to using using the actual features of a language so that they can pretend it's a different language doesn't seem rational to me. Using JavaScript as JavaScript with the lint enforced best practices does, and the styleguides appear to back this view.

lib/rules/sort-comp.js
nosilleg: Agreed. This is a concern for the code that loads the default config. 

This code is dealing with the finalised config. It should not undo a config based on assumptions about the environment.  eg the user may have their own `UNSAFE_` method that just happens to match the choice a new version of React implements. One would hope this is unlikely, but it would be bad practice to assume it impossible.  

Assuming we even wanted to try and undo the config, because custom groups are flattened in the config, and someone could overwrite the lifecycle group, we could only undo methods that were not contiguous with other lifecycle methods. Even then we would be assuming that the user used the lifecycle group when they may have hand coded the list.  We could inspect the raw config and deconstruct from there... <pauses> Why are we trying to undo the users config?

lib/rules/sort-comp.js
ljharb: ah lol, you're right about jslint, i misremembered.

Either way, avoiding switch statements is still a best practice (in JavaScript, obviously).

lib/rules/sort-comp.js
ljharb: i guess i'm confused why we need a separate list of lifecycle methods, if this is just "the user's config". should this not fall down to line 200?

nosilleg: Yes, line 162 is a functional subset of 200. However 162 is documented behaviour and 200 is not. As stated in the PR description, I've added a test to make sure the functionality isn't lost, but presumed it was better to code to expected behaviours.

Would you prefer that I comment L#200 making it explicit that it's undocumented functionality, or remove 162 and effectively hide (in my opinion) documented behaviour in the code for undocumented behaviour?  Of course the tests would catch if someone removed the code thinking they were removing the undocumented behaviour, but that feels too obfuscated to me.

lib/rules/sort-comp.js
nosilleg: Can you please provide sources that confirm it is best practice to avoid switch statements in JavaScript, or help me understand what is incorrect about my last message if you believe your previous sources were sufficient.

lib/rules/sort-comp.js
ljharb: Hmm.

Let's keep this as-is; thanks for your explanation.

lib/rules/sort-comp.js
```diff
@@ -131,84 +131,91 @@ module.exports = {
...
+          'UNSAFE_componentWillReceiveProps',
+          'shouldComponentUpdate',
+          'componentWillUpdate',
+          'UNSAFE_componentWillUpdate',
+          'getSnapshotBeforeUpdate',
+          'componentDidUpdate',
+          'componentDidCatch',
+          'componentWillUnmount',
+          'render'
+        ].includes(currentGroup)) {
```
ljharb: `includes` is not available on node 4; please use the `array-includes` package instead of relying on the native method.

nosilleg: Done.

nosilleg:
I'm just wanted to calibrate my expectations with this PR.  Is there a rough estimate on when it will be merged/released?  Is there anything more I can do to help progress it?

ljharb:

Issue #1817, opened 2018-06-08
jsx-sort-default-props: unreachable path?


hornta:
Can someone explain for me why it is neccessary to check for CallExpressions in here:
https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/jsx-sort-default-props.js#L133

I want to help and contribute to this project having full coverage.

ljharb:
cc @b0gok

b0gok:
@hornta you're right, it's not necessary in case of default props. I think it's an artifact from here (where it's necessary of course):
https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/sort-prop-types.js#L156

So you can safely remove this case from `jsx-sort-default-props`.
👍 for attentiveness.
PR is welcome.

Issue #1834, opened 2018-06-19
jsx-sort-default-props misnamed


webOS101:
Why does this rule have 'jsx' in the name?  It is not a jsx related rule.  Can this be renamed to be more consistent with the other rules?

ljharb:
The first step would be a semver-minor change, adding in `sort-default-props` without removing `jsx-sort-default-props` (but deprecating it).

Later, when it's time for a semver-major, we could then remove the jsx- one.

alexzherdev:
I'm working on the first part of it now. Do we also want to copy the tests? If there is a bug in the rule before next semver-major, will we be fixing it in two places and adding tests to two files then?

ljharb:
Yes, and yes.

PR #1861, opened 2018-06-29
[New] add `sort-default-props` and deprecate `jsx-sort-default-props`

alexzherdev:
Pt. 1 of #1834

### DIFF
--- a/lib/rules/sort-default-props.js
+++ b/lib/rules/sort-default-props.js
@@ -9,9 +9,6 @@
 const variableUtil = require('../util/variable');
 const docsUrl = require('../util/docsUrl');
 const report = require('../util/report');
-const log = require('../util/log');
-
-let isWarnedForDeprecation = false;
 
 // ------------------------------------------------------------------------------
 // Rule Definition
@@ -23,12 +20,11 @@ const messages = {
 
 module.exports = {
   meta: {
-    deprecated: true,
     docs: {
       description: 'Enforce defaultProps declarations alphabetical sorting',
       category: 'Stylistic Issues',
       recommended: false,
-      url: docsUrl('jsx-sort-default-props'),
+      url: docsUrl('sort-default-props'),
     },
     // fixable: 'code',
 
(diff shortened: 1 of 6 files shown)


alexzherdev:
Looks like I need to fix #1817 to improve coverage

lib/rules/jsx-sort-default-props.js
```diff
@@ -130,12 +134,6 @@ module.exports = {
             checkSorted(propTypesObject.properties);
           }
           break;
-        case 'CallExpression':
-          const innerNode = node.arguments && node.arguments[0];
```
ljharb: hm, why is this unused exactly?

Could this change be done in a separate commit/PR?

alexzherdev: See discussion in https://github.com/yannickcr/eslint-plugin-react/issues/1817#issuecomment-396329150
I removed it because coveralls was red and I needed to increase coverage to make it pass 🙂Otherwise I can absolutely make this change separately

ljharb: it'd be great to at least make it separate commits, even if it stays in this PR :-)

alexzherdev: @ljharb looks like this change isn't necessary for coverage now, so I'll submit a separate PR when this one is in.

alexzherdev:
Just rebased to resolve conflicts

VincentLanglet:
If my PR is merged before yours, you'll have to add my bug fix.
https://github.com/yannickcr/eslint-plugin-react/pull/2182

VincentLanglet:
@alexzherdev Since my PR https://github.com/yannickcr/eslint-plugin-react/pull/2182 was merged, you have to update the new sort-default-props, or you'll lose the fix I made.

Thanks ;)

amannn:
What's the status of this? Looks like this PR was approved and the only thing left to do is to copy over the bugfix mentioned above.

ljharb:
ping @alexzherdev

alexzherdev:
Done

lib/rules/sort-default-props.js
```diff
@@ -0,0 +1,172 @@
...
+      MemberExpression(node) {
+        if (!isDefaultPropsDeclaration(node)) {
+          return;
+        }
+
+        checkNode(node.parent.right);
+      }
+    };
+  }
+};
```
ljharb: needs a trailing newline

alexzherdev: Ah, was on a different laptop 👍

ljharb:
This looks good, but I'm going to cut a patch release before merging semver-minor items.

PR #1863, opened 2018-06-29
Make attribute "charset" valid

silvenon:
The attribute `charset` was considered an error in favor of `charSet`, but both of them result in valid HTML because HTML is case-insensitive. In other words, these two lines will be interpreted in the same way:

```jsx
<meta charSet="utf-8" />
<meta charset="utf-8" />
```

They will render different HTML (exactly like this JSX suggests), but browsers won't care. If nothing else, `charset` should be the more "correct" form of this attribute because it doesn't suggest a dash (`fillOpacity` = `fill-opacity`, `charSet` ≠ `char-set`).

https://github.com/facebook/react/issues/11492#issuecomment-342953964

### DIFF
--- a/tests/lib/rules/no-unknown-property.js
+++ b/tests/lib/rules/no-unknown-property.js
@@ -30,6 +30,8 @@ ruleTester.run('no-unknown-property', rule, {
     {code: '<App class="bar" />;'},
     {code: '<App for="bar" />;'},
     {code: '<App accept-charset="bar" />;'},
+    {code: '<meta charset="utf-8" />;'},
+    {code: '<meta charSet="utf-8" />;'},
     {code: '<App http-equiv="bar" />;'},
     {code: '<App xlink:href="bar" />;'},
     {code: '<App clip-path="bar" />;'},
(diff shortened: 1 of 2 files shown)


tests/lib/rules/no-unknown-property.js
```diff
@@ -30,6 +30,8 @@ ruleTester.run('no-unknown-property', rule, {
     {code: '<App class="bar" />;'},
     {code: '<App for="bar" />;'},
     {code: '<App accept-charset="bar" />;'},
+    {code: '<meta charset="utf-8" />;'},
+    {code: '<meta charSet="utf-8" />;'},
```
ljharb: i think we should be forcing `charset`; HTML also supports `CHARSET` and `cHaRsEt`, but lowercase is the very dominant convention.

silvenon: That was my initial thought, but `charSet` is not an unknown attribute, as the name of the rule implies. That said, I’ll modify the PR, so you decide. 😉

ljharb: hmm, that is true. I suppose it’s a slight overreach of this rule to force it.

silvenon: Yeah. As much as I would like for people to write `charset` instead of `charSet`, raising an error because of that would not be symmetrical with raising an error for `acceptcharset` in favor of `acceptCharset`, because only the latter would result into the correct form of the attribute – `accept-charset`.

I'll hold with the changes then until we agree what they should be.

ljharb: no need, i think this makes sense as-is

PR #1868, opened 2018-06-30
Add ignoreDOMComponents option to jsx-no-bind

### LINKED ISSUE #1238: Allow jsx-no-bind to ignore dom components
In my point of view the `jsx-no-bind` rule is mostly for `PureComponent`, although there is consideration on GC performance, it should not be significant

In many cases, we need to pass some arguments to an event handler, for example: 

```js
<li key={name} className={className} onClick={() => this.onClickTag(name, selected)}>{text}</li>
```

This will trigger the `jsx-no-bind` rule, here we have 2 concerns:

1. If we are using a custom component in above code, we can pass `name` and `selected` to component's `props` and call `this.props.onClick(this.props.name, this.props.selected)` to prevent such function bindings, however we cannot do this with DOM components
2. DOM components are never `PureComponent`, so a large reason for `jsx-no-bind` rule will not be a benefit for DOM components

For these reasons, I expect the `jsx-no-bind` can have a `ignoreDOMComponents` config to ignore function bindings on DOM components

alexzherdev:
Resolves #1238

### DIFF
--- a/docs/rules/jsx-no-bind.md
+++ b/docs/rules/jsx-no-bind.md
@@ -7,21 +7,22 @@ A `bind` call or [arrow function](https://developer.mozilla.org/en-US/docs/Web/J
 The following patterns are considered warnings:
 
 ```jsx
-<div onClick={this._handleClick.bind(this)}></div>
+<Foo onClick={this._handleClick.bind(this)}></Foo>
 ```
 ```jsx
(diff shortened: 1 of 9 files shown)


alexzherdev:
@ljharb I think the error on travis is some kind of flakiness

PR #1873, opened 2018-07-02
[Tests] `no-access-state-in-setstate`: passing test for “don't error if it's not a React Component"

kentcdodds:
unstated uses `this.setState` and we're getting false positives when using this rule.

I figured this would be better than an issue. Is this something we could add?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,12 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ### Fixed
 * [`display-name`]: Get rid of false position on component detection ([#2759] @iiison)
 
+### Changed
+* [`no-access-state-in-setstate`]: passing test for “don't error if it's not a React Component” ([#1873] @kentcdodds)
+
 [#2640]: https://github.com/yannickcr/eslint-plugin-react/pull/2640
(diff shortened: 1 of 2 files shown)


ljharb:
Absolutely - we should be using `Components.detect` in this rule: https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/no-access-state-in-setstate.js#L24

similarly to https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/no-deprecated.js#L43. Want to make that change? :-D

kentcdodds:
Neat! Perhaps I will. But if anyone else would like to make that happen before I can get around to it that'd be super :+1:

ljharb:
I've pushed up an initial WIP commit; it won't fix the problem quite yet. One will need to check [like is done here](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/no-deprecated.js#L164) to ensure that the setState call is inside a React component.

If anyone else wants to help, please *do not* submit a new PR - post a link to your commits here, and I'll happily pull them into this one.

ljharb:
This seems to have been fixed in #2711. I'll rebase this and land the test case to close it.

PR #1888, opened 2018-07-19
Possible jsx-no-allocation-in-props rule

alexzherdev:
cc @ljharb 

So... now that I know what future prospects are for `jsx-no-bind` (https://github.com/yannickcr/eslint-plugin-react/issues/1633#issuecomment-402611450) I actually have doubts that it makes sense to add this new rule on its own. The code here is basically the same as `jsx-no-bind`. Perhaps we could instead just update that rule to handle object creation. Thoughts?


docs/rules/jsx-no-allocation-in-props.md
```diff
@@ -0,0 +1,34 @@
...
+<Foo bar={[1, 2, 3]}></Foo>
+```
+```jsx
+<Foo bar={{ foo: 1 }}></Foo>
+```
+
+## Rule Options
+
+```js
+"react/jsx-no-allocation-in-props": [<enabled>, {
```
ljharb: array literals, and object literals, should probably have separate controls

README.md
```diff
@@ -150,6 +150,7 @@ Enable the rules that you would like to use.
 * [react/jsx-key](docs/rules/jsx-key.md): Validate JSX has key prop when in array or iterator
 * [react/jsx-max-depth](docs/rules/jsx-max-depth.md): Validate JSX maximum depth
 * [react/jsx-max-props-per-line](docs/rules/jsx-max-props-per-line.md): Limit maximum of props on a single line in JSX (fixable)
+* [react/jsx-no-allocation-in-props](docs/rules/jsx-no-allocation-in-props.md): Prevent usage of `[]` and `{}` in JSX props
```
ljharb: does this need a `jsx-` prefix? couldn't it also apply to `React.createElement`?

ljharb (commented):
I don't think the name "no bind" makes sense for objects and arrays :-/

docs/rules/jsx-no-allocation-in-props.md
alexzherdev: I actually couldn't come up with a use case where an array would be supplied as a prop to a DOM component, that probably is not a valid case?

alexzherdev:
Oh yeah, a new rule (if we go that way) would have a different name.

docs/rules/jsx-no-allocation-in-props.md
ljharb: for DOM components sure, but i meant, i should be able to configure this rule to only warn on object literals, or to only warn on array literals.

alexzherdev:
Removed `jsx` prefix, added `createElement` support and controls for arrays/objects

ljharb:
I don't think "allocations" is the best word to use in a memory-managed language, where allocations aren't user-observable.

alexzherdev:
Oh, for sure, this name leaves much to be desired. I guess this is more `no-object-or-array-literals-in-props` but that's quite a mouthful 😐 Although not really more than `no-redundant-should-component-update` 😄

benasher44:
I see this was updated recently. Any chance it'll ship anytime soon? Would love to see this one live!

ljharb:
@benasher44 it hasn't been updated since 2018; those 4 force pushes are to master, not this PR.

If you'd like to help, please comment a link to an updated branch (NOT a new PR) and i can update this one :-)

benasher44:
@ljharb just pushed benasher44/no-allocation-in-props. Only thing I'm having trouble with right now is getting this error running the test locally:

```
AssertionError [ERR_ASSERTION]: Parsers provided as strings to RuleTester must be absolute paths
```

I get this error running other tests locally though too, so I think I'm just missing some configuration in my local env/setup. Any idea what it might be?

benasher44:
Ah wait I think I figured that out. Hold please!

benasher44:
Sweet okay pushed again. Tests pass.

benasher44:
Sweet okay pushed again. Tests pass now!

benasher44:
Looks like last thing being discussed was naming. Are we good with `no-object-or-array-literals-in-props`?

benasher44:
@ljharb just checking in — happy to do the work if more is needed!

PR #1891, opened 2018-07-20
Add fixer for sort-prop-types

finnp:
This adds adds a fixer for the `react/sort-prop-types` rule. 

It fixes ordering issues including the `sortShapeProp` and the `ignoreCase` option.

~~It doesn't work yet for `callbacksLast` or `requiredFirst`, but that could be added later. (Right now those errors would just be ignored)~~

### DIFF
--- a/lib/rules/sort-prop-types.js
+++ b/lib/rules/sort-prop-types.js
@@ -20,6 +20,8 @@ module.exports = {
       url: docsUrl('sort-prop-types')
     },
 
+    fixable: 'code',
+
     schema: [{
       type: 'object',
       properties: {
@@ -71,6 +73,46 @@ module.exports = {
       );
     }
 
+    function getShapeProperties (node) {
+      return node.arguments && node.arguments[0] && node.arguments[0].properties;
+    }
+
+    function sorter(a, b) {
+      let aKey = getKey(a);
+      let bKey = getKey(b);
+      if (requiredFirst) {
+        if (isRequiredProp(a) && !isRequiredProp(b)) {
+          return -1;
+        }
+        if (!isRequiredProp(a) && isRequiredProp(b)) {
+          return 1;
+        }
+      }
+
+      if (callbacksLast) {
+        if (isCallbackPropName(aKey) && !isCallbackPropName(bKey)) {
+          return 1;
+        }
+        if (!isCallbackPropName(aKey) && isCallbackPropName(bKey)) {
+          return -1;
+        }
+      }
+
+      if (ignoreCase) {
+        aKey = aKey.toLowerCase();
+        bKey = bKey.toLowerCase();
+      }
+
+      if (aKey < bKey) {
+        return -1;
+      }
+      if (aKey > bKey) {
+        return 1;
+      }
+      return 0;
+    }
+
+
     /**
      * Checks if propTypes declarations are sorted
      * @param {Array} declarations The array of AST nodes being checked.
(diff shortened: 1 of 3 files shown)


finnp:
@ljharb Yeah, I noticed earlier that this is an issue. I added an additional testcase for this and adapted the fixing code to make sure this doesn't happen.

I also ended up adapting the fixing code to also work with the additional options as well.

lib/rules/sort-prop-types.js
```diff
@@ -83,6 +85,78 @@ module.exports = {
         return;
       }
 
+      function fix(fixer) {
+        function sortInSource (allNodes, source) {
```
ljharb: no space after the function name

lib/rules/sort-prop-types.js
```diff
@@ -83,6 +85,78 @@ module.exports = {
...
+            if (curr.type === 'ExperimentalSpreadProperty' || curr.type === 'SpreadElement') {
+              acc.push([]);
+            } else {
+              acc[acc.length - 1].push(curr);
+            }
+            return acc;
+          }, [[]]);
+
+          nodeGroups.forEach(nodes => {
+            const sortedAttributes = nodes.slice(0).sort((a, b) => {
```
ljharb: This sorting function should probably be defined at file level, since it’s pure?

lib/rules/sort-prop-types.js
```diff
@@ -83,6 +85,78 @@ module.exports = {
...
+              const attr = nodes[i];
+              let sortedAttrText = sourceCode.getText(sortedAttr);
+              if (sortShapeProp && isShapeProp(sortedAttr.value)) {
+                const attrSource = sortInSource(
+                  sortedAttr.value.arguments[0].properties,
+                  originalSource
+                );
+                sortedAttrText = attrSource.slice(sortedAttr.range[0], sortedAttr.range[1]);
+              }
+              source = `${source.substr(0, attr.range[0])}${sortedAttrText}${source.substr(attr.range[1])}`;
```
ljharb: please use slice over substring/substr

finnp:
@ljharb Thanks for the review, I incorporated your feedback.

lib/rules/sort-prop-types.js
```diff
@@ -83,6 +121,45 @@ module.exports = {
...
+          nodeGroups.forEach(nodes => {
+            const sortedAttributes = nodes.slice(0).sort(sorter);
+
+            for (let i = nodes.length - 1; i >= 0; i--) {
+              const sortedAttr = sortedAttributes[i];
+              const attr = nodes[i];
+              let sortedAttrText = sourceCode.getText(sortedAttr);
+              if (sortShapeProp && isShapeProp(sortedAttr.value)) {
+                const attrSource = sortInSource(
+                  sortedAttr.value.arguments[0].properties,
```
ljharb: does isShapeProp ensure that the value has a non-empty arguments? What about `PropTypes.shape()`?

lib/rules/sort-prop-types.js
```diff
@@ -83,6 +121,45 @@ module.exports = {
...
+            if (curr.type === 'ExperimentalSpreadProperty' || curr.type === 'SpreadElement') {
+              acc.push([]);
+            } else {
+              acc[acc.length - 1].push(curr);
+            }
+            return acc;
+          }, [[]]);
+
+          nodeGroups.forEach(nodes => {
+            const sortedAttributes = nodes.slice(0).sort(sorter);
```
ljharb: nbd but .slice() works too

lib/rules/sort-prop-types.js
```diff
@@ -83,6 +121,45 @@ module.exports = {
...
+            }
+          });
+          return source;
+        }
+
+        const source = sortInSource(declarations, context.getSourceCode().getText());
+
+        const rangeStart = declarations[0].range[0];
+        const rangeEnd = declarations[declarations.length - 1].range[1];
+        return fixer.replaceTextRange([rangeStart, rangeEnd], source.substr(rangeStart, rangeEnd - rangeStart));
```
ljharb: please use slice over substr

lib/rules/sort-prop-types.js
finnp: Good point, I added test cases for this. Also there was already a bug that cause the linting to fail on `PropTypes.shape()`. I included this fix: https://github.com/yannickcr/eslint-plugin-react/pull/1842

finnp:
@ljharb Good catch with `PropTypes.shape`. I added test cases and a fix.

Issue #1893, opened 2018-07-23
jsx-one-expression-per-line with CRLF on Windows


luissmg:
The rule always triggers, even when I have an expression per line, as the rule expects.

![imagem](https://user-images.githubusercontent.com/20660551/43066452-99d88c72-8e5c-11e8-86a3-4e15c502d4da.png)

Then I realized that I am on Windows and the line break is done with a `\r\n`(CRLF). What I did was to change it to what Linux and OSX do and make the line breack as `\n` (LF). The plugin then works as expected.

Is it supposed to work only with that type of line break?

ljharb:
I wouldn't expect anyone on any platform, including Windows, to use anything but normal LF, since CRLF breaks so many things.

luissmg:
Well, Git expects you, by default (at least on Windows), to use CRLF. If you don't do it, it automatically changes it. Yes, I know you can change that and add a configuration. But why? VSCode does the same. He uses the Windows CRLF. Unless you want to change that in all your files in your project, I think the rule should detect this.

ljharb:
Yes, you should be changing that in all files in all projects, and also using git config to ensure that you convert everything to LF on clone/commit/etc.

(I also think the rule should detect it, however)

luissmg:
Yes, I am not defending that what I am doing is the right or wrong thing to do. Just saying that are other possibilities.
Will add tests ASAP and merge them.

Istoman:
I have the same problem on Windows yet my EOL is set to LF.

dominic-p:
> Yes, you should be changing that in all files in all projects

I'm running into this myself, so I thought I would chime in. The default and recommended setting when installing Git on Windows is `core.autocrlf = true` which coverts line endings to CRLF on Windows and commits them as LF.

Unless I'm missing something, in that configuration, you are indeed expected to have CRLF in your files when editing on Windows.

![autocrlf](https://user-images.githubusercontent.com/1195080/59391357-e6afcd00-8d28-11e9-9e70-de0d3bedcdac.png)

ljharb:
@dominic-p the option you want to select is the second one, imo.

dominic-p:
@ljharb, that's understandable. I'm just pointing out that since it's the default and the Git recommendation, we're not talking about an edge case here. It might be worth supporting.

Thanks, btw, for all the work you do on this config. It's super helpful.

ljharb:
[...]

PR #1894, opened 2018-07-23
Detect line breaks in different OS

luissmg:
Check the following issue for reference.
#1893

### DIFF
--- a/lib/rules/jsx-one-expression-per-line.js
+++ b/lib/rules/jsx-one-expression-per-line.js
@@ -99,8 +99,8 @@ module.exports = {
             return;
           }
 
-          countNewLinesBeforeContent = (child.raw.match(/^\s*\n/g) || []).length;
-          countNewLinesAfterContent = (child.raw.match(/\n\s*$/g) || []).length;
+          countNewLinesBeforeContent = (child.raw.match(/^\s*(\r?\n)/g) || []).length;
+          countNewLinesAfterContent = (child.raw.match(/(\r?\n)\s*$/g) || []).length;
         }
 
         const startLine = child.loc.start.line + countNewLinesBeforeContent;
@@ -201,7 +201,7 @@ module.exports = {
 
         const nodeToReport = details.node;
         const descriptor = details.descriptor;
-        const source = details.source.replace(/(^ +| +(?=\n)*$)/g, '');
+        const source = details.source.replace(/(^ +| +(?=(\r?\n))*$)/g, '');
 
         const leadingSpaceString = details.leadingSpace ? '\n{\' \'}' : '';
         const trailingSpaceString = details.trailingSpace ? '{\' \'}\n' : '';
--- a/tests/lib/rules/jsx-one-expression-per-line.js
+++ b/tests/lib/rules/jsx-one-expression-per-line.js
@@ -80,6 +80,18 @@ ruleTester.run('jsx-one-expression-per-line', rule, {
     code: [
       '<App>',
       '  <Foo />',
+      '</App>'
+    ].join('\r\n')
+  }, {
+    code: [
+      '<App>',
+      '  <Foo />',
+      '</App>'
+    ].join('\r')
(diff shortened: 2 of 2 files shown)


ljharb (commented):
This seems a fine improvement, despite CRLF being a terrible idea to use, even on Windows.

However, can you please add some tests for this?

luissmg:
Hi. Adding the tests right now. Do you want me to add tests for every case that exists right now but just change the line break?
Asking because there are already 800 lines of tests for this rule.

ljharb:
@luissmg we definitely don't need that many tests :-) one or two should probably suffice.

luissmg:
I thought just the same but wanted to confirm first :)

luissmg:
@ljharb Sorry for taking so long on this. Added the tests!

ljharb (approved):
Thanks!

ljharb:
@luissmg altho it seems like the tests are failing

luissmg:
@ljharb I was checking and [this test](https://github.com/yannickcr/eslint-plugin-react/blob/9fe9e941c7d331db5cc87ee1d6988215577e6669/tests/lib/rules/jsx-one-expression-per-line.js#L153) that I added is exactly the same as the [one above](https://github.com/yannickcr/eslint-plugin-react/blob/9fe9e941c7d331db5cc87ee1d6988215577e6669/tests/lib/rules/jsx-one-expression-per-line.js#L136) that is passing, excepting for the type of line break.

Shouldn't it pass too or am I missing something?

ljharb:
I suspect the failing test ends up with a mix of `\r\n` and `\n`, which seems like a problem. Perhaps the autofixer should detect the linebreak type most commonly used in the file, or in the node it’s looking at, and use that instead of always outputting a LF?

grahammcculloch:
Hi. Would love to see this get merged :smile:  Currently making me sad about my Windows machine!

ljharb:
It needs a rebase and fixed tests.

luissmg:
@ljharb @grahammcculloch Doing that this weekend!

evertonfraga:
@luissmg any news? :)

lib/rules/jsx-one-expression-per-line.js
```diff
@@ -59,8 +59,8 @@ module.exports = {
               return;
             }
 
-            countNewLinesBeforeContent = (child.raw.match(/^ *\n/g) || []).length;
-            countNewLinesAfterContent = (child.raw.match(/\n *$/g) || []).length;
+            countNewLinesBeforeContent = (child.raw.match(/^ *[\r\n|\r|\n]/g) || []).length;
+            countNewLinesAfterContent = (child.raw.match(/[\r\n|\r|\n] *$/g) || []).length;
```
sindresorhus: I think you meant `(\r\n|\r|\n)` or `[\r\n]` since brackets will match anything inside, even pipes. The best regex for line endings is simply `\r?\n`. It will match both CRLF and LF. Matching only CR is not needed as it's not used anymore.

luissmg: Thank you so much for the tip! Will try to fix it this weekend

luissmg:
Hi guys. I will try to fix this in this weekend. Sorry for not replying any sooner.

ljharb:
@luissmg are you still interested in completing this PR?

luissmg:
Hi @ljharb! Sorry for letting this here in the air. I will try to push something this week. Will apply the suggestion from @sindresorhus and check if the tests are ok

luissmg:
Hey @ljharb, how are you? I took a look into this today. can you check if it is okay?

ljharb:
@luissmg unfortunately the tests don't fail, when your fix is removed.

luissmg:
@ljharb Hmmmm you are right... Could be that this problem is not a problem anymore?

ljharb:
It's entirely possible!

Issue #1901, opened 2018-07-26
react/prefer-stateless-function works different than described in docs


metreniuk:
For the `react/prefer-stateless-function` rule the docs for [ignorePureComponents](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prefer-stateless-function.md#ignorepurecomponents) option states:
 >When true the rule will ignore Components extending from React.PureComponent that use this.props or this.context.

Actually it also ignore every PureComponent that use **not only `props` or `context`**, but ignores every component that use `this`. If this is not intended I prepared a fix, otherwise it makes sense to update the docs.

PR #1902, opened 2018-07-26
Add ignorePureComponents check on 'this' usage

metreniuk:
This PR address the #1901 issue.

### DIFF
--- a/tests/lib/rules/prefer-stateless-function.js
+++ b/tests/lib/rules/prefer-stateless-function.js
@@ -349,6 +349,20 @@ ruleTester.run('prefer-stateless-function', rule, {
       code: `
         class Foo extends React.PureComponent {
           render() {
+            return <div>this.bar</div>;
+          }
+        }
+      `,
+      options: [{
+        ignorePureComponents: true
+      }],
+      errors: [{
+        message: 'Component should be written as a pure function'
+      }]
+    }, {
+      code: `
+        class Foo extends React.PureComponent {
+          render() {
             return <div>{this.props.foo}</div>;
           }
         }
--- a/lib/rules/prefer-stateless-function.js
+++ b/lib/rules/prefer-stateless-function.js
@@ -361,7 +361,7 @@ module.exports = {
           if (
             !has(list, component) ||
             hasOtherProperties(list[component].node) ||
-            list[component].useThis ||
+            (list[component].useThis && !ignorePureComponents) ||
             list[component].useRef ||
             list[component].invalidReturn ||
             list[component].hasChildContextTypes ||


tests/lib/rules/prefer-stateless-function.js
```diff
@@ -345,6 +345,20 @@ ruleTester.run('prefer-stateless-function', rule, {
       errors: [{
         message: 'Component should be written as a pure function'
       }]
+    }, {
+      code: `
+        class Foo extends React.PureComponent {
+          render() {
+            return <div>this.bar</div>;
```
ljharb: I'm a bit confused here; how can the linter statically know that there's no `bar = foo` defined in the class or the constructor?

metreniuk: I see the problem. I was confused by the docs:
>When true the rule will ignore Components extending from React.PureComponent that use this.props or this.context.

The example that you mentioned leads to the fact that this rule will ignore all `this`'s properties, not only `props` and `context`. Is it accurate?

ljharb: If you have a component that doesn't reference props or context, then there's other better mechanisms for anything that hangs off of `this` - so the rule *should* warn on it, and you should refactor to use an SFC.

alexzherdev: Is this line missing curlies around `this.bar`?

ljharb: good catch, yes.

metreniuk:
I've accidentally removed the fork from which I made the PR.  I will check if I could come with another one.

ljharb:
That's unfortunate.

ljharb:
@metreniuk any chance you'd be able to recreate the PR?

metreniuk:
@ljharb Yes, I will work on it in the begging of the week

ljharb:
Any update?

metreniuk:
Sorry again. I was looking into it and found a positive test case in the master for the issue addressed in this PR
https://github.com/jsx-eslint/eslint-plugin-react/blob/c88c7c497a75988497d14bc492aee9b8cf3b05a3/tests/lib/rules/prefer-stateless-function.js#L158-L167

Is this the intended behaviour or in this case we still want to warn the user to use SFC?

ljharb:
In that case, `this.bar` isn't valid anyways, so I'm not really sure. It's *probably* a bug, and should be `this.props.bar`, in which case we _would_ want to warn and suggest an SFC.

metreniuk:
I'm still confused. The docs says that if `instance property other than this.props and this.context` is found (ex: `this.bar`) then the rule should not warn. I think I misunderstood the rule in the first place. If you have a case that would make sense to cover in this rule, let me know.

ljharb:
yes, that's fair, and I agree with your interpretation.

PR #1903, opened 2018-07-26
[New] `no-children-prop`: Add `allowFunctions` option 

### LINKED ISSUE #1803: Proposal: alter no-children-prop to allow functions
If you use react router, is some situations you need to [pass a function as children](https://reacttraining.com/react-router/web/api/Route/children-func).
It would be nice to be able to use the same syntax as them.

```jsx
<Route path={to} children={({ match }) => (
  <li className={match ? 'active' : ''}>
    <Link to={to} {...rest}/>
  </li>
)} />
```
is easier to reason about than
```jsx
<Route path={to}>
  {({ match }) => (
    <li className={match ? 'active' : ''}>
      <Link to={to} {...rest}/>
    </li>
  )}
</Route>
```

There could be a config to allow functions (not the default).

alexzherdev:
Resolves #1803

### DIFF
--- a/docs/rules/no-children-prop.md
+++ b/docs/rules/no-children-prop.md
@@ -34,3 +34,29 @@ Examples of **correct** code for this rule:
 React.createElement("div", {}, 'Children')
 React.createElement("div", 'Child 1', 'Child 2')
 ```
+
+## Rule Options
+
+```js
+"react/no-children-prop": [<enabled>, {
+  "allowFunctions": <boolean> || false
+}]
+```
+
+### `allowFunctions`
+
+When `true`, and passing a function as `children`, it must be in prop position and not child position.
+
+The following patterns are considered warnings:
+
+```jsx
+<MyComponent>{data => data.value}</MyComponent>
+React.createElement(MyComponent, {}, data => data.value)
+```
+
+The following are **not** considered warnings:
+
+```jsx
+<MyComponent children={data => data.value} />
+React.createElement(MyComponent, { children: data => data.value })
+```
--- a/lib/rules/no-children-prop.js
+++ b/lib/rules/no-children-prop.js
@@ -40,15 +40,37 @@ module.exports = {
 
     messages: {
       nestChildren: 'Do not pass children as props. Instead, nest children between the opening and closing tags.',
-      passChildrenAsArgs: 'Do not pass children as props. Instead, pass them as additional arguments to React.createElement.'
+      passChildrenAsArgs: 'Do not pass children as props. Instead, pass them as additional arguments to React.createElement.',
+      nestFunction: 'Do not nest a function between the opening and closing tags. Instead, pass it as a prop.',
+      passFunctionAsArgs: 'Do not pass a function as an additional argument to React.createElement. Instead, pass it as a prop.'
     },
 
-    schema: []
+    schema: [{
+      type: 'object',
+      properties: {
+        allowFunctions: {
+          type: 'boolean',
+          default: false
+        }
+      },
+      additionalProperties: false
+    }]
   },
   create(context) {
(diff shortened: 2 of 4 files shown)


alexzherdev:
Sounds good 👍

ljharb:
My personal belief is that function children are an abomination that destroys the entire purpose for using jsx, and I never want them to exist in code I maintain.

alexzherdev:
Ok, I'll have to do my research 😃 But immediately, `forbid` would not even allow a `children` prop, I'm just not sure how that is different from e.g. `onClick={() => {}}` (which I know the downsides of, but that's already handled by a different rule)

ljharb:
Oh sure, I just want an enum value that's equivalent to "not specifying the option"

alexzherdev:
Ok, I don't think I'm on the same page. Let me go over it once again.

Current rule: `children` prop is a warning no matter what.
My PR:
  - `allowFunctions: false (or not specified)`: no change to current rule
  - `allowFunctions: true`: functions in the `children` prop are **not** a warning. Functions in the child position are a warning - isn't that what you were going for originally?

And then extending that, the enum would be the 4 possible combinations of either allowing or forbidding functions in a prop or in the child position. Not specifying the enum value retain the current behaviour of the rule, which would be equal to the `child` option: functions are allowed in the child position but not as a prop. `prop` would be what `allowFunctions: true` does in my PR.

ljharb:
ahh i see what you mean - ok, fair. then true/false is probably fine.

altho i can also see people asking for a third option, which is requiring that functions be in the child position.

alexzherdev:
But don't they already have it via the current rule. As soon as you do `children={() => {}}`, it will warn, thereby forcing you to move the function into the child position.

ljharb:
Right - but what if they want all non-function children in the prop position, and function children in the child position?

Or, all non-function children in one position or the other, but function children disallowed?

alexzherdev:
That sounds like a complicated schema, so maybe we start small to make sure those options are actually in demand? Also the rule being called `no-children-prop` kinda assumes we should be discouraging the children prop 🤔

ljharb:
ha, true. the main purpose is to force child position.

alexzherdev:
Ok, let me know if any other changes are necessary.

docs/rules/no-children-prop.md
```diff
@@ -34,3 +34,29 @@ The following patterns are **not** considered warnings:
...
+
+```js
+"react/no-children-prop": [<enabled>, {
+  "allowFunctions": <boolean> || false
+}]
+```
+
+### `allowFunctions`
+
+When `true`, passing a function in the children prop is preferred.
```
ljharb: let's say this instead: when true, and passing a function as `children`, it must be in prop position and not child position.

lib/rules/no-children-prop.js
```diff
@@ -36,15 +36,38 @@ module.exports = {
...
+        }
+      },
+      additionalProperties: false
+    }]
   },
   create: function(context) {
+    const configuration = context.options[0] || {};
+
+    function isFunction(node) {
+      return (node.type === 'ArrowFunctionExpression' || node.type === 'FunctionExpression')
```
ljharb: what about generators or async functions? is there a more generic way to catch all functions?

alexzherdev: According to AST explorer, both of these are `FunctionExpression`s with either `generator` or `async` flag equal to true.

ljharb: awesome, thanks. maybe test cases with these would be good tho :-D

G-Rath:
Any update on this? I'd love to see this feature land, since libraries like `react-router` seem to recommend this pattern.

PR #1909, opened 2018-07-30
Ignore class properties in destructuring-assignment

### LINKED ISSUE #1875: destructuring-assignment in class properties
First I'd like to just say that I like the new `destructuring-assignment` rule because I think it enforces a lot more clarity and explicitness. However I'm having problems with it throwing errors inside of class properties. 

```javascript
  state = {
    value: this.props.defaultValue,
  };

  ...
}
```
IMO this should be an exception and be totally valid or at least have the option to exclude it just from this use-case. I know that I could move it inside a `constructor` and destructure it there but I'd rather not do that.

alexandernanberg:
Fixes #1875

Please let me know if you would like me to add more tests or if anything can be improved

### DIFF
--- a/lib/rules/destructuring-assignment.js
+++ b/lib/rules/destructuring-assignment.js
@@ -22,12 +22,20 @@ module.exports = {
         'always',
         'never'
       ]
+    }, {
+      type: 'object',
+      properties: {
+        ignoreClassFields: {
+          type: 'boolean'
+        }
+      },
+      additionalProperties: false
     }]
   },
 
   create: Components.detect((context, components, utils) => {
     const configuration = context.options[0] || DEFAULT_OPTION;
-
+    const ignoreClassFields = context.options[1] && context.options[1].ignoreClassFields === true || false;
 
     /**
     * Checks if a prop is being assigned a value props.bar = 'bar'
@@ -82,7 +90,10 @@ module.exports = {
         !isAssignmentToProp(node)
       );
 
-      if (isPropUsed && configuration === 'always') {
+      if (
+        isPropUsed && configuration === 'always' &&
+        !(ignoreClassFields && node.parent.type === 'ClassProperty')
+      ) {
         context.report({
           node: node,
           message: `Must use destructuring ${node.object.property.name} assignment`
@@ -128,7 +139,10 @@ module.exports = {
           });
         }
 
-        if (classComponent && destructuringClass && configuration === 'never') {
+        if (
+          classComponent && destructuringClass && configuration === 'never' &&
+          !(ignoreClassFields && node.parent.type === 'ClassProperty')
+        ) {
           context.report({
             node: node,
             message: `Must never use destructuring ${node.init.property.name} assignment`
--- a/docs/rules/destructuring-assignment.md
+++ b/docs/rules/destructuring-assignment.md
@@ -85,3 +85,24 @@ const Foo = class extends React.PureComponent {
     return <div>{this.state.title}</div>;
   }
 };
+```
+
+## Rule Options
+
+```js
+...
+"react/destructuring-assignment": [<enabled>, "always", { "ignoreClassFields": <boolean> }]
+...
+```
+
+### `ignoreClassFields`
+
+When `true` the rule will ignore class field declarations.
+
+The following patterns are then considered okay and do not cause warnings:
+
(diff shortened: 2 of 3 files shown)


ljharb:
That's one of the points of the rule - if you need to rely on props values, it's supposed to force it to go in the constructor.

alexandernanberg:
Okay I get it!

How would you like the option to be implemented? Not sure that I see a clear way to do it well without a breaking change. I guess we could add another enum option like `ignoreClassProperties` but I feel that it would maybe be a bit ambiguous?

ljharb:
`allowInClassFields` seems reasonable to me, or `allow: ["class-fields"]` (where that's the only allowed value atm)

alexandernanberg:
Okay so I've now added the option, I went for another name because I thought it was more logical than using the "allow" word, but let me know if you want me to change it. I also a bit unsure if I wrote the schema correctly

So you would do this;
`["error", "always", { "ignoreClassFields": true }]`

lib/rules/destructuring-assignment.js
```diff
@@ -22,12 +22,20 @@ module.exports = {
...
+        }
+      },
+      additionalProperties: false
     }]
   },
 
   create: Components.detect((context, components, utils) => {
     const configuration = context.options[0] || DEFAULT_OPTION;
-
+    const ignoreClassFields = configuration === 'always' && context.options[1] && context.options[1].ignoreClassFields === true || false;
```
ljharb: when set to “never”, this should still apply - you could maybe have `foo = function () { const { a } = this.props; }.call(this)`, or when do expressions land, you could have `foo = do { const { a } = this.props; a; }`

alexandernanberg: Makes sense, I'll update my PR 👍

ljharb: shouldn't https://github.com/yannickcr/eslint-plugin-react/pull/1909/files#diff-d73ce54079fb0715f60db27176a09fb6R135 also be updated?

HenriBeck:
@ljharb should this be behind a flag because it doesn't have to do anything with destructuring assignments? I would agree that it would be a great rule to disallow props usage inside class properties but doing it here might be a little bit confusing for people.

lib/rules/destructuring-assignment.js
alexandernanberg: Yep it probably should. Sorry for missing that, I've never worked in AST or eslint-plugin projects before so it's all new to me. I've updated the PR again.

ljharb:
@HenriBeck the rule is to force, or forbid, destructuring assignment regardless of location inside a class body. That class fields make destructuring awkward might be an inconvenience to some, hence the need for an option - but personally, I like the side effect that if you're trying to do anything complex in a class field, it forces you to properly put it in a constructor.

HenriBeck:
@ljharb I agree with you that doing complicated things inside class properties isn't the best idea.
But forbidding it in this rule seems odd because class properties aren't about destructuring and should be able to disallow them whether or not you have this rule configured.

ljharb:
Methods aren't about destructuring either - the rule deals with any references to state, props, or context, no matter where they appear.

alexandernanberg:
Well shit, this does not work when assigning an object with properties containing `this.props` to a class field.

```js
  state = {
    // this will still throw an error
    bar: this.props.bar
  }
}
```

Did a quick try to fix this but didn't make any progress, can somebody either try to solve this or point me in the right direction? @ljharb

EvHaus:
That code you posted isn't valid JS. You cannot use `this.props` in a class property definition. You'd have to rewrite this to:

```js
  constructor (props) {
    super(props);

    this.state = {
      // this will still throw an error
      bar: props.bar
    }
  }
}
```

anna93:
Yup 7.11.1, the below still warns me about de-structuring props assignment:
 const sampleSchemaData = this.props.schemaData;

ljharb:
@anna93 that’s a variable, not a class field.

zslabs:
This still throws one for me using the same setup as @anna93:

```
class Input extends React.Component {
  id = `${this.props.name}-${generateUUID()}`;
 ...
```

ljharb:
@zslabs if with the option enabled you still get a warning there, please file a new issue

Issue #1908, opened 2018-07-30
default-props-match-prop-types doesn't detect properties in quotes


zech:
Hello,

same as in issue [#1201](https://github.com/yannickcr/eslint-plugin-react/issues/1201) last year the above mentioned rule doesn't detect quoted properties. The error for the second and third property is `defaultProp "undefined" defined for isRequired propType. (react/default-props-match-prop-types)`

```javascript
SomeComponent.propTypes = {
    "firstProperty": PropTypes.string.isRequired,
    "secondProperty": PropTypes.bool,
    "thirdProperty": PropTypes.func
};

SomeComponent.defaultProps = {
    "secondProperty": false,
    "thirdProperty": () => undefined
};
```
As before removing the quotes solves the error. But I don't think this should be consistent and a non issue.

alexzherdev:
This one will get automatically fixed when this rules uses the common detection logic from #1911.

PR #1911, opened 2018-08-03
Extract propTypes detection code

alexzherdev:
~~This still fails a bunch of tests, but wanted to put it up for preliminary high-level review early.~~

~~The reason tests are failing is because I extracted similar code from `prop-types` and `no-unused-prop-types` but those two have diverged and I need to reconcile two implementations into a single helper.~~

Next steps:

- [x] fix failing tests (this PR)
- [x] look at helper coverage when separately running tests for both rules and see if any one of them is missing test cases that they now support by virtue of using a unified implementation _(PR in my fork: alexzherdev/eslint-plugin-react#1)_
- [x] currently only **declared** propTypes detection is extracted. Could also extract **used** propTypes detection from `prop-types` and `no-unused-prop-types` ? _(PR in my fork: alexzherdev/eslint-plugin-react#2)_
- [x] `require-default-props` and `default-props-match-prop-types` also have some common functionality for requiredness and defaultProps detection - extract those? That would be a prerequisite to those rules using the common helper. _(PRs in my fork: alexzherdev/eslint-plugin-react#3, alexzherdev/eslint-plugin-react#4)_

### DIFF
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -9,11 +9,8 @@
 
 const has = require('has');
 const Components = require('../util/Components');
-const variable = require('../util/variable');
-const annotations = require('../util/annotations');
 const astUtil = require('../util/ast');
 const versionUtil = require('../util/version');
-const propsUtil = require('../util/props');
 const docsUrl = require('../util/docsUrl');
 
 // ------------------------------------------------------------------------------
@@ -57,35 +54,10 @@ module.exports = {
   },
 
   create: Components.detect((context, components, utils) => {
-    const defaults = {skipShapeProps: true};
     const sourceCode = context.getSourceCode();
-    const configuration = Object.assign({}, defaults, context.options[0] || {});
-    const skipShapeProps = configuration.skipShapeProps;
-    const customValidators = configuration.customValidators || [];
-    const propWrapperFunctions = new Set(context.settings.propWrapperFunctions || []);
     const checkAsyncSafeLifeCycles = versionUtil.testReactVersion(context, '16.3.0');
 
-    // Used to track the type annotations in scope.
-    // Necessary because babel's scopes do not track type annotations.
-    let stack = null;
-
     const UNUSED_MESSAGE = '\'{{name}}\' PropType is defined but prop is never used';
-
-    /**
-     * Helper for accessing the current scope in the stack.
-     * @param {string} key The name of the identifier to access. If omitted, returns the full scope.
-     * @param {ASTNode} value If provided sets the new value for the identifier.
-     * @returns {Object|ASTNode} Either the whole scope or the ASTNode associated with the given identifier.
-     */
-    function typeScope(key, value) {
-      if (arguments.length === 0) {
-        return stack[stack.length - 1];
-      } else if (arguments.length === 1) {
-        return stack[stack.length - 1][key];
-      }
-      stack[stack.length - 1][key] = value;
-      return value;
-    }
 
     /**
      * Check if we are in a lifecycle method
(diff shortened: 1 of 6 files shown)


lib/rules/no-unused-prop-types.js
```diff
@@ -893,7 +454,7 @@ module.exports = {
         return;
       }
 
-      (props || []).forEach(prop => {
+      Object.values(props || {}).forEach(prop => {
```
alexzherdev: `declaredPropTypes` used to be an array here, but an object in `prop-types`. Common implementation uses an object (keyed by prop name) for easier lookups

lib/rules/no-unused-prop-types.js
```diff
@@ -980,28 +509,31 @@ module.exports = {
     function handleFunctionLikeExpressions(node) {
       handleSetStateUpdater(node);
       markDestructuredFunctionArgumentsAsUsed(node);
-      markAnnotatedFunctionArgumentsAsDeclared(node);
+    }
+
+    function handleCustomValidators(component) {
+      const propTypes = component.declaredPropTypes;
```
alexzherdev: This piece of code deals with used propTypes, but it was nested inside of `markPropTypesAsDeclared` (which was extracted out).

lib/util/propTypes.js
```diff
@@ -0,0 +1,671 @@
...
+module.exports = {
+  instructions(context, components, utils) {
+    let stack = null;
+
+    // Used to track the type annotations in scope.
+    // Necessary because babel's scopes do not track type annotations.
+    const classExpressions = [];
+    const defaults = {skipShapeProps: true, customValidators: []};
+    const configuration = Object.assign({}, defaults, context.options[0] || {});
+    const customValidators = configuration.customValidators;
```
alexzherdev: These two config options come from two different rules. Put them here for now, maybe will need to consider if all rules using this functionality should take these as options:

lib/util/propTypes.js
```diff
@@ -0,0 +1,671 @@
...
+    const skipShapeProps = configuration.skipShapeProps;
+    const sourceCode = context.getSourceCode();
+    const propWrapperFunctions = new Set(context.settings.propWrapperFunctions || []);
```
ljharb: `new Set` takes null and undefined quietly, so i'm not sure the `|| []` is needed.

lib/util/propTypes.js
```diff
@@ -0,0 +1,671 @@
...
+
+    /**
+     * Helper for accessing the current scope in the stack.
+     * @param {string} key The name of the identifier to access. If omitted, returns the full scope.
+     * @param {ASTNode} value If provided sets the new value for the identifier.
+     * @returns {Object|ASTNode} Either the whole scope or the ASTNode associated with the given identifier.
+     */
+    function typeScope(key, value) {
+      if (arguments.length === 0) {
```
ljharb: is there another way to write this function so it doesn't use `arguments`?

lib/util/propTypes.js
```diff
@@ -0,0 +1,671 @@
...
+        // prevent problems with recursive types.
+        seen = new Set();
+      }
+      if (seen.has(annotation)) {
+        // This must be a recursive type annotation, so just accept anything.
+        return {};
+      }
+      seen.add(annotation);
+
+      switch (annotation.type) {
```
ljharb: not a huge deal, but it'd be great to avoid a switch statement

lib/util/propTypes.js
```diff
@@ -0,0 +1,671 @@
...
+
+module.exports = {
+  instructions(context, components, utils) {
+    let stack = null;
+
+    // Used to track the type annotations in scope.
+    // Necessary because babel's scopes do not track type annotations.
+    const classExpressions = [];
+    const defaults = {skipShapeProps: true, customValidators: []};
+    const configuration = Object.assign({}, defaults, context.options[0] || {});
```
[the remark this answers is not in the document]
alexzherdev: that's not compatible with old node versions though?

lib/util/propTypes.js
alexzherdev: without changing it too much, I could check for `key` and/or `value` to be undefined I guess

lib/util/propTypes.js
alexzherdev: hmm do you mean a map with handlers for each type? there's also a bunch of other huge switches in here (like in other parts of the codebase) which could also get the same treatment

lib/util/propTypes.js
alexzherdev: Actually I was partly wrong, `customValidators` is present in both `prop-types` and `no-unused-prop-types`. `skipShapeProps` comes from `no-unused-prop-types` only, because it can be hard to figure out that a prop access corresponds to a shape. Not sure it's needed in `prop-types`, will take a look.

One problem right now is `skipShapeProps` is `true` by default in `no-unused-prop-types`, whereas `prop-types` never ignores shape props. Will try to fix that by always building declaredPropTypes for shapes and checking the option when validating props.

alexzherdev:
@ljharb thanks for the review. I'm still changing things up here fixing tests, just wanted to make sure I'm not doing anything outlandish.

alexzherdev:
screen shot 2018-08-03 at 9 07 50 pm
screen shot 2018-08-03 at 9 08 05 pm

Github bug? 🤔(same comment listed twice, one doesn't have a reply option)

lib/util/propTypes.js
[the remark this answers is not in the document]
ljharb: oh right, if we're not using babel here then we have to stick to Object.assign

lib/util/propTypes.js
ljharb: an alternative is creating multiple functions rather than overloading this one?

lib/util/propTypes.js
ljharb: if/else, or a map, yes

alexzherdev: Not sure how 4-5 chained else-ifs are better than a switch? anyway will try out a map

ljharb: they don't suffer from unintentional fallthrough.

alexzherdev:
@ljharb I responded to your comments. Could you take a look at the items in the PR description to see if they make sense? If they do, should I continue modifying this PR, or would it get too big?

lib/rules/display-name.js
```diff
@@ -116,7 +117,7 @@ module.exports = {
       );
 
       const namedFunctionExpression = (
-        (node.type === 'FunctionExpression' || node.type === 'ArrowFunctionExpression') &&
+        astUtil.isFunctionLikeExpression(node) &&
```
ljharb: I think this kind of change could be peeled off into a separate pr, and landed first?

alexzherdev: Ah, you’re right 👍

alexzherdev:
For those other changes though all of them depend on this one to be merged first.

I already have a bunch of new tests for item 2, I can push those separately then if you’d like.

I’ll split out the function helper, and then this PR can be considered complete.

alexzherdev:
@ljharb by itself this PR won’t fix #1908 as it doesn’t touch that particular rule.
My comment in #1908 might be misleading though. This detection doesn’t deal with defaultProps per se, but it does have the means to handle literals as propType keys. So the broader point I was trying to make is that having centralized defaultProps handling would make things like that just work. And this PR will provide the base for that further work (if we can get other collabs to sign off on it)
Sorry for the confusion, I hope this is more clear now.

alexzherdev:
@yannickcr @lencioni @EvHaus any chance you could take a look? asking because I have a truckload of sweet refactoring elsewhere that all depends on this one (see PR description) and I really like how it turned out.

ljharb:
I'll merge tonight if no issues are raised.

PR #1915, opened 2018-08-05
convert test runner from mocha to jest

hellatan:
sample jest output:

screen shot 2018-08-05 at 2 25 18 pm

current mocha output:

screen shot 2018-08-05 at 2 43 24 pm

thoughts?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # gitignore
 
 lib-cov
+.idea
 *.seed
 *.log
 *.csv
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,0 +1,16 @@
+'use strict';
+module.exports = {
+  collectCoverage: true,
+  coverageDirectory: '<rootDir>/reports/coverage',
+  coverageThreshold: {
+    global: {
+      statements: 96.48,
+      branches: 94.07,
+      functions: 99.53,
+      lines: 96.44
+    }
+  },
+  testURL: 'https://localhost/',
+  testMatch: ['<rootDir>/tests/**/*.js']
+};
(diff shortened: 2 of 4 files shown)


hellatan:
> If anything, we’d drop mocha and switch to tape

ok, gotcha. I also noticed that on the appveyor the main index.js test file had a `forEach` inside a `should` function and it kept on failing the build b/c it was complaining about nest tests or something. I didn't want to dig in too deep so I'll close this PR if you want to go in the direction of tape instead.

> Jest too-aggressively drops node support

not sure what you mean by this. in what away does it drop support?

ljharb:
Jest no longer supports node 4, for example, but we must continue to do so.

ljharb:
@hellatan if you can submit a PR that fixes the appveyor build without moving off of mocha, that would be great :-)

hellatan:
> if you can submit a PR that fixes the appveyor build without moving off of mocha, that would be great :-)

@ljharb well, i think it was an issue with jest as the test runner 😄since you'd rather use `tape`, i think i'd rather spend my effort towards something that would be more likely merged than not.

Are you sure you'd want to move test runners though? If so, I'd be willing to look into it in my spare time.

ljharb:
I’m not sure we need to do it yet; it’s a significant change and this is a large project. Probably don’t invest the time just yet :-) we do have https://github.com/yannickcr/eslint-plugin-react/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22 tho

hellatan:
that's a lot of issues 😄... though I've seen worse 🤣

Issue #1885, opened 2018-07-15
Destructuring Inside an Async Function or Class Method Isn't Warned


trisys3:
I added a GitHub repo illustrating the bug: [trisys3/prop-types-async-issue](https://github.com/trisys3/prop-types-async-issue).

Basically, when I have an async function, whether class method or a function by itself, and I destructure `this.props`, I am not warned about the props, even if they aren't in the prop-types. If I do have those props inside prop-types and I have the rule about unused prop-types on,

I found that if I use a destructured prop I **stop** getting warned about that prop being unused. This indicates that the bug is with the using, but not with the un-using, at least in classes.

This only applies to class methods, though. I never get warned about props in an async function declaration/statement, whether undeclared or unused, whether I use them or not, whether I declare them or not, or neither, or both. Also, this applies whether or not I'm destructuring. Basically, async functions seem un-implemented.

I tried to be as simple as possible. I added as few ESLint rules & plugins as I needed to illustrate the issue; I didn't even add babel, though I did add babel-eslint. I also tried to be short & succinct with my function/variable/class names, though I may have been a little too cute. If you're still confused, please tell me.

Issue #1473, opened 2017-10-12
prop-types not detected in async functions


jseminck:
Encountered this while working on CodeSandbox.

When I stripped async from the function, then the prop-types rule started to complain about modalActions not being defined in the propTypes. When I re-added async, then the warning disappeared.

![async_props](https://user-images.githubusercontent.com/9586897/31481773-cdb48fbe-af2d-11e7-8b8f-6e38bfb6669f.gif)

ljharb:
Presumably we're just not checking the node type for async arrow function expressions in a class property (probably not non-arrows either?).

PR #1939, opened 2018-08-15
Additional cases for prop-types and no-unused-prop-types

### LINKED ISSUE #836: react/prop-types breaks with spread syntax
```
static propTypes = {
  unusedProp: PropTypes.object.isRequired,
  usedProp: PropTypes.func.isRequired,
};
```

In this case, if I'm defining two props out of which I'm not using the first one inside component I'm getting:

```
'unusedProp' PropType is defined but prop is never used
```

which is correct. But if I define something in these lines:

```
static propTypes = {
  ...someMoreProps,
  unusedProp: PropTypes.object.isRequired,
  usedProp: PropTypes.func.isRequired,
};
```

I'm not going to get the error and I should. I'm still not using 'unusedProp' anywhere inside component and I'm not getting the error now that I have defined spread operator.

If anyone is asking why do I have these 'someMoreProps' rest syntax that is really common with some modules like [redux-form](https://github.com/erikras/redux-form). This is HOC component that one uses to wrap it's own component to add form validation. Module exports large number of props that nobody likes to write every single time so you just import them from module and add inside your propTypes as spread operator:

```
import { reduxForm, propTypes } from 'redux-form';

static propTypes = {
  ....propTypes,
  unusedProp: PropTypes.object.isRequired,
  usedProp: PropTypes.func.isRequired,
};
```

If I'm seeing this correctly, there shouldn't be an error that some prop is not defined as it may be inside rest operator, but it should still error out for those defined outside spread operator which are not used in the app which isn't happening in our case.

alexzherdev:
So I was adding test cases to both rules to bring them to parity after the refactoring, but a couple of things didn't make sense.
In particular, having just one flag for ignoring props validation doesn't seem to be enough (and that was the case before the refactoring as well). For instance, let's say we encounter a spread in props e.g.
```js
type Props = {
  ...data,
  lastname: string
};
```
That means any `this.props.foo` usage is considered to have a corresponding propType and it's ignored in `prop-types` via `ignorePropsValidation` flag we set in the helper.
However, this shouldn't preclude us from being able to detect whether `lastname` is an unused propType in `no-unused-prop-types` (this would be ignored if we met a `{...props}` spread). But we can't rely on the same flag, because that flag means a different thing. So I'm proposing to split it into two.

Also see comments inline.

Resolves #836

### DIFF
--- a/lib/util/propTypes.js
+++ b/lib/util/propTypes.js
@@ -143,10 +143,10 @@ module.exports = function propTypesInstructions(context, components, utils) {
         }
       });
 
-      // nested object type spread means we need to ignore/accept everything in this object
-      if (containsObjectTypeSpread) {
-        return {};
-      }
+      // Mark if this shape has spread. We will know to consider all props from this shape as having propTypes,
+      // but still have the ability to detect unused children of this shape.
+      shapeTypeDefinition.containsSpread = containsObjectTypeSpread;
+
       return shapeTypeDefinition;
     },
 
@@ -669,7 +669,7 @@ module.exports = function propTypesInstructions(context, components, utils) {
     JSXSpreadAttribute: function(node) {
       const component = components.get(utils.getParentComponent());
       components.set(component ? component.node : node, {
-        ignorePropsValidation: true
+        ignoreUnusedPropTypesValidation: true
       });
     },
 
--- a/tests/lib/rules/no-unused-prop-types.js
+++ b/tests/lib/rules/no-unused-prop-types.js
@@ -1802,34 +1802,6 @@ ruleTester.run('no-unused-prop-types', rule, {
         '};'
       ].join('\n')
     }, {
-      code: [
-        'type Person = {',
-        '  ...data,',
-        '  lastname: string',
-        '};',
-        'class Hello extends React.Component {',
-        '  props: Person;',
-        '  render () {',
-        '    return <div>Hello {this.props.firstname}</div>;',
-        '  }',
-        '}'
-      ].join('\n'),
-      parser: 'babel-eslint'
-    }, {
-      code: [
-        'type Person = {|',
-        '  ...data,',
-        '  lastname: string',
-        '|};',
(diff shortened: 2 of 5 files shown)


lib/util/propTypes.js
```diff
@@ -143,10 +143,10 @@ module.exports = function propTypesInstructions(context, components, utils) {
         }
       });
 
-      // nested object type spread means we need to ignore/accept everything in this object
-      if (containsObjectTypeSpread) {
-        return {};
-      }
+      // Mark if this shape has spread. We will know to consider all props from this shape as having propTypes,
```
alexzherdev: Similar to the case with propTypes, if the shape contains a spread plus some properties e.g.
```js
type Props = {
  person: {
    ...$Exact<BasePerson>,
    lastname: string
  }
};
```
it would be unwise to throw away the information we have about `lastname` just because we have a spread. By setting an additional flag and leaving the children data in place, we can detect if `lastname` is being used.

tests/lib/rules/no-unused-prop-types.js
```diff
@@ -1801,34 +1801,6 @@ ruleTester.run('no-unused-prop-types', rule, {
         '  bar: PropTypes.bool',
         '};'
       ].join('\n')
-    }, {
```
alexzherdev: So these tests are now able to detect that `lastname` is not used. I understand this is a breaking change? Not insisting on any of these to go in, but seems like a good change to me.

alexzherdev:
Sure. Wanted to start with a smaller set to validate the idea.

alexzherdev:
Travis is flakey.

ljharb:
Indeed; don't worry about those errors, I habitually rerun them when i see them.

alexzherdev:
It’d be great if this could get merged before I push the used prop type refactoring, as they’re slightly interrelated.

ThiefMaster:
I think this also fixed #1885 and #1473

PR #1941, opened 2018-08-16
Fix changelog "Ignore class properties" contributor

alexandernanberg:
No big deal, just changing to the right contributor


alexzherdev:
Oops, didn't notice that either :-)

PR #1942, opened 2018-08-16
Extract defaultProps detection

alexzherdev:


lib/util/Components.js
```diff
@@ -696,7 +697,9 @@ function componentRule(rule, context) {
   const ruleInstructions = rule(context, components, utils);
   const updatedRuleInstructions = util._extend({}, ruleInstructions);
   const propTypesInstructions = propTypes(context, components, utils);
-  const allKeys = new Set(Object.keys(detectionInstructions).concat(Object.keys(propTypesInstructions)));
+  const defaultPropsInstructions = defaultProps(context, components, utils);
+  const allKeys = new Set(Object.keys(detectionInstructions).concat(Object.keys(propTypesInstructions))
+    .concat(Object.keys(defaultPropsInstructions)));
```
ljharb: this doesn't need to be two concat calls, it can be `arr1.concat(arr2, arr3)`

lib/util/defaultProps.js
```diff
@@ -0,0 +1,264 @@
...
+      return;
+    }
+
+    if (defaultProps === 'unresolved') {
+      markDefaultPropsAsUnresolved(component);
+      return;
+    }
+
+    const defaults = component.defaultProps || {};
+    const newDefaultProps = defaultProps.reduce((acc, prop) => {
```
ljharb: let's use the `object.fromentries` package here.

lib/util/Components.js
alexzherdev: Oh 🤦‍♂️

alexzherdev:
@ljharb FYI I have a final refactoring ready but blocked by this, which is to extract `isRequired` detection from `default-props-match-prop-types` and `require-default-props` and make it part of the propTypes detection.
This will be a huge win since this is what will enable these two rules to migrate to the common detection.

lib/rules/require-default-props.js
```diff
@@ -369,9 +310,8 @@ module.exports = {
     return {
       MemberExpression: function(node) {
         const isPropType = propsUtil.isPropTypesDeclaration(node);
-        const isDefaultProp = propsUtil.isDefaultPropsDeclaration(node);
 
-        if (!isPropType && !isDefaultProp) {
+        if (!isPropType) {
```
ljharb: i'm a bit confused why the previous propTypes refactor didn't remove `isPropType`, but this defaultProps refactor is able to remove `isDefaultProp`. can you elaborate?

alexzherdev: Sure.

The first extraction was based on two other rules that had the most comprehensive detection functionality. However, `isRequired` detection, which is necessary for the two rules in this current PR, was not part of it. I could have added it there, but that would have blown up an already big PR.

But it is coming in the follow-up PR I mentioned above. I just didn’t want to combine it with this one, again, to keep them focused and manageable in size.

See https://github.com/alexzherdev/eslint-plugin-react/commit/438c06e3f0d71d32123583121d8c3cd8bce7cd62 and https://github.com/alexzherdev/eslint-plugin-react/commit/2d1243247c9a3b6c3cd1346d6fc74cac85947ad4 for an exclusive peek 😄

ljharb: ah ok, gotcha. so all the isPropType stuff will be able to be removed in a followup.

ljharb:
Will merge later today if no objections.

PR #1945, opened 2018-08-18
Add state-in-constructor rule

### LINKED ISSUE #1810: Rule suggestion: no-state-in-constructor
**INCORRECT**
```javascript
  constructor(props) {
    super(props);

    this.state = {
      counter: props.initialValue
    };
  }
}
```

**CORRECT**
```javascript
  state = {
    counter: this.props.initialValue
  };
}
```

**Why?**
It may reduce boilerplate by not having to define a constructor just for initializing the state. 

What do you think about this?

lukyth:
This will resolve #1810. Comments are very welcomed. 😃


lib/rules/state-in-constructor.js
```diff
@@ -0,0 +1,59 @@
...
+            message: 'State initialization should be in a constructor'
+          });
+        }
+      },
+      AssignmentExpression(node) {
+        if (
+          option === 'never' &&
+          node.left.object.type === 'ThisExpression' &&
+          node.left.property.name === 'state' &&
+          node.parent.parent.parent.parent.kind === 'constructor' &&
```
alexzherdev: This is a bit brittle: assignment could be nested in other nodes (e.g. if conditions) in which case this sequence will not work. This should either be traversing parents in a loop, or scopes like https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/no-unused-prop-types.js#L264 (although not sure which one is better)

lib/rules/state-in-constructor.js
```diff
@@ -0,0 +1,59 @@
...
+          utils.isES6Component(node.parent.parent.parent.parent.parent.parent)
```
alexzherdev: Same here, but in this case maybe `utils.getParentES6Component` would work

lib/rules/state-in-constructor.js
```diff
@@ -0,0 +1,59 @@
...
+          context.report({
+            node,
+            message: 'State initialization should be in a constructor'
+          });
+        }
+      },
+      AssignmentExpression(node) {
+        if (
+          option === 'never' &&
+          node.left.object.type === 'ThisExpression' &&
```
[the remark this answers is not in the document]
ljharb: Let's try to add a test case that would fail here.

lib/rules/state-in-constructor.js
ljharb: Let's try to add a test case that would fail here.

lukyth:
@alexzherdev @ljharb Thank you so much for your reviews! I've addressed all of them with these commits 81d0e8b...b1024c0. Could you please check them out?

I also removed some test cases which I think are redundant (81d0e8b). I'm not sure if that's the right call. If you guys don't think so then I can bring them back.

krawaller:
Just to inject some energy, this looks sweet! Good work @lukyth! :)

lukyth:
Anything I can do to get this merged?

lukyth:
@EvHaus I've added this rule to the list in `README.md` (9415814). Sorry it took me so long. I just saw your comment here.

PR #1946, opened 2018-08-19
Extract used propTypes detection

### LINKED ISSUE #963: `prop-types` should cover `props` in `setState()` too
As specified in [React docs](https://facebook.github.io/react/docs/react-component.html#setstate):

> It's also possible to pass a function with the signature `function(state, props) => newState`.

So when we use something like `this.setState((state, props) => {})`, that `props` argument must be covered by rule `prop-types` too.

### LINKED ISSUE #1871: react/prop-types not working in getDerivedStateFromProps
When a property is used in `getDerivedStateFromProps` but not defined in `defaultProps` this doesn't give me a warning or error.

In the below case `baz` gives me a `react/prop-types` error, I expect this same error from `bar`.

```
class TestClass extends React.PureComponent {
	static getDerivedStateFromProps(props) {
		const { foo, bar } = props;
		return {
			foobar: foo(bar),
		};
	}

	render() {
		const { baz } = this.props;
		const { foobar } = this.state;
		
		return <div>{foobar}</div>;
	}
}

TestClass.defaultProps = {
	foo: PropTypes.func.isRequired,
};
```

alexzherdev:
~~Pushing for visibility.
This is failing one existing test, which will get fixed after #1939 is merged (also will need a rebase).~~

Resolves #963, resolves #1871.


lib/util/usedPropTypes.js
```diff
@@ -0,0 +1,520 @@
...
+            markPropTypesAsUsed(node.parent);
+          }
+        });
+      }
+    },
+
+    'Program:exit': function() {
+      const list = components.list();
+
+      for (const component in list) {
```
ljharb: `for..in` should be avoided; can this use normal array iteration methods?

alexzherdev: Sure. `Object.keys(list).forEach` perhaps?
There's about a dozen other files already doing the same, handle them all?

alexzherdev:
I went ahead and fixed other `for..in` occurrences; I can go back and undo those if necessary. Also rebased.
`prop-types` is 201 line now, `no-unused-prop-types` is 144 🎉

lib/util/usedPropTypes.js
ljharb: yes, that would be great!

lib/rules/boolean-prop-naming.js
```diff
@@ -248,7 +247,7 @@ module.exports = {
             }
           }
 
-          if (!has(list, component) || (list[component].invalidProps || []).length) {
+          if ((list[component].invalidProps || []).length) {
```
ljharb: maybe `if (list[component].invalidProps && list[component].invalidProps.length > 0)`?

lib/rules/display-name.js
```diff
@@ -216,12 +215,12 @@ module.exports = {
       'Program:exit': function() {
         const list = components.list();
         // Report missing display name for all components
-        for (const component in list) {
-          if (!has(list, component) || list[component].hasDisplayName) {
-            continue;
+        Object.keys(list).forEach(component => {
+          if (list[component].hasDisplayName) {
+            return;
```
ljharb: `Object.keys(list).filter(component => !list[component].hasDisplayName).forEach`?

lib/rules/no-multi-comp.js
```diff
@@ -61,15 +60,15 @@ module.exports = {
         const list = components.list();
         let i = 0;
 
-        for (const component in list) {
-          if (!has(list, component) || isIgnored(list[component]) || ++i === 1) {
-            continue;
+        Object.keys(list).forEach(component => {
```
ljharb: `(component, i)` and then we don't need to `++` it, nor have `let i = 0` above.

lib/rules/no-set-state.js
```diff
@@ -74,12 +73,12 @@ module.exports = {
 
       'Program:exit': function() {
         const list = components.list();
-        for (const component in list) {
-          if (!has(list, component) || isValid(list[component])) {
-            continue;
+        Object.keys(list).forEach(component => {
```
ljharb: `Object.keys(list).filter(component => !isValid(list[component])).forEach`?

lib/rules/no-unused-prop-types.js
```diff
@@ -490,104 +124,20 @@ module.exports = {
...
-        }
-      },
-
       'Program:exit': function() {
         const list = components.list();
         // Report undeclared proptypes for all classes
-        for (const component in list) {
-          if (!has(list, component) || !mustBeValidated(list[component])) {
-            continue;
+        Object.keys(list).forEach(component => {
```
ljharb: `Object.keys(list).filter(component => mustBeValidated(list[component])).forEach`?

lib/rules/prop-types.js
```diff
@@ -478,76 +181,20 @@ module.exports = {
...
-        }
-      },
-
       'Program:exit': function() {
         const list = components.list();
         // Report undeclared proptypes for all classes
-        for (const component in list) {
-          if (!has(list, component) || !mustBeValidated(list[component])) {
-            continue;
+        Object.keys(list).forEach(component => {
```
ljharb: `Object.keys(list).filter(component => mustBeValidated(list[component])).forEach`?

lib/rules/require-default-props.js
```diff
@@ -626,21 +625,17 @@ module.exports = {
         stack = null;
         const list = components.list();
 
-        for (const component in list) {
-          if (!has(list, component)) {
-            continue;
-          }
-
+        Object.keys(list).forEach(component => {
```
ljharb: `Object.keys(list).filter(component => list[component].propTypes).forEach`?

lib/rules/require-optimization.js
```diff
@@ -221,12 +220,12 @@ module.exports = {
         const list = components.list();
 
         // Report missing shouldComponentUpdate for all components
-        for (const component in list) {
-          if (!has(list, component) || list[component].hasSCU) {
-            continue;
+        Object.keys(list).forEach(component => {
```
ljharb: `Object.keys(list).filter(component => !list[component].hasSCU).forEach`

lib/util/Components.js
```diff
@@ -123,9 +124,9 @@ class Components {
     const usedPropTypes = {};
 
     // Find props used in components for which we are not confident
-    for (const i in this._list) {
-      if (!has(this._list, i) || this._list[i].confidence >= 2) {
-        continue;
+    Object.keys(this._list).forEach(i => {
```
ljharb: `Object.keys(this._list).filter(i => this._list[i].confidence < 2).forEach`?

lib/util/Components.js
```diff
@@ -142,21 +143,22 @@ class Components {
...
+        usedPropTypes[componentId] = mergeUsedPropTypes(usedPropTypes[componentId] || [], newUsedProps);
       }
-    }
+    });
 
     // Assign used props in not confident components to the parent component
-    for (const j in this._list) {
-      if (!has(this._list, j) || this._list[j].confidence < 2) {
-        continue;
+    Object.keys(this._list).forEach(j => {
```
ljharb: `Object.keys(this._list).filter(j => this._list[j].confidence >= 2).forEach`?

lib/util/Components.js
```diff
@@ -697,7 +692,9 @@ function componentRule(rule, context) {
   const ruleInstructions = rule(context, components, utils);
   const updatedRuleInstructions = util._extend({}, ruleInstructions);
   const propTypesInstructions = propTypes(context, components, utils);
-  const allKeys = new Set(Object.keys(detectionInstructions).concat(Object.keys(propTypesInstructions)));
+  const usedPropTypesInstructions = usedPropTypesUtil(context, components, utils);
+  const allKeys = new Set(Object.keys(detectionInstructions).concat(Object.keys(propTypesInstructions))
+    .concat(Object.keys(usedPropTypesInstructions)));
```
ljharb: `new Set(Object.keys(detectionInstructions).concat(Object.keys(propTypesInstructions), Object.keys(usedPropTypesInstructions)))`?

lib/util/usedPropTypes.js
```diff
@@ -0,0 +1,526 @@
...
+  );
+}
+
+/**
+ * Checks if the component must be validated
+ * @param {Object} component The component to process
+ * @returns {Boolean} True if the component must be validated, false if not.
+ */
+function mustBeValidated(component) {
+  return Boolean(
```
ljharb: let's use `!!` instead of `Boolean(`

lib/util/usedPropTypes.js
```diff
@@ -0,0 +1,526 @@
...
+
+  /**
+   * Check if we are in a class constructor
+   * @return {boolean} true if we are in a class constructor, false if not
+   */
+  function inComponentWillReceiveProps() {
+    let scope = context.getScope();
+    while (scope) {
+      if (
+        scope.block && scope.block.parent &&
```
ljharb: if we cant get the whole condition on one line, let's put each condition on its own line

lib/util/usedPropTypes.js
alexzherdev: already done in https://github.com/yannickcr/eslint-plugin-react/pull/1946/commits/b35ea0c72317d4123237af32759f60858343b682 😄

lib/util/usedPropTypes.js
```diff
@@ -0,0 +1,526 @@
...
+   **/
+  function inLifeCycleMethod() {
+    let scope = context.getScope();
+    while (scope) {
+      if (scope.block && scope.block.parent && scope.block.parent.key) {
+        const name = scope.block.parent.key.name;
+
+        if (LIFE_CYCLE_METHODS.indexOf(name) >= 0) {
+          return true;
+        } else if (checkAsyncSafeLifeCycles && ASYNC_SAFE_LIFE_CYCLE_METHODS.indexOf(name) >= 0) {
```
ljharb: no need for the `else`

lib/util/usedPropTypes.js
```diff
@@ -0,0 +1,526 @@
...
+   * Returns true if the given node is a React Component lifecycle method
+   * @param {ASTNode} node The AST node being checked.
+   * @return {Boolean} True if the node is a lifecycle method
+   */
+  function isNodeALifeCycleMethod(node) {
+    const nodeKeyName = (node.key || {}).name;
+
+    if (node.kind === 'constructor') {
+      return true;
+    } else if (LIFE_CYCLE_METHODS.indexOf(nodeKeyName) >= 0) {
```
ljharb: also here (ie, the `no-else-return` eslint rule)

lib/util/usedPropTypes.js
```diff
@@ -0,0 +1,526 @@
...
+    }
+
+    const component = components.get(utils.getParentComponent());
+    const usedPropTypes = component && component.usedPropTypes || [];
+    let ignoreUnusedPropTypesValidation = component && component.ignoreUnusedPropTypesValidation || false;
+
+    switch (type) {
+      case 'direct':
+        // Ignore Object methods
+        if (Object.prototype[name]) {
```
ljharb: `name in Object.prototype` would be better, so as not to trigger getters

lib/util/usedPropTypes.js
```diff
@@ -0,0 +1,526 @@
...
+        markPropTypesAsUsed(node);
+      }
+    },
+
+    ObjectPattern: function(node) {
+      // If the object pattern is a destructured props object in a lifecycle
+      // method -- mark it for used props.
+      if (isNodeALifeCycleMethod(node.parent.parent)) {
+        node.properties.forEach((property, i) => {
+          if (i === 0) {
```
ljharb: this is just:
```js
if (node.properties.length > 0) {
  markPropTypesAsUsed(node.parent);
} 
```
the forEach seems useless?

lib/util/usedPropTypes.js
```diff
@@ -0,0 +1,526 @@
...
+            markPropTypesAsUsed(node.parent);
+          }
+        });
+      }
+    },
+
+    'Program:exit': function() {
+      const list = components.list();
+
+      Object.keys(list).forEach(component => {
```
ljharb: `Object.keys(list).filter(component => mustBeValidated(list[component])).forEach`?

lib/util/usedPropTypes.js
alexzherdev: yeah, I wonder why this isn't warning

lib/rules/no-multi-comp.js
alexzherdev: Can't do this because we only want to `++i` if `isIgnored` is false and doesn't short-circuit. I.e. we want to start reporting from the _second_ non-ignored component.

alexzherdev: Actually found a way to achieve this, also using `filter`.

ThiefMaster:
I believe this change caused #2017.

The old logic was:

```js
    function isPropTypesUsage(node) {
      const isClassUsage = (
        (utils.getParentES6Component() || utils.getParentES5Component()) &&
        node.object.type === 'ThisExpression' && node.property.name === 'props'
      );
      const isStatelessFunctionUsage = node.object.name === 'props' && !isAssignmentToProp(node);
      const isNextPropsUsage = node.object.name === 'nextProps' && (inComponentWillReceiveProps() || inShouldComponentUpdate());
      return isClassUsage || isStatelessFunctionUsage || isNextPropsUsage;
}
```

and it got replaced with:

```js
  function isPropTypesUsage(node) {
    const isClassUsage = (
      (utils.getParentES6Component() || utils.getParentES5Component()) &&
      ((node.object.type === 'ThisExpression' && node.property.name === 'props')
      || isPropArgumentInSetStateUpdater(node))
    );
    const isStatelessFunctionUsage = node.object.name === 'props' && !isAssignmentToProp(node);
    return isClassUsage || isStatelessFunctionUsage || inLifeCycleMethod();
  }
```

PR #1956, opened 2018-08-22
[New] Support shorthand fragment syntax

### LINKED ISSUE #1694: react/prop-types doesn't work with fragment syntax
**Example:**

```javascript
import React from "react";

const ForAttendees = ({ page }) => (
  <>
    <section className="intro intro_attendees">{page}</section>
    <div className="grid--5col">barfoo</div>
  </>
);

export default ForAttendees;
```

`react/prop-types` does **not** capture this. If you replace the fragment syntax with divs, it detects the case.

alexzherdev:
Resolves #1694.

### DIFF
--- a/lib/util/jsx.js
+++ b/lib/util/jsx.js
@@ -9,7 +9,7 @@ const COMPAT_TAG_REGEX = /^[a-z]|\-/;
 
 /**
  * Checks if a node represents a DOM element.
- * @param {String} node - JSXOpeningElement to check.
+ * @param {object} node - JSXOpeningElement to check.
  * @returns {boolean} Whether or not the node corresponds to a DOM element.
  */
 function isDOMComponent(node) {
@@ -25,6 +25,16 @@ function isDOMComponent(node) {
   return COMPAT_TAG_REGEX.test(name);
 }
 
+/**
+ * Checks if a node represents a JSX element or fragment.
+ * @param {object} node - node to check.
+ * @returns {boolean} Whether or not the node if a JSX element or fragment.
+ */
+function isJSX(node) {
+  return ['JSXElement', 'JSXFragment'].indexOf(node.type) >= 0;
+}
+
 module.exports = {
-  isDOMComponent: isDOMComponent
+  isDOMComponent: isDOMComponent,
+  isJSX: isJSX
 };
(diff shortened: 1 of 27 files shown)


lib/util/jsx.js
```diff
@@ -25,6 +25,16 @@ function isDOMComponent(node) {
   return COMPAT_TAG_REGEX.test(name);
 }
 
+/**
+ * Checks if a node represents a JSX element or fragment.
+ * @param {object} node - node to check.
+ * @returns {boolean} Whether or not the node if a JSX element or fragment.
+ */
+function isJSX(node) {
```
alexzherdev: Suggestions for a better name welcome (`isJSXElementOrFragment` seemed too wordy)

tests/lib/rules/jsx-curly-brace-presence.js
```diff
@@ -32,6 +32,10 @@ ruleTester.run('jsx-curly-brace-presence', rule, {
     {
       code: '<App {...props}>foo</App>'
     },
+    {
+      code: '<>foo</>',
```
alexzherdev: This rule doesn't really care whether it's an element or a fragment, so I just added one simple valid test, and one invalid.

ljharb:
I'll wait to merge this til we've updated all the relevant rules.

alexzherdev:
Quick q: I noticed some of the rules are using selectors e.g. https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/jsx-curly-brace-presence.js#L249
Selectors were introduced in ESLint 3.18 and we seem to be supporting ^3.0.0, so I guess we should avoid using them?

ljharb:
Fair point; altho nobody's complained. I guess we should fix that.

alexzherdev:
I was asking mostly because this change could use the syntax for `'JSXElement, JSXFragment'`. I will avoid it then and fix existing cases separately afterwards.

alexzherdev:
I believe I covered what needed to be covered, perhaps even redundantly.

alexzherdev:
Added support in 4 more rules, plus some tests in other relevant rules.

bensaufley:
@ljharb what's the timeline for the release that will include this?

ljharb:

PR #1962, opened 2018-08-29
Group static lifecycle methods under lifecycle

metreniuk:
Since the lifecycle group is a special group that deals with state/props changes, grouping static lifecycle methods under lifecycle group helps to have this logic together. This looks like a good default behavior that follows the docs.
This PR enforces static lifecycle methods to be grouped under `lifecycle` group as [one of possible solutions suggested](https://github.com/yannickcr/eslint-plugin-react/issues/1839#issuecomment-416658030) by @ljharb .

### DIFF
--- a/tests/lib/rules/sort-comp.js
+++ b/tests/lib/rules/sort-comp.js
@@ -468,7 +468,7 @@ ruleTester.run('sort-comp', rule, {
     }]
   }, {
     code: [
-      '// static lifecycle methods should be grouped with lifecycle',
+      '// static lifecycle methods should be grouped under lifecycle',
       'class Hello extends React.Component {',
       '  constructor() {}',
       '  static getDerivedStateFromProps() {}',
--- a/tests/lib/rules/sort-comp.js
+++ b/tests/lib/rules/sort-comp.js
@@ -468,15 +468,7 @@ ruleTester.run('sort-comp', rule, {
     }]
   }, {
     code: [
-      '// static lifecycle methods can be grouped (with statics)',
-      'class Hello extends React.Component {',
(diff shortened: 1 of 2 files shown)


metreniuk:
@alexzherdev what do you think about this? Is it accurate to do so?

ljharb:
It's definitely a breaking change; it'd be ideal to ship this now under an option, and then we can flip the option in the next major if desired.

metreniuk:
Since the current behavior was intended to allow the same method to be part of multiple groups I am willing to keep this as it is. At the same time there could be different opinions about where a method should be (ex: `getDerivedStateFromProps` before/after state declaration).
What do you think about naming the option `preferLifecycle` that will be implemented under the behavior from this PR?

metreniuk:
@alexzherdev sure. I've added the new functionality under the `preferLifecycle` option (`false` by default) and added more tests.

tests/lib/rules/sort-comp.js
```diff
@@ -468,20 +468,21 @@ ruleTester.run('sort-comp', rule, {
...
     code: [
-      '// static lifecycle methods can be grouped (with lifecycle)',
+      '// static lifecycle methods should be grouped under lifecycle with the preferLifecycle set to true',
       'class Hello extends React.Component {',
       '  constructor() {}',
       '  static getDerivedStateFromProps() {}',
       '}'
-    ].join('\n')
+    ].join('\n'),
+    options: [{preferLifecycle: true}]
```
ljharb: instead of modifying this test, can we duplicate it and modify the duplicate?

ljharb (changes requested):

PR #1973, opened 2018-09-05
[New] `sort-prop-types`: support comments on prop types

### LINKED ISSUE #1940: sort-prop-types fix to account for comments
I know comments on prop types are not in favour here, but for those who do have them, would it be acceptable to support sorting comments together with their prop types? i.e. currently the rule would fix
```js
{
  /** z */
  z: PropTypes.string,
  /** a */
  a: PropTypes.any
}
```
to
```js
propTypes: {
  /** z */
  a: PropTypes.any,
  /** a */
  z: PropTypes.string
}
```

alexzherdev:
Resolves #1940

### DIFF
--- a/lib/util/comments.js
+++ b/lib/util/comments.js
@@ -1,0 +1,76 @@
+/**
+ * @fileoverview Utility functions for comments handling.
+ */
+'use strict';
+
+/**
+ * Backports sourceCode.getCommentsBefore() for ESLint 3
+ *
+ * @param {Object} nodeOrToken Node or token to get comments for.
+ * @param {Object} sourceCode The SourceCode object.
+ * @returns {Array} Array of comment tokens.
+ */
+function getCommentsBefore(nodeOrToken, sourceCode) {
+  const token = sourceCode.getFirstToken(nodeOrToken, {includeComments: true});
+  let previousComments = [];
+
+  // ESLint >=4.x
+  if (sourceCode.getCommentsBefore) {
+    previousComments = sourceCode.getCommentsBefore(token);
+  // ESLint 3.x
+  } else {
+    let currentToken = token;
+    do {
+      const previousToken = sourceCode.getTokenBefore(currentToken);
+      const potentialComment = sourceCode.getTokenBefore(currentToken, {includeComments: true});
+
+      if (previousToken !== potentialComment) {
+        previousComments.push(potentialComment);
+        currentToken = potentialComment;
+      } else {
+        currentToken = null;
+      }
+    } while (currentToken);
+    previousComments = previousComments.reverse();
+  }
+
+  return previousComments;
+}
+
+/**
+ * Backports sourceCode.getCommentsAfter() for ESLint 3
+ *
+ * @param {Object} nodeOrToken Node or token to get comments for.
+ * @param {Object} sourceCode The SourceCode object.
+ * @returns {Array} Array of comment tokens.
+ */
+function getCommentsAfter(nodeOrToken, sourceCode) {
+  const token = sourceCode.getLastToken(nodeOrToken, {includeComments: true});
+  let nextComments = [];
+
+  // ESLint >=4.x
+  if (sourceCode.getCommentsAfter) {
+    nextComments = sourceCode.getCommentsAfter(token);
+  // ESLint 3.x
+  } else {
+    let currentToken = token;
+    do {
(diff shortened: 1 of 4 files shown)


lib/util/comments.js
```diff
@@ -0,0 +1,76 @@
...
+  let previousComments = [];
+
+  // ESLint >=4.x
+  if (sourceCode.getCommentsBefore) {
+    previousComments = sourceCode.getCommentsBefore(token);
+  // ESLint 3.x
+  } else {
+    let currentToken = token;
+    do {
+      const previousToken = sourceCode.getTokenBefore(currentToken);
```
alexzherdev: Note: this code was extracted from `jsx-curly-spacing` and for some reason this line (or the next) returns `undefined` on ESLint 3.0.0 😕 Works well on 3.19.0.

tests/lib/rules/sort-prop-types.js
```diff
@@ -1533,5 +1533,1330 @@ ruleTester.run('sort-prop-types', rule, {
       '  }',
       '});'
     ].join('\n')
+  }, {
+    code: `
```
alexzherdev: Suggestions for more test cases are welcome.

alexzherdev:
Hmm appveyor actually failed (node 5 choked on args spread somewhere inside eslint), but there's a green check here.

lib/util/comments.js
ljharb: If we backported the eslint API, would that work?

alexzherdev: Looks like [both `getLocFromIndex` and `getIndexFromLoc`](https://github.com/eslint/eslint/blob/da9174e0798c1d785ddabb3ae405860fc5b89311/lib/util/source-code.js#L434-L504) are relying heavily on `lineStartIndices` and `lines` in the `SourceCode` class that get populated once in the constructor. The ESLint docs explicitly say [we can use `lines` (bottom of the section)](https://eslint.org/docs/developer-guide/working-with-rules#contextgetsourcecode) but there's no word about `lineStartIndices`. Potentially we can just copy over those functions, but the bigger issue with `getTokenBefore`/`getTokenAfter` remains, and those don't seem possible to be backported.

ljharb: Is there a way we could runtime-detect these APIs, and provide a reasonable degraded experience for older eslint users?

alexzherdev: I'm skeptical (not that we could detect them, but that there is a reasonable experience beyond just not applying fixes), but let me look into that.

ljharb: yeah that's what i meant - applying fewer fixes when they weren't good enough.

amannn:
Is there something missing for this PR to get merged? I also encountered the problem described in the referenced issue.

alexzherdev:
@amannn I will be tackling https://github.com/yannickcr/eslint-plugin-react/pull/1973#discussion_r247757333 within a few days

amannn:
@alexzherdev That's great to hear! Thank you so much for your help!

It's really cool to have auto fixable `jsx-sort-props` and `sort-prop-types`. After that, there's only auto fixing for `jsx-sort-default-props` missing for me to go all-in on sorting props 🙂.

DNR500:
Any resolution on this PR yet? Would be super useful for anyone thats tools like https://github.com/storybookjs/babel-plugin-react-docgen

alexzherdev:
Haven’t had time to get back to it yet. The issue in https://github.com/yannickcr/eslint-plugin-react/pull/1973#discussion_r215113449 is a tricky one to make sense of.

amannn:
@alexzherdev Not sure if that's an option, but since even the test suite on master fails for that version, would it be an option to bump the peer dependency of `eslint` to 3.18.0?

Would be a shame if the support for that version would block helpful contributions like this PR. ESLint 3.0.0 was released on July 1, 2016 – that's three years ago. Even node.js LTS releases have a stop of maintenance after 3 years.

ljharb:
@amannn no, that'd be a breaking change.

Surely there's a way we can handle eslint 3.0 - even if that means gracefully degrading to the behavior prior to this PR.

ljharb:
@alexzherdev I'm going to disable the entire autofix behavior of this rule before the next release unless we can find a way to get this PR in; it'd be great if you had time to update this.

ljharb:
ping @alexzherdev, any update?

ljharb:
Closing in favor of #3471.

PR #1978, opened 2018-09-11
[New] Support "detect" option for React version setting

### LINKED ISSUE #1955: Warning: React version not specified in eslint-plugin-react settings.
Hello,

Since I've updated this plugin, i've got a message like `Warning: React version not specified in eslint-plugin-react settings.`.

This is true, I didn't specify the version of react, and this happens due to https://github.com/yannickcr/eslint-plugin-react/pull/1857/commits/8738e5917709c0cba0e7dd45289356f8a1d77384#diff-4e51c5c91adee02670adaca34f7fd7fdR17

However, when I read the doc, I've got this:

`"version": "15.0", // React version, default to the latest React stable release`

It's said "by default, it uses the latest version of react".

So why displaying a warning if there's a default and this is what I want? as I keep my react often upgraded.

alexzherdev:
I think this resolves #1955.

### DIFF
--- a/lib/util/version.js
+++ b/lib/util/version.js
@@ -4,15 +4,35 @@
  */
 'use strict';
 
+const resolve = require('resolve');
 const log = require('./log');
 
 let warnedForMissingVersion = false;
+
+function detectReactVersion() {
+  try {
+    const reactPath = resolve.sync('react', {basedir: process.cwd()});
+    const react = require(reactPath);
+    return react.version;
+  } catch (e) {
+    if (e.code === 'MODULE_NOT_FOUND') {
+      log('Warning: React version was set to "detect" in eslint-plugin-react settings, ' +
+        'but the "react" package is not installed. Assuming latest React version for linting.');
+      return '999.999.999';
+    }
+    throw e;
+  }
+}
 
 function getReactVersionFromContext(context) {
   let confVer = '999.999.999';
   // .eslintrc shared settings (http://eslint.org/docs/user-guide/configuring#adding-shared-settings)
   if (context.settings.react && context.settings.react.version) {
-    confVer = context.settings.react.version;
+    let settingsVersion = context.settings.react.version;
+    if (settingsVersion === 'detect') {
+      settingsVersion = detectReactVersion();
+    }
+    confVer = settingsVersion;
   } else if (!warnedForMissingVersion) {
     log('Warning: React version not specified in eslint-plugin-react settings. ' +
       'See https://github.com/yannickcr/eslint-plugin-react#configuration.');
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ logs
 reports
 build
 node_modules
+!tests/**/node_modules
 npm-debug.log
 sftp-config.json
 
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "lint": "eslint ./",
     "pretest": "npm run lint",
     "test": "npm run unit-test",
-    "unit-test": "istanbul cover --dir reports/coverage node_modules/mocha/bin/_mocha tests/**/*.js"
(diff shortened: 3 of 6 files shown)


lib/util/version.js
```diff
@@ -12,7 +13,23 @@ function getReactVersionFromContext(context) {
...
   // .eslintrc shared settings (http://eslint.org/docs/user-guide/configuring#adding-shared-settings)
   if (context.settings.react && context.settings.react.version) {
-    confVer = context.settings.react.version;
+    let settingsVer = context.settings.react.version;
+    if (settingsVer === 'detect') {
+      try {
+        const reactPath = resolve.sync('react', {basedir: process.cwd()});
+        const react = require(reactPath);
+        settingsVer = react.version;
+      } catch (e) {
```
alexzherdev: Not sure what (if anything) needs to happen here, basically made a best guess.

lib/util/version.js
```diff
@@ -12,7 +13,23 @@ function getReactVersionFromContext(context) {
...
+        if (e.code === 'MODULE_NOT_FOUND') {
+          log('Warning: React version was set to "detect" in eslint-plugin-react settings, ' +
+            'but the "react" package is not installed. Assuming latest React version for linting.');
+          settingsVer = '999.999.999';
```
ljharb: this should probably set `warnedForMissingVersion = true` as well?

lib/util/version.js
```diff
@@ -12,7 +13,23 @@ function getReactVersionFromContext(context) {
   let confVer = '999.999.999';
   // .eslintrc shared settings (http://eslint.org/docs/user-guide/configuring#adding-shared-settings)
   if (context.settings.react && context.settings.react.version) {
-    confVer = context.settings.react.version;
+    let settingsVer = context.settings.react.version;
+    if (settingsVer === 'detect') {
+      try {
+        const reactPath = resolve.sync('react', {basedir: process.cwd()});
```
ljharb: i think this basedir is the default? probably best to specify it either way tho

lib/util/version.js
alexzherdev: it can, but even without that execution would never go into the branch where we warn because we do have a value in `context.settings.react.version`.

lib/util/version.js
```diff
@@ -12,7 +13,23 @@ function getReactVersionFromContext(context) {
...
+        const react = require(reactPath);
+        settingsVer = react.version;
```
alexzherdev: @ljharb does it make sense to cache this value for subsequent access?

alexzherdev:
> This LGTM but is there any way to test for this?

I was gonna ask 😄
I could maybe imagine a unit-level test for `getReactVersionFromContext`. That would have to mock `resolve.sync` and `require` results - if this seems reasonable, any particular suggestions for libs to use for that?

lib/util/version.js
ljharb: ah, fair enough

lib/util/version.js
ljharb: yes, absolutely - good call.

ljharb:
hm, i don't think there is a good way :-/ however, you could construct a test directory with a package.json and a node_modules and a react dir with its own package.json, and require that?

alexzherdev:
Ah, I think I saw babel do something like that 👍

ljharb:
The import plugin does it in its tests, as well.

alexzherdev:
I added a couple of tests.
I tried caching the version into a local variable in the module and it worked, but then I didn't know how to reset that between different tests 🤔

sebastian-nowak:
Any chance this PR will be merged soon?

tgdev:
This looks great. Any idea on when this might be published to NPM?

alexzherdev:
I'd love to see it published too 😄 but see https://github.com/yannickcr/eslint-plugin-react/pull/2026#issuecomment-433959053
> There’s no policy and no set timeline; things get released when maintainers have time to do so.

ljharb:

PR #1980, opened 2018-09-13
[New] Add `no-arrow-function-lifecycle` rule

ngtan:
## Overview
I would like to add one more rule no-arrow-function-lifecycle which checks that it is not necessary to use arrow function for lifecycle methods and it will affect a bit performance.

## Movitation
I did code review some projects and found that our teammates use arrow function for lifecycle methods constantly. Beside that, I also found a ticket related to this issue. See [react/issues/10810](https://github.com/facebook/react/issues/10810) for more details. 

Consider the below code 

```
  render = () => {
    return <div />;
  }
}
```

Do you need to use arrow function here? I believe that, you will have the same idea that it shouldn't.

I'm eagerly looking forward to seeing your comments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,10 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ### Added
 * [`no-unused-class-component-methods`]: Handle unused class component methods ([#2166][] @jakeleventhal @pawelnvk)
+* add [`no-arrow-function-lifecycle`] ([#1980][] @ngtan)
 
 [#2166]: https://github.com/yannickcr/eslint-plugin-react/pull/2166
+[#1980]: https://github.com/yannickcr/eslint-plugin-react/pull/1980
 
 ## [7.26.1] - 2021.09.29
 
@@ -3465,6 +3467,7 @@ If you're still not using React 15 you can keep the old behavior by setting the 
 [`no-access-state-in-setstate`]: docs/rules/no-access-state-in-setstate.md
 [`no-adjacent-inline-elements`]: docs/rules/no-adjacent-inline-elements.md
 [`no-array-index-key`]: docs/rules/no-array-index-key.md
+[`no-arrow-function-lifecycle`]: docs/rules/no-arrow-function-lifecycle.md
 [`no-children-prop`]: docs/rules/no-children-prop.md
 [`no-comment-textnodes`]: docs/rules/jsx-no-comment-textnodes.md
 [`no-danger-with-children`]: docs/rules/no-danger-with-children.md
--- a/README.md
+++ b/README.md
@@ -135,6 +135,7 @@ Enable the rules that you would like to use.
 |  |  | [react/no-access-state-in-setstate](docs/rules/no-access-state-in-setstate.md) | Reports when this.state is accessed within setState |
 |  |  | [react/no-adjacent-inline-elements](docs/rules/no-adjacent-inline-elements.md) | Prevent adjacent inline elements not separated by whitespace. |
 |  |  | [react/no-array-index-key](docs/rules/no-array-index-key.md) | Prevent usage of Array index in keys |
+|  | 🔧 | [react/no-arrow-function-lifecycle](docs/rules/no-arrow-function-lifecycle.md) | Lifecycle methods should be methods on the prototype, not class fields |
 | ✔ |  | [react/no-children-prop](docs/rules/no-children-prop.md) | Prevent passing of children as props. |
 |  |  | [react/no-danger](docs/rules/no-danger.md) | Prevent usage of dangerous JSX props |
 | ✔ |  | [react/no-danger-with-children](docs/rules/no-danger-with-children.md) | Report when a DOM element is using both children and dangerouslySetInnerHTML |
--- a/lib/rules/no-arrow-function-lifecycle.js
+++ b/lib/rules/no-arrow-function-lifecycle.js
@@ -1,0 +1,80 @@
+/**
+ * @fileoverview Lifecycle methods should be methods on the prototype, not class fields
+ * @author Tan Nguyen
+ */
+
+'use strict';
+
+const values = require('object.values');
+
+const Components = require('../util/Components');
+const astUtil = require('../util/ast');
+const docsUrl = require('../util/docsUrl');
+const lifecycleMethods = require('../util/lifecycleMethods');
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Lifecycle methods should be methods on the prototype, not class fields',
+      category: 'Best Practices',
+      recommended: false,
+      url: docsUrl('no-arrow-function-lifecycle')
+    },
+    schema: [],
+    fixable: 'code'
+  },
+
+  create: Components.detect((context, components, utils) => {
+    function getText(node) {
+      const params = node.value.params.map((p) => p.name);
+
+      if (node.type === 'Property') {
+        return `: function(${params.join(', ')}) `;
+      }
+
+      if (node.type === 'ClassProperty') {
+        return `(${params.join(', ')}) `;
+      }
+
+      return null;
+    }
+
+    /**
+     * @param {Array} properties list of component properties
+     */
+    function reportNoArrowFunctionLifecycle(properties) {
+      properties.forEach((node) => {
+        const propertyName = astUtil.getPropertyName(node);
+        const nodeType = node.value.type;
+        const isLifecycleMethod = (
+          node.static && !utils.isES5Component(node)
+            ? lifecycleMethods.static
+            : lifecycleMethods.instance
+        ).indexOf(propertyName) > -1;
+
+        if (nodeType === 'ArrowFunctionExpression' && isLifecycleMethod) {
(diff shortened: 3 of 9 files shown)


ngtan:
@ljharb Thanks for your reviews. I think I have addressed your comments already. Could you check it out again? If you have any concerns, just leave your comment.

lib/rules/no-arrow-function-lifecycle.js
```diff
@@ -0,0 +1,74 @@
...
+  'componentWillUnmount',
+  'render'
+];
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'It is not necessary to use arrow function for lifecycle methods',
+      category: 'Best Practices',
+      recommended: true,
```
alexzherdev: Just to confirm, do we want this to be recommended? that would be a semver-major change, right?

ljharb: ah yes, it would - let's set this to false.

lib/rules/no-arrow-function-lifecycle.js
```diff
@@ -0,0 +1,74 @@
...
+    function reportNoArrowFunctionLifecycle(properties) {
+      properties.forEach(node => {
+        const propertyName = astUtil.getPropertyName(node);
+        const nodeType = node.value && node.value.type;
+        const isLifecycleMethod = lifecycleMethods.indexOf(propertyName) !== -1;
+
+        if (nodeType === 'ArrowFunctionExpression' && isLifecycleMethod) {
+          context.report({
+            node,
+            message: '{{propertyName}} is unnecessary to use arrow function',
```
ljharb: let's have this message be more like `{{propertyName}} is a React lifecycle method, and should not be an arrow function. Use an instance method instead.`

lib/rules/no-arrow-function-lifecycle.js
ljharb: good call, let's set this to false

ngtan:
@ljharb I really keen on your idea but, in my point of view it should be fixed by creator to make sure that they won't repeat it in next time. 

Beside that, I also figure out some common cases:

```
  render = () => {
    return (<div />);
  }
}
```
```
  render = () => (<div />);
}
```
```
  componentWillReceiveProps = nextProps => {}
  shouldComponentUpdate = (nextProps = { blabla }, nextState) => (returned)
  render() {
    return (<div />);
  }
}
```
To be honest, I didn't find out a good way to replace them with instance method yet. Anyway, I believe I will have a one to help it automatically fixed later after it was merged.

ljharb:
`$methodName($argList) { return $methodBody; }`?

ngtan:
> `$methodName($argList) { return $methodBody; }`?

 I also thought about it. But, have idea for now.

lib/rules/no-arrow-function-lifecycle.js
```diff
@@ -0,0 +1,74 @@
...
+  'UNSAFE_componentWillReceiveProps',
+  'shouldComponentUpdate',
+  'componentWillUpdate',
+  'UNSAFE_componentWillUpdate',
+  'getSnapshotBeforeUpdate',
+  'componentDidUpdate',
+  'componentDidCatch',
+  'componentWillUnmount',
+  'render'
+];
```
petersendidit: Most of this list is already defined elsewhere, should we make this list shareable so its easier to maintain?

https://github.com/yannickcr/eslint-plugin-react/blob/7dce90a4f570847b33eaefddb8d9bb7d6eca9dc5/lib/rules/no-typos.js#L14-L29

https://github.com/yannickcr/eslint-plugin-react/blob/6b239a7c71eb2d7060edd5f48f74f138ff89736e/lib/rules/no-unused-prop-types.js#L23-L24

alexzherdev:
The autofix would be so easy if we had AST-driven fixers 🙁 (although I'm guessing there's valid reasons we don't?)

lib/rules/no-arrow-function-lifecycle.js
```diff
@@ -7,26 +7,7 @@
...
-  'shouldComponentUpdate',
-  'componentWillUpdate',
-  'UNSAFE_componentWillUpdate',
-  'getSnapshotBeforeUpdate',
-  'componentDidUpdate',
-  'componentDidCatch',
-  'componentWillUnmount',
-  'render'
-];
+const lifecycleMethods = require('../util/lifecycleMethods');
```
ljharb: can we also use this list in the other places in the codebase that use lifecycle methods?

ngtan: Sure

ngtan:
> The autofix would be so easy if we had AST-driven fixers 🙁 (although I'm guessing there's valid reasons we don't?)

Honestly, I'm working on this.

golopot:
I think `no-class-fields-lifecycle` is a better name. The rule should also report on non-arrow-function like: `render = function() {…}`.

dylmye:
Any update on this? Thanks

niki-timofe:
➕

wheeler:
I'm going to add to the motivation for this rule. In my experience the `render = () => { ... }` pattern breaks `react-hot-loader` (it will render version `n-1` of the component where `n` is supposed to be the latest). I would **LOVE** to have this rule (because I was about to try writing it myself before finding this).

@ngtan // @ljharb This has been idle for a while - but I don't see any outstanding change requests. I'm interested in helping move this forward. Do we need to write a fixer or could that come in a subsequent patch?

ljharb:
@ngtan any interest in finishing this PR?

ngtan:
Hi, It has been a long time. I will make time this week to finish the PR.

CHANGELOG.md
```diff
@@ -3457,6 +3457,12 @@ If you're still not using React 15 you can keep the old behavior by setting the
...
 [`static-property-placement`]: docs/rules/static-property-placement.md
+<<<<<<< HEAD
 [`style-prop-object`]: docs/rules/style-prop-object.md
 [`void-dom-elements-no-children`]: docs/rules/void-dom-elements-no-children.md
 [`wrap-multilines`]: docs/rules/jsx-wrap-multilines.md
+=======
+[`jsx-curly-newline`]: docs/rules/jsx-curly-newline.md
+[`jsx-no-useless-fragment`]: docs/rules/jsx-no-useless-fragment.md
+[`no-arrow-function-lifecycle`]: docs/rules/no-arrow-function-lifecycle.md
+>>>>>>> [New] add `no-arrow-function-lifecycle`
```
ljharb, line 3468: seems like a bad merge conflict?
```suggestion
[`style-prop-object`]: docs/rules/style-prop-object.md
[`void-dom-elements-no-children`]: docs/rules/void-dom-elements-no-children.md
[`wrap-multilines`]: docs/rules/jsx-wrap-multilines.md
[`no-arrow-function-lifecycle`]: docs/rules/no-arrow-function-lifecycle.md
```

README.md
```diff
@@ -115,6 +115,7 @@ Enable the rules that you would like to use.
 
 # List of supported rules
 
+<<<<<<< HEAD
```
ljharb, line 118: for this one, please run `npm run generate-list-of-rules` and commit the output

lib/rules/no-typos.js
```diff
@@ -7,12 +7,14 @@
...
 const Components = require('../util/Components');
 const docsUrl = require('../util/docsUrl');
+const lifecycleMethods = require('../util/lifecycleMethods');
 
 // ------------------------------------------------------------------------------
 // Rule Definition
 // ------------------------------------------------------------------------------
 
 const STATIC_CLASS_PROPERTIES = ['propTypes', 'contextTypes', 'childContextTypes', 'defaultProps'];
+<<<<<<< HEAD
```
ljharb, line 17: this file i think you'll have to address manually

ljharb (changes requested):

lib/util/lifecycleMethods.js
```diff
@@ -0,0 +1,26 @@
...
+ * @fileoverview lifecycle methods
+ * @author Tan Nguyen
+ */
+
+'use strict';
+
+module.exports = [
+  'getDefaultProps',
+  'getInitialState',
+  'getChildContext',
```
ljharb, line 11: some of these are additions that will affect `no-typos` also. could we add some tests that cover those?

ljharb (changes requested):

lib/rules/no-arrow-function-lifecycle.js
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     * @param {Array} properties list of component properties
+     */
+    function reportNoArrowFunctionLifecycle(properties) {
+      properties.forEach((node) => {
+        const propertyName = astUtil.getPropertyName(node);
+        const nodeType = node.value.type;
+        const isLifecycleMethod = lifecycleMethods.includes(propertyName);
```
ljharb, line 47: please use the `array-includes` package; node 4 doesn't have `.includes`

lib/util/lifecycleMethods.js
ngtan: @ljharb it already updated, please lemme know if anything needs to be updated

ljharb (approved):

PR #1984, opened 2018-09-16
jsx-wrap-multilines now catches single missing newlines

### LINKED ISSUE #1965: jsx-wrap-multilines wrong behavior of parens-new-line
Setting `{ return: 'parens-new-line' }` option should validate that there is a new line after `(` and before `)`.
So following code is the valid code:
```
return (
    <div>
        <div>example</div>
    <div>
)
```

But following code also considered as valid whereas as it isn't:
```
return (<div>
        <div>example</div>
    <div>
)
```

and this one also:
```
return (
    <div>
        <div>example</div>
    <div>)
```

So, the issue occurs when there is one parenthesis on a separate line and another one on the same line with the JSX code.

MrHen:
Fixes #1965. The only interesting question is how the fixer should treat the newlines. I chose to make it add a new line if there was exactly zero after or before the parens. This means:

```
  var Hello = createReactClass({
    render: function() {
      return (<div>
        <p>Hello {this.props.name}</p>
      </div>
      );
    }
  });
```

Fixes to:

```
  var Hello = createReactClass({
    render: function() {
      return (
<div>
        <p>Hello {this.props.name}</p>
      </div>
      );
    }
  });
```

----

And:

```
  var Hello = createReactClass({
    render: function() {
      return (<div>
        <p>Hello {this.props.name}</p>
      </div>

      );
    }
  });
```

Fixes to:

```
  var Hello = createReactClass({
    render: function() {
      return (
<div>
        <p>Hello {this.props.name}</p>
      </div>

      );
    }
  });
```

----

The alternative would be to make exactly _one_ newline regardless of how many there were before -- but I assumed such a behavior should be its own rule instead of attached to this rule.

### DIFF
--- a/lib/rules/jsx-wrap-multilines.js
+++ b/lib/rules/jsx-wrap-multilines.js
@@ -98,9 +98,19 @@ module.exports = {
       const previousToken = sourceCode.getTokenBefore(node);
       const nextToken = sourceCode.getTokenAfter(node);
 
-      return isParenthesised(node) &&
-        previousToken.loc.end.line === node.loc.start.line &&
-        node.loc.end.line === nextToken.loc.end.line;
+      if (!isParenthesised(node)) {
+        return false;
+      }
+
+      if (previousToken.loc.end.line === node.loc.start.line) {
+        return true;
+      }
+
+      if (node.loc.end.line === nextToken.loc.end.line) {
+        return true;
+      }
+
+      return false;
     }
 
     function isMultilines(node) {
@@ -151,7 +161,11 @@ module.exports = {
             report(node, MISSING_PARENS, fixer => fixer.replaceText(node, `(\n${sourceCode.getText(node)}\n)`));
           }
         } else if (needsNewLines(node)) {
-          report(node, PARENS_NEW_LINES, fixer => fixer.replaceText(node, `\n${sourceCode.getText(node)}\n`));
+          report(node, PARENS_NEW_LINES, fixer => {
+            const text = sourceCode.getText(node);
+            const fixed = text.replace(/^\n?((.|\n)*?)\n?$/, '\n$1\n');
+            return fixer.replaceText(node, fixed);
+          });
         }
       }
     }
--- a/lib/rules/jsx-wrap-multilines.js
+++ b/lib/rules/jsx-wrap-multilines.js
@@ -94,9 +94,8 @@ module.exports = {
         nextToken.value === ')' && nextToken.range[0] >= node.range[1];
     }
 
-    function needsNewLines(node) {
+    function needsOpeningNewLine(node) {
       const previousToken = sourceCode.getTokenBefore(node);
-      const nextToken = sourceCode.getTokenAfter(node);
 
       if (!isParenthesised(node)) {
         return false;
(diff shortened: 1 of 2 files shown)


tests/lib/rules/jsx-wrap-multilines.js
```diff
@@ -501,7 +525,7 @@ const ATTR_PAREN_NEW_LINE_AUTOFIX_FRAGMENT = `
 `;
 
 function addNewLineSymbols(code) {
-  return code.replace(/\(</g, '(\n<').replace(/>\)/g, '>\n)');
+  return code.replace(/\((\s*)</g, '($1\n<').replace(/>(\s*)\)/g, '>\n$1)');
```
ljharb: i know this code is already in master, but this makes the test cases very inscrutable :-/

i feel like i'd much rather see an explicit `output` example for each input.

lib/rules/jsx-wrap-multilines.js
```diff
@@ -98,9 +98,19 @@ module.exports = {
       const previousToken = sourceCode.getTokenBefore(node);
       const nextToken = sourceCode.getTokenAfter(node);
 
-      return isParenthesised(node) &&
-        previousToken.loc.end.line === node.loc.start.line &&
```
ljharb: it seems like the effective change here is changing `&&` to `||`?

tests/lib/rules/jsx-wrap-multilines.js
MrHen: Yeah, me too. I can add explicit output for the tests touched by this PR.

lib/rules/jsx-wrap-multilines.js
MrHen: Yes. I find mixing `&&` and `||` harder to read but if you prefer it that way I can switch it back.

ljharb: Nah this is good, just making sure i understand.

MrHen: Apparently I didn't quite understand how the nodes worked with whitespace. I had to break this up into separate functions since the newlines were getting attached to the previous token.

tests/lib/rules/jsx-wrap-multilines.js
```diff
@@ -86,6 +86,56 @@ const RETURN_PAREN_NEW_LINE = `
   });
 `;
 
+const RETURN_PAREN_NEW_LINE_OPENING = `
+  var Hello = createReactClass({
+    render: function() {
+      return (
+
```
ljharb: These extra newlines here seem odd - why is this allowed to be here?

MrHen: See my earlier comment: 
https://github.com/yannickcr/eslint-plugin-react/pull/1984#issue-215844164

My first attempts at fixing this caused extraneous newlines to disappear which seemed like an undesired side-effect of this particular rule. The extra newline isn't being added by the fix -- its in the original test condition. The test is making sure the extra newline is _not_ being stripped.

ljharb: That makes sense (although I’d want it stripped, this probably isn’t the rule for it)

MrHen:
Thanks for the review @ljharb. What's the next step to getting this merged? Anything I need to do?

Issue #1986, opened 2018-09-17
Should enforceDynamicLinks be react/jsx-no-target-blank?


atomcorp:
Maybe I'm missing something, but for the [jsx-no-target-blank](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md) rule it says: 

```markdown
There are two main options for the rule:

{"enforceDynamicLinks": "always"} enforces the rule if the href is a dynamic link (default)
{"enforceDynamicLinks": "never"} does not enforce the rule if the href is a dynamic link
```

This didn't do anything for me. `"react/jsx-no-target-blank": 2,` works fine, but it wasn't really clear if that is indeed what you're meant to do.

Thanks

atomcorp:
FWIW I'm using: 

``` json    
"eslint": "^5.6.0",
"eslint-plugin-react": "^7.11.1",
````

ljharb:
To set that option, configure the rule with `[2, { “enforceDynamicLinks”: “always” }]` instead of just 2.

atomcorp:
Thanks for the information!

Personally I find that doc page pretty unclear if you just quickly looking for the rule. I'll try and update it a little later on, similar to [this style](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md#rule-options)

PR #1988, opened 2018-09-18
Add full example in react/jsx-no-target-blank

atomcorp:
See #1986 

I found the docs a bit unclear, so I've added a fuller example, hopefully in line with other examples.

### DIFF
--- a/docs/rules/jsx-no-target-blank.md
+++ b/docs/rules/jsx-no-target-blank.md
@@ -15,35 +15,34 @@ This rule aims to prevent user generated links from creating security vulnerabil
 
 There are two main options for the rule:
 
-- `{"enforceDynamicLinks": "always"}` enforces the rule if the href is a dynamic link (default)
-- `{"enforceDynamicLinks": "never"}` does not enforce the rule if the href is a dynamic link
+* `{"enforceDynamicLinks": "always"}` enforces the rule if the href is a dynamic link (default)
+* `{"enforceDynamicLinks": "never"}` does not enforce the rule if the href is a dynamic link
 
 ```json
 "react/jsx-no-target-blank": [<enabled>, { enforceDynamicLinks: <enforce> }]
 ```
 
-- enabled: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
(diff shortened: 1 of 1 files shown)


docs/rules/jsx-no-target-blank.md
```diff
@@ -18,6 +18,13 @@ There are two main options for the rule:
 * `{"enforceDynamicLinks": "always"}` enforces the rule if the href is a dynamic link (default)
 * `{"enforceDynamicLinks": "never"}` does not enforce the rule if the href is a dynamic link
 
+```json
+"react/jsx-no-target-blank": [<enabled>, { enforceDynamicLinks: <enforce> }]
```
ljharb: the key needs to be quoted, since it's json

docs/rules/jsx-no-target-blank.md
```diff
@@ -18,6 +18,13 @@ There are two main options for the rule:
...
+```
+
+* enabled: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
```
ljharb: is this line necessary, since it's how every single eslint rule is configured?

docs/rules/jsx-no-target-blank.md
```diff
@@ -18,6 +18,13 @@ There are two main options for the rule:
...
+* enforce: optional string, defaults to "always"
```
ljharb: this seems redundant, since it's on lines 18-19

docs/rules/jsx-no-target-blank.md
atomcorp: It's used on other pages (I copied it from one). For example: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-optimization.md

docs/rules/jsx-no-target-blank.md
atomcorp: Again, I'm following how it works on some other pages

atomcorp:
Made that change. I looked at some of the other docs more carefully, and reordered the Rule Options so hopefully it makes a bit more sense.

PR #1989, opened 2018-09-19
Fix `no-this-in-sfc` rule behavior for arrow functions inside a class field

### LINKED ISSUE #1967: no-this-in-sfc false positives

sergei-startsev:
Fixed #1967. Adjusted component detection in `getParentStatelessComponent` for arrow functions inside a class field, e.g.:

```js
class Foo {
  bar() { 
    () => {
      this.something();
      return null;
    };
  }
}
```
See details in #1972.

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -469,7 +469,13 @@ function componentRule(rule, context) {
         const node = scope.block;
         const isClass = node.type === 'ClassExpression';
         const isFunction = /Function/.test(node.type); // Functions
-        const isMethod = node.parent && node.parent.type === 'MethodDefinition'; // Classes methods
+        const isArrowFunction = node.type === 'ArrowFunctionExpression';
+        let functionScope = scope;
+        if (isArrowFunction) {
+          functionScope = utils.getParentFunctionScope(scope);
+        }
+        const methodNode = functionScope && functionScope.block.parent;
+        const isMethod = methodNode && methodNode.type === 'MethodDefinition'; // Classes methods
         const isArgument = node.parent && node.parent.type === 'CallExpression'; // Arguments (callback, etc.)
         // Attribute Expressions inside JSX Elements (<button onClick={() => props.handleClick()}></button>)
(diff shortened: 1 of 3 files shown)


lib/util/Components.js
```diff
@@ -486,6 +492,23 @@ function componentRule(rule, context) {
...
 
+    /**
+     * Get a parent scope created by a FunctionExpression or FunctionDeclaration
+     * @param {Scope} scope The child scope
+     * @returns {Scope} A parent function scope
+     */
+    getParentFunctionScope(scope) {
+      scope = scope.upper;
+      while (scope) {
+        const type = scope.block.type;
```
ljharb: will scope always have an object `block` property?

lib/util/Components.js
```diff
@@ -486,6 +492,23 @@ function componentRule(rule, context) {
...
+        if (type === 'FunctionExpression' || type === 'FunctionExpression') {
```
ljharb: Is this meant to be checking the same thing twice?

lib/util/Components.js
sergei-startsev: I think yes, here's the [proof](https://github.com/eslint/eslint-scope/blob/3e69fdcf102082f3b3c419d0864d7ed90323f4ca/lib/scope.js#L650). There's also no extra checks in [the code above](https://github.com/yannickcr/eslint-plugin-react/pull/1989/files#diff-bafb3355a409c4ba076a9e3d609ad65bR470).

lib/util/Components.js
sergei-startsev: good catch, it should be `FunctionExpression` and `FunctionDeclaration`

sergei-startsev:
Fixed typo and added one more more test.

ljharb:
I don't see any tests of function declarations tho (arrows are only ever expressions)

PR #1994, opened 2018-09-23
[New] add rule to enforce fragment syntax

### LINKED ISSUE #1661: Rule Proposal: JSX Fragment shorthand
Hey there, apologies if this was brought up in the past but I'd like to propose a new rule for the [ReactFragment](https://reactjs.org/docs/fragments.html) feature in React 16. Specifically, the rule would allow you to select between either the [shorthand](https://reactjs.org/docs/fragments.html#short-syntax) version `<>` or the `<React.Fragment>` version.

alexzherdev:
Resolves #1661.


README.md
```diff
@@ -159,6 +159,7 @@ Enable the rules that you would like to use.
 * [react/jsx-one-expression-per-line](docs/rules/jsx-one-expression-per-line.md): Limit to one expression per line in JSX
 * [react/jsx-curly-brace-presence](docs/rules/jsx-curly-brace-presence.md): Enforce curly braces or disallow unnecessary curly braces in JSX
 * [react/jsx-pascal-case](docs/rules/jsx-pascal-case.md): Enforce PascalCase for user-defined JSX components
+* [react/jsx-prefer-fragment-shorthand](docs/rules/jsx-prefer-fragment-shorthand.md): Enforce standard or shorthand syntax for React fragments
```
ljharb: maybe “prefer” shouldn’t be in the rule name, since that’s only one of the settings

docs/rules/jsx-prefer-fragment-shorthand.md
```diff
@@ -0,0 +1,57 @@
...
+
+```js
+...
+"react/jsx-prefer-fragment-shorthand": [<enabled>, <mode>]
+...
+```
+
+### `always` mode
+
+This is the default mode. It will enforce the shorthand syntax for React fragments, with one exception. [Keys or attributes are not supported by the shorthand syntax][short_syntax], so the rule will not warn on standard-syntax fragments that use those.
```
ljharb: What attributes besides “key” are possible here?

lib/rules/jsx-prefer-fragment-shorthand.js
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    // --------------------------------------------------------------------------
+    // Public
+    // --------------------------------------------------------------------------
+
+    return {
+      JSXElement(node) {
+        const openingEl = node.openingElement;
+        if (elementType(openingEl) === `${reactPragma}.${fragmentPragma}`) {
```
ljharb: What if someone already has `<Fragment`? I’d expect to warn on that, if it was destructured from React.

README.md
alexzherdev: `jsx-fragment-syntax` of either `'shorthand'` or `'standard'`? (Also not sure what’s the better default)

docs/rules/jsx-prefer-fragment-shorthand.md
alexzherdev: Currently none, but the docs hint at event handlers in the future: https://reactjs.org/docs/fragments.html#keyed-fragments

lib/rules/jsx-prefer-fragment-shorthand.js
alexzherdev: Oh. That’s what I was trying to figure out in #1661, but I took https://github.com/yannickcr/eslint-plugin-react/issues/1661#issuecomment-423386464 to mean we don’t have to handle that for now?

README.md
ljharb: Technically React.Fragment isn’t syntax, it’s api - maybe `jsx-fragments` with either “syntax” or “element” or something.

I’d say the default should probably be “syntax” even though that requires babel 7.

lib/rules/jsx-prefer-fragment-shorthand.js
ljharb: I don’t think we have to handle it in the autofixer - but i think we do need to handle it in terms of warning.

README.md
alexzherdev: Ah, I see what you mean 👍

lib/rules/jsx-prefer-fragment-shorthand.js
alexzherdev: Also
```js
const { Fragment } = require('react');
```

lib/rules/jsx-fragments.js
```diff
@@ -0,0 +1,179 @@
...
+          });
+        }
+      },
+
+      ImportDeclaration(node) {
+        if (node.source && node.source.value === 'react') {
+          node.specifiers.forEach(spec => {
+            if (spec.imported && spec.imported.name === fragmentPragma) {
+              if (spec.local) {
+                fragmentNames.push(spec.local.name);
```
ljharb: Is it worth using a Set here instead of an array?

alexzherdev: Sure - that would communicate the intent better (and probably make no difference performance-wise given the data size here). Just haven't gotten used to the new collections

lib/rules/jsx-fragments.js
```diff
@@ -0,0 +1,179 @@
...
+    const openFragShort = '<>';
+    const closeFragShort = '</>';
+    const openFragLong = `<${reactPragma}.${fragmentPragma}>`;
+    const closeFragLong = `</${reactPragma}.${fragmentPragma}>`;
+
+    function reportOnReactVersion(node) {
+      if (!versionUtil.testReactVersion(context, '16.2.0')) {
+        context.report({
+          node,
+          message: 'Fragments are only supported starting from React v16.2'
```
EvHaus: Could we add something along the lines of:

```
Please disable the `jsx-fragments` rule or upgrade your version of React.
```

I think it's always a good idea to give users an action to do to help them resolve the error/warning.

ljharb: That sounds like a great improvement.

PR #1995, opened 2018-09-24
Fix `no-this-in-sfc` for class properties

### LINKED ISSUE #1960: Getting unrelated (react/no-this-in-sfc) violation error
I ran the following simple js program. But I got an error at line this.getA(); which is totally unrelated here. The error said "stateless functional components should not use this (react/no-this-in-sfc)"

```
class Demo {
    getA = () => 'a';

    hasBug = () => {
        this.getA();

        return null;
    };
}
```

What did you expect to happen?
No error expected

What actually happened? Please include the actual, raw output from ESLint.
Got this error - stateless functional components should not use this (react/no-this-in-sfc) for line  ```this.getA()```

sergei-startsev:
The PR fixes #1960 - `no-this-in-sfc` rule for class properties.

Class properties example:
```js
class Foo {
  bar = () => {
    this.something();
    return null;
  };
}
```

The corresponding AST: [link](https://astexplorer.net/#/gist/01be31cd9af1884c028099be843c3bab/ade0ae41ec5b2a03737618519741454bdaa0dc33).

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -457,14 +457,13 @@ function componentRule(rule, context) {
       while (scope) {
         const node = scope.block;
         const isFunction = /Function/.test(node.type); // Functions
-        const isArrowFunction = node.type === 'ArrowFunctionExpression';
+        const isArrowFunction = astUtil.isArrowFunction(node);
         let enclosingScope = scope;
         if (isArrowFunction) {
           enclosingScope = utils.getArrowFunctionScope(scope);
         }
-        const enclosingScopeType = enclosingScope && enclosingScope.block.type;
         const enclosingScopeParent = enclosingScope && enclosingScope.block.parent;
-        const isClass = enclosingScopeType === 'ClassDeclaration' || enclosingScopeType === 'ClassExpression';
+        const isClass = enclosingScope && astUtil.isClass(enclosingScope.block);
         const isMethod = enclosingScopeParent && enclosingScopeParent.type === 'MethodDefinition'; // Classes methods
         const isArgument = node.parent && node.parent.type === 'CallExpression'; // Arguments (callback, etc.)
         // Attribute Expressions inside JSX Elements (<button onClick={() => props.handleClick()}></button>)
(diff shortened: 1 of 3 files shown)


lib/util/Components.js
```diff
@@ -482,15 +483,16 @@ function componentRule(rule, context) {
...
+     * @param {Scope} scope Current scope
+     * @returns {Scope} An enclosing scope used by an arrow function
      */
-    getParentFunctionScope(scope) {
+    getArrowFunctionScope(scope) {
       scope = scope.upper;
       while (scope) {
         const type = scope.block.type;
-        if (type === 'FunctionExpression' || type === 'FunctionDeclaration') {
+        if (type === 'FunctionExpression' || type === 'FunctionDeclaration'
```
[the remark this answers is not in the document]
sergei-startsev: I didn't find `isFunction`, only `isFunctionLikeExpression` in `ast.js`, so extracted some util functions for reusing in `Components.js`

lib/util/Components.js
```diff
@@ -456,15 +456,15 @@ function componentRule(rule, context) {
...
         const node = scope.block;
-        const isClass = node.type === 'ClassExpression';
         const isFunction = /Function/.test(node.type); // Functions
-        const isArrowFunction = node.type === 'ArrowFunctionExpression';
-        let functionScope = scope;
+        const isArrowFunction = astUtil.isArrowFunction(node);
+        let enclosingScope = scope;
         if (isArrowFunction) {
-          functionScope = utils.getParentFunctionScope(scope);
+          enclosingScope = utils.getArrowFunctionScope(scope);
```
ljharb: this seems like it could be `const enclosingScope = isArrowFunction ? utils.getArrowFunctionScope(scope) : scope;`

tests/lib/rules/no-this-in-sfc.js
```diff
@@ -16,6 +16,8 @@ const ERROR_MESSAGE = 'Stateless functional components should not use this';
 const rule = require('../../../lib/rules/no-this-in-sfc');
 const RuleTester = require('eslint').RuleTester;
 
+require('babel-eslint');
```
alexzherdev: @ljharb do you know if this is really necessary? Not all of the test files that use this parser actually have the require.

ljharb: if you run those test files in isolation, do they pass?

sergei-startsev: I don't see any differences after removing `require('babel-eslint')`. See the proof #2004 . Also removed it here.

lib/util/Components.js
sergei-startsev: sure thing, fixed

PR #1996, opened 2018-09-25
Supporting async/await in `no-did-mount-setstate` rule

### LINKED ISSUE #1110: Should react/no-did-mount-set-state trigger on ES7 async componentDidMount?
The rule `react/no-did-mount-set-state` triggers error on async componentDidMount function for a deferred `setState`. Basically, `this.setState()` is being called here after an async action has been completed. It is not re-rendering straight after mounting, but only after some data has been retrieved from a server. Is the rule maybe catching the wrong intent?

```
async componentDidMount() {
    try {
      await someAsyncAction();
    } catch (e) {
      this.setState({errorMessage: 'Failed to load data!'});
    }
  }
```

idanen:
Pretty naive approach: checking if the call is in an `async` method and if an `await` is found before `setState` (going over tokens).

Fixes #1110.

### DIFF
--- a/docs/rules/no-did-mount-set-state.md
+++ b/docs/rules/no-did-mount-set-state.md
@@ -49,6 +49,20 @@ var Hello = createReactClass({
 });
 ```
 
+```jsx
+var Hello = createReactClass({
+  componentDidMount: async function() {
+    const newName = await getName();
+    this.setState({
+      name: newName
+    });
+  },
+  render: function() {
+    return <div>Hello {this.state.name}</div>;
+  }
+});
+```
+
 ## Rule Options
 
 ```js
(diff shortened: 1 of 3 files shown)


ljharb (changes requested):
Thanks, this is a great start, but I think it'd need lots more test cases.

idanen:
Sure. Do you have some in mind?

ljharb:
Things mentioned in the linked issue; `this.setState` defined in a callback function before the first await, but not invoked or passed to anything prior to the first `await`; things like that.

idanen:
That's much more difficult...
Do you think we should use a different "entry" than what is currently used for this rule?
I don't really have any experience or knowledge in AST so I think I'll need some help to continue

ljharb:
I feel like we may have to track all `await`s inside cDM, as well as all this.setState calls, and after they’re all tracked, figure out which ones are a problem.

It’s certainly a very hard problem.

PR #1997, opened 2018-09-25
Allow matching only beginning or ending of allowedFor items [forbid-component-props]

mikkelrd:


docs/rules/forbid-component-props.md
```diff
@@ -47,7 +47,7 @@ Each array element can either be a string with the property name or object speci
 ```js
 {
   "propName": "someProp",
-  "allowedFor": [SomeComponent, AnotherComponent]
+  "allowedFor": [SomeComponent, "AnotherComponent", "/^Svg/"]
```
ljharb: i have a strong aversion to allowing users to provide regexes; would allowing `*` be sufficient?

lib/rules/forbid-component-props.js
```diff
@@ -62,10 +62,21 @@ module.exports = {
       return [propName, whitelist];
     }));
 
+    function isAllowed(whitelist, tagName) {
+      return whitelist.some(item => {
+        if (item.startsWith('/')) {
+          const re = new RegExp(item.slice(1, item.length - 1));
```
ljharb: `item.slice(1, -1)`

docs/rules/forbid-component-props.md
mikkelrd: I think that would be flexible enough. May I ask why you don't like it more open-ended?

mikkelrd: Also, are you referring to allowing wildcard only at the beginning of the string? End? Both? Multiple? E.g. `*Icon*`

ljharb: I think at any position would be fine - altho i'm not sure how complex this makes the algorithm.

It'd also be fine to start with just "at the beginning" and/or "at the end", to be simpler, and allow use only of startsWith/endsWith.

ljharb: Escaping regular expressions is a very very hard problem, and passing user input into the RegExp constructor will likely be a bug farm, where the answer used to close the issue will be "that's not a valid regex", over and over.

lib/rules/forbid-component-props.js
```diff
@@ -62,10 +62,22 @@ module.exports = {
       return [propName, whitelist];
     }));
 
+    function isAllowed(whitelist, tagName) {
+      return whitelist.some(item => {
+        if (item.startsWith('*')) {
+          return tagName.endsWith(item.slice(1));
+        } else if (item.endsWith('*')) {
```
ljharb: is it worth supporting both, as well? ie, `*Foo*`?

ljharb:
ping @mikkelrd, are you still interested in completing this PR?

PR #2006, opened 2018-10-05
Update no-typos rule to check static lifecycle methods

bsonntag:
This updates the `"no-typos"` rule so that it checks for lifecycle methods that should be static, like `getDerivedStateFromProps`.

Closes https://github.com/yannickcr/eslint-plugin-react/issues/1999

### DIFF
--- a/lib/rules/no-typos.js
+++ b/lib/rules/no-typos.js
@@ -49,7 +49,8 @@ module.exports = {
       if (node.name !== 'isRequired') {
         context.report({
           node,
-          message: `Typo in prop type chain qualifier: ${node.name}`
+          message: 'Typo in prop type chain qualifier: {{name}}',
+          data: {name: node.name}
         });
       }
     }
@@ -58,7 +59,8 @@ module.exports = {
       if (node.name && !PROP_TYPES.some(propTypeName => propTypeName === node.name)) {
         context.report({
           node,
-          message: `Typo in declared prop type: ${node.name}`
+          message: 'Typo in declared prop type: {{name}}',
+          data: {name: node.name}
         });
       }
     }
@@ -151,7 +153,8 @@ module.exports = {
         if (method.toLowerCase() === nodeKeyName.toLowerCase() && method !== nodeKeyName) {
           context.report({
             node,
-            message: 'Typo in component lifecycle method declaration'
+            message: 'Typo in component lifecycle method declaration: {{actual}} should be {{expected}}',
+            data: {actual: nodeKeyName, expected: method}
           });
         }
       });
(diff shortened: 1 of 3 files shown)


lib/rules/no-typos.js
```diff
@@ -149,6 +150,17 @@ module.exports = {
       });
     }
 
+    function reportErrorIfStaticLifecycleMethod(node) {
+      STATIC_LIFECYCLE_METHODS.forEach(method => {
+        if (method.toLowerCase() === node.key.name.toLowerCase() && !node.static) {
```
ljharb: this line doesn't match line 144 above; i think you need to also check that `method !== node.key.name`?

tests/lib/rules/no-typos.js
```diff
@@ -1369,6 +1369,17 @@ ruleTester.run('no-typos', rule, {
     }, {
       message: 'Typo in declared prop type: objectof'
     }]
+  }, {
+    code: `
+      class Hello extends React.Component {
+        getDerivedStateFromProps() { }
```
ljharb: let's please add a "valid" rule for `static getDerivedStateFromProps`, and also an invalid one for `static getDERIVEDStateFromProps` :-)

lib/rules/no-typos.js
bsonntag: Did you mean `method === node.key.name`?

Also, does it make sense to do both checks? What we want to do here is check if the current node is a static lifecycle method.

I'm using `.toLowerCase()` so that this is caught sooner by the user. For example, the following code would give two errors (note: I should add a test for this):

```js
class Foo extends Component {
  GetDerivedStateFromProps() {}
}
```

One for the typo and other for the missing `static` keyword.

If instead we do `method === node.key.name`, the user will first only see the typo error, and after fixing it would see the missing `static` keyword error.

tests/lib/rules/no-typos.js
bsonntag: I added a test to check that `static getDerivedStateFromProps` is valid.

Regarding the second, there is already a test that checks `static GetDerivedStateFromProps` is invalid. Is it necessary to check `static getDERIVEDStateFromProps` too?

lib/rules/no-typos.js
ljharb: no, i mean `!==` - if the name is equal when lowercased, but *the same*, then it isn't a typo and shouldn't be reported.

tests/lib/rules/no-typos.js
ljharb: as long as the checks cover a) the proper spelling and capitalization, b) any miscapitalization, c) any misspelling, and then those three with 1) static and 2) instance, then we should be good.

lib/rules/no-typos.js
bsonntag: But I'm not reporting typos on the name here, I'm reporting missing `static` keywords.

bsonntag: Yes, the other check covers that.

ljharb:
@bsonntag i've rebased this; the tests are still failing. are you still interested in completing this PR?

bsonntag:
@ljharb not really, sorry.

PR #2008, opened 2018-10-05
Adding option to print node value

### LINKED ISSUE #2007: Printing Node value to the report
Hi guys, this is not a bug, is more like a new feature that I needed for a project I'm working on.

Basically I needed to detect and print to the report the hardcoded text found in a bunch of JSX templates so I did some tweaks to the jsx-no-literals rule to be able to print the node value.

I'm just leaving this here in case somebody else might need something like that, I'll be creating a PR for this. :)

jlgonzalezdev:
Basically I did some tweaks to the jsx-no-literals rule to be able to print the node value. :)

Fixes #2007.

### DIFF
--- a/lib/rules/jsx-no-literals.js
+++ b/lib/rules/jsx-no-literals.js
@@ -33,6 +33,7 @@ module.exports = {
 
   create: function(context) {
     const isNoStrings = context.options[0] ? context.options[0].noStrings : false;
+    const sourceCode = context.getSourceCode();
 
     const message = isNoStrings ?
       'Strings not allowed in JSX files' :
@@ -41,7 +42,7 @@ module.exports = {
     function reportLiteralNode(node) {
       context.report({
         node: node,
-        message: message
+        message: `${message}: “${sourceCode.getText(node).trim()}”`
       });
     }
 
--- a/tests/lib/rules/jsx-no-literals.js
+++ b/tests/lib/rules/jsx-no-literals.js
@@ -24,6 +24,14 @@ const parserOptions = {
 // Tests
 // ------------------------------------------------------------------------------
 
+function stringsMessage(str) {
+  return `Strings not allowed in JSX files: “${str}”`;
+}
+
+function jsxMessage(str) {
+  return `Missing JSX expression container around literal string: “${str}”`;
+}
+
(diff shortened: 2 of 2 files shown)


lib/rules/jsx-no-literals.js
```diff
@@ -25,23 +25,29 @@ module.exports = {
       properties: {
         noStrings: {
           type: 'boolean'
+        },
+        printNodeValue: {
```
ljharb: i do not think it makes any sense to add this as a configuration option; if it's useful, it should be in the message for everyone.

jlgonzalezdev:
Yes, Basically My use case is:
I need to print to the report the hardcoded text found in the JSX templates.
for something like  <button>SIGN IN</button>
I need an output like:
Missing JSX expression container around literal string: SIGN IN.

So, I need to print the node value("SIGN IN " in the example above).

I think that context.getSourceCode() instead of node.value would be the right approach for this, I'll be making some changes to my fork.

ljharb:
@jlgonzalezdev let's remove the option and just do this by default; i don't see any reason not to.

jlgonzalezdev:
@ljharb  ok, removing the option

lib/rules/jsx-no-literals.js
```diff
@@ -41,7 +42,7 @@ module.exports = {
     function reportLiteralNode(node) {
       context.report({
         node: node,
-        message: message
+        message: `${message}:${sourceCode.getText(node).trim()}`
```
ljharb: Let’s add a space after the colon, and it’d be good to have some kind of boundary markers around the text - backticks or curly quotes, maybe?

Separately, I’m not sure if the trimming is a good idea - if it’s a literal, the whitespace could be important.

jlgonzalezdev: > Let’s add a space after the colon, and it’d be good to have some kind of boundary markers around the text - backticks or curly quotes, maybe?

Sure, let's use it like this->   : "TEXT"

> Separately, I’m not sure if the trimming is a good idea - if it’s a literal, the whitespace could be important.

I think Not having the trimming really impact the readability of the output, in every case that I have work with this kind of output, I have needed to use the trimming. So, I woud like to keep it.

PR #2016, opened 2018-10-13
more friendly message for no-unescaped-entities

stevemao:


lib/rules/no-unescaped-entities.js
```diff
@@ -59,10 +71,18 @@ module.exports = {
         for (let j = 0; j < entities.length; j++) {
           for (let index = 0; index < rawLine.length; index++) {
             const c = rawLine[index];
-            if (c === entities[j]) {
+            if (typeof entities[j] === 'string') {
```
ljharb: what is this line for? i don't see any new tests.

stevemao: The structure of `DEFAULTS` changed to an array of objects. I'll add some tests.

stevemao:
BTW, why is `&` not in the default?

lib/rules/no-unescaped-entities.js
stevemao: I have added a test which takes an option, `forbid`.

stevemao:
@ljharb Do you also want to change the option to be `array` of

```js
{
  char,
  alternatives,
}
```

This would become a breaking change (change from `array` of `string`s)?

ljharb:
I definitely want to avoid a breaking change.

tests/lib/rules/no-unescaped-entities.js
```diff
@@ -186,7 +186,20 @@ ruleTester.run('no-unescaped-entities', rule, {
...
+        var Hello = createReactClass({
+          render: function() {
+            return <>foo & bar</>;
+          }
+        });
+      `,
+      parser: 'babel-eslint',
+      errors: [{message: 'HTML entities, \`&\` , must be escaped.'}],
+      options: [{
+        forbid: ['&']
```
ljharb: Can we add a forbid test that use the default parser?

stevemao: @ljharb done :)

docs/rules/no-unescaped-entities.md
```diff
@@ -83,4 +83,14 @@ Overwrite the default forbidden entities array `['>', '"', '\'', '}']` with your
...
+"react/no-unescaped-entities": ["error", {"forbid": [{
+  char: ">",
+  alternatives: ['&gt;']
+}, {
+  char: "}",
+  alternatives: ['&#125;']
+}]}],
 ```
+
+Where `char` is a special character and `alternatives` is the correct escapes.
```
stevemao: Also added docs

ljharb: Why multiple alternatives?

tests/lib/rules/no-unescaped-entities.js
ljharb: maybe also tests that cover providing custom alternatives?

lib/rules/no-unescaped-entities.js
```diff
@@ -28,10 +40,7 @@ module.exports = {
       type: 'object',
       properties: {
         forbid: {
-          type: 'array',
-          items: {
-            type: 'string'
```
ljharb: these should be tightly specified - either a string, or an object with the proper "char" + "alternatives" schema

stevemao: Yeah, I tried. I couldn't find a `Either` type with this lib.

ljharb:
(I'm going to hold off merging it for awhile while the 7.12 release settles down)

PR #2024, opened 2018-10-23
[new] Add no-update-forcing-props rule

### LINKED ISSUE #1633: New rule to avoid prop value with array or object creation in render
There is a rule for prop value with function creation (`bind`, `function(...)`, or arrow function): `react/jsx-no-bind`
But I can't find a rule for array (`[]`) or object (`{}`) creation.

It would be useful since it would help to avoid useless renders of pure components.

wojtekmaj:
Warns you if you have passed a prop with a value that will be [referentially unequal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators#Identity_strict_equality_()) on each render. This is bad for performance, as it will result in the garbage collector being invoked way more than is necessary. It may also cause unnecessary re-renders if a brand new object or function is passed as a prop to a component that uses reference equality check on the prop to determine if it should update.

Fixes #1633. Closes #1888.


damianlajara:
You should add your rule [here](https://github.com/yannickcr/eslint-plugin-react#jsx-specific-rules) as well.

ljharb:
Also see #1633 and #1888.

ljharb:
HTML elements aren't pure - React has to parse it and compare *every single time* anyways. It's only on class-based pure components that referential equality matters.

lifeiscontent:
@ljharb any updates on getting this in? I'd love to have this rule in create-react-app, I've got an app that has a ton of this happening and it would be nice to have the visual feedback because I unfortunately didn't realize this happened in react.

emattias:
Will this be able to detected when an object literal defined in the render function is hoisted by a Babel plugin and not mark that as a violation of this rule?

My main example of this is the css prop when using emotion: https://emotion.sh/docs/@emotion/babel-preset-css-prop

If it can't detect the hoisting by the Babel plugin then maybe this should have option where you can send in prop names to ignore.

ljharb:
@wojtekmaj are you still interested in completing this PR?

wojtekmaj:
@ljharb Not sure what's left to do here

lib/rules/jsx-no-update-forcing-props.js
```diff
@@ -0,0 +1,116 @@
...
+          type: 'boolean'
+        },
+        allowCallExpressions: {
+          default: false,
+          type: 'boolean'
+        },
+        ignoreRefs: {
+          default: false,
+          type: 'boolean'
+        }
```
ljharb, line 49: maybe rather than 7 booleans, 6 of which are named "allow" and default to false, and the 7th named "ignore" and defaults to true - we could use an object like:
```
{
  functions: 'allow',
  arrowFunctions: 'allow',
  objects: 'allow',
  arrays: 'allow',
  newExpressions: 'allow',
  callExpressions: 'allow',
  refs: 'ignore',
}
```
?

lib/rules/jsx-no-update-forcing-props.js
```diff
@@ -0,0 +1,116 @@
...
+      },
+      additionalProperties: false
+    }]
+  },
+
+  // eslint-disable-next-line
```
ljharb, line 55: what's being disabled here? can this be removed entirely?

lib/rules/jsx-no-update-forcing-props.js
```diff
@@ -0,0 +1,116 @@
...
+    const configuration = context.options[0] || {};
+
+    const typeToName = {
+      object: 'Objects',
+      array: 'Arrays',
+      func: 'Functions',
+      arrowFunc: 'Functions',
+      new: 'Constructor expressions',
+      call: 'Function calls'
+    };
```
ljharb, line 66: this object can be declared at module level, so it's only created once

lib/rules/jsx-no-update-forcing-props.js
```diff
@@ -0,0 +1,116 @@
...
+        return 'array';
+      } else if (!configuration.allowFunctionExpressions && nodeType === 'FunctionExpression') {
+        return 'func';
+      } else if (!configuration.allowArrowFunctionExpressions && nodeType === 'ArrowFunctionExpression') {
+        return 'arrowFunc';
+      } else if (!configuration.allowNewExpressions && nodeType === 'NewExpression') {
+        return 'new';
+      } else if (!configuration.allowCallExpressions && nodeType === 'CallExpression') {
+        return 'call';
+      }
```
ljharb, line 83:
```suggestion
      if (!configuration.allowObjectExpressions && nodeType === 'ObjectExpression') {
        return 'object';
      }
      if (!configuration.allowArrayExpressions && nodeType === 'ArrayExpression') {
        return 'array';
      }
      if (!configuration.allowFunctionExpressions && nodeType === 'FunctionExpression') {
        return 'func';
      }
      if (!configuration.allowArrowFunctionExpressions && nodeType === 'ArrowFunctionExpression') {
        return 'arrowFunc';
      }
      if (!configuration.allowNewExpressions && nodeType === 'NewExpression') {
        return 'new';
      }
      if (!configuration.allowCallExpressions && nodeType === 'CallExpression') {
        return 'call';
      }
```

lib/rules/jsx-no-update-forcing-props.js
```diff
@@ -0,0 +1,116 @@
...
+
+      return null;
+    }
+
+    function getErrorMessage(valueNodeType) {
+      return `${typeToName[valueNodeType]} must be defined outside attributes to avoid unnecessary updates.`;
+    }
```
ljharb, line 90: this function is only called once, can it be inlined?

ljharb (changes requested):
Oops, sorry for not reviewing this more thoroughly!

wojtekmaj:
Sorry @ljharb, had to re-create PR because I removed the fork during these 3 years :joy:

Please head to https://github.com/yannickcr/eslint-plugin-react/pull/3037.

ljharb:
Oof, that sucks :-/ means this PR is a permanent dangling ref on the repo.

PR #2026, opened 2018-10-24
Use `loc` version of `end`, to handle parsers that don't fill `name.end`.

HauptmannEck:
Not all parsers fill the `name.end` field in the AST structure so relying on it is less robust. 
Ref: https://github.com/babel/babel/issues/6681     
But the `range` syntax is widely supported (babel has a flag). 

I ran into this issue using the `typescript-eslint-parser` which does not add `start` and `end` to the nodes, but has range.

Here is an example to see the difference:
https://astexplorer.net/#/gist/c336be172990d7b467d1ed3dbc1882ee/d312c83cd24626d2e82ba791add48fe061b7cdc0

### DIFF
--- a/lib/rules/jsx-first-prop-new-line.js
+++ b/lib/rules/jsx-first-prop-new-line.js
@@ -45,7 +45,7 @@ module.exports = {
                 node: decl,
                 message: 'Property should be placed on a new line',
                 fix: function(fixer) {
-                  return fixer.replaceTextRange([node.name.end, decl.range[0]], '\n');
+                  return fixer.replaceTextRange([node.name.range[1], decl.range[0]], '\n');
                 }
               });
             }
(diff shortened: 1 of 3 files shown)


HauptmannEck:
@ljharb It should be possible, but I would need to add `typescript-eslint-parser` to the `devDependencies` to test that route. Sadly `babel-eslint` does not allow you to turn off the ranges with an option.

If it is ok to add a new package to `devDependencies` then I should be able to create a test

ljharb:
Sure, that seems reasonable.

HauptmannEck:
It ended up needed `typescript` as well to run the parser, I don't know if there is a better/lighter parser for showing this issue.

HauptmannEck:
@ljharb Quick question, what is the release policy for this repo? Is there a certain timeline for releases or does it just deped on a accumulation of changes? Just curious, as I am not blocked by this fix.

ljharb:
There’s no policy and no set timeline; things get released when maintainers have time to do so.

ljharb:

Issue #1422, opened 2017-09-09
`prop-types` fails with destructuring


MatthewHerbst:
The [`prop-types`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md) rule fails when presented with destructuring:

**package.json**
```json
{
  "dependencies": {
    "eslint": "^4.6.1",
    "eslint-plugin-react": "^7.3.0",
    "proptypes": "^1.1.0",
    "react": "^15.6.1"
  }
}
```

**.eslintrc**
```json
{
  "env": {
    "es6": true,
    "node": true
  },
  "parserOptions": {
    "sourceType": "module"
  },
  "plugins": ["eslint-plugin-react"],
  "rules": {
    "react/prop-types": 2
  }
}
```

**Examples.jsx**
```js
import React from 'react';
import PropTypes from 'proptypes';

// FAIL: Should error that `bar` is missing in propTypes
class DestructuredDeepMissing extends React.PureComponent {
  render () {
    const {
      foo: {
        bar,
      },
    } = this.props;

    const it = bar;
  }
}
DestructuredDeepMissing.propTypes = {
  foo: PropTypes.shape({}).isRequired,
};

// Interestingly, non-deep destructuring works just fine
// PASS: Should error
class DestructuredMissing extends React.PureComponent {
  render () {
    const {
      foo,
    } = this.props;

    const it = foo;
  }
}
DestructuredMissing.propTypes = {};
```
Related: #296
Possibly related: #1346 

Also possibly related is that `no-unused-prop-types` (without using Flow) fails on destructuring as well, which is behavior I was able to replicate (though there seem to be a plethora of tickets related to this, especially in the Flow context).

ljharb:
Shape props have known bugs; do you have skipShapeProps enabled?

MatthewHerbst:
@ljharb no - we use Relay, so basically 95% of our props are within a shape (`viewer.x`, or deeper). Enabling that option would be akin to turning this rule off for us. I know that `no-unused-prop-types` has issues with shapes since it mentions that in the description, but there is no such warning in the `prop-types` rule (though makes sense that the two rules would share some issues).

jseminck:
We have this test case:

```js
      code: [
        'class Hello extends React.Component {',
        '  render() {',
        '    this.props.a.b',
        '    return <div>Hello</div>;',
        '  }',
        '}',
        'Hello.propTypes = {',
[...]

PR #2029, opened 2018-10-26
Chore: Add tests for prop-types destructuring

sstern6:
Fixes and verifies: https://github.com/yannickcr/eslint-plugin-react/issues/1422

### DIFF
--- a/tests/lib/rules/prop-types.js
+++ b/tests/lib/rules/prop-types.js
@@ -2237,6 +2237,58 @@ ruleTester.run('prop-types', rule, {
       }]
     }, {
       code: [
+        'class Hello extends React.Component {',
+        '  render() {',
+        '    const { name, ...rest } = this.props',
+        '    return <div>Hello</div>;',
+        '  }',
+        '}'
+      ].join('\n'),
+      errors: [{
+        message: '\'name\' is missing in props validation',
+        line: 3,
+        column: 13,
+        type: 'Property'
+      }]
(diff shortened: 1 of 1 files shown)


MatthewHerbst:
@sstern6 Would it be possible to add a deep-destructuring example as well? For example:

```javascript
class MyComponent extends React.PureComponent {
  propTypes = {
    foo: PropTypes.shape({
      // TODO: fill this out
    }).isRequired,
  }

  render() {
    return <div>{this.props.foo.bar}</div>;
  }
}
```

Expected error would be: `\'bar\' is missing in props validation'`

I ask for this because I've seen cases where sometimes destructuring works on the outer layer but not when you go deeper.

MatthewHerbst:
@sstern6 #1422 actually reports the problem noted above, which is why I was hoping to see a deep-destructuring test in the test suite that purportedly tested it heh. Seems #1422 still isn't fixed though. I think the tests you have above are good to add regardless, but I don't think they should be said to fully fix and/or verify #1422

sstern6:
Youre right sorry about that, ill get on a fix for the deep destructuring, got caught up in the spread operator examples.

MatthewHerbst:
Amazing, thank you! If there's any way I can help, please let me know. I've taken a glance at it before and didn't really know where to start, but happy to look again.

MatthewHerbst:
Wow. The robots liked this being merged!

PR #2032, opened 2018-10-31
`jsx-fragments`: improve message text

alexzherdev:


lib/rules/jsx-fragments.js
```diff
@@ -47,7 +47,8 @@ module.exports = {
       if (!versionUtil.testReactVersion(context, '16.2.0')) {
         context.report({
           node,
-          message: 'Fragments are only supported starting from React v16.2'
+          message: 'Fragments are only supported starting from React v16.2. '
+            + 'Please disable the `react/jsx-fragments` rule or upgrade your version of React.'
```
ljharb: maybe it’s worth adding “in eslint settings”?

alexzherdev:
@ljharb our tests are broken due to https://github.com/eslint/typescript-eslint-parser/pull/526 (when I fix `typescript-eslint-parser` to 20.0.0 they pass). Is this an example of peerDep changes being semver-major?

ljharb:
Yes, technically that linked commit is a semver-major, and it'd be great if you filed an issue on that repo about it.

We'll have to peg to v20 of that dep in the meantime, then.

package.json
```diff
@@ -39,7 +39,7 @@
     "istanbul": "^0.4.5",
     "mocha": "^5.2.0",
     "typescript": "^3.1.3",
-    "typescript-eslint-parser": "^20.0.0"
+    "typescript-eslint-parser": "20.0.0"
```
ljharb: i think this is now fixed, so now we can use `^` again

alexzherdev: Right, but should we say `^20.1.1` now?

ljharb: yes, good call

PR #2043, opened 2018-11-12
Updating no-target-blank logic to only require no-referrer

### LINKED ISSUE #2022: target=_blank rel=noreferrer implies noopener
As it turns out, `<a target="_blank" rel="noreferrer">` actually implies noopener behavior.
See spec note here: https://html.spec.whatwg.org/multipage/links.html#link-type-noreferrer

This means adding/enforcing `noopener` here is redundant:
https://github.com/yannickcr/eslint-plugin-react/blob/72a71b302c5f36ee3dd440da43737afc21cc311e/lib/rules/jsx-no-target-blank.js#L38

seancrater:
Fixes #2022 

Just a minor change to only require `no-referrer` rather than both `noreferrer noopener`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ Enable the rules that you would like to use.
 * [react/jsx-no-duplicate-props](docs/rules/jsx-no-duplicate-props.md): Enforce no duplicate props
 * [react/jsx-no-literals](docs/rules/jsx-no-literals.md): Prevent using string literals in React component definition
 * [react/jsx-no-script-url](docs/rules/jsx-no-script-url.md): Forbid `javascript:` URLs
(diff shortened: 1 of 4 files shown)


seancrater:
@ljharb Probably should've noted that I only made this so it's available if the team decides to go through with it.

seancrater:
@ljharb Also shared what I found for research on the issue, I'm not sure if it's actually enough to draw a conclusion from but it seems like a pretty valid source to me.

ljharb:
Reopening per https://github.com/yannickcr/eslint-plugin-react/issues/2022#issuecomment-624339760

ljharb:
Rebased this, and kept all existing test cases rather than modifying them.

PR #2065, opened 2018-12-02
Change allowed propWrapperFunctions setting values

jomasti:
This change is adapting the `propWrapperFunctions` setting to allow for
using objects along side the already present strings. This will help
facilitate adding extra attributes to these objects like for exact prop
functions.

This was initially proposed [here](https://github.com/yannickcr/eslint-plugin-react/pull/1547#issuecomment-345545576).

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,10 +42,12 @@ You should also specify settings that will be shared across all the plugin rules
       "version": "15.0", // React version, default to the latest React stable release
       "flowVersion": "0.53" // Flow version
     },
-    "propWrapperFunctions": [ "forbidExtraProps" ] // The names of any functions used to wrap the
-                                                   // propTypes object, e.g. `forbidExtraProps`.
-                                                   // If this isn't set, any propTypes wrapped in
-                                                   // a function will be skipped.
+    "propWrapperFunctions": [
+        // The names of any function used to wrap propTypes, e.g. `forbidExtraProps`. If this isn't set, any propTypes wrapped in a function will be skipped.
+        "forbidExtraProps",
+        {"property": "freeze", "object": "Object"}
+        {"property": "myFavoriteWrapper"}
+    ]
   }
 }
 ```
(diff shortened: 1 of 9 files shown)


ljharb:
I’m confused; Object.freeze as a string should already work here.

ljharb:
ah, for the exact: true part tho. Got it.

README.md
```diff
@@ -42,10 +42,11 @@ You should also specify settings that will be shared across all the plugin rules
       "version": "15.0", // React version, default to the latest React stable release
       "flowVersion": "0.53" // Flow version
     },
-    "propWrapperFunctions": [ "forbidExtraProps" ] // The names of any functions used to wrap the
-                                                   // propTypes object, e.g. `forbidExtraProps`.
-                                                   // If this isn't set, any propTypes wrapped in
-                                                   // a function will be skipped.
+    "propWrapperFunctions": [                      // The names of any functions used to wrap the
+        "forbidExtraProps",                        // propTypesObject, e.g. `forbidExtraProps`
+        {"property": "freeze", "object": "Object"} // If this isn't set, any propTypes wrapped in
```
ljharb: This comment is meant to be all one line - let’s move all the comment bits to a single comment inside the array?

lib/util/propWrapper.js
```diff
@@ -0,0 +1,25 @@
...
+function isPropWrapperFunction(context, name) {
+  const propWrapperFunctions = getPropWrapperFunctions(context);
+  const splitName = name.split('.');
+  const found = Array.from(propWrapperFunctions).find(func => {
+    if (splitName.length === 2 && func.object === splitName[0] && func.property === splitName[1]) {
+      return true;
+    }
+    return name === func || func.property === name;
+  });
+  return Boolean(found);
```
ljharb: Instead of .find and this return, return .some directly?

PR #2075, opened 2018-12-06
Adjust `no-unsafe` rule to handle all unsafe life-cycle methods

sergei-startsev:
[The original PR](https://github.com/yannickcr/eslint-plugin-react/pull/2069) has been broken down to 2 parts, the changes related to `no-unsafe` rule have been moved to this PR.

## Summary
- `no-unsafe` rule has been adjusted to handle all unsafe life-cycle methods including their aliases. Developers who want to avoid using of legacy lifecycle methods can enable it today.
- Instructions on updating components have been adjusted in `no-unsafe` rule to be consistent with React runtime warnings.
- Documentation pages have been updated.
- Corresponding unit tests have been updated/added.

### DIFF
--- a/lib/rules/no-unsafe.js
+++ b/lib/rules/no-unsafe.js
@@ -1,5 +1,5 @@
 /**
- * @fileoverview Prevent usage of UNSAFE_ methods
+ * @fileoverview Prevent usage of unsafe lifecycle methods
  * @author Sergei Startsev
  */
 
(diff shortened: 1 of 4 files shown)


lib/rules/no-unsafe.js
```diff
@@ -31,16 +31,29 @@ module.exports = {
       return {};
     }
 
+    const unsafe = {};
+    unsafe.componentWillMount = unsafe.UNSAFE_componentWillMount = {
```
ljharb: instead of making an object and mutating it, let's create it all in one literal? the various objects can be stored in variables and assigned like `{ componentWillMount, UNSAFE_componentWillMount: componentWillMount }` etc.

sergei-startsev: fixed

sergei-startsev:
@ljharb The breaking changes are hidden behind `checkAliases` option.

PR #2085, opened 2018-12-13
Fix: no-array-index-key not working in React.Children methods

### LINKED ISSUE #2083: no-array-index-key not working in React.Children.map
Hi! This currently does not flag an error with `no-array-index-key`:

```jsx
React.Children.map(this.props.children, (child, index) => {
    const className = classnames(
                        child.props.className,
                        `${this.props.transition}-finished`
                    );
    return React.cloneElement(child, { key: index, className });
});
```

Specifically, I would expect the `key: index` part to trigger the `no-array-index-key` warning.

(I realize that there is probably a better way to write this (ie: not using `cloneElement`), but I noticed this recently and saw that it didn't flag the warning.

---

PS: Thanks for the awesome ESLint plugin! 🎉  Can't wait for the next release!

mcbook

himynameisdave:
The `no-array-index-key` was not flagging any errors when iterating via `React.Children` methods. This fixes that!

Fixes #2083

spin-skull

### DIFF
--- a/lib/rules/no-array-index-key.js
+++ b/lib/rules/no-array-index-key.js
@@ -47,6 +47,32 @@ module.exports = {
         && indexParamNames.indexOf(node.name) !== -1;
     }
 
+    function isUsingReactChildren(node) {
+      const callee = node.callee;
+      if (
+        !callee
+        || !callee.property
+        || !callee.object
+      ) {
+        return null;
+      }
+
+      const isReactChildMethod = ['map', 'forEach'].indexOf(callee.property.name) > -1;
+      if (!isReactChildMethod) {
+        return null;
+      }
+
+      const obj = callee.object;
+      if (obj && obj.name === 'Children') {
+        return true;
+      }
+      if (obj && obj.object && obj.object.name === 'React') {
+        return true;
+      }
+
+      return false;
+    }
+
     function getMapIndexParamName(node) {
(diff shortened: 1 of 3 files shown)


lib/rules/no-array-index-key.js
```diff
@@ -47,6 +47,32 @@ module.exports = {
...
+      const isReactChildMethod = ['map', 'forEach'].indexOf(callee.property.name) > -1;
+      if (!isReactChildMethod) {
+        return null;
+      }
+
+      const obj = callee.object;
+      if (obj && obj.name === 'Children') {
+        return true;
+      }
+      if (obj && obj.object && obj.object.name === 'React') {
```
ljharb: This should check the pragma setting, and not be hardcoded to “React”

lib/rules/no-array-index-key.js
```diff
@@ -132,24 +161,20 @@ module.exports = {
...
           if (!indexParamNames.length) {
             return;
           }
-
           const props = node.arguments[1];
 
           if (props.type !== 'ObjectExpression') {
             return;
           }
-
```
ljharb: Please restore all these blank lines, they help readability

himynameisdave:
@ljharb I've addressed your CR comments, please read through again and let me know if there is anything else you need me to change! 💪👌🤙

PR #2086, opened 2018-12-13
Handle JSX attribute indentation in jsx-indent

### LINKED ISSUE #2035: Closing braces being flushed all the way to the left for inline component props

* **ESLint Version:**
```
"eslint": "^5.7.0",
"eslint-config-airbnb": "^17.1.0",
"eslint-plugin-import": "^2.14.0",
"eslint-plugin-jsx-a11y": "^6.1.2",
"eslint-plugin-react": "^7.11.1",
```

* **Node Version:** v8.12.0
* **yarn Version:** 1.9.4

**What parser (default, Babel-ESLint, etc.) are you using?** babel-eslint

<details>

```js
{
  "root": true,
  "extends": "airbnb",
  "parser": "babel-eslint",
  "rules": {
    "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
    "react/prop-types": 0,
    "no-console": 0,
    "no-debugger": 0,
    "react/sort-comp": 0
  }
}
```

</details>

```js
const Component = () => (
  <View
    ListFooterComponent={
      <View
        rowSpan={3}
        placeholder="Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do"
      />
    }
  />
);
```

```bash
eslint --fix
```

It should be formatted like this instead:
```
const Component = () => (
  <View
    ListFooterComponent={(
      <View
        rowSpan={3}
        placeholder="Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do"
      />
    )}
  />
);
```
On the older versions, it will not flush the brackets to the left:
```
"eslint": "^4.13.0",
"eslint-config-airbnb": "^16.1.0",
"eslint-plugin-import": "^2.8.0",
"eslint-plugin-jsx-a11y": "^6.0.2",
"eslint-plugin-react": "^7.5.1",
```

Line 11's closing braces are being flushed to the left when `eslint` formats it.

jomasti:
Resolves #2035

### DIFF
--- a/lib/rules/jsx-indent.js
+++ b/lib/rules/jsx-indent.js
@@ -242,11 +242,23 @@ module.exports = {
       checkNodesIndent(node, peerElementIndent);
     }
 
+    function handleAttribute(node) {
+      if (node.value.type !== 'JSXExpressionContainer') {
+        return;
+      }
+      const nameIndent = getNodeIndent(node.name);
+      const lastToken = sourceCode.getLastToken(node.value);
+      const firstInLine = astUtil.getFirstNodeInLine(context, lastToken);
+      const indent = node.name.loc.start.line === firstInLine.loc.start.line ? 0 : nameIndent;
+      checkNodesIndent(firstInLine, indent);
+    }
+
     return {
       JSXOpeningElement: handleOpeningElement,
       JSXOpeningFragment: handleOpeningElement,
       JSXClosingElement: handleClosingElement,
       JSXClosingFragment: handleClosingElement,
+      JSXAttribute: handleAttribute,
       JSXExpressionContainer: function(node) {
         if (!node.parent) {
           return;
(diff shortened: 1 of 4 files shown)


ljharb:
This looks great, but since it will introduce new warnings, it’s a breaking change. Can we hide this new behavior behind an option?

lib/rules/jsx-indent.js
```diff
@@ -73,6 +81,7 @@ module.exports = {
     }
 
     const indentChar = indentType === 'space' ? ' ' : '\t';
+    const checkAttributes = context.options[1] && context.options[1].checkAttributes || false;
```
ljharb:
```suggestion
    const options = context.options[1] || {};
    const checkAttributes = options.checkAttributes || false;
```

jomasti:
> This looks great, but since it will introduce new warnings, it’s a breaking change. Can we hide this new behavior behind an option?

Updated

PR #2098, opened 2018-12-28
Fix no-unused-state crash

### LINKED ISSUE #2096: no-unused-state: Cannot read property 'add' of null
We have hundreds of files with JSX so I'm not sure which one is leading to this exception.

* `7.11.1`: No error.
* `v7.12.0`: The following error.

```
TypeError: Cannot read property 'add' of null
    at Object.CallExpression (./eslint-plugin-react/lib/rules/no-unused-state.js:269:31)
    at updatedRuleInstructions.(anonymous function) (./eslint-plugin-react/lib/util/Components.js:754:75)
    at listeners.(anonymous function).forEach.listener (./eslint/lib/util/safe-emitter.js:45:58)
    at Object.emit (./eslint/lib/util/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (./eslint/lib/util/node-event-generator.js:251:26)
    at NodeEventGenerator.applySelectors (./eslint/lib/util/node-event-generator.js:280:22)
    at NodeEventGenerator.enterNode (./eslint/lib/util/node-event-generator.js:294:14)
    at CodePathAnalyzer.enterNode (./eslint/lib/code-path-analysis/code-path-analyzer.js:632:23)
    at nodeQueue.forEach.traversalInfo (./eslint/lib/linter.js:750:28)
```

jomasti:
Don't assume `classInfo.aliases` is a set since there are some cases where
it is `null` for a `ClassProperty`.

Fixes #2096

### DIFF
--- a/tests/lib/rules/no-unused-state.js
+++ b/tests/lib/rules/no-unused-state.js
@@ -636,6 +636,64 @@ eslintTester.run('no-unused-state', rule, {
       }
       `,
       parser: 'babel-eslint'
+    }, {
+      code: `
+      var Foo = createReactClass({
+        getInitialState: function() {
+          return { initial: 'foo' };
+        },
+        handleChange: function() {
+          this.setState(state => ({
+            current: state.initial
+          }));
+        },
+        render() {
+          const { current } = this.state;
+          return <div>{current}</div>
+        }
+      });
+      `
+    }, {
+      code: `
+      var Foo = createReactClass({
+        getInitialState: function() {
+          return { initial: 'foo' };
+        },
+        handleChange: function() {
+          this.setState((state, props) => ({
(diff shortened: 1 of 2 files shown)


tests/lib/rules/no-unused-state.js
```diff
@@ -636,6 +636,64 @@ eslintTester.run('no-unused-state', rule, {
...
+          const { current } = this.state;
+          return <div>{current}</div>
+        }
+      });
+      `
+    }, {
+      // Don't error out
+      code: `
+      class Foo extends Component {
+        handleChange = function() {
```
jomasti: This and the next test case were what I could find that would reproduce the error. It seems like weird usages, but it won't error out.

PR #2099, opened 2018-12-28
Fix false positives inside lifecycle methods

### LINKED ISSUE #2094: v7.12.0 update throw is missing in props validation  (react/prop-types)
After updating to v7.12.0 appeared strange validation errors.

I have this propTypes:
```js
propTypes = {
    actions: PropTypes.object.isRequired,
};
```

and this code:
```js
componentWillReceiveProps (nextProps) {
    this.props.actions.doSomething();
}

componentWillUnmount () {
    this.props.actions.doSomething();
}
```

And then I have eslint error:  'doSomething' is missing in props validation  react/prop-type
on the line in componentWillReceiveProps. But the line in componentWillUnmount is ok. 
Strange that it says "'doSomething' is missing" and not "'actions.doSomething' is missing"

I tried to change "PropTypes.object" to "PropTypes.shape", still the same.

Error disappears if I change name "componentWillReceiveProps" to something else. 
If I add propType  "doSomething: PropTypes.func.isRequired" error disappears too but it doesn't make sense.

jomasti:
This will prevent triggering on any member expression inside a lifecycle
method. Instead, it will do the naive checking of the node names for
`props`, `nextProps`, and `prevProps` which is already done here.

Resolves #2094

### DIFF
--- a/lib/util/usedPropTypes.js
+++ b/lib/util/usedPropTypes.js
@@ -285,13 +285,14 @@ module.exports = function usedPropTypesInstructions(context, components, utils) 
    * @returns {Boolean} True if we are using a prop, false if not.
    */
   function isPropTypesUsage(node) {
+    const isThisPropsUsage = node.object.type === 'ThisExpression' && node.property.name === 'props';
+    const isPropsUsage = isThisPropsUsage || node.object.name === 'nextProps' || node.object.name === 'prevProps';
     const isClassUsage = (
       (utils.getParentES6Component() || utils.getParentES5Component()) &&
-      ((node.object.type === 'ThisExpression' && node.property.name === 'props')
-      || isPropArgumentInSetStateUpdater(node))
+      (isThisPropsUsage || isPropArgumentInSetStateUpdater(node))
     );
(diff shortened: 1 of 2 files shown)


lib/util/usedPropTypes.js
```diff
@@ -285,13 +285,14 @@ module.exports = function usedPropTypesInstructions(context, components, utils)
    * @returns {Boolean} True if we are using a prop, false if not.
    */
   function isPropTypesUsage(node) {
+    const isThisPropsUsage = node.object.type === 'ThisExpression' && node.property.name === 'props';
+    const isPropsUsage = isThisPropsUsage || node.object.name === 'nextProps' || node.object.name === 'prevProps';
```
alexzherdev: Thanks for the fix!
Later on it'd be great not to depend on the names (similar to #1829), but for now not crashing is the priority 👍

jomasti: Agreed!

paulmelnikow:
Hi! Thanks so much for your work on this plugin.

We're running into this issue at badges/shields#2609. Could this be released?

PR #2103, opened 2019-01-02
Link does not end with '.'

yoyo837:
Fix link hash `#`


yoyo837:
It prints to the console, such as `vscode` , and can't scroll to the anchored location after clicking the link.

PR #2110, opened 2019-01-03
[new] Add prefer-read-only-props rule

### LINKED ISSUE #1454: Add a rule prefer-covariant-props
With react and flow, it's very common to type props like this:

```javascript
type Props = {
  prop1: string,
  prop2: number
};
```

However this makes it possible to write to the props object, like this: `this.props.prop1 = 'foo'` which is to be avoided.
To force this with flow, the solution is to use covariant properties:
```javascript
type Props = {
  +prop1: string,
  +prop2: number
};
```

It would be useful to have an eslint rule to force this pattern.

### LINKED ISSUE #1869: Rule proposal: Require read-only props
When using flow require that you make all of the props read-only.

drx:
This adds a rule to enforce that flow types for props are read-only (covariant). This was requested in #1454 and #1869.

I added a fixer for the rule, which I tested on a decently large codebase at Flexport (5000+ files) with thousands of violations.

One interesting note is that `declaredPropTypes` doesn't recognize `foo` in `props: $ReadOnly<{foo: bar}>` as a prop. In this case that's OK, because that wouldn't give a warning anyway.

Fixes #1454. Fixes #1869.


ljharb:
It'd be great if we could separate out the fixes into a separate PR from the new rule.

lib/rules/prefer-read-only-props.js
```diff
@@ -0,0 +1,74 @@
...
+ * @fileoverview Require component props to be typed as read-only.
+ * @author Luke Zapart
+ */
+'use strict';
+
+const Components = require('../util/Components');
+const docsUrl = require('../util/docsUrl');
+
+function isFlowPropertyType(node) {
+  return node.type === 'ObjectTypeProperty';
```
ljharb: does this conflict with TS node names at all?

lib/rules/prefer-read-only-props.js
```diff
@@ -0,0 +1,74 @@
...
+
+      Object.keys(list).forEach(key => {
+        const component = list[key];
+
+        if (!component.declaredPropTypes) {
+          return;
+        }
+
+        Object.keys(component.declaredPropTypes).forEach(propName => {
+          const prop = component.declaredPropTypes[propName];
```
alexzherdev: So cool to see the propTypes detection in use 👍

alexzherdev:
@drx to clarify this bit:
> Worse yet, because declaredPropTypes is an object, those helper properties can overwrite actual props.

Do you mean that if we wrongly detect something as a prop, it can overwrite the metadata for an actual prop with the same name?

drx:
> It'd be great if we could separate out the fixes into a separate PR from the new rule.

Done! #2111


> @drx to clarify this bit:
> 
> > Worse yet, because declaredPropTypes is an object, those helper properties can overwrite actual props.
> 
> Do you mean that if we wrongly detect something as a prop, it can overwrite the metadata for an actual prop with the same name?

Yeah, if you have something like:

```jsx
type Props = {
  foo: string,
}

class Hello extends React.Component<Props> {
  renderHelper = ({foo}: {foo: string}) => {
    return <div />;
  }
  render() {
    return <div>{this.props.used}</div>;
  }
}
```

The metadata (including the `loc`) might be overwritten by the `renderHelper` `foo`, I have had this happen while fixing which is how I discovered this.

lib/rules/prefer-read-only-props.js
drx: Good question. I checked and AST that `typescript-eslint-parser` generates for `function Hello(props: {name: string}) {...}` is: `FunctionDeclaration <- TSTypeAnnotation <- TSTypeLiteral <- TSPropertySignature <- TSStringKeyword`, so it's safe.

drx:
Thanks!

PR #2114, opened 2019-01-03
update README

mohsinulhaq:
https://github.com/yannickcr/eslint-plugin-react/issues/1955#issuecomment-451182940

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,8 @@ You should also specify settings that will be shared across all the plugin rules
       "createClass": "createReactClass", // Regex for Component Factory to use,
                                          // default to "createReactClass"
       "pragma": "React",  // Pragma to use, default to "React"
-      "version": "15.0", // React version, default to the latest React stable release
+      "version": "detect", // React version. "detect" automatically picks the version you have installed.
+                           // You can also use `16.0`, `16.3`, etc, if you want to override the detected value.
       "flowVersion": "0.53" // Flow version
     },
     "propWrapperFunctions": [


README.md
```diff
@@ -39,7 +39,7 @@ You should also specify settings that will be shared across all the plugin rules
       "createClass": "createReactClass", // Regex for Component Factory to use,
                                          // default to "createReactClass"
       "pragma": "React",  // Pragma to use, default to "React"
-      "version": "15.0", // React version, default to the latest React stable release
+      "version": "detect", // React version, "detect" automatically picks the version from your project's package.json. "detect" will be the default value in future versions
```
ljharb: Let's make sure to document both "detect" and the explicit form, like so:
```suggestion
      "version": "detect", // React version, "detect" automatically picks the version from your project's package.json. You can also use `16.0`, `16.3`, etc, if you want to override the detection value.
```

mohsinulhaq:
is it ok now?

README.md
```diff
@@ -39,7 +39,9 @@ You should also specify settings that will be shared across all the plugin rules
       "createClass": "createReactClass", // Regex for Component Factory to use,
                                          // default to "createReactClass"
       "pragma": "React",  // Pragma to use, default to "React"
-      "version": "15.0", // React version, default to the latest React stable release
+      "version": "detect", // React version, "detect" automatically picks the version from your project's package.json.
```
alexzherdev: This is technically not entirely correct though. `package.json` (generally) defines a range, whereas "detect" will detect the exact version you have installed.

mohsinulhaq: yes, you are right, so do we actually check package lock?

alexzherdev: I'll defer the final wording to @ljharb though.

mohsinulhaq: okay that makes sense.

mohsinulhaq:
Thanks! This was my first OSS PR to a well-known library. 😃

PR #2115, opened 2019-01-03
Fix propType detection inside class bodies

drx:
I investigated it a bit more and it turns out #2111 has more instances than just `ClassProperties`. A legit example is JSX render props:

```jsx
type Props = {
  used: string,
}
class Hello extends React.Component<Props> {
  render() {
    return <QueryRenderer
      render={
        ({notAProp}: {notAProp: string}) => <div>{this.props.used}</div>
      }
    />;
  }
}
```

The `no-unused-prop-types` rule gives a warning that `notAProp` is not used, because it thinks `notAProp` is a prop of `Hello`.

I added two more tests. The problem seems to occur in many cases involving functions being declared inside class bodies. The nice thing is that all of them have a `ClassBody` node as an ancestor so I changed the previous fix to look for `ClassBody` instead, which actually feels more like a proper fix.

### DIFF
--- a/lib/util/propTypes.js
+++ b/lib/util/propTypes.js
@@ -68,15 +68,15 @@ function iterateProperties(context, properties, fn) {
 }
 
 /**
- * Checks if a node is inside a class property.
+ * Checks if a node is inside a class body.
  *
  * @param {ASTNode} node  the AST node being checked.
- * @returns {Boolean} True if the node has a ClassProperty ancestor, false if not.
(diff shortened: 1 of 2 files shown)


alexzherdev:
Thanks for the great fixes @drx! These helpers were extracted from two rules that had diverged slightly in how they detected propTypes, so it was tricky to de-duplicate them. It’s unfortunate that a few cases slipped through the cracks; they probably worked before but didn’t have test cases in the repo.

PR #2116, opened 2019-01-04
Allow custom components in jsx-no-target-blank

gbakernet:
# What is this?

The following PR allows for customising additional components used jsx-no-target-blank. 
Closes https://github.com/yannickcr/eslint-plugin-react/issues/1075

# Why do this

It's reasonable common for projects to implement `<Link>` or `<Hyperlink>` components and use them literally instead of `<a>` elements.

# How does it work

From the docs

### `components`

An array specifying the names of additional components that should also be validated.
Each array element can either be a string with the component name or object specifying the name and url attribute:

When using an object like so:

```js
{
  "name": "Link",
  "linkAttribute": "to"
}
```

The following patterns are considered errors:

```jsx
var Hello = <Link target='_blank' to="http://example.com/"></Link>
var Hello = <Link target='_blank' to={ dynamicLink }></Link>
```

The following patterns are **not** considered errors:

```jsx
var Hello = <Link target='_blank' rel='noopener noreferrer' to="http://example.com"></Link>
var Hello = <Link target='_blank' to="relative/path/in/the/host"></Link>
var Hello = <Link target='_blank' to="/absolute/path/in/the/host"></Link>
var Hello = <Link />
```

When you use a string such as `'Link'`, it's equivalent to:

```js
{
  "name": "Link",
  "linkAttribute": "href"
}
```


docs/rules/jsx-no-target-blank.md
```diff
@@ -14,14 +14,18 @@ This rule aims to prevent user generated links from creating security vulnerabil
 ## Rule Options
 ```json
 ...
-"react/jsx-no-target-blank": [<enabled>, { "enforceDynamicLinks": <enforce> }]
+"react/jsx-no-target-blank": [<enabled>, { "enforceDynamicLinks": <enforce> }, { "components": [<string>|<object>] }]
```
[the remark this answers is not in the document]
gbakernet: Updated

gbakernet:
@ljharb I'm not too familiar with all the specific rules. A quick search for `'a'` only yielded this rule, so I'm not sure what settings might look like given the additionally complexity of the `href` link attribute option.

ljharb:
I think the current schema for the components array would work fine as a plugin-level setting; we can enhance it in the future as needed.

gbakernet:
I've update the settings as I think you intend. Please let me know if you need any further changes.

docs/rules/jsx-no-target-blank.md
```diff
@@ -14,14 +14,18 @@ This rule aims to prevent user generated links from creating security vulnerabil
 ## Rule Options
 ```json
 ...
-"react/jsx-no-target-blank": [<enabled>, { "enforceDynamicLinks": <enforce> }]
+"react/jsx-no-target-blank": [<enabled>, { "enforceDynamicLinks": <enforce>, "components": [<string>|<object>] }]
```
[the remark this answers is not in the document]
gbakernet: Removed.

lib/rules/jsx-no-target-blank.js
```diff
@@ -64,13 +66,21 @@ module.exports = {
     const configuration = context.options[0] || {};
     const enforceDynamicLinks = configuration.enforceDynamicLinks || 'always';
 
+    const elements = new Map((DEFAULTS.concat(context.settings.linkComponents || [])).map(value => {
```
ljharb: i feel like we could make a separate helper to get the "array of components"?

gbakernet: Done.

gbakernet:
Is there anything else I can do to this PR?

docs/rules/jsx-no-target-blank.md
```diff
@@ -20,37 +20,61 @@ This rule aims to prevent user generated links from creating security vulnerabil
...
 ```
 
 The following patterns are **not** considered errors:
 
 ```jsx
-var Hello = <p target='_blank'></p>
-var Hello = <a target='_blank' rel='noopener noreferrer' href="http://example.com"></a>
-var Hello = <a target='_blank' href="relative/path/in/the/host"></a>
-var Hello = <a target='_blank' href="/absolute/path/in/the/host"></a>
+var Hello = <p target="_blank"></p>
```
EvHaus: Was there a specific reason for this change? The rest of the docs use single quotes `'` instead of double quotes `"`. Maybe we should keep things consistent?

docs/rules/jsx-no-target-blank.md
```diff
@@ -20,37 +20,61 @@ This rule aims to prevent user generated links from creating security vulnerabil
 
 * enabled: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
 * enforce: optional string, 'always' or 'never'
+* Link components can be something other than an `<a>`, see [shared settings](https://github.com/yannickcr/eslint-plugin-react/blob/master/README.md#configuration) for `linkComponents` configuration)
```
EvHaus: This is already duplicated on line 60 below. I would recommend removing this.

docs/rules/jsx-no-target-blank.md
```diff
@@ -20,37 +20,61 @@ This rule aims to prevent user generated links from creating security vulnerabil
...
 `{"enforceDynamicLinks": "never"}` does not enforce the rule if the href is a dynamic link
 
 When {"enforceDynamicLinks": "never"} is set, the following patterns are **not** considered errors:
 
 ```jsx
-var Hello = <a target='_blank' href={ dynamicLink }></a>
+var Hello = <a target='_blank' href={dynamicLink}></a>
+```
+
+### Link components
```
EvHaus: I found this doc section a little hard to understand as a first time user of the feature. Can I recommend the following changes instead?

```
### Custom link components

This rule supports the ability to use custom components for links, such as `<Link />` which is popular in libraries like `react-router`, `next.js` and `gatsby`. To enable this, define your custom link components in the global [shared settings](https://github.com/yannickcr/eslint-plugin-react/blob/master/README.md#configuration) under the `linkComponents` configuration area. Once configured, this rule will check those components as if they were `<a />` elements.



```

docs/rules/jsx-no-target-blank.md
ljharb: I made this change; i believe the linter was rrroring on something so i changed them all.

if we want consistency, we should lint our markdown, but also, double quotes is the most common/idiomatic style for jsx attributes in the ecosystem, so I’d prefer we go with that.

tests/util/linkComponents.js
```diff
@@ -0,0 +1,36 @@
+/* eslint-env mocha */
```
alexzherdev: I was going to say it might be reasonable to add this env as an [override](https://eslint.org/docs/user-guide/configuring#configuration-based-on-glob-patterns) for the `tests/util/*.js` glob, but apparently that feature is only available in ESLint v4.1.0+, and we have `"^3.0.0 || ^4.0.0 || ^5.0.0"` in our devDependencies.
Would it perhaps make sense to put an `.eslintrc` in `tests/util` to use [cascading](https://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy) instead?

docs/rules/jsx-no-target-blank.md
gbakernet: Update

tests/util/linkComponents.js
gbakernet: `.eslintrc` added to `tests/util`

PR #2125, opened 2019-01-08
Fix isPragmaComponentWrapper

### LINKED ISSUE #2124: Cannot read property 'name' of undefined in isPragmaComponentWrapper
Error existed in versions 7.12.0-7.12.1.
Version 7.12.2 was ok.
It's present again in version 7.12.3.

> TypeError: Cannot read property 'name' of undefined
>     at Object.isPragmaComponentWrapper (<project_root>/node_modules/eslint-plugin-react/lib/util/Components.js:405:66)

CrOrc:
In isPragmaComponentWrapper check name of callee object if awailable.

Fixes #2124

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -401,8 +401,8 @@ function componentRule(rule, context) {
       }
       const propertyNames = ['forwardRef', 'memo'];
       const calleeObject = node.callee.object;
-      if (calleeObject) {
-        return arrayIncludes(propertyNames, node.callee.property.name) && node.callee.object.name === pragma;
+      if (calleeObject && node.callee.property) {
+        return arrayIncludes(propertyNames, node.callee.property.name) && calleeObject.name === pragma;
       }
       return arrayIncludes(propertyNames, node.callee.name) && this.isDestructuredFromPragmaImport(node.callee.name);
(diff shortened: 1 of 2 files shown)


lib/util/Components.js
```diff
@@ -402,7 +402,7 @@ function componentRule(rule, context) {
       const propertyNames = ['forwardRef', 'memo'];
       const calleeObject = node.callee.object;
       if (calleeObject) {
-        return arrayIncludes(propertyNames, node.callee.property.name) && node.callee.object.name === pragma;
```
ljharb: This isn’t equivalent, because the first one has .property and not .object.

CrOrc: Maybe I should add check for both node.callee.property and node.callee.object?

ljharb: I’m not really sure - a regression test would help make that clear.

CrOrc: Added a regression test

rpellerin:
When will this be released?

ljharb:
[v7.12.4](https://github.com/yannickcr/eslint-plugin-react/releases/tag/v7.12.4) has been released.

PR #2146, opened 2019-01-29
[New]: `jsx-no-literals`: add `ignoreProps` option to ignore props validation

iiison:
- Update Tests Accordingly
- Update Rule Readme file
- Rule shouldn't allow strings in empty tags

### DIFF
--- a/docs/rules/jsx-no-literals.md
+++ b/docs/rules/jsx-no-literals.md
@@ -28,13 +28,14 @@ var Hello = <div>
 
 There are two options:
 
-* `noStrings` - Enforces no string literals used as children, wrapped or unwrapped.
+* `noStrings` (default: `false`) - Enforces no string literals used as children, wrapped or unwrapped.
 * `allowedStrings` - An array of unique string values that would otherwise warn, but will be ignored.
+* `ignoreProps` (default: `false`) - When `true` the rule ignores literals used in props, wrapped or unwrapped.
 
 To use, you can specify as follows:
 
 ```js
-"react/jsx-no-literals": [<enabled>, {"noStrings": true, "allowedStrings": ["allowed"]}]
+"react/jsx-no-literals": [<enabled>, {"noStrings": true, "allowedStrings": ["allowed"], "ignoreProps": false}]
 ```
 
 In this configuration, the following are considered warnings:
@@ -52,6 +53,19 @@ var Hello = <div>
   {'test'}
 </div>;
 ```
+
+```jsx
+var Hello = <div class='xx' />;
+```
+
+```jsx
+var Hello = <div class={'xx'} />;
+```
+
+```jsx
+var Hello = <div class={`xx`} />;
+```
+
 
 The following are **not** considered warnings:
 
@@ -75,6 +89,33 @@ var Hello = <div>allowed</div>
 var Hello = <div>
   allowed
 </div>;
+```
+
+```jsx
+// spread props object
+var Hello = <Text {...props} />
+```
+
+```jsx
+// use variable for prop values
+var Hello = <div class={xx} />
+```
+
+```jsx
+// cache
+class Comp1 extends Component {
+  asdf() {}
+
+  render() {
+    return (
+      <div onClick={this.asdf}>
+        {'asdjfl'}
+        test
+        {'foo'}
+      </div>
+    );
+  }
+}
 ```
 
 ## When Not To Use It
--- a/lib/rules/jsx-no-literals.js
+++ b/lib/rules/jsx-no-literals.js
@@ -12,6 +12,10 @@ const docsUrl = require('../util/docsUrl');
 // Rule Definition
 // ------------------------------------------------------------------------------
 
+function trimIfString(val) {
+  return typeof val === 'string' ? val.trim() : val;
+}
+
 module.exports = {
   meta: {
     docs: {
@@ -33,6 +37,9 @@ module.exports = {
           items: {
             type: 'string'
           }
+        },
+        ignoreProps: {
+          type: 'boolean'
         }
       },
       additionalProperties: false
@@ -40,11 +47,7 @@ module.exports = {
   },
 
   create(context) {
-    function trimIfString(val) {
-      return typeof val === 'string' ? val.trim() : val;
-    }
-
-    const defaults = {noStrings: false, allowedStrings: []};
+    const defaults = {noStrings: false, allowedStrings: [], ignoreProps: false};
     const config = Object.assign({}, defaults, context.options[0] || {});
     config.allowedStrings = new Set(config.allowedStrings.map(trimIfString));
 
@@ -52,10 +55,12 @@ module.exports = {
       'Strings not allowed in JSX files' :
       'Missing JSX expression container around literal string';
 
-    function reportLiteralNode(node) {
+    function reportLiteralNode(node, customMessage) {
+      const errorMessage = customMessage || message;
+
       context.report({
         node,
-        message: `${message}: “${context.getSourceCode().getText(node).trim()}”`
+        message: `${errorMessage}: “${context.getSourceCode().getText(node).trim()}”`
       });
     }
 
@@ -82,15 +87,44 @@ module.exports = {
       return standard && parent.type !== 'JSXExpressionContainer';
     }
 
+    function getParentAndGrandParentType(node) {
+      const parent = getParentIgnoringBinaryExpressions(node);
+      const parentType = parent.type;
+      const grandParentType = parent.parent.type;
+
+      return {
+        parent,
+        parentType,
+        grandParentType,
+        grandParent: parent.parent
+      };
+    }
+
+    function hasJSXElementParentOrGrandParent(node) {
+      const parents = getParentAndGrandParentType(node);
+      const parentType = parents.parentType;
+      const grandParentType = parents.grandParentType;
+
+      return parentType === 'JSXFragment' || parentType === 'JSXElement' || grandParentType === 'JSXElement';
+    }
+
     // --------------------------------------------------------------------------
     // Public
     // --------------------------------------------------------------------------
 
     return {
-
       Literal(node) {
-        if (getValidation(node)) {
+        if (getValidation(node) && (hasJSXElementParentOrGrandParent(node) || !config.ignoreProps)) {
           reportLiteralNode(node);
+        }
+      },
+
+      JSXAttribute(node) {
+        const isNodeValueString = node.value && node.value && node.value.type === 'Literal' && typeof node.value.value === 'string';
+
+        if (config.noStrings && !config.ignoreProps && isNodeValueString) {
+          const customMessage = 'Invalid prop value';
+          reportLiteralNode(node, customMessage);
         }
       },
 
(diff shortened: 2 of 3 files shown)


tests/lib/rules/jsx-no-literals.js
```diff
@@ -350,38 +385,39 @@ ruleTester.run('jsx-no-literals', rule, {
       options: [{noStrings: true}],
       errors: [{message: stringsMessage('`Test`')}]
     }, {
-      code: '<Foo bar={`Test`} />',
-      options: [{noStrings: true}],
-      errors: [{message: stringsMessage('`Test`')}]
```
ljharb: maybe instead of removing these tests entirely, we should add an option that makes them valid?

In other words, have the bug fix be "no longer error on props by default", but have an option to error on props, so those who *want* this behavior have a trivial means to get the errors back?

iiison: so, just to confirm, I shall introduce a new property in `options` to check string values in props. 

Will do it.

ljharb: Yep! Then you can keep all these tests, but add an option to make them pass.

iiison:
@ljharb added new option to validate props. I've set it to true by default so that behaviour stays the same, user can set the value to false in order to remove props validation.

One more thing, rule passes if prop is a strings literal(`<div className="abc" />`), but if the prop value is a template literal, then rule will throw error. This behaviour was there from starting, LMK if you want to change this.

iiison:
@ljharb the build is breaking from starting, it works fine in my local, is it because I am using ES6 syntax? Could you please help me if you are aware of the error..

docs/rules/jsx-no-literals.md
```diff
@@ -21,12 +21,15 @@ var Hello = <div>{'test'}</div>;
 
 ### Options
 
-There is only one option:
+There are two options:
 
 * `noStrings` - Enforces no string literals used as children, wrapped or unwrapped.
+* `validateProps` - Enforces no template literals used as props.
```
ljharb: why only template? i'd expect *any* literals to be detected by this option

docs/rules/jsx-no-literals.md
```diff
@@ -21,12 +21,15 @@ var Hello = <div>{'test'}</div>;
...
 
 To use, you can specify like the following:
 
+#### noStrings
+Rule is set to `false` by default
```
ljharb:
```suggestion
set to `false` by default
```

tests/lib/rules/jsx-no-literals.js
```diff
@@ -368,20 +437,6 @@ ruleTester.run('jsx-no-literals', rule, {
         {message: stringsMessage('`foo`')},
         {message: stringsMessage('\'bar\'')}
       ]
-    }, {
-      code: '<Foo bar={`foo` + `bar`} />',
-      options: [{noStrings: true}],
```
ljharb: why do these test cases need to be deleted? if it's behind an option, they should remain passing.

docs/rules/jsx-no-literals.md
iiison: I was trying to mimic the same behaviour as before, will detect string literals too.

tests/lib/rules/jsx-no-literals.js
iiison: these tests seemed duplicate to me, I didn't notice the bar was string literal in first and foo was string in another. will add them again.

iiison: Added tests.

iiison:
@ljharb I see you removed your comment to detect and disallow string literals in props, shall I add that feature or not..

ljharb:
@iiison i marked it resolved because you indicated you'd incorporate it. There's no reason for a rule called "literals" to differentiate between string and template.

iiison:
@ljharb my laptop was broken for some time, can you please provide the context of the last comment that you resolved.. I will fix it as soon as I get a little context of it.

ljharb:
@iiison see https://github.com/yannickcr/eslint-plugin-react/pull/2146#discussion_r254564767, i've unmarked it as resolved. basically, we want template and string literals checked.

iiison:
@ljharb if we don't allow strings in props, then you need to declare variables for all props like `type` attribute in `input` tag element and for a lot of other places too. I think this will complicate things. Let me know your thoughts.

ljharb:
@iiison sure, that's why you might not want to enable the option. There's no difference between `foo="bar"` and `foo={"bar"}` and ``` foo={`bar`} ```.

iiison:
@ljharb updated the PR. Added requested changes. But build is failing. The code is failing in few versions of Node(not sure why), can you please help me with that.

lib/rules/jsx-no-literals.js
```diff
@@ -25,24 +25,38 @@ module.exports = {
...
       additionalProperties: false
     }]
   },
 
-  create: function(context) {
-    const isNoStrings = context.options[0] ? context.options[0].noStrings : false;
-    const sourceCode = context.getSourceCode();
+  create: function({
+    report,
+    options: [configs = {}],
```
ljharb: does this syntax work in node 4?

lib/rules/jsx-no-literals.js
```diff
@@ -25,24 +25,38 @@ module.exports = {
...
         }
       },
       additionalProperties: false
     }]
   },
 
-  create: function(context) {
-    const isNoStrings = context.options[0] ? context.options[0].noStrings : false;
-    const sourceCode = context.getSourceCode();
+  create: function({
```
ljharb: let's try going back to the non-destructuring form.

lib/rules/jsx-no-literals.js
```diff
@@ -25,24 +25,38 @@ module.exports = {
...
+    options: [configs = {}],
+    getSourceCode
+  }) {
+    const {
+      noStrings = false,
+      validateProps = false
+    } = configs;
+    const isNoStrings = noStrings;
+    const shouldValidateProps = validateProps;
+    const sourceCode = getSourceCode();
```
ljharb: in particular, it seems that eslint 3 relies on the receiver (the `this` value) of `getSourceCode` being `context`.

iiison:
fixed all errors!

iiison:
@ljharb Let me know if you want to add more changes to the PR.

tests/lib/rules/jsx-no-literals.js
```diff
@@ -351,37 +366,49 @@ ruleTester.run('jsx-no-literals', rule, {
...
         {message: stringsMessage('`bar`')}
       ]
+    }, {
+      code: '<Foo bar={\'bar\'} />',
+      options: [{noStrings: true, validateProps: true}],
+      errors: [
+        {message: stringsMessage('\'bar\'')}
+      ]
+    }, {
+      code: '<Foo event={() => {}} />',
```
ljharb: this rule is about *string* literals. why would a function value ever trigger an error?

iiison: I suppose I went a little too far on this. Will remove the logic.

docs/rules/jsx-no-literals.md
```diff
@@ -53,6 +54,54 @@ var Hello = <div><Text {...message} /></div>
...
+```jsx
+var Hello = <div class={'xx'} />;
+```
+
+```jsx
+var Hello = <div class={`xx`} />;
+```
+
+```jsx
+var Hello = <div event={()=>{}} />;
```
ljharb: should this be removed from the readme?

iiison:
@ljharb I'm not sure why, but AppVeyor is failing out of nowhere. I just deleted 2 lines from the readme file, not sure why it is breaking.

ljharb:
I'm not sure either, looks like a jest error - either way, could you try rebasing on latest master instead of merging?

iiison:
The build was failing when I didn't merge the master in the branch. but will try the rebase as well.

iiison:
rebased with the master, still, the build is failing with the same error.

iiison:
@ljharb [build is also failing](https://ci.appveyor.com/project/yannickcr/eslint-plugin-react/builds/24159969) for the master

ljharb:
Blocked by #2253.

iiison:
@ljharb as #2253 is closed, can we merge this branch...

mauricecruz:
@iiison any interest in moving this forward? Happy to take a look if you're no longer vested in this fix.

iiison:
@mauricecruz Please go ahead. I won't be able to work on this for a while.

dbassett17:
@mauricecruz @ljharb @iiison Any progress on this or is it still blocked? If nobody else has the time I can probably find some time to finish this up soon.

tests/lib/rules/jsx-no-literals.js
```diff
@@ -415,33 +427,33 @@ ruleTester.run('jsx-no-literals', rule, {
       errors: [{message: stringsMessage('`Test`')}]
     }, {
       code: '<Foo bar={`Test`} />',
-      options: [{noStrings: true}],
+      options: [{noStrings: true, validateProps: true}],
       errors: [{message: stringsMessage('`Test`')}]
     }, {
       code: '<Foo bar={`${baz}`} />',
-      options: [{noStrings: true}],
+      options: [{noStrings: true, validateProps: true}],
```
ljharb: These test changes show that the option (that doesn't exist yet) seems to already be `true` by default prior to this PR. In that case, this option should probably be more like `ignoreProps: true` which defaults to false.

ljharb:
ping @iiison, are you still interested in completing this PR?

SalimBensiali:
@ljharb @iiison, any update on this? Are there any requested changes still pending? Happy to chip in.

ljharb:
@SalimBensiali if you want to post a link to your branch with the needed changes (do NOT post a new PR, please) then i'm happy to pull it in.

tests/lib/rules/jsx-no-literals.js
SalimBensiali: @ljharb just checking... Is this the only requested change pending?

ljharb: @Gasparila at the moment, yes - i'll have to give it a rereview after it's next updated

malyw: Just found this great react eslint rule to use in a big product and faced the same problem that it goes through the props, which was a bit surprising, but ok.
In short, the current rule behavior is to check props, facilitating it to be disabled by default may break some existing users functionalities, who may rely on this.
+1 , is should be either changed to ````validateProps`(`true` default)``` in the current naming, or switched to `ignoreProps`

malyw: Hey @ljharb @SalimBensiali
I created a PR to switch to `ignoreProps: true` and adjusted the code, docs and tests to pass: https://github.com/iiison/eslint-plugin-react/pull/1
It targets the current branch, can you please review when you have time, and let me know if you'd need me to merge it, or you'd prefer other way of having this arrived in the main repo

malyw:
Hey @ljharb @SalimBensiali @iiison 
I created a PR to switch to ignoreProps: true and adjusted the code, docs and tests to pass: https://github.com/iiison/eslint-plugin-react/pull/1
It targets the current branch, can you please review when you have time, and let me know if you'd need me to merge it, or you'd prefer other way of having this arrived in the main repo

PR #2158, opened 2019-02-11
Add jsx frag pragma support

### LINKED ISSUE #2156: jsx-uses-react rule does not detect jsxFrag rule
In babel, if one does not use `jsxFrag`, fragment syntax always requires `React` import (to use `React.Fragment`).
However, `jsx-uses-react` rule does not check this pragma and does not require `React` import when jsx pragma is set.
I think the rule should check `jsxFrag` pragma too and mark `React` as used if there is no `jsxFrag` pragma.

Ailrun:
Resolves #2156.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,10 @@ You should also specify settings that will be shared across all the plugin rules
     "react": {
       "createClass": "createReactClass", // Regex for Component Factory to use,
                                          // default to "createReactClass"
-      "pragma": "React",  // Pragma to use, default to "React"
+      "jsxPragma": "React.createElement" // JSX Pragma to use, default to "<settings.react.pragma>.createElement"
+      "jsxFragPragma": "React.Fragment" // JSX Fragment Pragma to use, default to "<settings.react.pragma>.<settings.react.fragment>"
+      "pragma": "React",  // *Deprecated* Pragma to use, default to "React"
+      "fragment": "Fragment", // *Deprecated* Element of Pragment to use as Fragment, default to "Fragment"
       "version": "detect", // React version. "detect" automatically picks the version you have installed.
                            // You can also use `16.0`, `16.3`, etc, if you want to override the detected value.
       "flowVersion": "0.53" // Flow version
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ You should also specify settings that will be shared across all the plugin rules
     "react": {
       "createClass": "createReactClass", // Regex for Component Factory to use,
                                          // default to "createReactClass"
+      "localName": "React" // Local name of React import, default to "<setttings.react.pragma>"
       "jsxPragma": "React.createElement" // JSX Pragma to use, default to "<settings.react.pragma>.createElement"
       "jsxFragPragma": "React.Fragment" // JSX Fragment Pragma to use, default to "<settings.react.pragma>.<settings.react.fragment>"
       "pragma": "React",  // *Deprecated* Pragma to use, default to "React"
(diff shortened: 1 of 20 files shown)


README.md
```diff
@@ -38,7 +38,11 @@ You should also specify settings that will be shared across all the plugin rules
     "react": {
       "createClass": "createReactClass", // Regex for Component Factory to use,
                                          // default to "createReactClass"
-      "pragma": "React",  // Pragma to use, default to "React"
+      "localName": "React" // Local name of React import, default to "<setttings.react.pragma>"
+      "jsxPragma": "React.createElement" // JSX Pragma to use, default to "<settings.react.pragma>.createElement"
+      "jsxFragPragma": "React.Fragment" // JSX Fragment Pragma to use, default to "<settings.react.pragma>.<settings.react.fragment>"
+      "pragma": "React",  // *Deprecated* Pragma to use, default to "React"
```
ljharb: i'm confused, why would this be deprecated?

README.md
```diff
@@ -38,7 +38,11 @@ You should also specify settings that will be shared across all the plugin rules
...
+      "fragment": "Fragment", // *Deprecated* Element of Pragment to use as Fragment, default to "Fragment"
```
ljharb: similarly, why deprecate this?

ljharb (changes requested):
The values in "settings" have no need to be the same as the inline jsdoc comment names; why deprecate the nice ones?

README.md
Ailrun: Since semantics changed, I think new settings should be provided.
If you think just breaking those (and probably update major version) is better, I will update this PR.

ljharb: How have the semantics changed (for users who haven't added the new pragma settings)?

Ailrun: I mean, if old users use `pragma` with `React`, originally rules will detect `<pragma>.createElement`. With `jsxPragma`, rules detect `<jsxPragma>` so that its behavior aligns with babel's.
Therefore, if I just use `pragma` and do not add new settings, it will break existing settings.

ljharb: Right but why would `jsxPragma` even exist in the common case for users of this plugin, if it doesn't do anything for them?

Ailrun: @ljharb Could you tell me your suggestion for the setting combination?
I don't get the right point I need to explain, so I think that would be easier to me.

Ailrun: @ljharb FYI, what I considered are
1. A case with custom element creator
    ```
    import { jsx } from 'my-jsx';
    import React from 'react';

    <><div /></>
    ```
1. A case with different local name for React
    ```
    import R from 'react';

    <><div /></>
    ```
1. A case with different library with Fragment support.
    ```
    import O from 'other-fragment-support-jsx-library';

    <><div /></>
    ```
1. Cases with mixes of others.

ljharb: OK, so. To support your first 3 cases:
1. currently not supported - `createElement` is hardcoded off of the `pragma`.
2. currently uses `pragma`; `localName` is fine to also support via an inline comment, but `pragma` should remain the only way to configure this in settings.
3. currently uses `fragment` in `getFragmentFromContext`, and then pairs that [here](https://github.com/yannickcr/eslint-plugin-react/blob/dd0757fbf7ab96770282a2a64c9b696856a47235/lib/rules/jsx-fragments.js#L43-L44) with `pragma` - if `jsxFragPragma` should override both, then that's fine to support as an inline comment, and maybe we could make the `fragment` setting be either a string (paired with pragma), or an object where either is supported?
4. without specifics i'm not sure what we need to do here.

It seems like we should maybe do these new features as separate PRs.

The first one seems to be 3 - ie, add support for an alternative schema in `fragment`, that inline `jsxFrag` comments hook into.

Ailrun:
I don't have enough time to manage this PR, so I close this.

ljharb:
If it’s ok, I’d prefer to keep it open in case someone else can help by posting a link to a branch (not an additional PR, please)

PR #2163, opened 2019-02-13
Update forbid-prop-types: handle prop-type functions with no args

### LINKED ISSUE #1673: Enhance forbid-prop-types to check for shape({})
Does it make sense to enhance the `forbid-prop-types` rule to check for cases using `shape` that do not actually add clarity to the code, for example `shape({})`.

From https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/forbid-prop-types.md

> By default this rule prevents vague prop types with more specific alternatives available

And:

> This rule encourages prop types that more specifically document their usage.

When using for example `shape({})` we are not more specifically documenting prop usage.

jack-lewin:
This PR aims to fix #1673 by adding a new option to the forbid-prop-types rule: `empty`.

This will catch prop types which are functions, such as `shape()` or `arrayOf()`, but are used with either no arguments, or empty arguments (`{}` or `[]`). These prop types don't add a huge amount of value in those scenarios, and should be avoided if possible.

Feedback appreciated 🙂

### DIFF
--- a/lib/rules/forbid-prop-types.js
+++ b/lib/rules/forbid-prop-types.js
@@ -13,6 +13,7 @@ const propWrapperUtil = require('../util/propWrapper');
 // Constants
 // ------------------------------------------------------------------------------
 
+const EMPTY_RULE = 'empty';
 const DEFAULTS = ['any', 'array', 'object'];
 
 // ------------------------------------------------------------------------------
@@ -53,9 +54,30 @@ module.exports = {
     const checkContextTypes = configuration.checkContextTypes || false;
     const checkChildContextTypes = configuration.checkChildContextTypes || false;
 
-    function isForbidden(type) {
+    function isEmptyArgument(arg) {
+      const isEmptyArray = arg.type === 'ArrayExpression' && arg.elements.length === 0;
+      const isEmptyObject = arg.type === 'ObjectExpression' && arg.properties.length === 0;
+
+      return isEmptyArray || isEmptyObject;
+    }
+
+    function isMissingArguments(type, value) {
       const forbid = configuration.forbid || DEFAULTS;
-      return forbid.indexOf(type) >= 0;
+      const propIsAFunction = propsUtil.isPropTypeAFunction(type);
+
+      if (forbid.indexOf(EMPTY_RULE) === -1 || !propIsAFunction) {
+        return false;
+      }
+
+      const hasArguments = value.parent && Array.isArray(value.parent.arguments) && value.parent.arguments.length > 0;
+      const noPropertiesInArguments = hasArguments && value.parent.arguments.every(isEmptyArgument);
+
+      return !hasArguments || noPropertiesInArguments;
+    }
+
+    function isForbidden(type, value) {
+      const forbid = configuration.forbid || DEFAULTS;
+      return forbid.indexOf(type) >= 0 || isMissingArguments(type, value);
     }
 
     function shouldCheckContextTypes(node) {
@@ -103,7 +125,7 @@ module.exports = {
         } else if (value.type === 'Identifier') {
           target = value.name;
         }
-        if (isForbidden(target)) {
+        if (isForbidden(target, value)) {
           context.report({
             node: declaration,
             message: `Prop type \`${target}\` is forbidden`
--- a/lib/util/props.js
+++ b/lib/util/props.js
@@ -5,6 +5,8 @@
 
 const astUtil = require('./ast');
 
+const PROP_TYPE_FUNCTIONS = ['object', 'shape', 'instanceOf', 'oneOf', 'oneOfType', 'objectOf', 'arrayOf', 'exact'];
+
 /**
  * Checks if the Identifier node passed in looks like a propTypes declaration.
  * @param {ASTNode} node The node to check. Must be an Identifier node.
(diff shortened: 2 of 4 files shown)


lib/rules/forbid-prop-types.js
```diff
@@ -103,7 +125,7 @@ module.exports = {
         } else if (value.type === 'Identifier') {
           target = value.name;
         }
-        if (isForbidden(target)) {
+        if (isForbidden(target, value)) {
           context.report({
             node: declaration,
             message: `Prop type \`${target}\` is forbidden`
```
jack-lewin: Would it be worth adding a new error message here?

lib/util/props.js
```diff
@@ -5,6 +5,8 @@
 
 const astUtil = require('./ast');
 
+const PROP_TYPE_FUNCTIONS = ['object', 'shape', 'instanceOf', 'oneOf', 'oneOfType', 'objectOf', 'arrayOf', 'exact'];
```
jack-lewin: I'm not sure whether there's a better way of checking this, instead of hard-coding the list?

lib/rules/forbid-prop-types.js
ljharb: yes - when the `empty` option is set, I think the message should be more specific.

lib/util/props.js
ljharb: something like
```js
Object.keys(PropTypes)
```
?

lib/util/props.js
```diff
@@ -63,10 +65,20 @@ function isRequiredPropType(propTypeExpression) {
   return propTypeExpression.type === 'MemberExpression' && propTypeExpression.property.name === 'isRequired';
 }
 
+/**
+ * Checks if the PropType is a function.
+ * @param {String} propType to check.
+ * @returns {Boolean} `true` if the PropType is a function, `false` if not.
+ */
+function isPropTypeAFunction(propType) {
+  return PROP_TYPE_FUNCTIONS.indexOf(propType) > -1;
```
ljharb: i'm a bit confused here; are we already sure by the time `propType` comes in that it's off of `prop-types`? Or if i have `import { shape } from 'airbnb-prop-types`, will that trigger it?

tests/lib/rules/forbid-prop-types.js
```diff
@@ -892,6 +1024,231 @@ ruleTester.run('forbid-prop-types', rule, {
...
+      '  propTypes: {',
+      '    retailer: PropTypes.exact({}),',
+      '  },',
+      '  render: function() {',
+      '    return <div />;',
+      '  }',
+      '});'
+    ].join('\n'),
+    options: [{
+      forbid: ['empty']
```
ljharb: this doesn't really make sense to me as a schema; i think it'd be better to have `forbidEmpty: true` alongside `forbid`.

lib/util/props.js
jack-lewin: At this point `propType` can be from anywhere - I hadn't considered the scenario of named custom validators like `airbnb-prop-types`. So yes, the example you mentioned would also trigger this error.

I'm not sure how to go about avoiding that 😕 Do you have any ideas?

ljharb: Track where it's imported/required from, and ensure it's `'prop-types'`?

lib/util/props.js
jack-lewin: That would include prop-types which are not functions, e.g. `string`.

Seeing as these are unlikely to change much, I think I'm happy with this, if you are?

lib/util/props.js
jack-lewin: I've given that a try 😄

lib/util/ast.js
```diff
@@ -148,6 +148,40 @@ function isClass(node) {
   return node.type === 'ClassDeclaration' || node.type === 'ClassExpression';
 }
 
+const DECL_OTHER = 'other',
+  DECL_REQUIRE = 'require',
```
ljharb: each variable should be declared by itself, please

lib/rules/forbid-prop-types.js
```diff
@@ -67,6 +69,21 @@ module.exports = {
     function isMissingArguments(type, value) {
       const propIsAFunction = propsUtil.isPropTypeAFunction(type);
 
+      // disregard custom prop-type imports (i.e. not the react prop-types)
```
ljharb: let's also be sure to support `React.PropTypes`

tests/lib/rules/forbid-prop-types.js
```diff
@@ -1458,4 +2025,5 @@ ruleTester.run('forbid-prop-types', rule, {
     }],
     errors: 1
   }]
+
```
ljharb:
```suggestion
```

jack-lewin:
I've added a lot of tests in this PR. Happy to cut them down if you think that would make it easier for other contributors to follow.

lib/rules/forbid-prop-types.js
```diff
@@ -13,6 +13,7 @@ const propWrapperUtil = require('../util/propWrapper');
 // Constants
 // ------------------------------------------------------------------------------
 
+const REACT_MODULE = 'react';
```
ljharb: This needs to come from the pragma, so it can be overridden - not hardcoded to "React".

jack-lewin: Sorry, I'm not sure I understand this.

From what I can see, the pragma util refers to instances where the code refers to an imported `React`, for example `React.Fragment`. I was using the example below as a point of reference:

https://github.com/yannickcr/eslint-plugin-react/blob/4a72e6a0784f7cc7f16f4574ce7f14de55540a73/lib/rules/jsx-fragments.js#L97-L122


I'm using this `REACT_MODULE` constant to determine when something is _being imported_ from React, for example `import { PropTypes } from React`. Do I need to make changes to the pragma util in order to support this? Or is there something else I've missed here?

Thanks for your help with this PR, so far 🙂

ljharb: ah, i see what you're saying. I suppose this is fine as-is, then.

jack-lewin:
Just made a couple of small changes to make sure this is backwards-compatible 🙂

lib/rules/forbid-prop-types.js
```diff
@@ -97,6 +139,7 @@ module.exports = {
           value.callee.type === 'MemberExpression'
         ) {
           value = value.callee;
+          value.parent = {arguments: declaration.value.arguments};
```
ljharb: is this mutation necessary?

lib/util/props.js
ljharb: True - but they do indeed change; I just added a new one a couple weeks ago (see #2188 where this technique actually prevented a bug)

ljharb:
@jack-lewin ping, are you still interested in completing this PR?

jack-lewin:
Sorry it's been so long, I'm not going to revisit this one - closing the PR 🧹

PR #2182, opened 2019-02-28
[jsx-sort-default-props] Bug fix for one case with spread

### LINKED ISSUE #2178: Add option to ignore spread operator in sort-default-props
With `foo = { bar: baz }`.
```
{ ...foo, bar } = { bar }
```
And
```
{ bar, ...for } = { baz }
```

So I think, an option to ignore spread operator in sort default props will be great.
Correct example
```
{ a: 'a', c: 'c', ...foo, b: 'b', d: 'd' }
```
Incorrect example
```
{ a: 'a', ...foo, c: 'c', b: 'b', d: 'd' }
``` 

Plus, it allow to develop a fixer (asked [here](https://github.com/yannickcr/eslint-plugin-react/issues/2062)) without breaking the code.

VincentLanglet:
Closes #2178

I thought an option was needed, but instead it was a bug

### DIFF
--- a/lib/rules/jsx-sort-default-props.js
+++ b/lib/rules/jsx-sort-default-props.js
@@ -96,7 +96,7 @@ module.exports = {
      */
     function checkSorted(declarations) {
       declarations.reduce((prev, curr, idx, decls) => {
-        if (/SpreadProperty$/.test(curr.type)) {
+        if (/Spread(?:Property|Element)$/.test(curr.type)) {
           return decls[idx + 1];
         }
 
--- a/tests/lib/rules/jsx-sort-default-props.js
+++ b/tests/lib/rules/jsx-sort-default-props.js
@@ -326,6 +326,25 @@ ruleTester.run('jsx-sort-default-props', rule, {
       'First.propTypes = propTypes;',
       'First.defaultProps = defaultProps;'
     ].join('\n')
+  }, {
+    code: [
+      'const defaults = {',
+      '  b: "b"',
+      '};',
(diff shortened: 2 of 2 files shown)


tests/lib/rules/jsx-sort-default-props.js
```diff
@@ -314,12 +314,17 @@ ruleTester.run('jsx-sort-default-props', rule, {
...
+      '  b: "b"',
+      '};',
       'const First = (props) => <div />;',
       'export const propTypes = {',
       '    a: PropTypes.any,',
+      '    b: PropTypes.string,',
       '    z: PropTypes.string,',
       '};',
       'export const defaultProps = {',
+      '    ...defaults,',
```
ljharb: instead of modifying this test, could you add a new one?

VincentLanglet: Done

lib/rules/jsx-sort-default-props.js
```diff
@@ -96,7 +96,7 @@ module.exports = {
      */
     function checkSorted(declarations) {
       declarations.reduce((prev, curr, idx, decls) => {
-        if (/SpreadProperty$/.test(curr.type)) {
+        if (/SpreadProperty$/.test(curr.type) || /SpreadElement$/.test(curr.type)) {
```
ljharb:
```suggestion
        if (/Spread(?:Property|Element)$/.test(curr.type)) {
```

VincentLanglet:
@ljharb Done :)

PR #2191, opened 2019-03-08
[new] Add jsx-props-no-spreading rule

### LINKED ISSUE #1094: Rule idea: no spreading props
One thing we've noticed tends to make code a little opaque is when we destructure the entire props for a sub-component:

```jsx
function MyComponent(props) {
  return <MySubComponent {...props} />
}
```

It would be great to have an eslint rule to force expanding individual props that should be passed down:

```jsx
function MyComponent(props) {
  return <MySubComponent one_prop={props.one_prop} two_prop={props.two_prop} />
}
```

ashbhir:
This adds a rule to disallow JSX props spreading as it reduces both the readability and maintainability of the code. Fixes #1094.

So, instead of passing props like this:
```
function MyComponent(props) {
  return <MySubComponent {...props} />
}
```

This rule will enforce passing props like this:
```
function MyComponent(props) {
  return <MySubComponent one_prop={props.one_prop} two_prop={props.two_prop} />
}
```

### DIFF
--- a/docs/rules/jsx-props-no-spreading.md
+++ b/docs/rules/jsx-props-no-spreading.md
@@ -1,0 +1,109 @@
+# Disallow JSX props spreading (react/jsx-props-no-spreading)
+
+Enforces that there is no spreading for any JSX attribute. This enhances readability of code by being more explicit about what props are received by the component. It is also good for maintainability by avoiding passing unintentional extra props and allowing react to emit warnings when invalid HTML props are passed to HTML elements.
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```jsx
+<App {...props} />
+<MyCustomComponent {...props} some_other_prop={some_other_prop} />
+<img {...props} />
+```
+
+The following patterns are **not** considered warnings:
+
+```jsx
+const {src, alt} = props;
+const {one_prop, two_prop} = otherProps;
+<MyCustomComponent one_prop={one_prop} two_prop={two_prop} />
+<img src={src} alt={alt} />
+```
+
+
+## Rule Options
+
+```js
+...
+"react/jsx-props-no-spreading": [{
+    "html": "ignore" / "enforce",
+    "custom": "ignore" / "enforce",
+    "exceptions": [<string>]
+}]
+...
+```
+
+### html
+
+`html` set to `ignore` will ignore all html jsx tags like `div`, `img` etc. Default is set to `enforce`.
+
+The following patterns are **not** considered warnings when `html` is set to `ignore`:
+
+```jsx
+<img {...props} />
+```
+
+The following patterns are still considered warnings:
+
+```jsx
+<MyCustomComponent {...props} />
+```
+
+### custom
+
+`custom` set to `ignore` will ignore all custom jsx tags like `App`, `MyCustomComponent` etc. Default is set to `enforce`.
+
+The following patterns are **not** considered warnings when `custom` is set to `ignore`:
+
+```jsx
+<MyCustomComponent {...props} />
+```
+
+The following patterns are still considered warnings:
+```jsx
+<img {...props} />
+```
+
+### exceptions
+
+An "exception" will always flip the resulting html or custom setting for that component - ie, html set to `ignore`, with an exception of `div` will enforce on an `div`; custom set to `enforce` with an exception of `Foo` will ignore `Foo`.
+
+```js
+{ "exceptions": ["Image", "img"] }
+```
+
+The following patterns are **not** considered warnings:
+
+```jsx
+const {src, alt} = props;
+<Image {...props} />
+<img {...props} />
+```
+
+The following patterns are considered warnings:
+```jsx
+<MyCustomComponent {...props} />
+```
+
+```js
+{ "html": "ignore", "exceptions": ["MyCustomComponent", "img"] }
+```
+
+The following patterns are **not** considered warnings:
+
+```jsx
+const {src, alt} = props;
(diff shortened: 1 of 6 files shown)


ljharb:
also hopefully like this:
```jsx
function MyComponent({ one_prop, two_prop }) {
  return <MySubComponent one_prop={one_prop} two_prop={two_prop} />
}
```

docs/rules/jsx-props-no-spreading.md
```diff
@@ -0,0 +1,24 @@
...
+The following patterns are **not** considered warnings:
+
+```jsx
+const {one_prop, two_prop} = props;
+<App one_prop={one_prop} two_prop={two_prop} />
+```
+
+## When Not To Use It
+
+If you are not using JSX or have lots of props to be passed or don't worry much about maintainability.
```
ljharb: "don't worry much about maintainability" is a bit snarky and passive aggressive. Let's reword.

docs/rules/jsx-props-no-spreading.md
```diff
@@ -0,0 +1,24 @@
+# Disallow JSX props spreading (react/jsx-props-no-spreading)
+
+Enforces that there is no spreading for any JSX attribute. This enhances readability of code and hence leads to better maintainability.
```
ljharb: Let's elaborate here on why it helps maintainability - this should talk about a) avoiding passing unintentional extra props, b) being explicit about what props are received and written, c) avoiding react warnings when invalid HTML props are passed to HTML elements, etc.

lib/rules/jsx-props-no-spreading.js
```diff
@@ -0,0 +1,33 @@
...
+      description: 'Prevent JSX prop spreading',
+      category: 'Best Practices',
+      recommended: false,
+      url: docsUrl('jsx-props-no-spreading')
+    }
+  },
+
+  create: function(context) {
+    return {
+      JSXSpreadAttribute: function(node) {
```
ljharb:
```suggestion
      JSXSpreadAttribute: function (node) {
```

lib/rules/jsx-props-no-spreading.js
```diff
@@ -0,0 +1,33 @@
...
+  meta: {
+    docs: {
+      description: 'Prevent JSX prop spreading',
+      category: 'Best Practices',
+      recommended: false,
+      url: docsUrl('jsx-props-no-spreading')
+    }
+  },
+
+  create: function(context) {
```
ljharb:
```suggestion
  create: function (context) {
```

docs/rules/jsx-props-no-spreading.md
ashbhir: Yes, very much agree. Updated

ashbhir:
> I think this will need an `exceptions` array option as well, since spreading props is actually necessary for HOCs.
> 
> Additionally, I think it would be ideal to provide another option that controls whether this rule applies to custom component elements, or HTML elements, or both.

Added option `allowedFor` for catering exceptions, namely: `HTMLTags`, `Components` and selective custom components.

ljharb:
hmm. Thanks for adding that! We probably should have discussed the schema a bit more :-)

Let's do something like this:
```js
{
  html: 'enforce'/'ignore',
  custom: 'enforce'/'ignore',
  exceptions: ['ComponentName', 'OtherComponentName'],
}
```
and make sure it errors if `html` and `custom` are both false. An "exception" will always flip the resulting `html` or `custom` setting for that component - ie, `html` set to `ignore`, with an exception of `div` will enforce on a div; `custom` set to `enforce` with an exception of `Foo` will ignore Foo.

ashbhir:
> hmm. Thanks for adding that! We probably should have discussed the schema a bit more :-)
> 
> Let's do something like this:
> 
> ```js
> {
>   html: 'enforce'/'ignore',
>   custom: 'enforce'/'ignore',
>   exceptions: ['ComponentName', 'OtherComponentName'],
> }
> ```
> 
> and make sure it errors if `html` and `custom` are both false. An "exception" will always flip the resulting `html` or `custom` setting for that component - ie, `html` set to `ignore`, with an exception of `div` will enforce on a div; `custom` set to `enforce` with an exception of `Foo` will ignore Foo.

Sure, made the respective changes. 
`and make sure it errors if `html` and `custom` are both false`. How do we do this in schema ?

ljharb:
That's a good question :-) it'll probably have to be something like:
```
allOf: [
  {
    oneOf: [
      bothEnforce,
      htmlEnforceCustomIgnore,
      htmlIgnoreCustomEnforce
    ],
  },
  exceptionsSchema
]
```

ashbhir:
> That's a good question :-) it'll probably have to be something like:
> 
> ```
> allOf: [
>   {
>     oneOf: [
>       bothEnforce,
>       htmlEnforceCustomIgnore,
>       htmlIgnoreCustomEnforce
>     ],
>   },
>   exceptionsSchema
> ]
> ```

I think the `allOf` won't work as then one has to mention all properties. But one may just want to mention the exceptions assuming others to be set as default.
Also, do we really need to restrict both `ignore` mode? What if one wants to ignore everything barring some very important Custom Components. For them the configuration will be:
```
{
    html: 'ignore',
    custom: 'ignore',
    exceptions: ['CustomComponent', 'img']
}
```

ashbhir:
> ah true, that's a good point.
> 
> I'd like the schema to be as explicit as possible. If both are "ignore", then `exceptions` must not be empty.

@ljharb made the requested changes.

PR #2193, opened 2019-03-10
[new] add static-property-placement rule

### LINKED ISSUE #1807: Experimental features?
I have made a rule that enforces you to specify `propTypes` and `defaultProps` as static class properties. I don't know if I should make a PR for this since it depends on the experimental feature. We are planning on using it in our codebase, which is large. Would it be of interest here?

INCORRECT
`
class MyComp extends Component {}
MyComp.propTypes = {};
MyComp.defaultProps = {}
`

CORRECT
`
class MyComp extends Component {
  static propTypes = {};
  static defaultProps = {};
}
`

### LINKED ISSUE #1923: Add rule to require `static propTypes = ..` instad of `Foo.propTypes = ..` for class-based components
With such a rule the following would be invalid:

```javascript
class Foo extends React.Component {}
Foo.propTypes = {...}
Foo.defaultProps = {...}
```

and this would be valid:

```javascript
    static propTypes = {...};
    static defaultProps = {...};
}

function Bar() {
    return <>...</>;
}
Bar.propTypes = {...};
```

dmason30:
This adds a rule discussed in these two issues:

* Fixes #1923
* Fixes #1807

I have based the options on @ljharb suggestions from both issues. This is my first attempt at writing a rule so I expect plenty of feedback. 
* `static public field` (`ClassProperty`)
* `static getter` {`MethodDefinition`)
* `property assignment` (`MemberExpression`)

```
...
"react/static-property-placement": [<enabled>, <string>, {
  childContextTypes: <string>,
  contextTypes: <string>,
  contextType: <string>,
  defaultProps: <string>,
  displayName: <string>,
  propTypes: <string>,
}]
...
```

- Also I have never worked with TypeScript/Flow so I am not sure what (if any) additional changes are needed?

- **This rule only applies to ES6 classes.** 
-- Ignored: Stateless functional components can only ever have a `MemberExpression` to declare any of the above properties. 
-- Ignored: The `React.createClass/createReactClass` can only ever have a `Property` to declare any of the above properties. 

- This does not have `autofix` either but I feel that is a "nice to have" option and could come later.

### DIFF
--- a/docs/rules/static-property-placement.md
+++ b/docs/rules/static-property-placement.md
@@ -1,0 +1,185 @@
+# Enforces where React component static properties should be positioned. (static-property-placement)
+
+This rule allows you to enforce where `childContextTypes`, `contextTypes`, `contextType`, `defaultProps`, `displayName`,
+and `propTypes` are declared in an ES6 class.
+
+
+## Rule Details
+
+By default, this rule will check for and warn about declaring any of the above properties outside of the class body.
+
+There are three key options are `static public field`, `static getter`, and `property assignment`. 
+
+### When `static public field` is enabled (default):
+
+Examples of **incorrect** code for this rule:
+
+```js
+class MyComponent extends React.Component {
+  static get childContextTypes() { /*...*/ }  
+  static get contextTypes() { /*...*/ }  
+  static get contextType() { /*...*/ }  
+  static get displayName() { /*...*/ }  
+  static get defaultProps() { /*...*/ }   
+  static get propTypes() { /*...*/ }
+}
+```
+
+```js
+class MyComponent extends React.Component { /*...*/ }
+MyComponent.childContextTypes = { /*...*/ };
+MyComponent.contextTypes = { /*...*/ };
+MyComponent.contextType = { /*...*/ };
+MyComponent.displayName = "Hello";
+MyComponent.defaultProps = { /*...*/ };
+MyComponent.propTypes = { /*...*/ };
+```
+
+Examples of **correct** code for this rule:
+
+```js
+class MyComponent extends React.Component {
+  static childContextTypes = { /*...*/ };  
+  static contextTypes = { /*...*/ };  
+  static contextType = { /*...*/ };  
+  static displayName = "Hello";  
+  static defaultProps = { /*...*/ };  
+  static propTypes = { /*...*/ };
+}
+```
+
+### When `static getter` is enabled:
+
+Examples of **incorrect** code for this rule:
+
+```js
+class MyComponent extends React.Component {
+  static childContextTypes = { /*...*/ };  
+  static contextTypes = { /*...*/ };  
+  static contextType = { /*...*/ };  
+  static displayName = "Hello";  
+  static defaultProps = { /*...*/ };  
+  static propTypes = { /*...*/ };
+}
+```
+
+```js
+class MyComponent extends React.Component { /*...*/ }
+MyComponent.childContextTypes = { /*...*/ };
+MyComponent.contextTypes = { /*...*/ };
+MyComponent.contextType = { /*...*/ };
+MyComponent.displayName = "Hello";
+MyComponent.defaultProps = { /*...*/ };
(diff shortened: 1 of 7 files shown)


dmason30:
This is "Ready for review", sorry I opened as a draft.

docs/rules/static-property-placement.md
```diff
@@ -0,0 +1,173 @@
+# Enforces where React component static properties should be positioned. (static-property-placement)
+
+This rule allows you to enforce where `childContextTypes`, `contextTypes`, `defaultProps`, `displayName`
```
ljharb: Don't forget the new `contextType`.
```suggestion
This rule allows you to enforce where `childContextTypes`, `contextTypes`, `contextType`, `defaultProps`, `displayName`,
```

docs/rules/static-property-placement.md
```diff
@@ -0,0 +1,173 @@
...
+  childContextTypes: "static public field",
+  contextTypes: "static public field",
+  displayName: "static public field",
+}]
+...
+```
+
+Based on the above configuration `defaultProps` and `propTypes` will both follow the `property assignment` rule.
+
+The `childContextTypes`, `contextTypes` and `displayName` will  enforce the `static public field` rule.
```
ljharb:
```suggestion
`childContextTypes`, `contextTypes`, and `displayName` will  enforce the `static public field` rule.
```

docs/rules/static-property-placement.md
```diff
@@ -0,0 +1,173 @@
...
+
+## When Not To Use It
+
+If you have no placement preference, although you should choose one so that your code is consistent.
```
ljharb:
```suggestion
If you have no placement preference for React's static class properties.
```

lib/rules/static-property-placement.js
```diff
@@ -0,0 +1,144 @@
...
+  childContextTypes: propsUtil.isChildContextTypesDeclaration,
+  contextTypes: propsUtil.isContextTypesDeclaration,
+  displayName: node => propsUtil.isDisplayNameDeclaration(astUtil.getPropertyNameNode(node))
+};
+
+const classProperties = Object.keys(propertiesToCheck);
+
+const schemaProperties = {};
+classProperties.map(property => {
+  schemaProperties[property] = {enum: POSITION_SETTINGS};
```
ljharb: Please don't use `.map` for side effects; instead of this, let's do:
```js
const fromEntries = require('object.fromentries');
const schemaProperties = fromEntries(classProperties.map(property => [property, {enum: POSITION_SETTINGS}]))l
```

lib/rules/static-property-placement.js
```diff
@@ -0,0 +1,144 @@
...
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Defines where React component static properties should be positioned.',
+      category: 'Stylistic Issues',
+      recommended: false,
+      url: docsUrl('static-property-placement')
+    },
+    fixable: null, // or 'code' or 'whitespace'
```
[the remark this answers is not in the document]
dmason30: I'm not really sure where to start, Everything, I have done so far has been based on looking at and learning from other rules. The fixable code in the other rules looks a lot like trial and error and I could possibly hack something together which may not be the most efficient. 

I did mention in the OP that making it fixable is _nice to have_ but I feel more experienced hands with this codebase would know how to write an _fixer_ for this in the best way.

dmason30:
Added the `static getter` option - Great Idea 🎉 

I await your feedback. 😄

lib/rules/static-property-placement.js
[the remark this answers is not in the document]
ljharb: It’s fine if we want to try to do that as a followup; the real value in this rule imo is if it can autofix tho :-)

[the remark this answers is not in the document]
dmason30: I think leaving it as a follow up is best. I make no promises that it will be me 😄

dmason30:
@ljharb LGTM?

PR #2203, opened 2019-03-17
[Fix] `jsx-curly-brace-presence`: accept multiline template string

### LINKED ISSUE #1592: jsx-curly-brace-presence `never` breaks multi-line strings
Using eslint-preset-airbnb, the [`jsx-curly-brace-presence` rule](https://github.com/airbnb/javascript/blob/44dbd0bdc41d08eb5de8ad698099ae44240f4b0d/packages/eslint-config-airbnb/rules/react.js#L355) breaks multi-line strings given as props, e.g.:

```jsx
<div css={`
  background: #009688;
  transition: background 0.5s;
`}>
  Hello world!
</div>
```

Gets transformed into

```jsx
<div css="
  background: #009688;
  transition: background 0.5s;
">
  Hello world!
</div>
```

By `eslint --fix`, which is invalid. Multi-line strings given as props shall not be flagged as mistakes.

golopot:
Fixes #1592

Multi-line template string is no longer considered error.
```javascript
<App prop={`
   a
   b
`} />
```

### DIFF
--- a/tests/lib/rules/jsx-curly-brace-presence.js
+++ b/tests/lib/rules/jsx-curly-brace-presence.js
@@ -241,6 +241,44 @@ ruleTester.run('jsx-curly-brace-presence', rule, {
     {
       code: ['<a a={"start\\', '\\', 'end"}/>'].join('/n'),
       options: ['never']
+    },
+    {
+      code: `
+        <App prop={\`
+          a
+          b
+        \`} />
+      `,
+      options: ['never']
+    },
+    {
+      code: `
+        <App prop={\`
+          a
+          b
+        \`} />
+      `,
+      options: ['always']
+    },
+    {
+      code: `
+        <App>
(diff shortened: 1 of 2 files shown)


tests/lib/rules/jsx-curly-brace-presence.js
```diff
@@ -241,6 +241,26 @@ ruleTester.run('jsx-curly-brace-presence', rule, {
...
       options: ['never']
+    },
+    {
+      code: `
+        <App prop={\`
+          a
+          b
+        \`} />
+      `,
+      options: ['never']
```
ljharb: can we add similar test cases for the "always" option?

golopot: Added

golopot:
thanks

PR #2206, opened 2019-03-17
test(jsx-curly-spacing): test against fixed in the past bug

### LINKED ISSUE #998: jsx-curly-spacing looks into functions
It produces such warning (or error, It's not important)
```
/foo/src/components/Bar.js
  75:93  error  There should be no space before '}'  react/jsx-curly-spacing
```
On this line:
```
onLayout={() => { /* dummy callback to fix android bug with component measuring */ }}
```

I've got to debug this rule, this is what I've noticed.
for this function:
```
isSpaceBetweenTokens(first, second) {
        const text = this.text.slice(first.range[1], second.range[0]);

        return /\s/.test(text.replace(/\/\*.*?\*\//g, ""));
    }
```

these arguments are passed:
__first__:
```
{
  "type": "Block",
  "value": " dummy callback to fix android bug with component measuring ",
  "start": 1722,
  "end": 1786,
  "loc": {
    "start": {
      "line": 75,
      "column": 26
    },
    "end": {
      "line": 75,
      "column": 90
    }
  },
  "range": [
    1722,
    1786
  ]
}
```

__second__:
```
{
  "type": "Punctuator",
  "value": "}",
  "start": 1788,
  "end": 1789,
  "loc": {
    "start": {
      "line": 75,
      "column": 92
    },
    "end": {
      "line": 75,
      "column": 93
    }
  },
  "range": [
    1788,
    1789
  ]
}
```

and `text` becomes --> `" }"` ... but that _space_ belongs to function, not for jsx curly brace

version is `eslint-plugin-react@6.8.0`

ColCh:
Fixes #998

### DIFF
--- a/tests/lib/rules/jsx-curly-spacing.js
+++ b/tests/lib/rules/jsx-curly-spacing.js
@@ -680,6 +680,9 @@ ruleTester.run('jsx-curly-spacing', rule, {
   }, {
     code: '<>{bar} {baz}</>;',
     parser: 'babel-eslint'
+  }, {
+    code: '<div onLayout={() => { /* dummy callback to fix android bug with component measuring */ }} />',
+    parser: 'babel-eslint'
   }],
 
   invalid: [{


tests/lib/rules/jsx-curly-spacing.js
```diff
@@ -2281,5 +2281,14 @@ ruleTester.run('jsx-curly-spacing', rule, {
...
     }]
+  }, {
+    code: [
+      '<div onLayout={() => { /* dummy callback to fix android bug with component measuring */ }} />'
+    ].join('\n'),
+    output: [
+      '<div onLayout={() => { /* dummy callback to fix android bug with component measuring */ }} />'
+    ].join('\n'),
+    options: ['never', {allowMultiline: true}],
+    errors: []
```
ljharb: If this has no errors, it should be in the “valid” section, not the “invalid” section

ColCh: oh, sorry! didn't noticed that there is "valid" section

removed that test and added test into valid section

PR #2207, opened 2019-03-19
[Fix]`no-unknown-property`: fix case like `<Foo.bar>`

### LINKED ISSUE #2204: react/no-unknown-property incorrect `for` on none `label` tag detection
I have a `Table.Index` component with `for` prop.

the rule enforces me to use `htmlFor` instead and incorrectly auto replaces it when I run `eslint --fix`

golopot:
Fixes #2204

### DIFF
--- a/lib/rules/no-unknown-property.js
+++ b/lib/rules/no-unknown-property.js
@@ -168,6 +168,20 @@ function getTagName(node) {
 }
 
 /**
+ * Test wether the tag name for the JSXAttribute is
+ * something like <Foo.bar />
+ * @param {JSXAttribute} node - JSXAttribute being tested.
+ * @returns {Boolean} result
+ */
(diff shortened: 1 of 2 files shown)


lib/rules/no-unknown-property.js
```diff
@@ -167,6 +167,20 @@ function getTagName(node) {
...
 }
 
+/**
+ * Test wether the tag name for the JSXAttribute is
+ * something like <Foo.bar />
+ * @param {JSXAttribute} node - JSXAttribute being tested.
+ * @returns {Boolean} result
+ */
+function tagNameHasDot(node) {
+  return Boolean(
```
ljharb:
```suggestion
  return !!(
```

golopot: Updated

PR #2210, opened 2019-03-22
[Fix]`self-closing-comp`: stop reporting single-line spaces

### LINKED ISSUE #1717: Spaces get condensed by --fix when they are alone in a <span>
We often need to add spacing between two elements in JSX. For example:

```jsx
<GenericTextComponent type='test' />
<span>{' '}</span>
<GenericTextComponent type='other' />
```

Applying `--fix` turns the code into this:
```jsx
<GenericTextComponent type='test' />
<span />
<GenericTextComponent type='other' />
```
which is not the same thing.

I think I've traced down what's happening. The code above gives off this error:
```
error  Curly braces are unnecessary here  react/jsx-curly-brace-presence
```

Which, when corrected (by removing the curly braces and quotes), gives off another error:
```
error  Empty components are self-closing  react/self-closing-comp
```

which `--fix` happily turns into the `<span />` above.

Versions:
* `eslint@4.18.2`
* `eslint-plugin-react@7.7.0`

golopot:
Closes #1717

React does not trim JSXText if it is single-line spaces.
In case like `<span>    </span>`, the spaces are left as is.

### DIFF
--- a/tests/lib/rules/self-closing-comp.js
+++ b/tests/lib/rules/self-closing-comp.js
@@ -38,6 +38,10 @@ ruleTester.run('self-closing-comp', rule, {
         </Hello>
       `
     }, {
+      code: 'var HelloJohn = <Hello name="John"> </Hello>;'
+    }, {
+      code: 'var HelloJohn = <Hello name="John">        </Hello>;'
+    }, {
       code: 'var HelloJohn = <div>&nbsp;</div>;'
     }, {
       code: 'var HelloJohn = <div>{\' \'}</div>;'
@@ -55,6 +59,12 @@ ruleTester.run('self-closing-comp', rule, {
           <Hello name="John" />
         </Hello>
       `,
+      options: []
+    }, {
+      code: 'var HelloJohn = <div> </div>;',
+      options: []
+    }, {
+      code: 'var HelloJohn = <div>        </div>;',
       options: []
     }, {
       code: 'var HelloJohn = <div>&nbsp;</div>;',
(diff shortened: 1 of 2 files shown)


tests/lib/rules/self-closing-comp.js
```diff
@@ -37,6 +37,10 @@ ruleTester.run('self-closing-comp', rule, {
           <Hello name="John" />
         </Hello>
       `
+    }, {
+      code: 'var HelloJohn = <Hello name="John"> </Hello>;'
+    }, {
+      code: 'var HelloJohn = <Hello name="John">        </Hello>;'
```
ljharb: this does get collapsed to a single space, though.

tests/lib/rules/self-closing-comp.js
```diff
@@ -56,6 +60,12 @@ ruleTester.run('self-closing-comp', rule, {
         </Hello>
       `,
       options: []
+    }, {
+      code: 'var HelloJohn = <div> </div>;',
+      options: []
+    }, {
+      code: 'var HelloJohn = <div>        </div>;',
```
ljharb: as does this.

tests/lib/rules/self-closing-comp.js
golopot: I tried this on jsfiddle and didn't find it collapsed.

ljharb: hm, it should - the browser itself does that.

golopot: Yay I mean react rendered html string does not collapse spaces in this case. Browser does.

PR #2211, opened 2019-03-22
[New] add `jsx-no-padded-children`

### LINKED ISSUE #2045: Rule proposal: jsx-padded-children
Something similar to [`padded-blocks`](https://eslint.org/docs/rules/padded-blocks) where it disallows code similar to this:
```html
<div>

  <img />
</div>
```

Depending on the setting the following should **not** throw a warning:
```html
// jsx-padded-children: [2, "always"]
<div>

  <img />

</div>

// jsx-padded-children: [2, "never"]
<div>
  <img />
</div>
```

Not sure if anyone ever would pick up `always`, that's why I'm also down for `jsx-no-padded-children` or something like that

golopot:
Fixes #2045 

The name of the rule mimics that of eslint rule `padded-blocks`. Don't know weather it is good or not.

I am concerned of the code duplication in functions `checkPaddingTop` and `checkPaddingBottom`, but cannot find a good way to solve it.

### DIFF
--- a/docs/rules/jsx-no-padded-children.md
+++ b/docs/rules/jsx-no-padded-children.md
@@ -1,0 +1,40 @@
+# Disallow padding blank lines in a jsx element (react/jsx-no-padded-children)
+
+**Fixable:** This rule is automatically fixable using the `--fix` flag on the command line.
+
+## Rule Details
+
+This rule will disallow padding blank lines in the start and end of a jsx element.
+
+The following patterns are considered warnings:
+
+```jsx
+var Hello = <div>
+  test
+
+</div>;
+```
+
+```jsx
+var Hello = <div>
+  
+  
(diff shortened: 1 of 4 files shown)


golopot:
I am very unclear about how this rule should be designed.
1. Maybe it should be named `jsx-padded-children`?
2. Maybe it should be named  `jsx-children-newline`, and do what other `*-newline` rules do (`function-parens-newline`, `array-bracket-newline`, `object-curly-newline`). However since the concern of #2045 is about disallow empty line, this rule add an option to disallow empty lines, that is, only zero or one newline is allowed.

ljharb:
`jsx-children-newline` seems like a better name and model.

golopot:
Closed this because I couldn't find a satisfactory solution.

PR #2220, opened 2019-03-28
Add `react/no-array-index-key` rule example to documentation

dmmarmol:
Improve documentation of this particular rule by adding a use case example.

### DIFF
--- a/docs/rules/no-array-index-key.md
+++ b/docs/rules/no-array-index-key.md
@@ -6,6 +6,19 @@ The `key` is used by React to [identify which items have changed, are added, or 
 
 It's a bad idea to use the array index since it doesn't uniquely identify your elements. In cases where the array is sorted or an element is added to the beginning of the array, the index will be changed even though the element representing that index may be the same. This results in unnecessary renders.
 
+## Example
+
+Add to your eslint configuration file (usually `.estlintrc`)
+
+```
+{
+	"rules": {
+		"react/no-array-index-key": 0 // 0 = off, 1 = warn, 2 = error
+	}
+}
+
+```
+
 ## Rule Details
 
 The following patterns are considered warnings:


docs/rules/no-array-index-key.md
```diff
@@ -6,6 +6,19 @@ The `key` is used by React to [identify which items have changed, are added, or
...
 It's a bad idea to use the array index since it doesn't uniquely identify your elements. In cases where the array is sorted or an element is added to the beginning of the array, the index will be changed even though the element representing that index may be the same. This results in unnecessary renders.
 
+## Example
+
+Add to your eslint configuration file (usually `.estlintrc`)
+
+```
+{
+	"rules": {
+		"react/no-array-index-key": 0 // 0 = off, 1 = warn, 2 = error
```
ljharb: generally we don't document things like this, because it's basics of how eslint itself works. It's assumed that users of an eslint plugin know how eslint works.

dmmarmol: I see your point but, as an eslint user, sometimes I found that there're many rules that either exptects a `string` like `error` or `warn`, while others could expect `numbers`, `booleans` or even `objects` or `arrays`. This made me waste some time until I found some example in the web about the correct usage of a specific rule.

I personally think that documenting this doesn't distract the user and also helps on document how this specific rule works.

ljharb: If so, then let's also audit every other rule doc, and make sure they all document themselves the same way?

ljharb: (do note tho, *every rule without exception* supports 0/1/2 or "off"/"warn"/"error"; the variability is only in additional configuration arguments)

dmmarmol: I'd love to see every rule documented, but as you know there're a lot of them and it's better to start at some point if you truly consider it.
On the other side (and since I don't consider myself a an advanced eslint user), I wasn't fully aware of that rule (my bad for don't reading the `eslint` docs :hand:).
As a conclusion, I see that you have a valid point about being redudant adding this as a documentation, but maybe it could help some begginers or anxiety devs like me :).

ljharb: I definitely think reducing confusion is good; but i don't think doing it piecemeal is likely to be effective.

Issue #529, opened 2016-04-05
Rule proposal: disallow curlies around string literals


lencioni:
Bad:

``` jsx
function MyComponent() {
  return (
    <div className={'MyComponent'}>
      Hi mom
    </div>
  )
}
```

Good: 

``` jsx
function MyComponent() {
  return (
    <div className='MyComponent'>
      Hi mom
    </div>
  )
}
```

yannickcr:
So, the opposite of [jsx-no-literals](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-literals.md) ? If yes we could make a `jsx-literals` rule with `always`/`never` options.

lencioni:
Yes, I think so. For the `never` case, there are probably some exceptions when used outside of props (e.g. `{' '}` is often used).

ljharb:
I'd think this is `jsx-string-literals-in-props` or something - I don't think we'd want this rule to legislate text used outside of props.

lencioni:
Yeah, that's fair. I could see folks wanting different things in props vs outside props, so having separate rules seems like a decent way to achieve that.

yannickcr:
Mmm, apparently the current `jsx-no-literals` rule only apply outside props.

ljharb:
Seems closed by #1349.

PR #2228, opened 2019-04-08
[New] add `jsx-quoted-string-props` rule

rkusa:
Implements #529. Enforce quoted props for string-typed property values.

Bad:

```jsx
<div id={'foobar'} />
```

Good:

```jsx
<div id="foobar" />
```

Since this is the first ESLint rule I've ever implemented, feel free to be picky, feedback is highly welcome!


golopot:
It seemed that this feature is already implemented in `jsx-curly-brace-presence`.

https://github.com/yannickcr/eslint-plugin-react/blob/35de81b7b47f92b3a1b20e34269df15de462adb6/tests/lib/rules/jsx-curly-brace-presence.js#L411-L418

rkusa:
Even better - Thanks for pointing the existing one out!

PR #2230, opened 2019-04-09
[Fix] - Fix sorting props with numeric keys

### LINKED ISSUE #2212: `TypeError: currentPropName.toLowerCase is not a function` in sort-prop-types
The rule `sort-prop-types` (with `ignoreCase: true`) will produce this error when using `PropTypes.shape` with a number as a key.

Here's an example to reproduce this issue

```
import React from 'react';

export default function Component({ numberToNameMap }) {
  return <div>{numberToNameMap[0]}</div>;
}

Component.propTypes = {
  numberToNameMap: PropTypes.shape({
    0: PropTypes.string.isRequired,
  }).isRequired,
};
```

The error occurs at this line https://github.com/yannickcr/eslint-plugin-react/blob/v7.12.4/lib/rules/sort-prop-types.js#L192.

Using a string instead of the number as the key when defining the propTypes avoid this issue
```
import React from 'react';

export default function Component({ numberToNameMap }) {
  return <div>{numberToNameMap[0]}</div>;
}

Component.propTypes = {
  numberToNameMap: PropTypes.shape({
    '0': PropTypes.string.isRequired,
  }).isRequired,
};
```

I'm not sure if using a number instead of a string as a key is "idiomatic" React; I didn't see an example of this in the React documentation either way.

pawelnvk:
Error occurs when some of props has keys declared as numbers. Object keys shouldn't be declared as numbers. Nevertheless, error should not occur.

Fixes #2212.

### DIFF
--- a/lib/rules/sort-prop-types.js
+++ b/lib/rules/sort-prop-types.js
@@ -85,6 +85,10 @@ module.exports = {
       return node.arguments && node.arguments[0] && node.arguments[0].properties;
     }
 
+    function toLowerCase(item) {
+      return String(item).toLowerCase();
+    }
+
     function sorter(a, b) {
       let aKey = getKey(a);
       let bKey = getKey(b);
@@ -107,8 +111,8 @@ module.exports = {
       }
 
       if (ignoreCase) {
-        aKey = aKey.toLowerCase();
-        bKey = bKey.toLowerCase();
+        aKey = toLowerCase(aKey);
+        bKey = toLowerCase(bKey);
       }
 
       if (aKey < bKey) {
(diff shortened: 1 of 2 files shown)


lib/rules/sort-prop-types.js
```diff
@@ -85,6 +85,10 @@ module.exports = {
       return node.arguments && node.arguments[0] && node.arguments[0].properties;
     }
 
+    function toLowerCase(item) {
+      return item.toString().toLowerCase();
```
ljharb:
```suggestion
      return String(item).toLowerCase();
```
(always use `String(x)`, never `x.toString()`)

PR #2233, opened 2019-04-10
implement flow version detection

jedwards1211:
not sure why you guys implemented React version detection without doing the same for Flow


ljharb:
Looks good, but let’s add tests (just like react detection has)

jedwards1211:
yup, done

PR #2234, opened 2019-04-10
Validate boolean prop names nested deeper

pawelnvk:
Prop names nested inside of `shape` weren't validated. Issue shouldn't occur.


ljharb:
This looks good, but I’m concerned about suddenly causing more warnings for people. Can we put this behind an option?

pawelnvk:
Ok. Now it should work with option.

PR #2239, opened 2019-04-13
[New] add `no-unused-class-component-methods` 

### LINKED ISSUE #2166: react/no-unused-class-comp-methods
I'd like to have a rule to detect unused `React` class methods. The rule would filter out `React` lifecycle methods, take the other ones and check if they are used anywhere in the class itself.

Example:
```js
class Component extends React.Component {
    // error: unused method
    someMethod() {
        // something...
    }

    makeApiCall() {
        // do api call...
    }

    handleClick() {
        this.makeApiCall();
    }

    render() {
         return (<button onClick={this.handleClick}>Do it!</button>);
    }
}
```

The motivation behind this rule is that it's actually a bad practice to reference `React.Component` instance methods outside of the instance itself, so it would make sense to check if any methods inside the class are being used or are some kind of unused leftover.

pawelnvk:
Display errors in case class component contains unused methods.

Fixes #2166.

### DIFF
--- a/docs/rules/no-unused-class-component-methods.md
+++ b/docs/rules/no-unused-class-component-methods.md
@@ -1,0 +1,34 @@
+# Prevent declaring unused methods of component class (react/no-unused-class-component-methods)
+
+Warns you if you have defined a method or property but it is never being used anywhere.
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```jsx
+class Foo extends React.Component {
+  handleClick() {}
+  render() {
+    return null;
+  }
+}
+```
+
+The following patterns are **not** considered warnings:
+
+```jsx
+class Foo extends React.Component {
+  static getDerivedStateFromError(error) {
+    return { hasError: true };
+  }
+  action() {}
+  componentDidMount() {
+    this.action();
+  }
+  render() {
+    return null;
+  }
+}
+});
+```
--- a/lib/rules/no-unused-class-component-methods.js
+++ b/lib/rules/no-unused-class-component-methods.js
@@ -1,0 +1,244 @@
+/**
+ * @fileoverview Prevent declaring unused methods and properties of component class
+ * @author Paweł Nowak, Berton Zhu
+ */
+
+'use strict';
+
+const Components = require('../util/Components');
+const docsUrl = require('../util/docsUrl');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+const LIFECYCLE_METHODS = new Set([
+  'constructor',
+  'componentDidCatch',
+  'componentDidMount',
+  'componentDidUpdate',
+  'componentWillMount',
+  'componentWillReceiveProps',
+  'componentWillUnmount',
+  'componentWillUpdate',
+  'getSnapshotBeforeUpdate',
+  'render',
+  'shouldComponentUpdate',
+  'UNSAFE_componentWillMount',
+  'UNSAFE_componentWillReceiveProps',
+  'UNSAFE_componentWillUpdate'
+]);
+
+const ES6_LIFECYCLE = new Set([
+  'state'
+]);
+
+const ES5_LIFECYCLE = new Set([
+  'getInitialState',
+  'getDefaultProps',
+  'mixins'
+]);
+
+function isKeyLiteralLike(node, property) {
+  return property.type === 'Literal'
+     || (property.type === 'TemplateLiteral' && property.expressions.length === 0)
+     || (node.computed === false && property.type === 'Identifier');
+}
+
+// Descend through all wrapping TypeCastExpressions and return the expression
+// that was cast.
+function uncast(node) {
+  while (node.type === 'TypeCastExpression') {
+    node = node.expression;
+  }
+  return node;
+}
+
+// Return the name of an identifier or the string value of a literal. Useful
+// anywhere that a literal may be used as a key (e.g., member expressions,
+// method definitions, ObjectExpression property keys).
+function getName(node) {
+  node = uncast(node);
+  const type = node.type;
+
+  if (type === 'Identifier') {
+    return node.name;
+  }
+  if (type === 'Literal') {
+    return String(node.value);
+  }
+  if (type === 'TemplateLiteral' && node.expressions.length === 0) {
+    return node.quasis[0].value.raw;
+  }
+  return null;
+}
+
+function isThisExpression(node) {
+  return uncast(node).type === 'ThisExpression';
+}
+
+function getInitialClassInfo(node, isClass) {
+  return {
+    classNode: node,
+    isClass,
+    // Set of nodes where properties were defined.
+    properties: new Set(),
+
+    // Set of names of properties that we've seen used.
+    usedProperties: new Set(),
+
+    inStatic: false
+  };
+}
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Prevent declaring unused methods of component class',
+      category: 'Best Practices',
+      recommended: false,
+      url: docsUrl('no-unused-class-component-methods')
+    },
+    schema: [
+      {
+        type: 'object',
+        additionalProperties: false
+      }
+    ]
+  },
+
+  create: Components.detect((context, components, utils) => {
+    let classInfo = null;
+
+    // Takes an ObjectExpression node and adds all named Property nodes to the
+    // current set of properties.
+    function addProperty(node) {
+      classInfo.properties.add(node);
+    }
+
+    // Adds the name of the given node as a used property if the node is an
+    // Identifier or a Literal. Other node types are ignored.
+    function addUsedProperty(node) {
+      const name = getName(node);
+      if (name) {
+        classInfo.usedProperties.add(name);
+      }
+    }
+
+    function reportUnusedProperties() {
+      // Report all unused properties.
+      for (const node of classInfo.properties) { // eslint-disable-line no-restricted-syntax
+        const name = getName(node);
+        if (
+          !classInfo.usedProperties.has(name)
+           && !LIFECYCLE_METHODS.has(name)
+           && (classInfo.isClass ? !ES6_LIFECYCLE.has(name) : !ES5_LIFECYCLE.has(name))
+        ) {
+          const className = (classInfo.classNode.id && classInfo.classNode.id.name) || '';
+
+          context.report({
+            node,
+            message: `Unused method or property "${name}"${className ? ` of class "${className}"` : ''}`
+          });
+        }
+      }
+    }
+
+    function exitMethod() {
+      if (!classInfo || !classInfo.inStatic) {
+        return;
+      }
+
+      classInfo.inStatic = false;
+    }
+
+    return {
+      ClassDeclaration(node) {
+        if (utils.isES6Component(node)) {
+          classInfo = getInitialClassInfo(node, true);
+        }
+      },
+
+      ObjectExpression(node) {
+        if (utils.isES5Component(node)) {
+          classInfo = getInitialClassInfo(node, false);
+        }
+      },
+
+      'ClassDeclaration:exit'() {
+        if (!classInfo) {
+          return;
+        }
+        reportUnusedProperties();
+        classInfo = null;
+      },
+
+      'ObjectExpression:exit'(node) {
+        if (!classInfo || classInfo.classNode !== node) {
+          return;
+        }
+        reportUnusedProperties();
+        classInfo = null;
+      },
+
+      Property(node) {
+        if (!classInfo || classInfo.classNode !== node.parent) {
+          return;
+        }
+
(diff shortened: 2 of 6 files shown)


lib/rules/no-unused-class-component-methods.js
```diff
@@ -0,0 +1,152 @@
...
+
+const Components = require('../util/Components');
+const astUtil = require('../util/ast');
+const docsUrl = require('../util/docsUrl');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+const internalMethods = [
```
ljharb: what about unused static methods?

lib/rules/no-unused-class-component-methods.js
```diff
@@ -0,0 +1,152 @@
...
+    schema: [
+      {
+        type: 'object',
+        additionalProperties: false
+      }
+    ]
+  },
+
+  create: Components.detect((context, components, utils) => {
+    const isNotComponent = node => (
```
ljharb: shouldn't `Components.detect` give you a confidence value you can use here, without having to explicitly check component types?

lib/rules/no-unused-class-component-methods.js
```diff
@@ -0,0 +1,152 @@
...
+      !utils.isES5Component(node) &&
+      !utils.isES6Component(node) &&
+      !utils.isCreateElement(node)
+    );
+    const filterAllMethods = node => {
+      const isMethod = node.type === 'MethodDefinition';
+      const isArrowFunction = (
+        node.type === 'ClassProperty' &&
+        node.value.type === 'ArrowFunctionExpression'
```
ljharb: what about async functions, async arrow functions, generators, and async generators?

tests/lib/rules/no-unused-class-component-methods.js
```diff
@@ -0,0 +1,237 @@
...
+          }
+        }
+      `
+    },
+    {
+      code: `
+        class Foo extends React.Component {
+          action() {}
+          componentDidMount() {
+            const action = this.action;
```
golopot: What about use by object destructuring?

lib/rules/no-unused-class-component-methods.js
```diff
@@ -0,0 +1,152 @@
...
+            break;
+          case 'VariableDeclarator':
+            getThisExpressions(subnode.init);
+            break;
+          case 'MemberExpression':
+            if (subnode.object.type !== 'ThisExpression') {
+              return;
+            }
+
+            methods = methods.filter(method =>
```
golopot: `this.foo = bar` will make foo falsely considered used.

pawelnvk:
Thanks for feedback. Currently I'm taking some time off, but I'll definitely get back to it next week. I'm super excited about it!

ha404:
Would this also account for functional components?

golopot:
@ha404 It shouldn't. `no-unused-vars` should suffice in the case of functional components.

ha404:
@golopot but you can add static properties to functional components too.

```javascript
const SomeComponent = () => ...

SomeComponent.propTypes = ...
SomeComponent.anythingYouWant = ...
```

lib/rules/no-unused-class-component-methods.js
pawelnvk: Should we be concerned about static methods? They doesn't seem to be closely connected to component. I feel that static methods could be used outside of component and in that case it would be harder to determine if method is used..

lib/rules/no-unused-class-component-methods.js
pawelnvk: You are correct! I'll fix it.

pawelnvk: TBH I have some problem. `Components.detect` adds components to components list, so I can access it with `component.list()` yet still I would need to filter only components that I'm interested in. Wouldn't that be redundant?

lib/rules/no-unused-class-component-methods.js
ljharb: In a react component, yes - it's not meant to be used as a class.

(An instance method could also be used outside of the component)

Perhaps maybe an option to control checking of static methods?

lib/rules/no-unused-class-component-methods.js
ljharb: no, because that way the same logic used throughout the plugin is doing the initial filtering.

lib/rules/no-unused-class-component-methods.js
pawelnvk: That might be a good idea. I'll try to tackle it after basic version.

lib/rules/no-unused-class-component-methods.js
```diff
@@ -0,0 +1,162 @@
...
+      return isMethod || isArrowFunction || isFunctionExpression;
+    };
+    const checkMethods = node => {
+      const properties = astUtil.getComponentProperties(node);
+      let methods = properties
+        .filter(property => (
+          filterAllMethods(property) &&
+          internalMethods.indexOf(astUtil.getPropertyName(property)) === -1
+        ));
+      const getThisExpressions = subnode => {
```
ljharb: maybe this function would be useful moved out to a util, and unit-tested?

Kiwka:
Is there anything I can do to help to get this to mergeable state?

ljharb:
@Kiwka if you post a link here to a branch that addresses the review comments and test failures here (please do NOT create a new PR) then I can pull in your commits, and we can definitely get this in.

valtido:
Looks good, can we look at those tests, please? desperado for this... :)  (y)

meowtec:
How about use `public` keyword (only for eslint-typescript parser) as an alternative method to disable this rule?

**not** considered warnings:
```
class ClassPublicMemberTest extends React.Component {
  public foo() {}
  render() {
    return <SomeComponent />;
  },
}
```

ljharb:
@meowtec everything is public unless it's a private field (ie, `#foo`, the language proposal), and a React component shouldn't **have** anything that's *intended* for public use.

meowtec:
@ljharb
TS `public` or jsdoc `@public` is more explicit.
> a React component shouldn't have anything that's intended for public use.
I don't think so. Why not we forbid the use of `ref`?

ljharb:
@meowtec this isn't a "think" thing; react components are consumed via React.createElement, not via construction and instances. `ref` is a) not accessing your component's instance methods, and b) something that react itself plans to deprecate in react-dom in an upcoming semver-major, so I'm not sure how that's relevant.

meowtec:
@ljharb 
Can you give an example that `ref` is used on react component (not host element) **but without** accessing any instance method or property?

ljharb:
@meowtec I'm not sure what you mean - `ref` is primarily used for accessing raw DOM elements, and although you *can* use it to access class component instances, it's a terrible idea to do that.

meowtec:
@ljharb I agree accessing class component instances is terrible in the majority of cases.
I mean we should not forbid it here - we should have an independent rule like `no-non-host-ref` to do that.

tests/lib/rules/no-unused-class-component-methods.js
```diff
@@ -0,0 +1,388 @@
...
+              </div>
+            );
+          }
+        }
+      `
+    },
+    {
+      code: `
+        class Foo extends React.Component {
+          property = {}
```
meowtec: Why is it valid?

meowtec: Could it also check unused property?

ljharb:
This isn't forbidding it. It's recognizing that since it's never valid to access React component instances outside of the `class`, anything not used inside of the class is thus unused.

If you insist on following a bad practice, you can use an eslint override comment to disable the rule on any methods you want to be considered "used".

tests/lib/rules/no-unused-class-component-methods.js
ljharb: That's a good idea; the rule should check both properties and methods.

meowtec:
Still in progress?

tests/lib/rules/no-unused-class-component-methods.js
pawelnvk: IMO it's out of basic scope. Rule name is `no-unused-class-component-methods`. I feel we are trying to solve all issues with that single rule and this can overcomplicate it quickly.

ljharb:
ping @pawelnvk

lib/rules/no-unused-class-component-methods.js
```diff
@@ -0,0 +1,152 @@
...
+        if (!methods.length) {
+          return;
+        }
+
+        switch (subnode.type) {
+          case 'ClassProperty':
+          case 'JSXAttribute':
+          case 'MethodDefinition':
+            getThisExpressions(subnode.value);
```
valtido: Consider using the Object switch methodology rather than an actual switch, in my view, it's a lot faster than the switch. 

See this link for a reference. 
https://stackoverflow.com/questions/37730199/switch-vs-object-lookup-performance-since-jsperf-is-down

ljharb: Faster or not, it's always cleaner :-)

meowtec:
It seams that this PR has been abandoned.

I have (re)written this rule with another implementation and use it in our project which has over 100k lines of code.

see here https://github.com/meowtec/eslint-plugin-react-ext/blob/master/lib/rules/no-unused-class-property.js

ljharb:
@meowtec if you'd be able to make those changes in a *branch* on a fork - not a PR - then I can look into pulling them into this PR.

meowtec:
@ljharb My branch is at https://github.com/meowtec/eslint-plugin-react/tree/unused-component-methods-improve

What are difference:
 - The implementation inspired by `no-unused-state`
 - Check **both** properties and methods.
 - Remove the support for the legacy ES5 style `React.createClass`. If necessary I will add it back.
 - Static methods are just ignored.
 - More test cases and bugfix.

@pawelnvk Please take a look because I am not sure if any important things was missing.

ref: https://github.com/yannickcr/eslint-plugin-react/pull/2239#issuecomment-544415516

ljharb:
Yes, please add createClass support back; we'll be supporting it for the foreseeable future.

docs/rules/no-unused-class-component-methods.md
```diff
@@ -0,0 +1,31 @@
...
+```jsx
+class Foo extends React.Component {
+  handleClick() {}
+  render() {
+    return null;
+  }
+}
+```
+
+The following patterns are **not** considered warnings:
```
ljharb: let's add some unused static methods here, so it's clear.

tests/lib/rules/no-unused-class-component-methods.js
ljharb: ping on this one

docs/rules/no-unused-class-component-methods.md
ljharb: cc @meowtec

tests/lib/rules/no-unused-class-component-methods.js
ljharb: cc @meowtec

docs/rules/no-unused-class-component-methods.md
meowtec: > cc @meowtec

should all static methods be list here?

ljharb: Is there more than one? Either way we'd only need at least one as an example.

jakeleventhal:
Bumping

leusrox:
I tried to use it in our project which has over 500 components, it works good 👍

ljharb:
@pawelnvk are you still interested in landing this PR? if so, it could use a rebase and there's some review comments to address :-)

jakeleventhal:
@pawelnvk are you going to do this?
If not, I might do it @ljharb

ljharb:
@jakeleventhal that'd be great! however, please do NOT open a new PR; please comment here with a link to your branch, and I'll pull in the changes.

jakeleventhal:
@ljharb https://github.com/jakeleventhal/eslint-plugin-react/tree/jakeleventhal/no-unused-class-component-methods

ljharb:
@jakeleventhal thanks! Pulling it in now. I do note that it's still missing support for the static `getDerivedStateFromProps` method.

jakeleventhal:
@ljharb im a bit unfamiliar with this repo. do you think you could just take this over since you're pulling in the change anyway? i would appreciate it

jakeleventhal:
@ljharb also, does this cover hooks?

jakeleventhal:
@ljharb bump

ljharb:
Sure, I can try to add it.

No, this doesn't cover hooks; the rule only has to do with class components. SFCs don't have methods.

jakeleventhal:
@ljharb okay, thank you. I will leave it to you to finish then i suppose

ljharb (approved):

PR #2240, opened 2019-04-14
New: jsx-curly-newline rule

### LINKED ISSUE #1493: jsx-curly-spacing should support "consistent" option
Similar to how other eslint rules provide a "consistent" option for square brackets, curly braces, and parens, there should be a similar option for the curly braces in JSX. Basically, the absence or presence of a newline after the opening curly brace should be matched at the closing curly brace.

golopot:
fixes #1493

This rule is based on eslint core rule [function-paren-newline](https://eslint.org/docs/rules/function-paren-newline). Its options is similar to that of `function-paren-newline`, `array-bracket-newline`,  and `object-curly-newline`. However there is one new option `multiline-lax` (default), which is the same as `multiline` but tolerate cases like
```jsx
<div>
{
  foo
}
</div>
```
, that is, single line expressions are allowed to have both newlines.

Separately, since this commit is modified copy of the eslint core rule [function-paren-newline.js](https://github.com/eslint/eslint/blob/9ffdde5a95f45fac0d87a9f47ed42591f6ae42a6/lib/rules/function-paren-newline.js), I need help on how to deal with the license things.

### DIFF
--- a/docs/rules/jsx-curly-newline.md
+++ b/docs/rules/jsx-curly-newline.md
@@ -1,0 +1,149 @@
+# Enforce linebreaks in curly braces in JSX attributes and expressions. (react/jsx-curly-newline)
+
+Many style guides require or disallow newlines inside of jsx curly expressions.
+
+**Fixable:** This rule is automatically fixable using the `--fix` flag on the command line.
+
+## Rule Details
+
+This rule enforces consistent linebreaks inside of curlies of jsx curly expressions.
+
+## Rule Options
+
+This rule accepts either an object option:
+
+```ts
+{
+  multiline: "consistent" | "forbid" | "require", // default to 'consistent'
+  singleline: "consistent" | "forbid" | "require", // default to 'consistent'
+}
+```
+Option `multiline` takes effect when the jsx expression inside the curlies occupies multiple lines.
+
+Option `singleline` takes effect when the jsx expression inside the curlies occupies a single line.
+
+* `consistent` enforces either both curly braces have a line break directly inside them, or no line breaks are present.
+* `forbid` disallows linebreaks directly inside curly braces.
+* `require` enforces the presence of linebreaks directly inside curlies.
+
+or a string option:
+
+* `consistent` (default) is an alias for `{ multiline: "consistent", singleline: "consistent" }`.
+* `never` is an alias for `{ multiline: "forbid", singleline: "forbid" }`
+
+or an
+
+### consistent (default)
+
+When `consistent` or `{ multiline: "consistent", singleline: "consistent" }` is set, the following patterns are considered warnings:
+
+```jsx
+<div>
+  { foo
+  }
+</div>
+
+<div>
+  {
+    foo }
+</div>
+
+<div>
+  { foo &&
+    foo.bar
+  }
+</div>
+```
+
+The following patterns are **not** warnings:
+
+```jsx
+<div>
+  { foo }
+</div>
+
+<div>
+  {
+    foo
+  }
+</div>
+```
+
+### never
+
+When `never` or `{ multiline: "forbid", singleline: "forbid" }` is set, the following patterns are considered warnings:
+
+```jsx
+<div>
+  {
+    foo &&
+    foo.bar
+  }
+</div>
+
+<div>
+  {
+    foo
+  }
+</div>
+
+<div>
+  { foo
+  }
+</div>
+```
+
+The following patterns are **not** warnings:
+
+```jsx
+<div>
+  { foo &&
+    foo.bar }
+</div>
+
+<div>
+  { foo }
+</div>
+```
+
+## require
+
+When `{ multiline: "require", singleline: "require" }` is set, the following patterns are considered warnings:
+
+```jsx
+<div>
+  { foo &&
+    foo.bar }
+</div>
+
+<div>
+  { foo }
+</div>
+
+<div>
+  { foo
+  }
+</div>
+```
+
+The following patterns are **not** warnings:
+
+```jsx
+<div>
+  {
+    foo &&
+    foo.bar
+  }
+</div>
+
+<div>
+  {
+    foo
+  }
+</div>
+```
+
+
+## When Not To Use It
+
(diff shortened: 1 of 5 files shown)


docs/rules/jsx-curly-newline.md
```diff
@@ -0,0 +1,125 @@
...
+
+This rule enforces consistent linebreaks inside of curlies of jsx curly expressions.
+
+## Rule Options
+
+This rule has a string option:
+
+* `consistent` enforces either both of linebreaks around curlies are present, or none are present.
+* `multiline` enforces that when the contained expression spans multiple line, require both linebreaks. When the contained expression is single line, disallow both line breaks.
+* `multiline-lax` (default) enforces that when the contained expression spans multiple line, require both linebreaks. When the contained expression is single line, disallow inconsistent line break.
```
ljharb: this seems like `consistent-multiline` would be a better name

golopot: I think that consistent multiline sound like a consistent version of multiline, but that is not true. Because multipline is already consistent.

golopot: I think multilinelax is fit because it is like multiline, but more permissive when the expression is single line.

ljharb: Looking at eslint core, there's a number of rules that have an object option with a `multiline` boolean, and a `consistent` boolean - if we only allowed multiline, multiline + consistent, and consistent, that seems like it'd cover it more clearly?

golopot: `multiline + consistent = multiline`. `multiline` already implied consistent.

golopot: in `multiline` either both newline are present, or both newline are absent.  So a code that is valid in `multiline` is also valid in `consistent`. Consistent is the "weakest" option.

ljharb: I don't find the current schema clear.

It sounds like you want the following characteristics controllable:
1. force both linebreaks on a multiline expression, or not
1. disallow inconsistent linebreaks on a one-line expression, or disallow any linebreaks on a one-line expression.

So, what about:
```js
"multiline": "consistent" // or "ignore" or "forbid" or "require"
"single-line": "consistent" // or "ignore" or "forbid" or "require"
```
That way any setting can be configured without a subjective word like "lax", and it's extensible since it's an object option.

golopot: This might be good. I was trying to think in the box of precedent eslint core rules, and tries to follow the convention established there.

golopot: I don't like the option schema
> ```js
> "multiline": "consistent" // or "ignore" or "forbid" or "require"
> "single-line": "consistent" // or "ignore" or "forbid" or "require"
> ```
because there are 4*4 combinations, therefore many test cases to write. But in practice only a few of them make sense, so instead might as well just pick the few good options and give them a string name like `multiline`. It also saves users time by reducing their choices.

ljharb: Which don't make sense? We can use schema validation to only permit the desired combinations.

golopot: Now I feel I was wrong thinking most of them don't make sense. However I think it is safe to not support the `ignore` option (just use `consistent`).

golopot: How about this option schema?
```ts
"never" |
"consistent" |
{
  "single-line": "require" | "forbid" | "consistent" ,
  "multiline": "require" | "forbid" | "consistent" ,
}
```

ljharb: That seems good; let's update the PR and see what the test cases and docs for each combination look like :-D

golopot: Now I feel I am ok to live with the simple option schema `"never" | "consistent"`, dropping the multiline option. Are you interested in this direction?

docs/rules/jsx-curly-newline.md
```diff
@@ -0,0 +1,149 @@
...
+{
+  multiline: "consistent" | "forbid" | "require", // default to 'consistent'
+  singleline: "consistent" | "forbid" | "require", // default to 'consistent'
+}
+```
+Option `multiline` takes effect when the jsx expression inside the curlies occupies multiple lines.
+
+Option `singleline` takes effect when the jsx expression inside the curlies occupies a single line.
+
+* `consistent` enforces either both of linebreaks around curlies are present, or none are present.
```
ljharb: i think "around curlies" might not be the proper term, since that implies "surrounding the curlies" (outside) as opposed to "inside". Maybe:
```suggestion
* `consistent` enforces either both curly braces have a line break directly inside them, or no line breaks are present.
```

docs/rules/jsx-curly-newline.md
```diff
@@ -0,0 +1,149 @@
...
+* `forbid` disallows linebreaks around curlies.
```
ljharb:
```suggestion
* `forbid` disallows linebreaks directly inside curly braces.
```

docs/rules/jsx-curly-newline.md
```diff
@@ -0,0 +1,149 @@
...
+* `require` enforces the presence of linebreaks aounrd curlies.
```
ljharb:
```suggestion
* `require` enforces the presence of linebreaks directly inside curlies.
```

docs/rules/jsx-curly-newline.md
```diff
@@ -0,0 +1,149 @@
...
+  {
+    foo
+  }
+</div>
+```
+
+
+## When Not To Use It
+
+You can turn this rule off if you are not concerned with the consistency around the linebreaks inside of JSX attributes or expressions.
```
ljharb:
```suggestion
You can turn this rule off if you are not concerned with the consistency of padding linebreaks inside of JSX attributes or expressions.
```

golopot:
I forgot to add this rule to the README.

ljharb:
@golopot a followup PR would be great

PR #2242, opened 2019-04-17
Support '.tsx' filename extension by default

kripod:
As TypeScript is spreading and official ESLint support is given through the [typescript-eslint](https://github.com/typescript-eslint/typescript-eslint) project, I thought it would be convenient if the `.tsx` file extension was supported out of the box.


ljharb:
Hmm. I can see how this would help a TS user, but I'm not sure it's a sensible default - a non-TS user would actively want `tsx` files to be prohibited.

kripod:
@ljharb I don't think that a non-TS user is willing to create or transpile `tsx` files at all. We cannot prevent them from creating a file of a given kind anyway.

ljharb:
I agree with you - that's why i'm saying that the linter rule should *error* on a tsx file for a non-TSX user.

PR #2243, opened 2019-04-18
Allow jsx-pascal-case to ignore using regexps

soundasleep:
We want to ignore a consistent theme of component naming in our project, where the offending component names match a regexp, without having to individually label each of the components, or
to individually eslint-disable each of the instances.

This should be backwards compatible with existing rule sets, and matches how the `camelcase` rule works in eslint (which also supports regexps).

### DIFF
--- a/lib/rules/jsx-pascal-case.js
+++ b/lib/rules/jsx-pascal-case.js
@@ -62,7 +62,12 @@ module.exports = {
         const isPascalCase = PASCAL_CASE_REGEX.test(name);
         const isCompatTag = jsxUtil.isDOMComponent(node);
         const isAllowedAllCaps = allowAllCaps && ALL_CAPS_TAG_REGEX.test(name);
-        const isIgnored = ignore.indexOf(name) !== -1;
+        const isIgnored = ignore.find(stringOrRegexp => {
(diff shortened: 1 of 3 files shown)


lib/rules/jsx-pascal-case.js
```diff
@@ -62,7 +62,12 @@ module.exports = {
         const isPascalCase = PASCAL_CASE_REGEX.test(name);
         const isCompatTag = jsxUtil.isDOMComponent(node);
         const isAllowedAllCaps = allowAllCaps && ALL_CAPS_TAG_REGEX.test(name);
-        const isIgnored = ignore.indexOf(name) !== -1;
+        const isIgnored = ignore.find(stringOrRegexp => {
+          if (stringOrRegexp instanceof RegExp) {
```
ljharb: i don't think rules should be required to have a JS config to make them useful (camelcase rule notwithstanding); so it'd only ever get strings, not regex instances.

Also, `instanceof` is never reliable and should never be used; use https://npmjs.com/is-regex if you need to check for this.

ljharb (changes requested):
Instead of a regex, would a glob work?

ljharb:
I've converted this to a draft pending @soundasleep's response.

PR #2245, opened 2019-04-19
update creating class component with es6 class in rules/prop-types

Elecweb:
I think nowadays people don't use `createReactClass` as much and in official react document's example use es6 class for create class component. so I think it's more conformable for people that are newcomer in react.

### DIFF
--- a/docs/rules/prop-types.md
+++ b/docs/rules/prop-types.md
@@ -9,20 +9,20 @@ It can warn other developers if they make a mistake while reusing the component 
 The following patterns are considered warnings:
 
 ```jsx
-var Hello = createReactClass({
-  render: function() {
+class Hello extends React.Component {
+  render() {
     return <div>Hello {this.props.name}</div>;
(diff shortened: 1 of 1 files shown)


docs/rules/prop-types.md
```diff
@@ -9,20 +9,20 @@ It can warn other developers if they make a mistake while reusing the component
...
-});
+}
 
-var Hello = createReactClass({
-  propTypes: {
-    firstname: PropTypes.string.isRequired
-  },
-  render: function() {
+class Hello extends React.Component {
+  static propTypes = {
```
ljharb: this syntax is still stage 3; i don't think docs should be using it yet.

ljharb: Now it's stage 4, so i guess it's fine.

ljharb:
The current docs contain examples of all kinds of components; so i think this PR should be closed.

PR #2246, opened 2019-04-20
[fix]`prop-types`: fix case with destructuring and default param

### LINKED ISSUE #2241: prop-types not enforced with destructured props and default empty object param
```
// complains about 'baz' missing from prop-types validation
const Component = ({ foo, bar, baz }) => (
  // ...JSX
);

// doesn't complain about 'baz' missing from prop-types validation
const Component = ({ foo, bar, baz } = {}) => (
  // ...JSX
);

Component.propTypes = {
  foo: PropTypes.string,
  bar: PropTypes.string,
};
```

Since it's generally best-practice to provide defaults when destructuring, it would be nice if passing an empty object as a default param didn't break prop-types validation.

golopot:
fixes #2241

### DIFF
--- a/lib/util/usedPropTypes.js
+++ b/lib/util/usedPropTypes.js
@@ -316,16 +316,17 @@ module.exports = function usedPropTypesInstructions(context, components, utils) 
         break;
       case 'ArrowFunctionExpression':
       case 'FunctionDeclaration':
-      case 'FunctionExpression':
+      case 'FunctionExpression': {
         if (node.params.length === 0) {
           break;
         }
         type = 'destructuring';
-        properties = node.params[0].properties;
-        if (inSetStateUpdater()) {
-          properties = node.params[1].properties;
-        }
+        const propParam = inSetStateUpdater() ? node.params[1] : node.params[0];
+        properties = propParam.type === 'AssignmentPattern'
+          ? propParam.left.properties
+          : propParam.properties;
         break;
+      }
       case 'VariableDeclarator':
         for (let i = 0, j = node.id.properties.length; i < j; i++) {
           // let {props: {firstname}} = this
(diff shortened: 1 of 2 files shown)


lib/util/usedPropTypes.js
```diff
@@ -421,7 +422,15 @@ module.exports = function usedPropTypesInstructions(context, components, utils)
    *   FunctionDeclaration, or FunctionExpression
    */
   function markDestructuredFunctionArgumentsAsUsed(node) {
-    const destructuring = node.params && node.params[0] && node.params[0].type === 'ObjectPattern';
+    const param = inSetStateUpdater()
```
ljharb:
```suggestion
    const param = node.params && inSetStateUpdater() ? node.params[1] : node.params[0];
```

lib/util/usedPropTypes.js
```diff
@@ -421,7 +422,15 @@ module.exports = function usedPropTypesInstructions(context, components, utils)
    *   FunctionDeclaration, or FunctionExpression
    */
   function markDestructuredFunctionArgumentsAsUsed(node) {
-    const destructuring = node.params && node.params[0] && node.params[0].type === 'ObjectPattern';
+    const param = node.params && inSetStateUpdater() ? node.params[1] : node.params[0];
+      ? node.params && node.params[1]
+      : node.params && node.params[0];
```
ljharb:
```suggestion
```

lib/util/usedPropTypes.js
```diff
@@ -421,7 +422,15 @@ module.exports = function usedPropTypesInstructions(context, components, utils)
    *   FunctionDeclaration, or FunctionExpression
    */
   function markDestructuredFunctionArgumentsAsUsed(node) {
-    const destructuring = node.params && node.params[0] && node.params[0].type === 'ObjectPattern';
+    const param = node.params && inSetStateUpdater() ? node.params[1] : node.params[0];
+      ? node.params && node.params[1]
```
ljharb:
```suggestion
```

golopot:
Thank you.

PR #2250, opened 2019-04-24
Make all jsx-prop-sort lints fixable

guliashvili:
Not all the jsx-prop-sort lints were fixable.
There were cases when existing fix would cause other lints to trigger.
For example, in case when normal props were not sorted but callback functions were in the end(As they are supposed to be), sorting lint fix would mix callbacks props in between the other props.

```
Case:
ignoreCase = false;
callbacksLast = true;
shorthandFirst = false;
shorthandLast = false;
noSortAlphabetically = false;
reservedFirst = false;
--
<div a z onClick=y />

Existing lint fix would transform it to
<div a onClick=y z />

However, this also triggers another lint about callbacksLast.
```

Updated the tests accordingly.

### DIFF
--- a/lib/rules/jsx-sort-props.js
+++ b/lib/rules/jsx-sort-props.js
@@ -34,11 +34,9 @@ function contextCompare(a, b, options) {
   if (options.reservedFirst) {
     const aIsReserved = isReservedPropName(aProp, options.reservedList);
     const bIsReserved = isReservedPropName(bProp, options.reservedList);
-    if ((aIsReserved && bIsReserved) || (!aIsReserved && !bIsReserved)) {
-      // pass
-    } else if (aIsReserved && !bIsReserved) {
+    if (aIsReserved && !bIsReserved) {
       return -1;
-    } else {
+    } else if (!aIsReserved && bIsReserved) {
       return 1;
     }
   }
@@ -46,22 +44,18 @@ function contextCompare(a, b, options) {
   if (options.callbacksLast) {
     const aIsCallback = isCallbackPropName(aProp);
     const bIsCallback = isCallbackPropName(bProp);
-    if ((aIsCallback && bIsCallback) || (!aIsCallback && !bIsCallback)) {
-      // pass
-    } else if (aIsCallback && !bIsCallback) {
+    if (aIsCallback && !bIsCallback) {
       return 1;
-    } else {
+    } else if (!aIsCallback && bIsCallback) {
       return -1;
     }
   }
 
   if (options.shorthandFirst || options.shorthandLast) {
     const shorthandSign = options.shorthandFirst ? -1 : 1;
-    if (!a.value && !b.value) {
-      // pass
-    } else if (!a.value) {
+    if (!a.value && b.value) {
       return shorthandSign;
-    } else {
+    } else if (a.value && !b.value) {
       return -shorthandSign;
     }
   }
--- a/lib/rules/jsx-sort-props.js
+++ b/lib/rules/jsx-sort-props.js
@@ -27,22 +27,54 @@ function isReservedPropName(name, list) {
   return list.indexOf(name) >= 0;
 }
 
-function propNameCompare(a, b, options) {
-  if (options.ignoreCase) {
-    a = a.toLowerCase();
-    b = b.toLowerCase();
-  }
+function contextCompare(a, b, options) {
+  let aProp = propName(a);
+  let bProp = propName(b);
+
   if (options.reservedFirst) {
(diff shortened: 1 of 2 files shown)


lib/rules/jsx-sort-props.js
```diff
@@ -27,22 +27,54 @@ function isReservedPropName(name, list) {
...
-    const bIsReserved = isReservedPropName(b, options.reservedList);
+    const aIsReserved = isReservedPropName(aProp, options.reservedList);
+    const bIsReserved = isReservedPropName(bProp, options.reservedList);
     if ((aIsReserved && bIsReserved) || (!aIsReserved && !bIsReserved)) {
-      return a.localeCompare(b);
+      // pass
     } else if (aIsReserved && !bIsReserved) {
       return -1;
+    } else {
+      return 1;
```
ljharb: this doesn't need the final else; it can be de-nested

lib/rules/jsx-sort-props.js
```diff
@@ -27,22 +27,54 @@ function isReservedPropName(name, list) {
...
+
+  if (options.callbacksLast) {
+    const aIsCallback = isCallbackPropName(aProp);
+    const bIsCallback = isCallbackPropName(bProp);
+    if ((aIsCallback && bIsCallback) || (!aIsCallback && !bIsCallback)) {
+      // pass
+    } else if (aIsCallback && !bIsCallback) {
+      return 1;
+    } else {
+      return -1;
```
ljharb: also here

lib/rules/jsx-sort-props.js
```diff
@@ -27,22 +27,54 @@ function isReservedPropName(name, list) {
...
+    } else {
+      return 1;
+    }
+  }
+
+  if (options.callbacksLast) {
+    const aIsCallback = isCallbackPropName(aProp);
+    const bIsCallback = isCallbackPropName(bProp);
+    if ((aIsCallback && bIsCallback) || (!aIsCallback && !bIsCallback)) {
+      // pass
```
ljharb:
```suggestion
      return 0;
```

lib/rules/jsx-sort-props.js
```diff
@@ -27,22 +27,54 @@ function isReservedPropName(name, list) {
...
+  let bProp = propName(b);
+
   if (options.reservedFirst) {
-    const aIsReserved = isReservedPropName(a, options.reservedList);
-    const bIsReserved = isReservedPropName(b, options.reservedList);
+    const aIsReserved = isReservedPropName(aProp, options.reservedList);
+    const bIsReserved = isReservedPropName(bProp, options.reservedList);
     if ((aIsReserved && bIsReserved) || (!aIsReserved && !bIsReserved)) {
-      return a.localeCompare(b);
+      // pass
```
ljharb:
```suggestion
      return 0;
```

lib/rules/jsx-sort-props.js
```diff
@@ -27,22 +27,54 @@ function isReservedPropName(name, list) {
...
+      return 1;
+    } else {
+      return -1;
+    }
+  }
+
+  if (options.shorthandFirst || options.shorthandLast) {
+    const shorthandSign = options.shorthandFirst ? -1 : 1;
+    if (!a.value && !b.value) {
+      // pass
```
ljharb:
```suggestion
      return 0;
```

lib/rules/jsx-sort-props.js
```diff
@@ -27,22 +27,54 @@ function isReservedPropName(name, list) {
...
+    } else if (!a.value) {
+      return shorthandSign;
+    } else {
```
ljharb: also here

guliashvili:
@ljharb Thank you for so fast review Jordan.
I simplified code a bit by removing unnecessary if branches. Please let me know if there is anything else I can change.

guliashvili:
@ljharb @yannickcr 
Looks like before, new tags were cut frequently but not anymore. Is there any bug or something stopping you from cutting the new tag(I can work on that also). 

Would appreciate if you cut the new tag.

ljharb:
Yes, tests are failing on master right now; see #2253

guliashvili:
Thanks @ljharb for letting me know. 
Looks like new tag is cut thank you for that. 
https://github.com/yannickcr/eslint-plugin-react/compare/v7.13.0...master
Not sure how but 7.13 only contains 5 commits. Was wondering how it's possible. Can we cut new tag for the current master?

ljharb:
v7.13.0 is released.

guliashvili:
Thank you @ljharb . Appreciated  :)

PR #2259, opened 2019-05-03
Remove no-render-return-value from recommended

bz2:
This rules is about a potential future change to a ReactDOM interface
not any actual problem, and is the slowest `plugin:react/recommended`
by a factor of about 3.

Timings from one of our projects:

```
Rule                           | Time (ms) | Relative
:------------------------------|----------:|--------:
import/namespace               |  5505.442 |    19.3%
indent                         |  4169.809 |    14.6%
react/no-render-return-value   |  3402.119 |    11.9%
import/named                   |  1252.616 |     4.4%
react/no-deprecated            |  1165.668 |     4.1%
react/display-name             |  1126.994 |     4.0%
react/no-direct-mutation-state |  1109.049 |     3.9%
react/sort-comp                |   691.149 |     2.4%
react/no-string-refs           |   680.091 |     2.4%
import/default                 |   622.367 |     2.2%
```

Removing the rule does save 4s from a 60s run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,6 @@ The rules enabled in this configuration are:
 * [react/no-direct-mutation-state](docs/rules/no-direct-mutation-state.md)
 * [react/no-find-dom-node](docs/rules/no-find-dom-node.md)
 * [react/no-is-mounted](docs/rules/no-is-mounted.md)
-* [react/no-render-return-value](docs/rules/no-render-return-value.md)
 * [react/no-string-refs](docs/rules/no-string-refs.md)
 * [react/no-unescaped-entities](docs/rules/no-unescaped-entities.md)
 * [react/no-unknown-property](docs/rules/no-unknown-property.md)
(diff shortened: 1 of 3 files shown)


bz2:
Appveyor failed but only one platform and looks spurious:
```
Environment: nodejs_version=8; Platform: x64
Build started
Fetching repository commit (873816e)...An error occurred while sending the request.
```

ljharb:
I don't think "slowest" should be a disqualifier from whether it's recommended; it's a lint run in CI.

ljharb:
Separately, this is an actual problem for future migrations, so it seems useful to keep recommended.

As for it being slow, it doesn't seem like it should be - making it faster would be great tho!

yannickcr:
Closing since performance problem was likely fixed in #2264

bz2:
I still think this is not a good rule to enable by default, but thanks @golopot for making it less painful.

There's no harm in code just failing when an actual migration happens. I have two instances of `ReactDOM.render()` in my entire codebase, this isn't a thing it makes sense to check every line in the project for.

PR #2261, opened 2019-05-05
[New] add `jsx-no-useless-fragment` rule

### LINKED ISSUE #2042: Rule proposal: detect unnecessary use of fragments
The idea behind fragments is to group a list of children without adding extra nodes to the DOM. 
Fragments are completely redundant when wrapping a single child, so the rule should detect examples like:

```.js
<>
  <div />
</>
```

```.js
<Fragment>
  <div />
</Fragment>
```

and flag them as a warning/error.

Example of valid usage of fragments:
```.js
<>
  <div />
  <div />
</>
```
(any case where `children.length > 1`)

golopot:
fixes #2042

This rule disallows fragments with less than 2 children.

### DIFF
--- a/docs/rules/jsx-no-useless-fragment.md
+++ b/docs/rules/jsx-no-useless-fragment.md
@@ -1,0 +1,54 @@
+# Disallow unnecessary fragments (react/jsx-no-useless-fragment)
+
+A fragment is redundant if it contains only one child, or if it is the child of a html element, and is not a [keyed fragment](https://reactjs.org/docs/fragments.html#keyed-fragments).
+
+**Fixable:** This rule is sometimes automatically fixable using the `--fix` flag on the command line.
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```jsx
+<>{foo}</>
+
+<><Foo /></>
+
+<p><>foo</></p>
+
+<></>
+
+<Fragment>foo</Fragment>
+
+<React.Fragment>foo</React.Fragment>
+
+<section>
+  <>
+    <div />
+    <div />
+  </>
+</section>
+```
+
+The following patterns are **not** considered warnings:
+
+```jsx
+<>
+  <Foo />
+  <Bar />
+</>
+
+<>foo {bar}</>
+
+<> {foo}</>
+
+const cat = <>meow</>
+
+<SomeComponent>
+  <>
+    <div />
+    <div />
+  </>
+</SomeComponent>
+
+<Fragment key={item.id}>{item.value}</Fragment>
+```
--- a/lib/rules/jsx-no-useless-fragment.js
+++ b/lib/rules/jsx-no-useless-fragment.js
@@ -1,0 +1,210 @@
+/**
+ * @fileoverview Disallow useless fragments
+ */
+
+'use strict';
+
+const pragmaUtil = require('../util/pragma');
+const jsxUtil = require('../util/jsx');
+const docsUrl = require('../util/docsUrl');
+
+function isJSXText(node) {
+  return !!node && (node.type === 'JSXText' || node.type === 'Literal');
+}
+
+/**
+ * @param {string} text
+ * @returns {boolean}
+ */
+function isOnlyWhitespace(text) {
+  return text.trim().length === 0;
+}
+
+/**
+ * @param {ASTNode} node
+ * @returns {boolean}
+ */
+function isNonspaceJSXTextOrJSXCurly(node) {
+  return (isJSXText(node) && !isOnlyWhitespace(node.raw)) || node.type === 'JSXExpressionContainer';
+}
+
+/**
+ * Somehow fragment like this is useful: <Foo content={<>ee eeee eeee ...</>} />
+ * @param {ASTNode} node
+ * @returns {boolean}
+ */
+function isFragmentWithOnlyTextAndIsNotChild(node) {
+  return node.children.length === 1 &&
+    isJSXText(node.children[0]) &&
+    !(node.parent.type === 'JSXElement' || node.parent.type === 'JSXFragment');
+}
+
+/**
+ * @param {string} text
+ * @returns {string}
+ */
+function trimLikeReact(text) {
+  const leadingSpaces = /^\s*/.exec(text)[0];
+  const trailingSpaces = /\s*$/.exec(text)[0];
+
+  const start = leadingSpaces.includes('\n') ? leadingSpaces.length : 0;
+  const end = trailingSpaces.includes('\n') ? text.length - trailingSpaces.length : text.length;
+
+  return text.slice(start, end);
+}
+
+/**
+ * Test if node is like `<Fragment key={_}>_</Fragment>`
+ * @param {JSXElement} node
+ * @returns {boolean}
+ */
+function isKeyedElement(node) {
+  return node.type === 'JSXElement' &&
+    node.openingElement.attributes &&
+    node.openingElement.attributes.some(jsxUtil.isJSXAttributeKey);
+}
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    fixable: 'code',
+    docs: {
+      description: 'Disallow unnecessary fragments',
+      category: 'Possible Errors',
+      recommended: false,
+      url: docsUrl('jsx-no-useless-fragment')
+    },
+    messages: {
+      NeedsMoreChidren: 'Fragments should contain more than one child - otherwise, there‘s no need for a Fragment at all.',
+      ChildOfHtmlElement: 'Passing a fragment to an HTML element is useless.'
+    }
+  },
+
+  create(context) {
+    const reactPragma = pragmaUtil.getFromContext(context);
+    const fragmentPragma = pragmaUtil.getFragmentFromContext(context);
+
+    /**
+     * Test whether a node is an padding spaces trimmed by react runtime.
+     * @param {ASTNode} node
+     * @returns {boolean}
+     */
+    function isPaddingSpaces(node) {
+      return isJSXText(node) &&
+        isOnlyWhitespace(node.raw) &&
+        node.raw.includes('\n');
+    }
+
+    /**
+     * Test whether a JSXElement has less than two children, excluding paddings spaces.
+     * @param {JSXElement|JSXFragment} node
+     * @returns {boolean}
+     */
+    function hasLessThanTwoChildren(node) {
+      if (!node || !node.children || node.children.length < 2) {
+        return true;
+      }
+
+      return (
+        node.children.length -
+        (+isPaddingSpaces(node.children[0])) -
+        (+isPaddingSpaces(node.children[node.children.length - 1]))
+      ) < 2;
+    }
+
+    /**
+     * @param {JSXElement|JSXFragment} node
+     * @returns {boolean}
+     */
+    function isChildOfHtmlElement(node) {
+      return node.parent.type === 'JSXElement' &&
+        node.parent.openingElement.name.type === 'JSXIdentifier' &&
+        /^[a-z]+$/.test(node.parent.openingElement.name.name);
+    }
+
+    /**
+     * @param {JSXElement|JSXFragment} node
+     * @return {boolean}
+     */
+    function isChildOfComponentElement(node) {
+      return node.parent.type === 'JSXElement' &&
+        !isChildOfHtmlElement(node) &&
+        !jsxUtil.isFragment(node.parent, reactPragma, fragmentPragma);
+    }
+
+    /**
+     * @param {ASTNode} node
+     * @returns {boolean}
+     */
+    function canFix(node) {
+      // Not safe to fix fragments without a jsx parent.
+      if (!(node.parent.type === 'JSXElement' || node.parent.type === 'JSXFragment')) {
+        // const a = <></>
+        if (node.children.length === 0) {
+          return false;
+        }
+
+        // const a = <>cat {meow}</>
+        if (node.children.some(isNonspaceJSXTextOrJSXCurly)) {
+          return false;
+        }
+      }
+
+      // Not safe to fix `<Eeee><>foo</></Eeee>` because `Eeee` might require its children be a ReactElement.
+      if (isChildOfComponentElement(node)) {
+        return false;
+      }
+
+      return true;
+    }
+
+    /**
+     * @param {ASTNode} node
+     * @returns {Function | undefined}
+     */
+    function getFix(node) {
+      if (!canFix(node)) {
+        return undefined;
+      }
+
+      return function fix(fixer) {
+        const opener = node.type === 'JSXFragment' ? node.openingFragment : node.openingElement;
+        const closer = node.type === 'JSXFragment' ? node.closingFragment : node.closingElement;
+        const childrenText = context.getSourceCode().getText().slice(opener.range[1], closer.range[0]);
+
+        return fixer.replaceText(node, trimLikeReact(childrenText));
+      };
+    }
+
+    function checkNode(node) {
+      if (isKeyedElement(node)) {
+        return;
+      }
+
+      if (hasLessThanTwoChildren(node) && !isFragmentWithOnlyTextAndIsNotChild(node)) {
+        context.report({
+          node,
+          messageId: 'NeedsMoreChidren',
+          fix: getFix(node)
+        });
+      }
+
+      if (isChildOfHtmlElement(node)) {
+        context.report({
+          node,
+          messageId: 'ChildOfHtmlElement',
+          fix: getFix(node)
+        });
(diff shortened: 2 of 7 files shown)


lib/rules/jsx-no-useless-fragment.js
```diff
@@ -0,0 +1,94 @@
...
+
+  // <Fragment>
+  if (
+    name.type === 'JSXIdentifier'
+    && name.name === 'Fragment'
+  ) {
+    return true;
+  }
+
+  // <*.Fragment>
```
ljharb: we shouldn't check anything dot Fragment; we should follow the pragma settings here.

lib/rules/jsx-no-useless-fragment.js
```diff
@@ -0,0 +1,94 @@
...
+ * @param {JSXElement} node
+ * @returns {boolean}
+ */
+function isFragment(node) {
+  const name = node.openingElement.name;
+
+  // <Fragment>
+  if (
+    name.type === 'JSXIdentifier'
+    && name.name === 'Fragment'
```
ljharb: similarly, i think this is a pragma setting

ljharb:
Should this rule be autofixable?

golopot:
Autofixing `<>foo</>`, `<>{foo}</>` might break people's code if they use `PropTypes.element`.

Case like `<><Foo /></>` looks ok to be autofixable.

ljharb:
hm, maybe if there's cases that aren't safe to autofix, those should be separately controllable by an option?

golopot:
Now fragments in html element are reported.

Example:
```js
<section>
  <>
    <div />
    <div />
  </>
</section>
```

golopot:
Now a fragment is fixable if it is not an outermost element, except for when it might add extra spaces.

```js
// fixing by simply removing fragments will add a space between "git" and "hub"
<div>
  git<>
    hub
  </>
</div>
```

ljharb:
Couldn't that be fixed into:
```jsx
<div>
  git{`
    hub
  `}
</div>
```
or
```jsx
<div>
  git{' hub '}
</div>
```
?

golopot:
I think it is ok to leave these cases without a fix. Your fixed code both render "git hub", but my code renders "github".

ljharb:
ah, `git{'hub'}` or just `github` then? but sure, leaving it without a fix is fine too.

lib/rules/jsx-no-useless-fragment.js
```diff
@@ -0,0 +1,167 @@
...
+      return false;
+    }
+
+    /**
+     * Test whether a node is an padding spaces trimmed by react runtime.
+     * @param {ASTNode} node
+     * @returns {boolean}
+     */
+    function isPaddingSpaces(node) {
+      return (node.type === 'JSXText' || node.type === 'Literal')
```
captbaritone: Could use `isJSXText(node)` here.

captbaritone:
Hey! I'm curious what the status of this PR is. I just wrote a similar rule myself before I saw this PR.

If you plan on coming back and getting this across the finish line, I'll just add one more thing to consider: Fragments can be given props. Specifically `key`.

If you don't plan on coming back to this, I'd be happy to open a PR with the one I've written once we having it running for a little while and I'm confident there are no major issues.

ljharb:
@captbaritone please don't open a second PR; it'd be better to update this one.

docs/rules/jsx-no-useless-fragment.md
```diff
@@ -0,0 +1,50 @@
+# Disallow unnecessary fragments (react/jsx-no-useless-fragment)
+
+A fragment is redundant if it contains only one child, or if it is the child of a html element.
```
ljharb: Seems like we need to also add prose and examples and tests showing that a fragment is not useless if it has a `key` prop and also appears in a context where a key prop makes sense (ie, inside an array literal or returned from a function)

lib/rules/jsx-no-useless-fragment.js
```diff
@@ -0,0 +1,138 @@
...
+    const fragmentPragma = pragmaUtil.getFragmentFromContext(context);
+
+    /**
+     * Test whether a node is an padding spaces trimmed by react runtime.
+     * @param {ASTNode} node
+     * @returns {boolean}
+     */
+    function isPaddingSpaces(node) {
+      return isJSXText(node) &&
+        /^\s*$/.test(node.raw) &&
```
ljharb: `node.raw.trim().length === 0`? that covers a wider range of "whitespace"

lib/rules/jsx-no-useless-fragment.js
```diff
@@ -0,0 +1,138 @@
...
+        node.raw.includes('\n');
+    }
+
+    /**
+     * Test whether a JSXElement has less than two children, excluding paddings spaces.
+     * @param {JSXElement|JSXFragment} node
+     */
+    function hasLessThanTwoChildren(node) {
+      if (node.children.length < 2) {
```
ljharb:
```suggestion
      if (!node || !node.children || node.children.length < 2) {
```

lib/rules/jsx-no-useless-fragment.js
```diff
@@ -0,0 +1,138 @@
...
+        return true;
+      }
+
+      return (
+        node.children.length -
+        Number(isPaddingSpaces(node.children[0])) -
```
ljharb:
```suggestion
       isPaddingSpaces(node.children[0]) -
```

lib/rules/jsx-no-useless-fragment.js
```diff
@@ -0,0 +1,138 @@
...
+        Number(isPaddingSpaces(node.children[node.children.length - 1]))
```
ljharb:
```suggestion
       isPaddingSpaces(node.children[node.children.length - 1])
```

lib/rules/jsx-no-useless-fragment.js
```diff
@@ -0,0 +1,138 @@
...
+      if (!node.parent.children) {
+        return false;
+      }
+
+      const i = node.parent.children.indexOf(node);
+      const previousChild = node.parent.children[i - 1];
+      const nextChild = node.parent.children[i + 1];
+
+      return (
+        (!isJSXText(previousChild) || /\s$/.test(previousChild.value)) &&
```
ljharb: probably we should have a function for the "is only whitespace" check, so they can all use the same logic

lib/rules/jsx-no-useless-fragment.js
golopot: tsc does not allow me to use boolean in minus operator.

lib/rules/jsx-no-useless-fragment.js
golopot: ditto

docs/rules/jsx-no-useless-fragment.md
golopot: I pushed a version just let pass any keyed fragment.

captbaritone: If a fragment has only a single JSXElement child, it is still useless even if it does have a key.

For example:

```
<React.Fragment key={someVar}>
  <div />
</React.Fragment>
```
Should be refactored/fixed to:

```
<div key={someVar} />
```

lib/rules/jsx-no-useless-fragment.js
ljharb: (－‸ლ) then tsc is incorrect; at the very least, let's use `+` instead of `Number()`.

docs/rules/jsx-no-useless-fragment.md
ljharb: That's not entirely true; something might be cloning the element.

lib/rules/jsx-no-useless-fragment.js
golopot: ✅

captbaritone:
Another case which is worth considering:

```
<>
  In a case like this, fragments offer a nice way to split long
  strings across multiple lines. Using a fragment here, while
  technically useless, is probably nicer than using a quoted
  string.
</>
```

ljharb:
@captbaritone given that you can use backticks for the same effect, it seems like using a fragment there is strictly worse than the alternative (making an object instead of a string literal)

tests/lib/rules/jsx-no-useless-fragment.js
```diff
@@ -0,0 +1,172 @@
...
+            <b>hub</b>.
+          </>
+        </section>`,
+      output: null,
+      errors: [{messageId: 'ChildOfHtmlElement'}],
+      parser: parsers.BABEL_ESLINT
+    },
+    {
+      code: '<div>a <>{""}{""}</> a</div>',
+      output: null,
```
ljharb: this one seems autofixable, since it's the child of an html element?

tests/lib/rules/jsx-no-useless-fragment.js
```diff
@@ -0,0 +1,172 @@
...
+    },
+    {
+      code: '<>foo</>',
+      output: null,
+      errors: [{messageId: 'NeedsMoreChidren'}],
+      parser: parsers.BABEL_ESLINT
+    },
+    {
+      code: '<><div/></>',
+      output: null,
```
ljharb:
```suggestion
      output: '<div />',
```

tests/lib/rules/jsx-no-useless-fragment.js
```diff
@@ -0,0 +1,172 @@
...
+      code: '<Fragment />',
+      errors: [{messageId: 'NeedsMoreChidren'}]
+    },
+    {
+      code: `
+        <React.Fragment>
+          <Foo />
+        </React.Fragment>
+      `,
+      errors: [{messageId: 'NeedsMoreChidren'}]
```
ljharb:
```suggestion
      output: '<Foo />',
      errors: [{messageId: 'NeedsMoreChidren'}]
```

captbaritone:
@ljharb from a technical/performance perspective, I think you’re right, but from a readability/developer experience, perspective I think there’s a reasonable argument for fragments. With back ticks, new lines and indentation can be annoying write cleanly. For example, I don’t think Prettier will word-wrap strings in back ticks for you (since it can’t know if it’s safe) where as I believe it will word wrap JSX for you since it knows line breaks are safe in white space. (I’m on mobile so I haven’t actually checked)

So, forcing back ticks may also force users to manually handle line wrapping, which can be a bit annoying.

golopot:
Using fragments for linebreak-less strings can help avoid `max-len`. Also makes the string formattable by prettier.   

```js
<Author
  name="eeee"
  description="eee  eeeeee eee eeeee. eeeeeeee eeee  eeeeeeeeeeeee eeee. eeeeeee eee eeeeeeeeee ee eee eeeeeee eee."
/>;
```

```js
<Author
  name="eeee"
  description={
    <>
      eee eeeeee eee eeeee. eeeeeeee eeee eeeeeeeeeeeee eeee. eeeeeee eee
      eeeeeeeeee ee eee eeeeeee eee.
    </>
  }
/>;
```

ljharb:
`max-len` should just be wholesale disabled anyways :-) but ok, fair

tests/lib/rules/jsx-no-useless-fragment.js
golopot: Now fixed.

tests/lib/rules/jsx-no-useless-fragment.js
golopot: Now fixed.

tests/lib/rules/jsx-no-useless-fragment.js
golopot: Now it fixes.

golopot:
I have rewrite the fixer to fixes more cases.

lib/rules/jsx-no-useless-fragment.js
```diff
@@ -0,0 +1,198 @@
...
+/**
+ * Test if node is like `<Fragment key={_}>_</Fragment>`
+ * @param {JSXElement} node
+ * @returns {boolean}
+ */
+function isKeyedElement(node) {
+  return node.type === 'JSXElement' &&
+    node.openingElement.attributes &&
+    node.openingElement.attributes.some(attribute => (
+      attribute.type === 'JSXAttribute' &&
```
ljharb: should this predicate be moved to utils?

lib/rules/jsx-no-useless-fragment.js
```diff
@@ -0,0 +1,198 @@
...
+
+    /**
+     * @param {ASTNode} node
+     * @returns {boolean}
+     */
+    function canFix(node) {
+      // Fragments that are child elements can always be fixed
+      if (!(node.parent.type === 'JSXElement' || node.parent.type === 'JSXFragment')) {
+        // const a = <></>
+        if (node.children.length === 0) {
```
ljharb: this is covered implicitly by `[].some(…)` always returning false, so this block can be removed

lib/rules/jsx-no-useless-fragment.js
```diff
@@ -0,0 +1,198 @@
...
+     * @param {ASTNode} node
+     * @returns {((fixer: object) => object) | undefined}
+     */
+    function getFix(node) {
+      if (!canFix(node)) {
+        return undefined;
+      }
+
+      return function fix(fixer) {
+        const opener = node.type === 'JSXFragment' ? node.openingFragment : node.openingElement;
```
ljharb: should this use `utils.isFragment(node, …)`? same below

tests/lib/rules/jsx-no-useless-fragment.js
```diff
@@ -0,0 +1,205 @@
...
+        react: {
+          pragma: 'SomeReact',
+          fragment: 'SomeFragment'
+        }
+      },
+      errors: [{messageId: 'NeedsMoreChidren'}]
+    },
+    {
+      code: '<Eeee><>foo</></Eeee>',
+      output: '<Eeee>foo</Eeee>',
```
ljharb: actually this one might not be safe, if `Eeee` requires its children be a ReactElement; i think we can only autofix these for html elements, not custom ones.
```suggestion
      output: null,
```

lib/rules/jsx-no-useless-fragment.js
golopot: This is a hard question. The function is not reused elsewhere and its purpose is very tied to this rule, so I think no.

lib/rules/jsx-no-useless-fragment.js
golopot: Actually this cannot be removed otherwise tests fails. When `node.children` is an empty array, current behavior returns false. With the block removed, it returns true.

lib/rules/jsx-no-useless-fragment.js
golopot: no, because `utils.isFragment` accepts `<Fragment>`, while we only want `<>` .

tests/lib/rules/jsx-no-useless-fragment.js
golopot: indeed

lib/rules/jsx-no-useless-fragment.js
ljharb: i meant the `.some` callback, specifically

lib/rules/jsx-no-useless-fragment.js
ljharb: That doesn't make sense; `[].some(() => true)` returns `false`.

lib/rules/jsx-no-useless-fragment.js
ljharb: maybe we should make a `isFragmentSyntax`, and also use it inside `isFragment`?

lib/rules/jsx-no-useless-fragment.js
golopot: done

lib/rules/jsx-no-useless-fragment.js
golopot: There is little gain to change it to `isFragmentSyntax(node)`. `node.type === 'JSXFragment'` is expressive, short, and has zero ambiguity. And we ready do `node.type === '???'` in the codebase countless times.

lib/rules/jsx-no-useless-fragment.js
golopot: I confirmed again that test fails when the if statement is removed. https://travis-ci.org/golopot/eslint-plugin-react/jobs/582110265

ljharb: ohhh i get what i'm confused about. the `.some` will return false, preventing the early exit, and then will fall through to the `return true`. sorry for the noise :-)

PR #2265, opened 2019-05-08
[Refactor]: replace util._extend with Object.assign

golopot:


lib/rules/sort-comp.js
```diff
@@ -58,7 +57,7 @@ const defaultConfig = {
 function getMethodsOrder(userConfig) {
   userConfig = userConfig || {};
 
-  const groups = util._extend(defaultConfig.groups, userConfig.groups);
+  const groups = Object.assign(defaultConfig.groups, userConfig.groups);
```
ljharb: this should probably not mutate the default groups - this might be a bug in the rule itself.
```suggestion
  const groups = Object.assign({}, defaultConfig.groups, userConfig.groups);
```

lib/util/Components.js
```diff
@@ -110,7 +109,7 @@ class Components {
       // preserving original array so it can be merged later on.
       copyUsedPropTypes = this._list[id].usedPropTypes && this._list[id].usedPropTypes.slice();
     }
-    this._list[id] = util._extend(this._list[id], props);
+    this._list[id] = Object.assign(this._list[id], props);
```
ljharb: this should either be `Object.assign(this._list[id], props);` or `this._list[id] = Object.assign({}, this._list[id], props);` - iow, if it's mutating, it shouldn't reassign.

lib/util/Components.js
```diff
@@ -97,23 +96,24 @@ class Components {
    * @param {Object} props Additional properties to add to the component.
    */
   set(node, props) {
-    while (node && !this._list[getId(node)]) {
+    let component;
+    while (!(component = this._list[getId(node)])) {
```
ljharb: i'd prefer to avoid using assignment in an expression position; it'd be better to duplicate the assignment than to do this imo.

golopot: That make sense.

golopot:
Thanks.

PR #2267, opened 2019-05-10
Build: add type check using typescript --checkJs

golopot:


golopot:
This errors are found by typescript's `--checkJs` [feature](https://github.com/Microsoft/TypeScript/wiki/Type-Checking-JavaScript-Files). It pull types from jsdoc and typecheck. It is great. CI can be setup by running `tsc`.

ljharb:
Let's add an `npm run-script` for that, then, and a separate travis.yml job that does it, and the appropriate dev deps.

tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    "module": "commonjs",     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
+    "lib": ["es2016"],        /* Specify library files to be included in the compilation. */
+    "allowJs": true,          /* Allow javascript files to be compiled. */
+    "checkJs": true,          /* Report errors in .js files. */
+    "noEmit": true,           /* Do not emit outputs. */
+
+    /* Strict Type-Checking Options */
+    // "strict": true,                           /* Enable all strict type-checking options. */
+    // "noImplicitAny": false,                 /* Raise error on expressions and declarations with an implied 'any' type. */
+    // "strictNullChecks": true,              /* Enable strict null checks. */
```
golopot: The flag `strictNullChecks` is left unenabled because there is about 20 places where I need to either use `@ts-ignore` or write very awkward codes.

golopot:
It might make sense to add [eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc). The errors it catches includes: 1. Missing or superfluous jsdoc param. 2. Unmatched Jsdoc param name and actual name. 3. Enforce presence of types annotation.

lib/rules/jsx-closing-bracket-location.js
```diff
@@ -198,6 +198,7 @@ module.exports = {
       }
       const openingLine = sourceCode.lines[opening.line - 1];
       const openingStartOfLine = {
+        // @ts-ignore
```
ljharb: why?

lib/rules/jsx-closing-bracket-location.js
```diff
@@ -274,7 +275,7 @@ module.exports = {
               case 'tag-aligned':
               case 'line-aligned':
                 return fixer.replaceTextRange([cachedLastAttributeEndPos, node.range[1]],
-                  `\n${getIndentation(tokens, expectedLocation, correctColumn)}${closingTag}`);
+                  `\n${getIndentation(tokens, expectedLocation, correctColumn || 0)}${closingTag}`);
```
ljharb: can the signature of `getIndentation` instead be changed so that the last arg is optional?

lib/rules/no-unknown-property.js
```diff
@@ -193,12 +193,9 @@ function getStandardName(name) {
   if (SVGDOM_ATTRIBUTE_NAMES[name]) {
     return SVGDOM_ATTRIBUTE_NAMES[name];
   }
-  let i;
-  const found = DOM_PROPERTY_NAMES.some((element, index) => {
```
ljharb: i'd rather keep this the same and allow undefined

lib/rules/no-unused-state.js
```diff
@@ -101,7 +101,7 @@ module.exports = {
           parent &&
           parent.type === 'MethodDefinition' && (
             parent.static && parent.key.name === 'getDerivedStateFromProps' ||
-            classMethods.indexOf(parent.key.name !== -1)
+            classMethods.indexOf(parent.key.name) !== -1
```
ljharb: this seems like a legit bug, can we fix this with a regression test in another PR, first?

lib/types.d.ts
```diff
@@ -0,0 +1,13 @@
+import * as eslint from 'eslint';
+import * as estree from 'estree';
```
ljharb: please enable synthetic imports and esmoduleinterop, TS's module system is broken otherwise - `import * as ` is incorrect here.

lib/util/linkComponents.js
```diff
@@ -3,6 +3,7 @@
  */
 'use strict';
 
+/** @type {any[]}} */
```
ljharb: `string[]`?

lib/util/propTypes.js
```diff
@@ -172,6 +174,7 @@ module.exports = function propTypesInstructions(context, components, utils) {
     UnionTypeAnnotation: function(annotation, parentName, seen) {
       const unionTypeDefinition = {
         type: 'union',
+        /** @type {any} */
```
ljharb: lets avoid any use of any :-)

golopot: Tsc does not accept `any[]` and `any[]|true`.

lib/util/linkComponents.js
golopot: Tsc does not accept `string[]`. The official type declaration requires that in `a.concat(b)`, `a` and `b` must have the same type.

lib/types.d.ts
golopot: I had enabled esmoduleinterop. Does the import statements need to be changed? I looked into the `@types/eslint/index.d.ts` file, and it has name exports.

lib/rules/jsx-closing-bracket-location.js
golopot: I missed to clean up this after disabling  `--strictNullCheck`. Tsc cannot infer that `/^\s*/.exec(…)` is never null.

lib/rules/no-unknown-property.js
golopot: Updated

lib/rules/jsx-closing-bracket-location.js
golopot: updated

golopot:
I don't know why it this is closed. Maybe I did something wrong with force push. Apologies.

lib/util/propTypes.js
ljharb: `unknown[]`? `unknown[] | true`? you can certainly do any of those in TS, altho maybe not in jsdoc.

lib/types.d.ts
ljharb: `import eslint from 'eslint'` is the proper form; `import * as` should almost never be used.

lib/rules/jsx-closing-bracket-location.js
```diff
@@ -153,10 +153,11 @@ module.exports = {
      * Get the characters used for indentation on the line to be matched
      * @param {Object} tokens Locations of the opening bracket, closing bracket and last prop
      * @param {String} expectedLocation Expected location for the closing bracket
-     * @param {Number} correctColumn Expected column for the closing bracket
+     * @param {Number} [correctColumn] Expected column for the closing bracket. Defualt to 0
```
ljharb:
```suggestion
     * @param {Number} [correctColumn] Expected column for the closing bracket. Default to 0
```

package.json
```diff
@@ -7,8 +7,9 @@
   "scripts": {
     "coveralls": "cat ./reports/coverage/lcov.info | coveralls",
     "lint": "eslint ./",
-    "pretest": "npm run lint",
+    "pretest": "npm run lint && npm run type-check",
```
ljharb:
```suggestion
    "postlint": "npm run type-check",
    "pretest": "npm run lint",
```

lib/util/linkComponents.js
ljharb: when is this array concatted with something that's not also a `string[]`?

golopot: In lines bellow it is concatted with `setting.something`.

lib/util/propTypes.js
golopot: Tsc does not accept `unknown[]` `unknown[]|true` also.

lib/util/linkComponents.js
ljharb: which should be a `string[]` per the schema, no?

lib/util/propTypes.js
ljharb: just in jsdoc? it certainly works in TS.

lib/util/linkComponents.js
golopot: By schema it could be `{name: something, linkAttribute: something}[]`

ljharb: ah - then should the type be `(string | { name: string, linkAttribute: string })[]`?

lib/util/propTypes.js
golopot: `unionTypeDefinition.children` is at a time assigned `true`, so it cannot be `unknown[]`. `unionTypeDefinition.children.length` is used, so it cannot be `unknown[]|true`.

lib/util/linkComponents.js
golopot: But `settings.linkComponents` is not validated, it could be anything.

lib/types.d.ts
golopot: Changed to default import.

lib/rules/jsx-closing-bracket-location.js
golopot: Removed.

lib/util/propTypes.js
ljharb: then that implies there's a bug, or else type narrowing would have removed `true` prior to the `.length` access?

lib/util/linkComponents.js
ljharb: if it's not that schema tho, the plugin won't work properly, so i think we can assume that's its type.

lib/rules/destructuring-assignment.js
```diff
@@ -127,7 +127,7 @@ module.exports = {
           handleSFCUsage(node);
         }
         if (classComponent) {
-          handleClassUsage(node, classComponent);
+          handleClassUsage(node);
```
ljharb: This is a harmless extra arg being passed - good to clean up. semver-patch.

lib/rules/no-unknown-property.js
```diff
@@ -193,7 +193,7 @@ function getStandardName(name) {
   if (SVGDOM_ATTRIBUTE_NAMES[name]) {
     return SVGDOM_ATTRIBUTE_NAMES[name];
   }
-  let i;
+  let i = -1;
```
ljharb: This seems to be useless, but I assume tells tsc what the type of `i` is. It could also be set to `0`, but `-1` is clearer semantically.

lib/rules/no-unused-state.js
ljharb: Still hoping to see this peeled out, with a test, into a separate PR.

lib/rules/static-property-placement.js
```diff
@@ -102,17 +102,17 @@ module.exports = {
...
-     * @param expectedRule
+     * @param {ASTNode} node
+     * @param {string} expectedRule
      */
     function reportNodeIncorrectlyPositioned(node, expectedRule) {
       // Detect if this node is an expected property declaration adn return the property name
       const name = classProperties.find(propertyName => {
         if (propertiesToCheck[propertyName](node)) {
-          return propertyName;
+          return !!propertyName;
```
ljharb: Assuming the return value of `reportNodeIncorrectlyPositioned` is always used in a truthy/falsy position, this just returns `true` instead of "truthy"…

lib/rules/static-property-placement.js
```diff
@@ -102,17 +102,17 @@ module.exports = {
...
         }
 
-        return null;
+        return false;
```
ljharb: and `false` instead of "falsy".

lib/types.d.ts
ljharb:
```suggestion
import estree from 'estree';
```

lib/rules/no-unused-state.js
golopot: PR #2274 .

lib/rules/no-unknown-property.js
golopot: If `i` is not initialized, it could potentially be `undefined`, that would make tsc reject `DOM_PROPERTY_NAMES[i]`.

ljharb: `[undefined]` on an object without that property is safely undefined, that's just the type system being overly strict :-)

golopot:
Thanks for the thorough review!

PR #2271, opened 2019-05-12
Make propTypes one word consistently

coryhouse:


README.md
```diff
@@ -128,7 +128,7 @@ Enable the rules that you would like to use.
 * [react/no-unescaped-entities](docs/rules/no-unescaped-entities.md): Prevent invalid characters from appearing in markup
 * [react/no-unknown-property](docs/rules/no-unknown-property.md): Prevent usage of unknown DOM property (fixable)
 * [react/no-unsafe](docs/rules/no-unsafe.md): Prevent usage of unsafe lifecycle methods
-* [react/no-unused-prop-types](docs/rules/no-unused-prop-types.md): Prevent definitions of unused prop types
+* [react/no-unused-prop-types](docs/rules/no-unused-prop-types.md): Prevent definitions of unused propTypes
```
ljharb: Actually, now that I look at it, there's tons of usages of the phrase "prop types". `propTypes` is the property under which prop types are stored. I'm not sure there's anything to fix here.

coryhouse: Prop types isn't a term. It doesn't exist in React docs anywhere. I agree there are other pages that should be updated but I focused on this one. 

I noticed this issue because I searched for propTypes to find rules related to it, but some had spaces so my search missed them.

ljharb: React docs aren't the sole authority of what is or is not a term; community usage trumps the official docs, and "prop types" is definitely a term that describes the concept of typing props.

PR #2273, opened 2019-05-12
Update no-unused-prop-types.md

coryhouse:
Using latest syntax, eliminated SFC since that's a dated term. Fixed grammar.

### DIFF
--- a/docs/rules/no-unused-prop-types.md
+++ b/docs/rules/no-unused-prop-types.md
@@ -1,44 +1,49 @@
-# Prevent definitions of unused prop types (react/no-unused-prop-types)
+# Prevent definitions of unused propTypes (react/no-unused-prop-types)
 
-Warns you if you have defined a prop type but it is never being used anywhere.
+Warns if a propType isn't being used.
 
 ## Rule Details
 
 The following patterns are considered warnings:
 
 ```jsx
-var Hello = createReactClass({
-  propTypes: {
-    name: PropTypes.string
-  },
-  render: function() {
+class Hello extends React.Component {
+  render() {
     return <div>Hello Bob</div>;
   }
 });
 
-var Hello = createReactClass({
-  propTypes: {
-    firstname: PropTypes.string.isRequired,
-    middlename: PropTypes.string.isRequired, // middlename is never used below
-    lastname: PropTypes.string.isRequired
-  },
-  render: function() {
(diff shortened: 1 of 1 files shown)


docs/rules/no-unused-prop-types.md
```diff
@@ -86,12 +90,12 @@ AComponent.propTypes = {
   bProp: PropTypes.string // bProp is defined but never used
 };
 ```
-A suggested fix is to assign a bProp to a variable outside of the SFC.
```
ljharb: i prefer to keep the term SFC while it's still in common usage in the community. Please revert these.

docs/rules/no-unused-prop-types.md
```diff
@@ -154,5 +158,5 @@ AComponent.propTypes = {
 };
 ```
 
-Using Intermediate variables might be desired and unavoidable for more complex props structure.
-Like for shape prop types. To avoid false positive in this case make sure `skipShapeProps` is set to `true`.
+Using intermediate variables might be desired and unavoidable for more complex props structure.
+Like for shape propTypes. To avoid false positive in this case make sure `skipShapeProps` is set to `true`.
```
ljharb:
```suggestion
For example, for shape propTypes, to avoid a false positive, make sure `skipShapeProps` is set to `true`.
```

docs/rules/no-unused-prop-types.md
coryhouse: There's no mention of SFC in the react docs anymore. Or stateless function. https://www.dropbox.com/s/u51xabyw7wfuqqy/Screenshot%202019-05-12%2008.12.19.png?dl=0. And even if one does know the history and what SFC means, the warnings apply to function components with state (via Hooks) too.

ljharb: React docs are irrelevant; it’s still a widely used term in the community. A component with a useState hook is still an SFC.

EvHaus:
I've addressed the feedback and rebased against the latest changes in this PR. https://github.com/yannickcr/eslint-plugin-react/pull/2450

Closing this.

ljharb:
@EvHaus rather than opening another PR, collaborators can force push directly to PR branches.

EvHaus:
My bad @ljharb. Good to know for next time. Is this good to merge?

PR #2284, opened 2019-05-18
[fix] `static-property-placement`: fix bug with static property refs inside methods

### LINKED ISSUE #2283: [static-property-placement] Violation on accessing static variable
Example:
```typescript
class MyComponent extends React.Component<MyProps, MyState> {
	public static displayName = "MyComponent ";
	public myMethod() {
		console.log(MyComponent.displayName);
	}
}
```

In the example, line 4 is flagged as a static-property-placement violation even though it simply accesses the property.

dmason30:
Fixes #2283

### DIFF
--- a/lib/rules/static-property-placement.js
+++ b/lib/rules/static-property-placement.js
@@ -83,10 +83,10 @@ module.exports = {
     // ----------------------------------------------------------------------
 
     /**
-     * Checks if we are declaring function in class
-     * @returns {Boolean} True if we are declaring function in class, false if not.
+      * Checks if we are declaring context in class
+      * @returns {Boolean} True if we are declaring context in class, false if not.
      */
-    function isFunctionInClass () {
+    function isContextInClass() {
       let blockNode;
       let scope = context.getScope();
       while (scope) {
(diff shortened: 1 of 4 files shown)


lib/rules/static-property-placement.js
```diff
@@ -129,9 +129,16 @@ module.exports = {
       ClassProperty: node => reportNodeIncorrectlyPositioned(node, STATIC_PUBLIC_FIELD),
 
       MemberExpression: node => {
+        // If definition type is undefined then it must not be a defining expression or if the definition is inside a
+        // class body then skip this node.
+        if ((node.parent.right || {}).type === 'undefined' || isContextInClass()) {
```
ljharb:
```suggestion
        const { right } = node.parent;
        if (!right || right.type === 'undefined' || isContextInClass()) {
```

tests/lib/rules/static-property-placement.js
```diff
@@ -1072,6 +1072,35 @@ ruleTester.run('static-property-placement', rule, {
           }
         }
       `].join('\n')
+    },
+    // ------------------------------------------------------------------------------
+    // edge cases
+    // ------------------------------------------------------------------------------
+    {
+      // Do not error if property assignement inside class function
```
ljharb:
```suggestion
      // Do not error if property assignment is inside a class function
```

lib/rules/static-property-placement.js
```diff
@@ -129,9 +129,17 @@ module.exports = {
       ClassProperty: node => reportNodeIncorrectlyPositioned(node, STATIC_PUBLIC_FIELD),
 
       MemberExpression: node => {
+        // If definition type is undefined then it must not be a defining expression or if the definition is inside a
+        // class body then skip this node.
+        const { right } = node.parent;
```
ljharb: whoops, looks like this breaks in node 4? if so:
```suggestion
        const right = node.parent.right;
```

lib/rules/static-property-placement.js
```diff
@@ -86,7 +86,7 @@ module.exports = {
      * Checks if we are declaring function in class
```
jp7837: update?
```suggestion
      * Checks if we are declaring context in class
```

lib/rules/static-property-placement.js
```diff
@@ -86,7 +86,7 @@ module.exports = {
...
      * @returns {Boolean} True if we are declaring function in class, false if not.
```
jp7837:
```suggestion
      * @returns {Boolean} True if we are declaring context in class, false if not.
```

jp7837:
Buddy test pass on my local code :)

PR #2286, opened 2019-05-23
[Refactor]: remove deprecated context.report calls

golopot:


golopot:
This commit removes the deprecated `context.report(node, message, data, ...others)` calls with the  name parameter style api calls.

You can check that there is no omission by a project-wise search of `context.report\([^{]`.

PR #2288, opened 2019-05-25
[Refactor]: remove unused codes in util/propTypes

golopot:
`UnionTypeDefinition.children` had a inconvenient type `unknown[]|true`. It turned out `children === true` never happens, so the related codes can be removed.

### DIFF
--- a/lib/util/propTypes.js
+++ b/lib/util/propTypes.js
@@ -170,22 +170,9 @@ module.exports = function propTypesInstructions(context, components, utils) {
       /** @type {UnionTypeDefinition} */
       const unionTypeDefinition = {
         type: 'union',
-        children: []
+        children: annotation.types.map(type => buildTypeAnnotationDeclarationTypes(type, parentName, seen))
       };
-      for (let i = 0, j = annotation.types.length; i < j; i++) {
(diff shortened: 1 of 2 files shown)


lib/util/propTypes.js
```diff
@@ -151,22 +151,9 @@ module.exports = function propTypesInstructions(context, components, utils) {
       /** @type {UnionTypeDefinition} */
       const unionTypeDefinition = {
         type: 'union',
-        children: []
+        children: annotation.types.map(type => buildTypeAnnotationDeclarationTypes(type, parentName, seen))
       };
-      for (let i = 0, j = annotation.types.length; i < j; i++) {
-        const type = buildTypeAnnotationDeclarationTypes(annotation.types[i], parentName, seen);
-        // keep only complex type
```
ljharb: can you track down why this was added in the first place? it seems like something that might be important.

golopot: `children === true` comes from b8e0a4ec85024c2d5860a723b0da377374ad6649 . But it does not help me understand why it is needed.

ljharb: cc @phpnode any insight here?

phpnode: no recollection, sorry :(

PR #2291, opened 2019-05-27
Remove jsx-space-before-closing rule

jseminck:
This rule was deprecated over 2 years ago: https://github.com/yannickcr/eslint-plugin-react/commit/050508681a0a46d1e147e828c3debd04b9e81ef6

Would it make sense to remove this rule from the code-base after this much time?


ljharb:
If we're making a breaking change anyways, then sure, but absolutely not until that time.

PR #2292, opened 2019-05-27
Change the reporting for no-unused-prop-types

### LINKED ISSUE #2287: no-unused-prop-types: change reported node to property key
The rule [no-unused-prop-types](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unused-prop-types.md) currently reports the property **value** of the unused propType. I think it is more appropriate to report the property **key**.

```js
// current behavior
Foo.propTypes = {
  neverUsed: PropType.number,
//           ~~~~~~~~~~~~~~~
// PropType is defined but never used. 
}

// proposed behavior
Foo.propTypes = {
  neverUsed: PropType.number,
//~~~~~~~~~
// PropType is defined but never used. 
}
```

jseminck:
Fixes #2287

This change seems rather trivial, which made me wonder if there was ever a reason why it was not done this way? 🤔 

The code used to report only on `prop.node` value, then it would report like this:

Screen Shot 2019-05-27 at 08 02 12

However, this logic was changed in this commit to report only on the `node.value` instead. But this commit doesn't really explain why and seems mostly about another rule instead. I'm thinking that it may have been done by accident?
https://github.com/yannickcr/eslint-plugin-react/commit/2e4f91afdc8a089936a2ff49e6b4c83cccb3432a

So now I question whether we should report back to the full `node` or change it to `node.key`.

Otherwise, I tried it out on a local project and seemed to work fine, all tests also continue to pass once the column context values were adjusted.

Before:
Screen Shot 2019-05-27 at 07 44 58

After:
Screen Shot 2019-05-27 at 07 44 42

### DIFF
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -104,7 +104,7 @@ module.exports = {
 
         if (prop.node && !isPropUsed(component, prop)) {
           context.report({
-            node: prop.node.value || prop.node,
+            node: prop.node.key || prop.node,
             message: UNUSED_MESSAGE,
             data: {
               name: prop.fullName
--- a/tests/lib/rules/no-unused-prop-types.js
+++ b/tests/lib/rules/no-unused-prop-types.js
@@ -3229,7 +3229,7 @@ ruleTester.run('no-unused-prop-types', rule, {
       errors: [{
         message: '\'unused\' PropType is defined but prop is never used',
         line: 3,
-        column: 13
+        column: 5
       }]
     }, {
       code: [
@@ -3245,7 +3245,7 @@ ruleTester.run('no-unused-prop-types', rule, {
       errors: [{
         message: '\'name\' PropType is defined but prop is never used',
         line: 3,
-        column: 11
+        column: 5
       }]
     }, {
       code: [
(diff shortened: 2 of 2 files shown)


lib/rules/no-unused-prop-types.js
```diff
@@ -104,7 +104,7 @@ module.exports = {
 
         if (prop.node && !isPropUsed(component, prop)) {
           context.report({
-            node: prop.node.value || prop.node,
+            node: prop.node.key || prop.node,
```
jseminck: I wonder if the fallback to `prop.node` here is actually needed? If the fallback is used then it will report like this:

Screen Shot 2019-05-27 at 08 02 12

golopot: Maybe it is for flow object type? It seems that babel-eslint does not provide `key` in `ObjectTypeProperty` nodes [demo](https://astexplorer.net/#/gist/6e58a5503fe6641226fa05d744793698/bac2bea3e2ca012ba38e5950646a7d69648587f4).

ljharb:
@jseminck it'd be great to add a test case that fails when the `prop.node` fallback is removed - that way we can be confident about keeping it or not :-)

ljharb:
@jseminck are you no longer interested in completing this PR?

golopot:
> it'd be great to add a test case that fails when the prop.node fallback is removed - that way we can be confident about keeping it or not :-)

There are already 34+ test cases that will fail when prop.node fallback is removed. The prop.node fallback is needed for propTypes written with flow types.

This pr is ready to go after a rebase and fixing a test case from master.

PR #2301, opened 2019-06-07
[New] `prop-types`: handle variables defined as props

### LINKED ISSUE #442: Assignment of props to a variable allows bypass of prop-types rule
This case will not be detected by prop-types rule:

``` js
var props = this.props
console.log(props.something)
```

Nor this:

``` js
var {props} = this
console.log(props.something)
```

I can understand why this may be difficult/impossible if props is being assigned multiple times or within a something dynamic like a conditional but perhaps it's possible to detect these two simple cases. 

~~According to #298, it appears that there might (?) be some support for `var props = this.props` but I couldn't get this working in my code.~~

edit: According to the [prop-types tests](https://github.com/yannickcr/eslint-plugin-react/blob/e716f99f35579039be1c115d7648701ce90ceb56/tests/lib/rules/prop-types.js#L790-L801), it appears this is known to not work :disappointed::

``` js
    }, {
      // Reassigned props are ignored
      code: [
        'export class Hello extends Component {',
        '  render() {',
        '    const props = this.props;',
        '    return <div>Hello {props.name.firstname} {props[\'name\'].lastname}</div>',
        '  }',
        '}'
      ].join('\n'),
      parser: 'babel-eslint'
    }, {
```

[Source](https://github.com/yannickcr/eslint-plugin-react/blob/e716f99f35579039be1c115d7648701ce90ceb56/tests/lib/rules/prop-types.js#L790-L801)

Semi-Related Issues for reference:

https://github.com/yannickcr/eslint-plugin-react/issues/27
https://github.com/yannickcr/eslint-plugin-react/pull/36
https://github.com/yannickcr/eslint-plugin-react/issues/104
https://github.com/yannickcr/eslint-plugin-react/pull/112
https://github.com/yannickcr/eslint-plugin-react/pull/124
https://github.com/yannickcr/eslint-plugin-react/issues/136
https://github.com/yannickcr/eslint-plugin-react/issues/298

Aside: Benefit of this pattern is that it removes risk of unexpected `this`, is more terse than `this.props.something` and keeps props nested under a 'props' namespace thus prop values cannot be confused/shadowed by other vars.

---

Repro case:
#### index.js

``` js
import React from 'react'

class ShouldFail extends React.Component {
  render () {
    const props = this.props
    console.log(props.something)
  }
}
```
#### package.json

``` js
{
  "name": "eslint-react-prop-types-test",
  "version": "1.0.0",
  "main": "index.js",
  "scripts": {
    "test": "eslint index.js"
  },
  "dependencies": {
    "babel-eslint": "^4.1.8",
    "eslint": "^2.0.0",
    "eslint-plugin-react": "^3.16.1"
  }
}
```
#### .eslintrc

``` js
{
  "parser": "babel-eslint",
  "plugins": [
    "react"
  ],
  "rules": {
    "react/prop-types": 2
  }
}

```

### LINKED ISSUE #833: Referencing with PropTypes.shape
When referencing shaped PropTypes, it is producing `react/no-unused-prop-types`.

See below:

``` js
const myComponent = ({ bar }) => (
  <div>
    {bar.a}
    {bar.n}
    {bar.c}
  </div>
);

myComponent.propTypes = {
  bar: PropTypes.shape({
    a: PropTypes.string, // react-no-unused-prop-types
    b: PropTypes.string, // react-no-unused-prop-types
    c: PropTypes.string, // react-no-unused-prop-types
  }),
};
```

Also errors for:

``` js
const myComponent = (props) => {
  const bar = props.bar; // produces errors
  // const { bar } = props; // produce errors
  return (
    <div>
      {bar.a}
      {bar.n}
      {bar.c}
    </div>
  );
}
```

We would have to use it directly to avoid errors:

``` js
const myComponent = (props) => {
  return (
    <div>
      {props.bar.a}
      {props.bar.n}
      {props.bar.c}
    </div>
  );
}
```

### LINKED ISSUE #1002: react/no-unused-prop-types false positive on destructured props

### LINKED ISSUE #1116: [Stateless Components + Object props + Flow]: Proptype is defined but prop is never used
Consider the following contrived example:

```jsx
type PropsFlowType = {|
  ...
  primaryAction: {
    action: () => void, // here
    icon: string, // here
    title: string, // here
  },
  ...
|};

const StatelessComponent = ({ primaryAction, ... }: PropsFlowType) => (
  ...
  <SomeComponent
    action={primaryAction.action}
    icon={primaryAction.icon}
    title={primaryAction.title}
  />
  ...
);
```

Eslint will complain at the lines marked `// here` that these values are not used, even though they are in-fact used. Perhaps a compatibility issue with Flow?

### LINKED ISSUE #1257: [prop-types] Throw `is missing in props validation` when reassign `this.props` to `props`
Hello!

Got an strange behavior when assign `this.props` to `props`, it will throw an error message `is missing in props validation`.
And seems it only happens when use `const props = this.props`.

eslint: 4.0.0
eslint-plugin-react: 7.1.0

This piece of code will throw an error.
```js
const Card = React.createClass({
  propTypes: {
    isReturning: PropTypes.bool,
    date: PropTypes.string
  },
  render() {
    const props = this.props;
    return (
      <div>
          {props.isReturning}
          {props.departureDate}
      </div>
    )
  }
});
```
```bash
error  'isReturning' is missing in props validation    react/prop-types
error  'departureDate' is missing in props validation    react/prop-types
```

This is no error.
```js
const Card = React.createClass({
  propTypes: {
    isReturning: PropTypes.bool,
    date: PropTypes.string
  },
  render() {
    const {isReturning, departureDate} = this.props;
    return (
      <div>
          {isReturning}
          {departureDate}
      </div>
    )
  }
});
```

This is no error.
```js
const Card = React.createClass({
  propTypes: {
    isReturning: PropTypes.bool,
    date: PropTypes.string
  },
  render() {
    return (
      <div>
          {this.props.isReturning}
          {this.props.departureDate}
      </div>
    )
  }
});
```

Even this is no error.
```js
const Card = React.createClass({
  propTypes: {
    isReturning: PropTypes.bool,
    date: PropTypes.string
  },
  render() {
    const test = this.props;
    return (
      <div>
          {test.isReturning}
          {test.departureDate}
      </div>
    )
  }
});
```

### LINKED ISSUE #1764: False positive on `react/no-unused-prop-types`
I came across some code looking like this where `myProp` was reportedly not being used, even though it was.

```
render() {
    const { props } = this;
    const { myProp } = props;
    someFunction(myProp)
}

MyComponent.propTypes = {
  myProp: PropTypes.number,
};
```

 It seems the double desctructure confused the rule because a change to the code below fixed it.
```
render() {
    const { myProp } = this.props;
    someFunction(myProp)
}
```

I think it may be similar to  https://github.com/yannickcr/eslint-plugin-react/issues/933, except this one is not such an anti-pattern. It looks very strange on my reduced version, but in a context with more variables can be more natural.

golopot:
Fixes #442, fixes #1002, fixes #1257, fixes #1764. (`let {props} = this`)
Fixes #833, fixes #1116 (`const Foo = ({ a }) => <>{ a.b }</>`)

This pr fixed cases where a variable is defined as `props.**.*` For example:

```js
// cache props
let props = this.props;
use(props.a);

// cache a property of props
let {a} = props;
use(a.b)

// param destructuring
const Foo = ({ a }) => <>{ a.b }</>
```

In order to tell weather `foo.a` is a props usage, we need to tell whether `foo` is defined as `props.**.*`. For that purpose a Map is used.

### DIFF
--- a/lib/util/usedPropTypes.js
+++ b/lib/util/usedPropTypes.js
@@ -15,6 +15,46 @@ const ast = require('./ast');
 const LIFE_CYCLE_METHODS = ['componentWillReceiveProps', 'shouldComponentUpdate', 'componentWillUpdate', 'componentDidUpdate'];
 const ASYNC_SAFE_LIFE_CYCLE_METHODS = ['getDerivedStateFromProps', 'getSnapshotBeforeUpdate', 'UNSAFE_componentWillReceiveProps', 'UNSAFE_componentWillUpdate'];
 
+function createPropVariables() {
+  /** @type {Map<string, string[]>} Maps the variable to its definition. `props.a.b` is stored as `['a', 'b']` */
+  let propVariables = new Map();
+  let hasBeenWritten = false;
+  const stack = [{propVariables, hasBeenWritten}];
+  return {
+    pushScope() {
+      // popVariables is not copied until first write.
+      stack.push({propVariables, hasBeenWritten: false});
+    },
+    popScope() {
+      stack.pop();
+      propVariables = stack[stack.length - 1].propVariables;
+      hasBeenWritten = stack[stack.length - 1].hasBeenWritten;
+    },
+    /**
+     * Add a variable name to the current scope
+     * @param {string} name
+     * @param {string[]} allNames Example: `props.a.b` should be formatted as `['a', 'b']`
+     */
+    set(name, allNames) {
+      if (!hasBeenWritten) {
+        // copy on write
+        propVariables = new Map(propVariables);
+        stack[stack.length - 1].propVariables = propVariables;
+        stack[stack.length - 1].hasBeenWritten = true;
+      }
+      return propVariables.set(name, allNames);
+    },
+    /**
+     * Get the definition of a variable.
+     * @param {string} name
+     * @returns {string[]} Example: `props.a.b` is represented by `['a', 'b']`
+     */
+    get(name) {
+      return propVariables.get(name);
+    }
+  };
+}
+
 /**
  * Checks if the string is one of `props`, `nextProps`, or `prevProps`
  * @param {string} name The AST node being checked.
@@ -230,6 +270,8 @@ function isPropTypesUsageByMemberExpression(node, context, utils, checkAsyncSafe
 module.exports = function usedPropTypesInstructions(context, components, utils) {
   const checkAsyncSafeLifeCycles = versionUtil.testReactVersion(context, '16.3.0');
 
+  const propVariables = createPropVariables();
+
   /**
    * Mark a prop type as used
    * @param {ASTNode} node The AST node being marked.
@@ -260,6 +302,14 @@ module.exports = function usedPropTypesInstructions(context, components, utils) 
           ) {
             node.parent.id.parent = node.parent; // patch for bug in eslint@4 in which ObjectPattern has no parent
             markPropTypesAsUsed(node.parent.id, allNames);
+          }
+
+          // const a = props.a
+          if (
+            node.parent.type === 'VariableDeclarator' &&
+            node.parent.id.type === 'Identifier'
+          ) {
+            propVariables.set(node.parent.id.name, allNames);
           }
           // Do not mark computed props as used.
           type = name !== '__COMPUTED_PROP__' ? 'direct' : null;
@@ -314,6 +364,7 @@ module.exports = function usedPropTypesInstructions(context, components, utils) 
           const propName = ast.getKeyValue(context, properties[k]);
 
           if (propName) {
+            propVariables.set(propName, parentNames.concat([propName]));
             usedPropTypes.push({
               allNames: parentNames.concat([propName]),
               name: propName,
@@ -371,6 +422,7 @@ module.exports = function usedPropTypesInstructions(context, components, utils) 
    *   FunctionDeclaration, or FunctionExpression
    */
   function handleFunctionLikeExpressions(node) {
+    propVariables.pushScope();
     handleSetStateUpdater(node);
     markDestructuredFunctionArgumentsAsUsed(node);
   }
@@ -392,6 +444,11 @@ module.exports = function usedPropTypesInstructions(context, components, utils) 
 
   return {
     VariableDeclarator(node) {
+      // let props = this.props
+      if (isThisDotProps(node.init) && isInClassComponent(utils) && node.id.type === 'Identifier') {
+        propVariables.set(node.id.name, []);
+      }
+
       // Only handles destructuring
       if (node.id.type !== 'ObjectPattern') {
         return;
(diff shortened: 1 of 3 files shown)


lib/util/usedPropTypes.js
```diff
@@ -15,6 +15,46 @@ const ast = require('./ast');
...
+    /**
+     * Add a variable name to the current scope
+     * @param {string} name
+     * @param {string[]} allNames Example: `props.a.b` should be formatted as `['a', 'b']`
+     */
+    set(name, allNames) {
+      if (!hasBeenWritten) {
+        // copy on write
+        propVariables = new Map(propVariables);
+        stack[stack.length - 1].propVariables = propVariables;
```
ljharb:
```suggestion
        Object.assign(stack[stack.length - 1], { propVariables, hasBeenWritten: true });
```

lib/util/usedPropTypes.js
```diff
@@ -314,6 +364,7 @@ module.exports = function usedPropTypesInstructions(context, components, utils)
           const propName = ast.getKeyValue(context, properties[k]);
 
           if (propName) {
+            propVariables.set(propName, parentNames.concat([propName]));
```
ljharb:
```suggestion
            propVariables.set(propName, parentNames.concat(propName));
```
(unless `propName` could be an array)

lib/util/usedPropTypes.js
```diff
@@ -429,6 +497,12 @@ module.exports = function usedPropTypesInstructions(context, components, utils)
 
     FunctionExpression: handleFunctionLikeExpressions,
 
+    'FunctionDeclaration:exit': propVariables.popScope,
+
+    'ArrowFunctionExpression:exit': propVariables.popScope,
+
+    'FunctionExpression:exit': propVariables.popScope,
```
ljharb: do these not need to retain their receiver? (ie, `(...args) => propVariables.popScope(...args)`)

golopot: No because `createPropVariables` uses closure pattern.

ljharb: In that case, could we immediately destructure out propVariables rather than using it as a namespace?

ljharb:
I'll merge now; we can add more test cases in a followup.

PR #2304, opened 2019-06-10
[Tests] make tests pass in eslint@6

golopot:
Eslint 6.0.0-rc.0 is released 10 hours ago. This pr mostly fixes these two kinds of errors:

> A fatal parsing error occurred: Parsing error: sourceType 'module' is not supported when ecmaVersion < 2015. Consider adding `{ ecmaVersion: 2015 }` to the parser options.

> Parsers provided as strings to RuleTester must be absolute paths.

### DIFF
--- a/tests/lib/rules/no-unused-prop-types.js
+++ b/tests/lib/rules/no-unused-prop-types.js
@@ -4376,7 +4376,8 @@ ruleTester.run('no-unused-prop-types', rule, {
         line: 11,
         column: 8
       }]
-    }, { // None of the props are used issue #1162
+    }, {
+      // None of the props are used issue #1162
       code: [
         'import React from "react"; ',
         'var Hello = React.createReactClass({',
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ env:
   global:
     - TEST=true
   matrix:
-    - ESLINT=next
+    - ESLINT=^6.0.0-0
     - ESLINT=5
     - ESLINT=4
 after_success:
@@ -32,9 +32,8 @@ matrix:
     - node_js: '4'
       env: ESLINT=5
     - node_js: '4'
-      env: ESLINT=next
+      env: ESLINT=^6.0.0-0
     - node_js: '6'
-      env: ESLINT=next
+      env: ESLINT=^6.0.0-0
   allow_failures:
     - node_js: '11'
-    - env: ESLINT=next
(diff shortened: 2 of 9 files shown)


tests/lib/rules/no-unused-prop-types.js
```diff
@@ -4376,7 +4376,7 @@ ruleTester.run('no-unused-prop-types', rule, {
         line: 11,
         column: 8
       }]
-    }, { // None of the props are used issue #1162
+    }, {// None of the props are used issue #1162
```
ljharb:
```suggestion
    }, { // None of the props are used issue #1162
```

golopot:
> Can we make eslint 6 explicitly tested in travis.yml, and not allowed to fail?

We cannot yet replace `eslint@next` with `eslint@6` because the version `eslint@6` does not exists.
```
$ npm i eslint@6
npm ERR! code ETARGET
npm ERR! notarget No matching version found for eslint@6
npm ERR! notarget In most cases you or one of your dependencies are requesting
npm ERR! notarget a package version that doesn't exist.
```

> Also, can we fix appveyor too?

I believe appveyor just randomly fails. At least it passed in my fork.

tests/lib/rules/no-unused-prop-types.js
golopot: This is changed by eslint@6. Without this change the pretest will fail.

ljharb:
@golopot true, but we can use `@^6.0.0-0`.

golopot:
Appveyor still fails at `npm install`.

Maybe this pr can wait until the release of eslint 6, which I think will be in 2 weeks.

tests/lib/rules/no-unused-prop-types.js
ljharb: what lint rule complains about this?

golopot: object-curly-spacing

ljharb:
```suggestion
    }, {
      // None of the props are used issue #1162
```
this, then ^

petersendidit:
I believe `>6.0.0-rc` would work for installing version 6 before its out of a prerelease version.

ljharb:
@petersendidit yes, but then that would include v7 as well in the future. `^6.0.0-0` is the proper semver range.

PR #2316, opened 2019-06-21
fix(jsx-key): handle shorthand fragment syntax

kaykayehnn:
In eslint/eslint#9664 eslint added support for the shorthand fragment syntax (`<></>`) as a new node type: `JSXFragment`. The jsx-key rule hasn't been updated to handle it correctly so in some cases where it should report errors it doesn't. Here is an example

```js
[1, 2, 3].map(x => <>{x}</>);
```

This PR fixes this and adds some tests. The error messages could probably be improved and maybe we should add an example to the docs?

### DIFF
--- a/lib/rules/jsx-key.js
+++ b/lib/rules/jsx-key.js
@@ -31,6 +31,11 @@ module.exports = {
           node,
           message: 'Missing "key" prop for element in iterator'
         });
+      } else if (node.type === 'JSXFragment') {
+        context.report({
+          node,
+          message: 'Missing "key" prop for element in iterator. Shorthand fragment syntax does support providing keys'
+        });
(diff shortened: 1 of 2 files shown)


kaykayehnn:
Yeah, this would be a breaking change. I added it behind a checkFragmentShorthand option.
It would make sense that by default this option should be enabled. Maybe the option could be removed and be the default in the next major?

yannickcr:
Merged, thank you 🙂 

(and yes it will be definitely set as the default in next major)

kaykayehnn:
Great! I was planning to update the docs and tweak the message a bit, I should have left a comment about that. Is it ok if I make a follow up PR for that?

PR #2331, opened 2019-06-27
Minor speedup - Replace regexp usage with property comparison

steelbrain:
Getting source code and then creating a regexp and testing against it seems excessive when we already have access to AST. This minor change saves 500ms in a big project, your milage may vary


steelbrain:
The differences between outputs of different parsers are too many 🤦‍♂

PR #2340, opened 2019-07-02
fix: remove unnecessary schema from jsx-no-comment-textnodes

bradzacher:
I noticed that there was a schema defining an empty object - it's an unnecessary definition, so remove it.

### DIFF
--- a/lib/rules/jsx-no-comment-textnodes.js
+++ b/lib/rules/jsx-no-comment-textnodes.js
@@ -20,11 +20,7 @@ module.exports = {
       url: docsUrl('jsx-no-comment-textnodes')
     },
 
-    schema: [{
-      type: 'object',
-      properties: {},
-      additionalProperties: false
-    }]
+    schema: []
   },
 
   create(context) {


lib/rules/jsx-no-comment-textnodes.js
```diff
@@ -20,11 +20,7 @@ module.exports = {
       url: docsUrl('jsx-no-comment-textnodes')
     },
 
-    schema: [{
-      type: 'object',
-      properties: {},
-      additionalProperties: false
```
ljharb: this seems like it'd be a breaking change, in case someone had already been providing an empty object - also, the purpose of this schema is to make it easy for people to add new properties (without being forced to suddenly add the object)

bradzacher: as per #2341 - it's breaking in the strictest sense.

I personally don't see the point in keeping an invalid schema in here. Keeping it for the just in case seems a bit silly, considering that it has been waiting for just in case for 3 years?

PR #2341, opened 2019-07-02
fix: remove unnecessary schema from require-render-return

bradzacher:
I noticed that there was a schema defining an empty object - it's an unnecessary definition, so remove it.

### DIFF
--- a/lib/rules/require-render-return.js
+++ b/lib/rules/require-render-return.js
@@ -21,7 +21,7 @@ module.exports = {
       recommended: true,
       url: docsUrl('require-render-return')
     },
-    schema: [{}]
+    schema: []
   },
 
   create: Components.detect((context, components, utils) => {


lib/rules/require-render-return.js
```diff
@@ -21,7 +21,7 @@ module.exports = {
       recommended: true,
       url: docsUrl('require-render-return')
     },
-    schema: [{}]
+    schema: []
```
ljharb: this seems like it'd be a breaking change, in case someone had already been providing an empty object.

bradzacher: it technically would be breaking in the strictest sense.
Though reviewing the history, this has been an empty object for 3 years, so I doubt there's some legacy config people have left over.

The problem is that this schema value doesn't just allow an empty object, it allows literally value.

ljharb: yes - i agree that's a problem, however, that means that removing this would cause eslint to crash on a config that was previously working :-/

PR #2364, opened 2019-07-31
[Feat]: jsx-pascal-case allowAllCaps allows underscore

TylerR909:
Hello, I hope this finds you well.

My team uses a custom tool for Localization and by convention we've adopted an all-caps snake case for "Translatable Strings" like `<TS_USER_INPUT />` or `<TS_ADD_BUTTON />` to facilitate that.

Our ESLint configs fell out of date as we moved to adopt TypeScript, during which we've written hundreds of these new components. Now that I'm trying to refresh and re-enable them, the `jsx-pascal-case` rule is throwing tons of errors. I like what it's trying to do, though, so I don't want to turn it off. If I had more time I'd try to make `ignore` accept Regex patterns like `/^TS_[A-Z0-9]+$/` but I figure enabling Snake Case with the `allowAllCaps` rule is a simple enough change for most people as it doesn't seem like a breaking change to me.

I hope it meets with your approval! Thanks for your time.

### DIFF
--- a/lib/rules/jsx-pascal-case.js
+++ b/lib/rules/jsx-pascal-case.js
@@ -14,7 +14,7 @@ const jsxUtil = require('../util/jsx');
 // ------------------------------------------------------------------------------
 
 const PASCAL_CASE_REGEX = /^([A-Z0-9]|[A-Z0-9]+[a-z0-9]+(?:[A-Z0-9]+[a-z0-9]*)*)$/;
-const ALL_CAPS_TAG_REGEX = /^[A-Z0-9]+$/;
+const ALL_CAPS_TAG_REGEX = /^[A-Z0-9]+([A-Z0-9_]*[A-Z0-9]+)?$/;
 
 // ------------------------------------------------------------------------------
 // Rule Definition
@@ -65,10 +65,13 @@ module.exports = {
         const isIgnored = ignore.indexOf(name) !== -1;
 
         if (!isPascalCase && !isCompatTag && !isAllowedAllCaps && !isIgnored) {
-          context.report({
-            node,
-            message: `Imported JSX component ${name} must be in PascalCase`
-          });
+          let message = `Imported JSX component ${name} must be in PascalCase`;
+
+          if (allowAllCaps) {
+            message += ' or SCREAMING_SNAKE_CASE';
+          }
+
+          context.report({node, message});
         }
       }
     };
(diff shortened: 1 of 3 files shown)


lib/rules/jsx-pascal-case.js
```diff
@@ -14,7 +14,7 @@ const jsxUtil = require('../util/jsx');
 // ------------------------------------------------------------------------------
 
 const PASCAL_CASE_REGEX = /^([A-Z0-9]|[A-Z0-9]+[a-z0-9]+(?:[A-Z0-9]+[a-z0-9]*)*)$/;
-const ALL_CAPS_TAG_REGEX = /^[A-Z0-9]+$/;
+const ALL_CAPS_TAG_REGEX = /^[A-Z0-9_]+$/;
```
ljharb: i think we should ensure that it's `A_Z` or `A` but not `A_` or `_A` or `__` or `A__B`, with tests.

TylerR909: Great catch, thanks! I added that. Don't really want to get into RegEx voodoo but this one seems simple enough: https://regex101.com/r/kd5Fma/1/

tests/lib/rules/jsx-pascal-case.js
```diff
@@ -70,5 +70,17 @@ ruleTester.run('jsx-pascal-case', rule, {
...
     code: '<YMCA />',
     errors: [{message: 'Imported JSX component YMCA must be in PascalCase'}]
+  }, {
+    code: '<_TEST_COMPONENT />',
+    options: [{allowAllCaps: true}],
+    errors: [{message: 'Imported JSX component _TEST_COMPONENT must be in PascalCase'}]
+  }, {
+    code: '<TEST_COMPONENT_ />',
+    options: [{allowAllCaps: true}],
+    errors: [{message: 'Imported JSX component TEST_COMPONENT_ must be in PascalCase'}]
```
ljharb: seems like the message, when `allowAllCaps`, should say "in PascalCase or SCREAMING_SNAKE_CASE"?

lib/rules/jsx-pascal-case.js
```diff
@@ -65,10 +65,13 @@ module.exports = {
...
 
         if (!isPascalCase && !isCompatTag && !isAllowedAllCaps && !isIgnored) {
-          context.report({
-            node,
-            message: `Imported JSX component ${name} must be in PascalCase`
-          });
+          let message = `Imported JSX component ${name} must be in PascalCase`;
+
+          if (allowAllCaps) {
+            message = message.concat(' or SCREAMING_SNAKE_CASE');
```
ljharb:
```suggestion
            message += ' or SCREAMING_SNAKE_CASE';
```
?

PR #2375, opened 2019-08-06
[Fix] `no-unused-prop-types`: false positive with callback

### LINKED ISSUE #2350: react/no-unused-prop-types false positive in useEffect with destructuring
It seems that destructuring props inside of useEffect causes the `react/no-unused-prop-types` error to be triggered despite the prop being used. In the example below, calling the prop as `prop.loggedIn` instead of first destructuring it will fix the issue and make the error go away.

```
useEffect(() => {
	const { loggedIn, isAdmin } = props;
	.....
}, []);
```

golopot:
Fixes #2350

This pr fixes `getParentStatelessComponent` when the current traversed node is like a callback (is a function and is an argument). The function `getParentStatelessComponent` is revised to use a more whitelist-ish approach.

Fixes false positive like:
```js
function Foo(props) {
  useEffect(() => {
    const { a } = props;
    document.title = a;
  });
  
  return <p/>;
}
  Foo.propTypes = {
  a: PropTypes.string,
}
```

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -577,6 +577,56 @@ function componentRule(rule, context) {
     },
 
     /**
+     * @param {ASTNode} node
+     * @returns {boolean}
+     */
+    isInAllowedPositionForComponent(node) {
+      switch (node.parent.type) {
+        case 'VariableDeclarator':
+        case 'AssignmentExpression':
+        case 'Property':
+        case 'ReturnStatement':
+        case 'ExportDefaultDeclaration': {
+          return true;
+        }
+        case 'SequenceExpression': {
+          return utils.isInAllowedPositionForComponent(node.parent) &&
+            node === node.parent.expressions[node.parent.expressions.length - 1];
+        }
+        default:
+          return false;
+      }
+    },
+
+    /**
+     * Get node if node is a stateless component, or node.parent in cases like
+     * `React.memo` or `React.forwardRef`. Otherwise returns `undefined`.
+     * @param {ASTNode} node
+     * @returns {ASTNode | undefined}
+     */
+    getStatelessComponent(node) {
+      if (node.type === 'FunctionDeclaration') {
+        if (utils.isReturningJSXOrNull(node)) {
+          return node;
+        }
+      }
+
+      if (node.type === 'FunctionExpression' || node.type === 'ArrowFunctionExpression') {
+        if (utils.isInAllowedPositionForComponent(node) && utils.isReturningJSXOrNull(node)) {
+          return node;
+        }
+
+        // Case like `React.memo(() => <></>)` or `React.forwardRef(...)`
+        const pragmaComponentWrapper = utils.getPragmaComponentWrapper(node);
+        if (pragmaComponentWrapper) {
+          return pragmaComponentWrapper;
+        }
+      }
+
+      return undefined;
+    },
+
+    /**
      * Get the parent stateless component node from the current scope
      *
      * @returns {ASTNode} component node, null if we are not in a component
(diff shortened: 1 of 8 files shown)


tests/lib/rules/no-this-in-sfc.js
```diff
@@ -217,13 +217,6 @@ ruleTester.run('no-this-in-sfc', rule, {
       return <div onClick={onClick}>{this.props.foo}</div>;
     }`,
     errors: [{message: ERROR_MESSAGE}, {message: ERROR_MESSAGE}]
-  }, {
-    code: `
-    () => {
-      this.something();
-      return null;
-    }`,
-    errors: [{message: ERROR_MESSAGE}]
```
golopot: This test case is moot. It is an unused expression.

lib/util/Components.js
```diff
@@ -512,43 +512,35 @@ function componentRule(rule, context) {
...
-      while (scope) {
-        if (astUtil.isFunction(scope.block) || astUtil.isClass(scope.block)) {
-          return scope;
+        if (node.type === 'FunctionExpression' || node.type === 'ArrowFunctionExpression') {
+          switch (node.parent.type) {
+            case 'VariableDeclarator':
+            case 'AssignmentExpression':
+            case 'Property':
+            case 'ReturnStatement':
+            case 'ExportDefaultDeclaration': {
```
golopot: These cases are
```js
const Foo = () => <p />;

module.exports = function Foo() {};

module.exports = {
  Foo() {},
};

function hof() {
  return function Foo() {};
}

export default function Foo() {}
```

tests/lib/rules/no-this-in-sfc.js
ljharb: i'm not sure why; it's still a usage of `this` that someone might expect to have side effects.

lib/util/Components.js
ljharb: Can we add explicit test cases for each of them?

tests/lib/rules/no-this-in-sfc.js
golopot: I mean the arrow function is not reference-able and the program is equivalent to no-op, therefore it should never happen in real world code.

ljharb: ah fair, since the arrow function won't be recognized as a component.

lib/util/Components.js
golopot: Currently the tests cases for them are scattered in several rules, do you want a set of test cases centralized in, say, `tests/rules/prop-types.js`?

ljharb: it'd probably be ideal for every rule that uses component detection to have all of these cases, tbh

golopot:
The tests are broken by the merge of #2184, which added a test case: https://github.com/yannickcr/eslint-plugin-react/blob/7bc55ccb6ed2c32d381f09d2f43a77ca60afaae3/tests/lib/rules/no-multi-comp.js#L345-L349

Do we want to support components defined in comma operator? (IMO, no.)

ljharb:
Yes, we should support components in any expression position.

lib/util/Components.js
```diff
@@ -581,45 +581,62 @@ function componentRule(rule, context) {
...
-        // Attribute Expressions inside JSX Elements (<button onClick={() => props.handleClick()}></button>)
-        const isJSXExpressionContainer = node.parent && node.parent.type === 'JSXExpressionContainer';
-        const pragmaComponentWrapper = this.getPragmaComponentWrapper(node);
-        if (isFunction && pragmaComponentWrapper) {
-          return pragmaComponentWrapper;
+      /**
+       * @param {ASTNode} node
+       * @returns {boolean}
+       */
+      function isInAllowedPositiionForComponent(node) {
```
ljharb: Also, this function should be moved up to module level, so it's not recreated for each invocation of this method.
```suggestion
      function isInAllowedPositionForComponent(node) {
```

lib/util/Components.js
```diff
@@ -581,45 +581,62 @@ function componentRule(rule, context) {
...
+        switch (node.parent.type) {
+          case 'VariableDeclarator':
+          case 'AssignmentExpression':
+          case 'Property':
+          case 'ReturnStatement':
+          case 'ExportDefaultDeclaration': {
+            return true;
+          }
+          case 'SequenceExpression': {
+            return isInAllowedPositiionForComponent(node.parent) &&
```
ljharb:
```suggestion
            return isInAllowedPositionForComponent(node.parent) &&
```

lib/util/Components.js
```diff
@@ -581,45 +581,62 @@ function componentRule(rule, context) {
...
+          if (utils.isReturningJSXOrNull(node)) {
+            return node;
+          }
         }
-        // Return the node if it is a function that is not a class method and is not inside a JSX Element
-        if (isFunction && !isMethod && !isJSXExpressionContainer && utils.isReturningJSXOrNull(node)) {
-          return node;
+
+        if (node.type === 'FunctionExpression' || node.type === 'ArrowFunctionExpression') {
+          if (isInAllowedPositiionForComponent(node) && utils.isReturningJSXOrNull(node)) {
```
ljharb:
```suggestion
          if (isInAllowedPositionForComponent(node) && utils.isReturningJSXOrNull(node)) {
```

lib/util/Components.js
```diff
@@ -581,45 +581,62 @@ function componentRule(rule, context) {
...
-          return null;
+      }
+
+      /**
+       * Get node if node is a stateless component, or node.parent in cases like
+       * `React.memo` or `React.forwardRef`. Otherwise returns `undefined`.
+       * @param {ASTNode} node
+       * @returns {ASTNode | undefined}
+       */
+      function getStatelessComponent(node) {
```
ljharb: this function should also be hoisted up to module level

golopot: I 100% prefer hoisted functions, but in this case, hoisting up this function will move it very far away from its usage (because the functions are in a giant `utils` object literal), which I do not feel good about.

ljharb: Perhaps i shouldn't have used the word "hoisted" :-) i don't like hoisting, but i think functions should be defined as far away as possible to maximize reuse.

golopot: Hoisting is a word with overloaded meanings. I mean `hoisted` as `not nested function`. 

I feel functions dealing with the same topic should be co-located close together, so that reader do not need to jump around to understand a functionality.

ljharb: I think that's fine when the two positions are equivalent; in this case it will impact performance since it will be creating a new function many times, instead of reusing the same one from before.

PR #2380, opened 2019-08-12
Add allowed strings, jsx-no-literals

benhollander:


lib/rules/jsx-no-literals.js
```diff
@@ -26,6 +26,9 @@ module.exports = {
       properties: {
         noStrings: {
           type: 'boolean'
+        },
+        allowedStrings: {
+          type: 'array'
```
ljharb: this array should probably be required to be unique, and its contents of type string

lib/rules/jsx-no-literals.js
```diff
@@ -55,6 +59,9 @@ module.exports = {
     }
 
     function getValidation(node) {
+      if (allowedStrings && allowedStrings.indexOf(node.value) > -1) {
```
ljharb: should we make a `new Set` of it on line 40?

tests/lib/rules/jsx-no-literals.js
```diff
@@ -385,6 +413,21 @@ ruleTester.run('jsx-no-literals', rule, {
...
+    }, {
+      /* eslint-disable no-useless-escape */
+      code: `
+        class Comp1 extends Component {
+          render() {
+            return <div bar={\'foo\'}>asdf</div>
+          }
+        }
+      `,
+      /* eslint-enable no-useless-escape */
```
ljharb: no need for this override.
```suggestion
```

tests/lib/rules/jsx-no-literals.js
```diff
@@ -385,6 +413,21 @@ ruleTester.run('jsx-no-literals', rule, {
         {message: stringsMessage('\'foo\'')},
         {message: stringsMessage('`bar`')}
       ]
+    }, {
+      /* eslint-disable no-useless-escape */
```
ljharb:
```suggestion
```

tests/lib/rules/jsx-no-literals.js
```diff
@@ -385,6 +413,21 @@ ruleTester.run('jsx-no-literals', rule, {
...
+      code: `
+        class Comp1 extends Component {
+          render() {
+            return <div bar={\'foo\'}>asdf</div>
```
ljharb:
```suggestion
            return <div bar={'foo'}>asdf</div>
```

lib/rules/jsx-no-literals.js
benhollander: Sure, that would be slightly cleaner

lib/rules/jsx-no-literals.js
benhollander: updated

tests/lib/rules/jsx-no-literals.js
benhollander: updated

cainlevy:
I'd love to use this. Is there a timeline for new releases?

ljharb:
v7.15.0 is released.

cainlevy:
@benhollander is this working as you hoped? we're also dealing with `&nbsp;` (and other punctuation) but the `allowedStrings` option seems to be confounded by leading or trailing whitespace.

benhollander:
@cainlevy it does work for my use case, but I can see that [this line](https://github.com/yannickcr/eslint-plugin-react/pull/2380/files#diff-7098207e1843605155f2cb108eb603efR66) doesn't take whitespace into account, since it's doing a straight string comparison.

Options:
1. We could make that function strip whitespace from `node.value` before comparing it to the list of strings coming from the config. That would likely result in some results that users might not expect.
2. We could all for passing a regex instead of a straight string. That seems like it would result in false positives/negatives if people didn't write perfect regexes, which seems likely to me.

Other ideas?

cainlevy:
1. My first instinct for the immediate issue is also `trim`. Can you say more about a situation where it might be surprising?

    It's worth noting here that the [error message trims the value](https://github.com/benhollander/eslint-plugin-react/blob/2c6430d098a037128cc4645b9d87c5c6c44409e9/lib/rules/jsx-no-literals.js#L50-L55), which creates confusion about what should be added to `allowedStrings`. For example, this error message from my project was not fixed by adding `allowedStrings: ["&nbsp;"]`:

    ```
    Missing JSX expression container around literal string: “&nbsp;”         react/jsx-no-literals
    ```

2. We're migrating from tslint. The equivalent rule there is `jsx-use-translation-function` and its options are `allow-punctuation` and `allow-htmlentities`. Those options trade abstractness for usability and may take care of most of what people want from `allowedStrings` and `allowedRegexps` with less fiddling. Thoughts?

PR #2386, opened 2019-08-17
Added old context to no-deprecated rule

### LINKED ISSUE #2382: no-legacy rule
Hey, https://reactjs.org/docs/legacy-context.html  will be removed in the next major release. But now no rules which prohibit use.
Is it worth making a separate rule no-legacy? Or add conditions to no-deprecated?

Tom910:
In the 16.6.0 react did be deprecation old context https://reactjs.org/blog/2018/10/23/react-v-16-6.html#deprecations-in-strictmode and remove to next major release
fixes #2382


ljharb:
These are only deprecated in Strict Mode; they are decidedly not deprecated in non-Strict mode yet.

PR #2388, opened 2019-08-21
react/prop-types: allow ignoring by glob

a-x-:
example
```js
{
  'react/prop-types': ['warn', { ignore: ["/_path$/", "className"] },
}
```

# todo
- [ ] add test

### DIFF
--- a/lib/rules/prop-types.js
+++ b/lib/rules/prop-types.js
@@ -56,11 +56,14 @@ module.exports = {
 
     /**
      * Checks if the prop is ignored
-     * @param {String} name Name of the prop to check.
-     * @returns {Boolean} True if the prop is ignored, false if not.
+     * @param {String} name Name of the prop to check. Might be a regexp e.g. "/_path$/"
+     * @returns {Boolean} true if the prop is ignored, false otherwise.
      */
     function isIgnored(name) {
-      return ignored.indexOf(name) !== -1;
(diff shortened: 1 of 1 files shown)


lib/rules/prop-types.js
```diff
@@ -56,11 +56,14 @@ module.exports = {
...
      * Checks if the prop is ignored
-     * @param {String} name Name of the prop to check.
-     * @returns {Boolean} True if the prop is ignored, false if not.
+     * @param {String} name Name of the prop to check. Might be a regexp e.g. "/_path$/"
+     * @returns {Boolean} true if the prop is ignored, false otherwise.
      */
     function isIgnored(name) {
-      return ignored.indexOf(name) !== -1;
+      return ignored.reduce((res, pattern) => {
+        if (pattern[0] === '/') return res || new RegExp(pattern).test(name);
```
ljharb: I am strongly against allowing any user data to be used in `new RegExp`; this is also the source of a significant percentage of the CVEs for the JS ecosystem.

It might be OK to add a glob, however.

ljharb (changes requested):
This would need tests, documentation, and a schema change.

First, however, please elaborate on the use case - why is this complexity something we'd want to add?

ljharb:
@a-x- are you interested in updating this PR?

PR #2391, opened 2019-08-23
2381-fix camelCase error in jsx-sort-props

### LINKED ISSUE #2381: react/jsx-sort-props gets confused when props are sorted with --fix but one has camelCase
I found this weird issue where props were sorted automatically via the `--fix` option but the last one had a camel case.
The linter would then complain about the camel cased prop being expected before the other.

Example:
```tsx
<Test aa aB />
```

image

The configuration for this was 
```js
'react/jsx-sort-props': [
            'error',
            {
                callbacksLast: true,
                reservedFirst: true,
                shorthandFirst: true,
            },
        ],
```

Adding the flag `ignoreCase: true` removed the error, but obviously that still means there is a bug somewhere.

tanmoyopenroot:
#### PR checklist

- [x] Addresses an existing issue: fixes #2381 
- [x] bugfix
- [x] Includes tests
- [ ] Documentation update

#### Overview of change:
use `localCompare` rather than `comparison operator`

### DIFF
--- a/lib/rules/jsx-sort-props.js
+++ b/lib/rules/jsx-sort-props.js
@@ -342,7 +342,7 @@ module.exports = {
             }
           }
 
-          if (!noSortAlphabetically && currentPropName < previousPropName) {
+          if (!noSortAlphabetically && previousPropName.localeCompare(currentPropName) > 0) {
             context.report({
               node: decl.name,
               message: 'Props should be sorted alphabetically',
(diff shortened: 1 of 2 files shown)


tanmoyopenroot:
Hi @ljharb ,
As it is approved, Can you merge it?

ljharb:
Thanks for the ping; it's in my queue. I should be able to get to it tonight.

PR #2399, opened 2019-09-03
Fix false positives for display-name

### LINKED ISSUE #2269: False positive display-name when using forwardRef

* **ESLint Version:** 5.16.0
* **Node Version:** 10.12.0
* **npm Version:** 6.9.0

**What parser (default, Babel-ESLint, etc.) are you using?** babel-eslint 10.0.1

<details>

```js
module.exports = {
    "env": {
        "browser": true,
        "es6": true
    },
    "extends": [
        "eslint:recommended",
        "plugin:react/recommended"
    ],
    "parserOptions": {
        "ecmaVersion": 9,
        "sourceType": "module",
        "ecmaFeatures": {
            "jsx": true
        }
    },
    "rules": {
        "indent": [
            "error",
            2,
            { "SwitchCase": 1 }
        ],
        "linebreak-style": [
            "error",
            "unix"
        ],
        "no-console": "warn",
        "quotes": [
            "error",
            "single"
        ],
        "semi": [
            "error",
            "always"
        ]
    }
};
```

</details>

```js
import React from 'react';

const Paragraph = React.forwardRef(function Paragraph({ style }, ref) {
  // const finalStyle = Object.assign({}, style);
  return <div ref={ref} style={style} />;
});

Paragraph.propTypes = {
  style: PropTypes.object
};

export default Paragraph;
export const proptype = Paragraph.PropTypes;
```

```bash
npx eslint <file>
```

Nothing as the file is valid

```
  4:19  error  Component definition is missing display name  react/display-name

✖ 1 problem (1 error, 0 warnings)
```

**Fun fact!** If you uncomment the line with Object.assign, the false positive disappears...

### LINKED ISSUE #2324: display-name triggered on default export using React.memo
This issue was originally reported by @jwalton in #2133. Confirmed that issue exists in 7.13.0 and 7.14.1 (we are not using typescript in our project).

> simple solution there, use a default export :-D

In latest build, this appears to be broken for a default export (although... I'm in a typescript project, so this *might* be a problem specific to linting .tsx files):

```ts
export default React.memo(function LabelControl(props: ControlProps) {
    const control = props.control as OnDemandLabelControl;

    return (
        <ControlWrapper>
            <Translate tagName="h2" message={control.title} />
        </ControlWrapper>
    );
});
```

triggers react/display-name, but:

```ts
function LabelControl(props: ControlProps) {
    const control = props.control as OnDemandLabelControl;

    return (
        <ControlWrapper>
            <Translate tagName="h2" message={control.title} />
        </ControlWrapper>
    );
}

export default React.memo(LabelControl);
```

does not.

_Originally posted by @jwalton in https://github.com/yannickcr/eslint-plugin-react/issues/2133#issuecomment-502449263_

BPScott:
Wrapping a named function declaration with a React.memo or
React.forwardRef will no longer throw false positive errors.

Check how React devtools exposes the displayNames of components wrapped in memo and forwardRef in https://codesandbox.io/s/awesome-paper-hh0wh?module=App.js

The following no longer raise a linting warning (as react can determine displayNames for them):

```jsx
import React from 'react'

export const ComponentWithMemo = React.memo(function Component({ world }) {
 return <div>Hello {world}</div>
})

export const ComponentWithForwardRef = React.forwardRef(function Component({ world }) {
  return <div>Hello {world}</div>
})
```

Using either an unnamed Function declaration (`function() {/*...*/}`) or arrow functions continue to raise a warning, as a displayName can not be inferred.

Nesting memo and forwardref (`const Component = React.memo(React.forwardRef(function Component() { /*...*/}))` is not currently handled by react so continue to raise an error in that case (see facebook/react#16722)

Fixes #2324, Fixes #2269

### DIFF
--- a/lib/rules/display-name.js
+++ b/lib/rules/display-name.js
@@ -90,13 +90,13 @@ module.exports = {
       const namedClass = (
         (node.type === 'ClassDeclaration' || node.type === 'ClassExpression') &&
         node.id &&
-        node.id.name
+        !!node.id.name
       );
 
       const namedFunctionDeclaration = (
         (node.type === 'FunctionDeclaration' || node.type === 'FunctionExpression') &&
         node.id &&
-        node.id.name
+        !!node.id.name
       );
 
       const namedFunctionExpression = (
@@ -200,6 +200,31 @@ module.exports = {
           return;
         }
         markDisplayNameAsDeclared(node);
+      },
+
+      CallExpression(node) {
+        if (!utils.isPragmaComponentWrapper(node)) {
+          return;
+        }
+
+        if (node.arguments.length > 0 && astUtil.isFunctionLikeExpression(node.arguments[0])) {
+          // Skip over React.forwardRef declarations that are embeded within
+          // a React.memo i.e. React.memo(React.forwardRef(/* ... */))
+          // This means that we raise a single error for the call to React.memo
+          // instead of one for React.memo and one for React.forwardRef
+          const isWrappedInAnotherPragma = utils.getPragmaComponentWrapper(node);
+
+          if (
+            !isWrappedInAnotherPragma &&
+            (ignoreTranspilerName || !hasTranspilerName(node.arguments[0]))
+          ) {
+            return;
+          }
+
+          if (components.get(node)) {
+            markDisplayNameAsDeclared(node);
+          }
+        }
       },
 
       'Program:exit': function () {
--- a/tests/lib/rules/display-name.js
+++ b/tests/lib/rules/display-name.js
@@ -457,6 +457,22 @@ ruleTester.run('display-name', rule, {
     `
   }, {
     code: `
+      import React from 'react'
+
+      const ComponentWithMemo = React.memo(function Component({ world }) {
+        return <div>Hello {world}</div>
+      })
+    `
+  }, {
+    code: `
+      import React from 'react'
+
+      const ForwardRefComponentLike = React.forwardRef(function ComponentLike({ world }, ref) {
+        return <div ref={ref}>Hello {world}</div>
+      })
+    `
+  }, {
+    code: `
       function F() {
         let items = [];
         let testData = [{a: "test1", displayName: "test2"}, {a: "test1", displayName: "test2"}];
@@ -681,6 +697,94 @@ ruleTester.run('display-name', rule, {
       };
     `,
     parser: parsers.BABEL_ESLINT,
+    errors: [{
+      message: 'Component definition is missing display name'
+    }]
+  }, {
+    code: `
+      import React from 'react'
+
+      const ComponentWithMemo = React.memo(({ world }) => {
+        return <div>Hello {world}</div>
+      })
+    `,
+    errors: [{
+      message: 'Component definition is missing display name'
+    }]
+  }, {
+    code: `
+      import React from 'react'
+
+      const ComponentWithMemo = React.memo(function() {
+        return <div>Hello {world}</div>
+      })
+    `,
+    errors: [{
+      message: 'Component definition is missing display name'
+    }]
+  }, {
+    code: `
+      import React from 'react'
+
+      const ForwardRefComponentLike = React.forwardRef(({ world }, ref) => {
+        return <div ref={ref}>Hello {world}</div>
+      })
+    `,
+    errors: [{
+      message: 'Component definition is missing display name'
+    }]
+  }, {
+    code: `
+      import React from 'react'
+
+      const ForwardRefComponentLike = React.forwardRef(function({ world }, ref) {
+        return <div ref={ref}>Hello {world}</div>
+      })
+    `,
+    errors: [{
+      message: 'Component definition is missing display name'
+    }]
+  }, {
+    // Only trigger an error for the outer React.memo
+    code: `
+      import React from 'react'
+
+      const MemoizedForwardRefComponentLike = React.memo(
+        React.forwardRef(({ world }, ref) => {
+          return <div ref={ref}>Hello {world}</div>
+        })
+      )
+    `,
+    errors: [{
+      message: 'Component definition is missing display name'
+    }]
+  }, {
+    // Only trigger an error for the outer React.memo
+    code: `
+      import React from 'react'
+
+      const MemoizedForwardRefComponentLike = React.memo(
+        React.forwardRef(function({ world }, ref) {
+          return <div ref={ref}>Hello {world}</div>
+       })
+      )
+    `,
+    errors: [{
+      message: 'Component definition is missing display name'
+    }]
+  }, {
+    // React does not handle the result of forwardRef being passed into memo
+    // ComponentWithMemoAndForwardRef gets shown as Memo(Anonymous)
+    // See https://github.com/facebook/react/issues/16722
+    code: `
+      import React from 'react'
+
+      const MemoizedForwardRefComponentLike = React.memo(
+        React.forwardRef(function ComponentLike({ world }, ref) {
+          return <div ref={ref}>Hello {world}</div>
+        })
+      )
+    `,
     errors: [{
       message: 'Component definition is missing display name'
     }]


lib/rules/display-name.js
```diff
@@ -90,13 +90,13 @@ module.exports = {
       const namedClass = (
         (node.type === 'ClassDeclaration' || node.type === 'ClassExpression') &&
         node.id &&
-        node.id.name
+        !!node.id.name
```
BPScott: This and the below addition of `!!` aren't strictly needed but I spotted every other variable in this block returns a boolean so I figured these should too

tests/lib/rules/display-name.js
```diff
@@ -672,6 +688,94 @@ ruleTester.run('display-name', rule, {
...
+    // right if you squint at it
+    // In this test the plugin identifies two components instead of one.
+    // React is unable to work out the displayName for this code so we're
+    // reporting the right thing, just doing it twice.
+    // I think ideally this should only report one error.
+    code: `
+      import React from 'react'
+
+      export const ComponentWithMemoAndForwardRef = React.memo(
+        React.forwardRef(({ world }) => {
```
ljharb: the thing inside `forwardRef` is not a component, and should not be identified as such. `forwardRef` callbacks take props and `ref`, which isn't what components take.

tests/lib/rules/display-name.js
```diff
@@ -672,6 +688,94 @@ ruleTester.run('display-name', rule, {
...
+    parser: parsers.BABEL_ESLINT,
+    errors: [{
+      message: 'Component definition is missing display name', line: 4
+    }, {
+      message: 'Component definition is missing display name', line: 5
+    }]
+  }, {
+    // React does not currently set a displayName on the memo function when you
+    // pass it a value from forwardRef.
+    // Hopefully eventually that will be fixed, but until then flagging this
```
ljharb: this shouldn't be fixed because it's not broken; forwardRef is a component but does not accept one; it accepts a callback that takes `props` and `ref`.

BPScott: I think I'm referring to a slightly different issue. I'm referring how React.memo handles accepting a Component vs the result of calling forwardRef.

When inspecting the MemoFunction and  MemoThenForwardRefFunction 
 items on https://codesandbox.io/s/awesome-paper-hh0wh?module=App.js I end up with the following in my react devtools:

Screen Shot 2019-09-02 at 11 28 43 PM

Note that calling React.memo with a function component defined inline

```jsx
const MemoFunction = React.memo(function MemoFunction() {
  return <div>MemoFunction</div>;
});
```

Results in in the name `MemoFunction [memo]` in the component tree.

However calling React.memo with an invocation of React.forwardRef (which returns a Component)

```jsx
const MemoThenForwardRefFunction = React.memo(
  React.forwardRef(function MemoThenForwardRefFunction() {
    return <div>MemoThenForwardRefFunction</div>;
  })
);
```

Results in the name `Anonymous` in the component tree  - Given that the result of an unwrapped call to React.forwardRef has a meaningful name, I would have expected that to have been picked up by React.memo

tests/lib/rules/display-name.js
BPScott: Gotcha, thanks for the clarification :)

In that case I think there is something off with the Component detection, as this testcase identifies two components (as found by doing a `console.log(components.list())` in the `Program:exit` function in the display-name rule). It sounds like it should only identify one - possibly the outer call to `React.memo`. This also occurs on master too.

I don't understand how the internals of component detection work to be confident in modifying it so only a single Component is discovered. It looks like 98d4bf31 did some work towards this.

tests/lib/rules/display-name.js
BPScott: I've raised this upstream as https://github.com/facebook/react/issues/16722

BPScott:
Heya @ljharb, I've gave this a kick again and I think it is ready for review. I've updated the naming in the tests.

The functions that you pass into `forwardRef` are trixy - they're not quite components as you say, but you can set a `displayName` on them and that gets honored like regular components. Because they act like components in that sense I think it makes sense to use the existing componentt detection logic for them.

I'm torn on what to do when we have a nested `React.memo(React.forwardRef(/*...*/))`. It's marked as an error for now, but we can go back and revisit later.

tests/lib/rules/display-name.js
```diff
@@ -672,6 +688,101 @@ ruleTester.run('display-name', rule, {
...
+    }]
+  }, {
+    code: `
+      import React from 'react'
+
+      const ForwardRefComponentLike = React.forwardRef(function({ world }, ref) {
+        return <div ref={ref}>Hello {world}</div>
+      })
+    `,
+    parser: parsers.BABEL_ESLINT,
```
ljharb: does this need babel-eslint?

tests/lib/rules/display-name.js
```diff
@@ -672,6 +688,101 @@ ruleTester.run('display-name', rule, {
...
+    }]
+  }, {
+    code: `
+      import React from 'react'
+
+      const ForwardRefComponentLike = React.forwardRef(({ world }, ref) => {
+        return <div ref={ref}>Hello {world}</div>
+      })
+    `,
+    parser: parsers.BABEL_ESLINT,
```
ljharb: does this need babel-eslint?

tests/lib/rules/display-name.js
```diff
@@ -672,6 +688,101 @@ ruleTester.run('display-name', rule, {
...
+    }]
+  }, {
+    code: `
+      import React from 'react'
+
+      const ComponentWithMemo = React.memo(function() {
+        return <div>Hello {world}</div>
+      })
+    `,
+    parser: parsers.BABEL_ESLINT,
```
ljharb: does this need babel-eslint?

tests/lib/rules/display-name.js
```diff
@@ -672,6 +688,101 @@ ruleTester.run('display-name', rule, {
...
     }]
+  }, {
+    code: `
+      import React from 'react'
+
+      const ComponentWithMemo = React.memo(({ world }) => {
+        return <div>Hello {world}</div>
+      })
+    `,
+    parser: parsers.BABEL_ESLINT,
```
ljharb: does this need babel-eslint?

tests/lib/rules/display-name.js
```diff
@@ -672,6 +688,101 @@ ruleTester.run('display-name', rule, {
...
+    code: `
+      import React from 'react'
+
+      const MemoizedForwardRefComponentLike = React.memo(
+        React.forwardRef(({ world }, ref) => {
+          return <div ref={ref}>Hello {world}</div>
+        })
+      )
+    `,
+    parser: parsers.BABEL_ESLINT,
```
ljharb: does this need babel-eslint?

tests/lib/rules/display-name.js
```diff
@@ -672,6 +688,101 @@ ruleTester.run('display-name', rule, {
...
+    code: `
+      import React from 'react'
+
+      const MemoizedForwardRefComponentLike = React.memo(
+        React.forwardRef(function({ world }, ref) {
+          return <div ref={ref}>Hello {world}</div>
+       })
+      )
+    `,
+    parser: parsers.BABEL_ESLINT,
```
ljharb: does this need babel-eslint?

tests/lib/rules/display-name.js
```diff
@@ -672,6 +688,101 @@ ruleTester.run('display-name', rule, {
...
+    code: `
+      import React from 'react'
+
+      const MemoizedForwardRefComponentLike = React.memo(
+        React.forwardRef(function ComponentLike({ world }, ref) {
+          return <div ref={ref}>Hello {world}</div>
+        })
+      )
+    `,
+    parser: parsers.BABEL_ESLINT,
```
ljharb: does this need babel-eslint?

tests/lib/rules/display-name.js
BPScott: Nope, I've removed them all. (thanks for spotting the cargo culting as I copied from other tests)

BPScott:
Removed the unneeded parser settings in tests

fefrei:
> Nesting memo and forwardref (const Component = React.memo(React.forwardRef(function Component() { /*...*/})) is not currently handled by react so continue to raise an error in that case (see facebook/react#16722)

@BPScott: FYI, it looks like this React issue was [solved in November 2019](https://github.com/facebook/react/pull/17274), so it might make sense to allow this now, too 🙂

ljharb:
@fefrei please feel free to file an issue or a PR!

PR #2408, opened 2019-09-11
[New]`sort-comp`: add static-variables option

### LINKED ISSUE #2405: react/sort-comp `static-variables`
The order:
```
"instance-variables",
"constructor",
"static-methods",
"lifecycle",
"everything-else"
"render",                   
```

Is mistaking with **static members**.

For example:
```
private static const wrong_var = 1;
private b;

constructor(props: any) {
        super(props);

        this.state = {
            ....
        };
}
```
Throws: 
> error  wrong_var should be placed after constructor  react/sort-comp

Which is wrong....

Can you please add **"static-variables"** as option in the rule? 
Expect to be something like:
```
"static-variables",
"instance-variables",
"constructor",
"static-methods",
"lifecycle",
"everything-else"
"render",                   
```

vedadeepta:
- [x] Fixes #2405 
- [x] Includes test 
- [x] Includes minor docs update 

#### Overview of change:
Added option `static-variables` for checking ordering of static variables.
Default order config
```
 order: [
    'static-variables',
    'static-methods',
    'lifecycle',
    'everything-else',
    'render'
  ]
```

### DIFF
--- a/lib/rules/sort-comp.js
+++ b/lib/rules/sort-comp.js
@@ -148,8 +148,12 @@ module.exports = {
           if (method.typeAnnotation) {
             methodGroupIndexes.push(groupIndex);
           }
+        } else if (currentGroup === 'static-variables') {
+          if (method.staticVariable) {
+            methodGroupIndexes.push(groupIndex);
+          }
         } else if (currentGroup === 'static-methods') {
-          if (method.static) {
+          if (method.staticMethod) {
             methodGroupIndexes.push(groupIndex);
           }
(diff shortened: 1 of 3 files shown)


lib/rules/sort-comp.js
```diff
@@ -15,6 +15,7 @@ const docsUrl = require('../util/docsUrl');
 
 const defaultConfig = {
   order: [
+    'static-variables',
```
ljharb: changing the default config is a semver-major change; it shouldn't be added here at this time.
```suggestion
```

PR #2409, opened 2019-09-12
[fix]`jsx-curly-brace-presence`: report unnecessary curly braces on multline child expressions

### LINKED ISSUE #2201: react/jsx-curly-brace-presence {"children": "never"} doesn't error on multiline expressions
```
"eslint": "^4.10.0",
"eslint-plugin-react": "^7.10.0",
```

Given this configuration:

```
'react/jsx-curly-brace-presence': ['error', { props: 'never', children: 'never' }],
```

Unnecessary brackets will be detected when the start tag, children, and end tag are all on the same line, like:

```jsx
<div className={styles.myRateValuePercentSign}>{'%'}</div>
```

but they won't be detected when they're on different lines, like:

```jsx
<div className={styles.myRateValuePercentSign}>
    {'%'}
</div>
```

Is this the intended behavior?

Screen Shot 2019-03-15 at 10 59 45 AM
Screen Shot 2019-03-15 at 10 59 24 AM

vedadeepta:
- [x] Fixes #2201 
- [x] Includes test 

#### Overview of change:
Checks for more than one non-whitespace literal children of `JsxElement` instead of directly checking for `chidlren.length !== 1`.

### DIFF
--- a/lib/rules/jsx-curly-brace-presence.js
+++ b/lib/rules/jsx-curly-brace-presence.js
@@ -104,7 +104,7 @@ module.exports = {
     function containsWhitespaceExpression(child) {
       if (child.type === 'JSXExpressionContainer') {
         const value = child.expression.value;
-        return value ? !(/\S/.test(value)) : false;
+        return value ? jsxUtil.isWhiteSpaces(value) : false;
       }
       return false;
     }
@@ -206,7 +206,7 @@ module.exports = {
     }
 
     function isWhiteSpaceLiteral(node) {
-      return node.type && node.type === 'Literal' && node.value && !(/\S/.test(node.value));
+      return node.type && node.type === 'Literal' && node.value && jsxUtil.isWhiteSpaces(node.value);
     }
 
     function getAdjacentSiblings(node, children) {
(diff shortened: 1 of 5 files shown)


lib/rules/jsx-curly-brace-presence.js
```diff
@@ -205,17 +205,44 @@ module.exports = {
...
+      }
+      if (node === children[children.length - 1] && children[children.length - 2]) {
+        return [children[children.length - 2]];
+      }
+      return [];
+    }
+
+    function hasAdjacentJsxExpressionContainers(node, children) {
+      const childrenExcludingWhitespaceLiteral = children.filter(child => !isWhiteSpaceLiteral(child));
+      const adjSiblings = getAdjacentSiblings(node, childrenExcludingWhitespaceLiteral);
```
taion: I don't think this check is quite right. We now get stuff like:

```diff
           <div>
-            Confirm{' '}
+            Confirm
             <button type="button" onClick={this.onClickYes}>
```

But this is not the same markup and it's not desirable here.

PR #2410, opened 2019-09-15
fix[jsx-handler-name]: add option to disable property

### LINKED ISSUE #2400: react/jsx-handler-names: cannot opt-out of handler prefix independent of prop prefix
### Issue

I only want to enforce `eventHandlerPropPrefix`, but I cannot selectively opt out of `eventHandlerPrefix`.

---

#### 1. Rule always plugs in a default value 'handle' into `eventHandlerPrefix` when supplied value is an empty string `''` (falsy)

**Problematic code:**

```js
// will always be set to 'handle' when falsy
const eventHandlerPrefix = configuration.eventHandlerPrefix || 'handle';
```
[source](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/jsx-handler-names.js#L39
)

---

#### 2. Rule prevents "whitelist" JS regex workarounds noted in #1687 because a capital letter must succeed the allowed prefix.

**Example Config:** I want to whitelist `submit` `select` verbatim and `handle*` prefix:

```js
"react/jsx-handler-names": ['error', {
  "eventHandlerPrefix": '(handle|submit|select)',
  "eventHandlerPropPrefix": 'on',
}],
```

**Problematic code:**

```js
// [A-Z] succeeding eventHandlerPrefix prevents handler whitelist
const EVENT_HANDLER_REGEX = new RegExp(`^((props\\.${eventHandlerPropPrefix})|((.*\\.)?${eventHandlerPrefix}))[A-Z].*$`);
```

[source](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/jsx-handler-names.js#L42)

tanmoyopenroot:
#### PR checklist

- [x] Addresses an existing issue: fixes #2400 
- [x] enhancement
- [x] Includes tests
- [ ] Documentation update

#### Overview of change:
```javascript
    schema: [{
      oneOf: [
        {
          type: 'object',
          properties: {
            eventHandlerPrefix: {type: 'string'},
            eventHandlerPropPrefix: {type: 'string'}
          },
          additionalProperties: false
        }, {
          type: 'object',
          properties: {
            eventHandlerPrefix: {type: 'string'},
            eventHandlerPropPrefix: {
              type: 'boolean',
              enum: [false]
            }
          },
          additionalProperties: false
        }, {
          type: 'object',
          properties: {
            eventHandlerPrefix: {
              type: 'boolean',
              enum: [false]
            },
            eventHandlerPropPrefix: {type: 'string'}
          },
          additionalProperties: false
        }
      ]
    }]
```

### DIFF
--- a/lib/rules/jsx-handler-names.js
+++ b/lib/rules/jsx-handler-names.js
@@ -21,26 +21,59 @@ module.exports = {
     },
 
     schema: [{
-      type: 'object',
-      properties: {
-        eventHandlerPrefix: {
-          type: 'string'
-        },
-        eventHandlerPropPrefix: {
-          type: 'string'
+      oneOf: [
+        {
+          type: 'object',
+          properties: {
+            eventHandlerPrefix: {type: 'string'},
+            eventHandlerPropPrefix: {type: 'string'}
+          },
+          additionalProperties: false
+        }, {
+          type: 'object',
+          properties: {
+            eventHandlerPrefix: {type: 'string'},
+            eventHandlerPropPrefix: {
+              type: 'boolean',
+              enum: [false]
+            }
+          },
+          additionalProperties: false
+        }, {
+          type: 'object',
+          properties: {
+            eventHandlerPrefix: {
+              type: 'boolean',
+              enum: [false]
+            },
+            eventHandlerPropPrefix: {type: 'string'}
+          },
+          additionalProperties: false
         }
-      },
-      additionalProperties: false
+      ]
     }]
   },
 
   create(context) {
-    const configuration = context.options[0] || {};
-    const eventHandlerPrefix = configuration.eventHandlerPrefix || 'handle';
-    const eventHandlerPropPrefix = configuration.eventHandlerPropPrefix || 'on';
+    function isPrefixDisabled(prefix) {
+      return prefix === false;
+    }
 
-    const EVENT_HANDLER_REGEX = new RegExp(`^((props\\.${eventHandlerPropPrefix})|((.*\\.)?${eventHandlerPrefix}))[A-Z].*$`);
-    const PROP_EVENT_HANDLER_REGEX = new RegExp(`^(${eventHandlerPropPrefix}[A-Z].*|ref)$`);
+    const configuration = context.options[0] || {};
+
+    const eventHandlerPrefix = isPrefixDisabled(configuration.eventHandlerPrefix) ?
+      null :
+      configuration.eventHandlerPrefix || 'handle';
+    const eventHandlerPropPrefix = isPrefixDisabled(configuration.eventHandlerPropPrefix) ?
+      null :
+      configuration.eventHandlerPropPrefix || 'on';
+
+    const EVENT_HANDLER_REGEX = !eventHandlerPrefix ?
+      null :
+      new RegExp(`^((props\\.${eventHandlerPropPrefix || ''})|((.*\\.)?${eventHandlerPrefix}))[A-Z].*$`);
+    const PROP_EVENT_HANDLER_REGEX = !eventHandlerPropPrefix ?
+      null :
+      new RegExp(`^(${eventHandlerPropPrefix}[A-Z].*|ref)$`);
 
     return {
       JSXAttribute(node) {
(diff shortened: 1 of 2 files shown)


lib/rules/jsx-handler-names.js
```diff
@@ -24,23 +24,42 @@ module.exports = {
       type: 'object',
       properties: {
         eventHandlerPrefix: {
-          type: 'string'
+          type: ['string', 'boolean']
```
ljharb: can this be made to prevent use of `true`? perhaps `oneOf: [{ type: 'string' }, false]` or similar?

lib/rules/jsx-handler-names.js
```diff
@@ -24,23 +24,42 @@ module.exports = {
...
         }
       },
       additionalProperties: false
     }]
   },
 
   create(context) {
+    function isPrefixDisabled(prefix) {
+      return (
+        typeof prefix === 'boolean' &&
```
ljharb: this line is redundant with the one after it
```suggestion
```

lib/rules/jsx-handler-names.js
```diff
@@ -24,23 +24,42 @@ module.exports = {
...
-    const eventHandlerPropPrefix = configuration.eventHandlerPropPrefix || 'on';
+    const eventHandlerPrefix = isPrefixDisabled(configuration.eventHandlerPrefix) ?
+      null :
+      configuration.eventHandlerPrefix || 'handle';
+    const eventHandlerPropPrefix = isPrefixDisabled(configuration.eventHandlerPropPrefix) ?
+      null :
+      configuration.eventHandlerPropPrefix || 'on';
+
+    if (!eventHandlerPrefix && !eventHandlerPropPrefix) {
+      throw new Error('Both \'eventHandlerPrefix\' and \'eventHandlerPropPrefix\' cannot be false');
```
ljharb:
```suggestion
      throw new TypeError('Both \'eventHandlerPrefix\' and \'eventHandlerPropPrefix\' cannot be false');
```

however, this should be enforced via the rule schema, not via runtime JS.

In other words, the schema should allow:
 - string or false
 - false or string
 - string or string

but not "false or false", and no `true`s.

lib/rules/jsx-handler-names.js
tanmoyopenroot: Done

lib/rules/jsx-handler-names.js
tanmoyopenroot: Done

lib/rules/jsx-handler-names.js
tanmoyopenroot: Done

lib/rules/jsx-handler-names.js
```diff
@@ -24,23 +24,56 @@ module.exports = {
...
-    const PROP_EVENT_HANDLER_REGEX = new RegExp(`^(${eventHandlerPropPrefix}[A-Z].*|ref)$`);
+    const eventHandlerPrefix = isPrefixDisabled(configuration.eventHandlerPrefix) ?
+      null :
+      configuration.eventHandlerPrefix || 'handle';
+    const eventHandlerPropPrefix = isPrefixDisabled(configuration.eventHandlerPropPrefix) ?
+      null :
+      configuration.eventHandlerPropPrefix || 'on';
+
+    if (!eventHandlerPrefix && !eventHandlerPropPrefix) {
+      throw new TypeError('Both \'eventHandlerPrefix\' and \'eventHandlerPropPrefix\' cannot be false');
```
ljharb: i'd also like this to be enforced by the schema, eg one of these three:
1. string / string
1. string / false
1. false / string

tanmoyopenroot: In the schema definition,
Only allowed  combination are:
1: string / string 
2: string / false
3: false/ string

```javascript
        eventHandlerPrefix: {
          oneOf: [
            {
              type: 'string'
            },
            {
              type: 'boolean',
              enum: [false]
            }
          ]
        },
        eventHandlerPropPrefix: {
          oneOf: [
            {
              type: 'string'
            },
            {
              type: 'boolean',
              enum: [false]
            }
          ]
        }
```

ljharb: right, i'm suggesting we change the schema so that's not possible.

instead of two schemas, each with two variants (total 4), three schemas (total 3).

tanmoyopenroot: So, Is this ok?
```javascript
    schema: [{
      oneOf: [
        {
          type: 'object',
          properties: {
            eventHandlerPrefix: {type: 'string'},
            eventHandlerPropPrefix: {type: 'string'}
          },
          additionalProperties: false
        }, {
          type: 'object',
          properties: {
            eventHandlerPrefix: {type: 'string'},
            eventHandlerPropPrefix: {
              type: 'boolean',
              enum: [false]
            }
          },
          additionalProperties: false
        }, {
          type: 'object',
          properties: {
            eventHandlerPrefix: {
              type: 'boolean',
              enum: [false]
            },
            eventHandlerPropPrefix: {type: 'string'}
          },
          additionalProperties: false
        }
      ]
    }]
```

ljharb: That looks great!

tanmoyopenroot: Done

ljharb: that means that this TypeError can never happen, so it can be removed?

tanmoyopenroot: Oops, True.

tanmoyopenroot: Done

Issue #690, opened 2016-07-18
Rule proposal: Warn against using an arrow function/anon fn when declaring a component


tleunen:
Component should be named, and arrow functions are anonymous functions, so they produce unnamed components.

Even if Babel currently transforms an anonymous function into a named function, and even if Chrome currently set a name (`fn.name`) when assigning it to a variable, I believe it would be best to add a rule to warn against the use of arrow functions.

For example, when using jsdom and enzyme to test components built with arrow functions, we can get errors messages like this one because the component name was not set: `expected the node in <??? /> to have...`.

So having a specific rule rule would help waning against arrow and anonymous function in general would help for this.

ljharb:
Perhaps it could be more generic and be configurable to require/allow specific function kinds for SFCs.

gaearon:
>Even if Babel currently transforms an anonymous function into a named function, and even if Chrome currently set a name (fn.name) when assigning it to a variable, I believe it would be best to add a rule to warn against the use of arrow functions.

Why? Per spec, the name *will* be assigned, so it’s fine.
(FWIW the official position is that arrow functions work well as components, and we might use this pattern in the documentation.)

ljharb:
@gaearon but in practice it is not always assigned. In general, explicit > implicit, and a linter rule to require explicit names by disallowing arrow functions is appropriate.

gaearon:
Can you help me understand in which situations it wouldn’t get assigned?

ljharb:
@gaearon In the Babel repl, this function has no name:
```js
export default (props) => {
  return <div {...props} />;
}
```
which thus would necessitate:
```js
const Foo = (props) => {
  return <div {...props} />;
};
export default Foo;
```
but export-defaulting one thing only from a file is also a best practice.

In node 4 and later, if you do `const foo = () => {};`, `foo.name` is the empty string - I don't know the exact range of node and Chrome versions that this applies to, since the latest Chrome does have name inference, but there are *many* implementations in the wild that support arrow functions but not name inference.

bkozera:
+1 from me.

Reasons:
- Airbnb suggests that in their [guide](https://github.com/airbnb/javascript/tree/master/react#class-vs-reactcreateclass-vs-stateless) (unfortunately not explaining why)
- Makes it easy to have consistency when writing stateless components across team (could have also opposite version forcing fat arrow)
[...]

PR #2414, opened 2019-09-18
React/declarations for function components

Stefanwullems:
This is just draft of how I would like it to work. Any feedback would be much appreciated.

This pull request was inspired by #690

Tasks:
- [x] warn when component is not defined in the preferred way
- [x] Autofix component to preferred component
- [x] able to configure preference both named and unnamed components and autofix both accordingly
- [x] support autofix for typescript
  - [x] handle components with type arguments
  - [x] ignore function expressions and arrow functions of which the variable declaration is explicitly typed 
- [x] docs

### DIFF
--- a/docs/rules/function-component-definition.md
+++ b/docs/rules/function-component-definition.md
@@ -1,0 +1,254 @@
+# Enforce a specific function type for function components (react/function-component-definition)
+
+This option enforces a specific function type for function components.
+
+**Fixable:** This rule is automatically fixable using the `--fix` flag on the command line.
+
+## Rule Details
+
+This rule is aimed to enforce consistent function types for function components. By default it prefers function declarations for named components and function expressions for unnamed components.
+
+The following patterns are considered warnings:
+
+```jsx
+// function expression for named component
+var Component = function (props) {
+  return <div>{props.content}</div>;
+};
+
+// arrow function for named component
+var Component = (props) => {
+  return <div>{props.content}</div>;
+};
+
+// arrow function for unnamed component
+function getComponent() {
+  return (props) => {
+    return <div>{props.content}</div>;
+  };
+}
+```
+
+## Rule Options
+
+This rule takes an options object as a second parameter where the preferred function type for components can be specified. The first property of the options object is `"namedComponents"` which can be `"function-declaration"`, `"function-expression"`, or `"arrow-function"` and has `'function-declaration'` as its default. The second property is `"unnamedComponents"` that can be either `"function-expression"` or `"arrow-function"`, and has `'function-expression'` as its default.
+
+```js
+...
+"react/function-component-definition": [<enabled>, {
+  "namedComponents": "function-declaration" | "function-expression" | "arrow-function",
+  "unnamedComponents": "function-expression" | "arrow-function"
+}]
+...
+```
+
+The following patterns are considered warnings:
+
+```jsx
+// only function declarations for named components
+// [2, { "namedComponents": "function-declaration" }]
+var Component = function (props) {
+  return <div />;
+};
+
+var Component = (props) => {
+  return <div />;
+};
+
+// only function expressions for named components
+// [2, { "namedComponents": "function-expression" }]
+function Component (props) {
+  return <div />;
+};
+
+var Component = (props) => {
+  return <div />;
+};
+
+// only arrow functions for named components
+// [2, { "namedComponents": "arrow-function" }]
+function Component (props) {
+  return <div />;
+};
+
+var Component = function (props) {
+  return <div />;
+};
+
+// only function expressions for unnamed components
+// [2, { "unnamedComponents": "function-expression" }]
+function getComponent () {
+  return (props) => {
+    return <div />;
+  };
+}
+
+// only arrow functions for unnamed components
+// [2, { "unnamedComponents": "arrow-function" }]
+function getComponent () {
+  return (props) => {
+    return <div />;
+  };
+}
+
+```
+
+The following patterns are **not** warnings:
+
+```jsx
+
+// only function declarations for named components
+// [2, { "namedComponents": "function-declaration" }]
+function Component (props) {
+  return <div />;
+}
+
+// only function expressions for named components
+// [2, { "namedComponents": "function-expression" }]
+var Component = function (props) {
+  return <div />;
+};
+
+// only arrow functions for named components
+// [2, { "namedComponents": "arrow-function" }]
+var Component = (props) => {
+  return <div />;
+};
+
+// only function expressions for unnamed components
+// [2, { "unnamedComponents": "function-expression" }]
+function getComponent () {
+  return (props) => {
+    return <div />;
+  };
+}
+
+// only arrow functions for unnamed components
+// [2, { "unnamedComponents": "arrow-function" }]
+function getComponent () {
+  return (props) => {
+    return <div />;
+  };
+}
+
+```
+
+## Unfixable patterns
+
+There is one unfixable pattern in JavaScript.
+
+It has to do with the fact that this is valid syntax:
+
+```js
+export default function getComponent () {
+  return <div />;
+}
+```
+
+While these are not:
+
+```js
+export default var getComponent = () => {
+  return <div />;
+}
+
+export default var getComponent = function () {
+  return <div />;
+}
+```
+
+These patterns have to be manually fixed.
+
+## Heads up, TypeScript users
+
+Note that the autofixer is somewhat constrained for TypeScript users.
+
+The following patterns can **not** be autofixed in TypeScript:
+
+```tsx
+// function expressions and arrow functions that have type annotations cannot be autofixed to function declarations
+// [2, { "namedComponents": "function-declaration" }]
+var Component: React.FC<Props> = function (props) {
+  return <div />;
+};
+
+var Component: React.FC<Props> = (props) => {
+  return <div />;
+};
+
+// function components with one unconstrained type parameter cannot be autofixed to arrow functions because the syntax conflicts with jsx
+// [2, { "namedComponents": "arrow-function" }]
+function Component<T>(props: Props<T>) {
+  return <div />;
+};
+
+var Component = function <T>(props: Props<T>) {
+  return <div />;
+};
+
+// [2, { "unnamedComponents": "arrow-function" }]
+function getComponent() {
+  return function <T>(props: Props<T>) => {
+    return <div />;
+  }
+}
+```
+
+Type parameters do not produce syntax conflicts if either there are multiple type parameters or, if there is only one constrained type parameter.
+
+The following patterns can be autofixed in TypeScript:
+
+```tsx
+// autofix to function expression with type annotation
+// [2, { "namedComponents": "function-expression" }]
+var Component: React.FC<Props> = (props) => {
+  return <div />;
+};
+
+// autofix to arrow function with type annotation
+// [2, { "namedComponents": "function-expression" }]
+var Component: React.FC<Props> = function (props) {
+  return <div />;
+};
+
+// autofix to named arrow function with one constrained type parameter
+// [2, { "namedComponents": "arrow-function" }]
+function Component<T extends {}>(props: Props<T>) {
+  return <div />;
+}
+
+var Component = function <T extends {}>(props: Props<T>) {
+  return <div />;
+};
+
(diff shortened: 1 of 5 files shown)


Stefanwullems:
#690

index.js
```diff
@@ -9,6 +9,7 @@ const allRules = {
   'button-has-type': require('./lib/rules/button-has-type'),
   'default-props-match-prop-types': require('./lib/rules/default-props-match-prop-types'),
   'destructuring-assignment': require('./lib/rules/destructuring-assignment'),
+  'declarations-for-function-components': require('./lib/rules/declarations-for-function-components'),
```
ljharb: let's keep the PR as a draft until we settle on a name :-)

lib/rules/declarations-for-function-components.js
```diff
@@ -0,0 +1,118 @@
...
+        }
+      },
+
+      'Program:exit': function () {
+        const list = components.list();
+        // Report that component is arrow function
+        Object.keys(list)
+          .filter(component => list[component].isFunctionalComponent)
+          .filter(component => !list[component].isMethodDefinition)
+          .filter(component => !list[component].isDeclaredFunctionalComponent)
```
ljharb: might as well combine these three filters into one, so it's 3 times more efficient

lib/rules/declarations-for-function-components.js
```diff
@@ -0,0 +1,118 @@
...
+      ArrowFunctionExpression(node) {
+        if (components.get(node)) {
+          markComponentAsFunctionalComponent(node);
+        }
+      },
+
+      FunctionDeclaration(node) {
+        if (components.get(node)) {
+          markComponentAsFunctionalComponent(node);
+          markComponentAsDeclaredFunctionComponent(node);
```
ljharb: why does `markComponentAsDeclaredFunctionComponent` not implicitly call `markComponentAsFunctionalComponent`?

index.js
Stefanwullems: Something not specific to the different types of function definitions maybe. Maybe something like `standard-function-component-definition`. The function definition type can be specified in the options then with default being arrow function (since it's the most popular I think).

Stefanwullems:
@ljharb can you point me to an example of an implemented autofix?

Stefanwullems:
Does autofix take typescript into account at all? I imagine autofixing a functional component that is defined like so:
```
const Component: React.FC<Props> = (props) => {
   return null
} 
```
would lead to unintentional side effects. Should eslint-plugin-react care about this? @ljharb

ljharb:
That’s a fair point; if the types can be fixed, great, if not, best to skip auto fixing entirely when there’s type annotations.

lib/rules/function-component-definition.js
```diff
@@ -0,0 +1,112 @@
...
+      url: docsUrl('function-component-definition')
+    },
+
+    schema: [{
+      enum: ['arrow-function', 'function-declaration', 'function-expression']
+    }]
+  },
+
+  create: Components.detect((context, components) => {
+    const configuration = context.options[0] || 'arrow-function';
```
ljharb: the default should not be `arrow-function`; it's a bad practice to have arrow function components. The default should probably be `function expression` - but that raises the question: should there be an option for "function declaration or expression"? It should be totally fine to have either; the primary contention is "arrow or not arrow".

tests/lib/rules/function-component-definition.js
```diff
@@ -0,0 +1,129 @@
...
+    `,
+    options: ['function-declaration']
+  }, {
+    code: `
+      const Hello = function (props) {
+        return <div/>;
+      }
+    `,
+    options: ['function-expression']
+  }],
```
golopot: need test cases for valid `const Foo = React.memo(function Foo() { returns <p />})` when option is `function-declaration`.

lib/rules/function-component-definition.js
golopot: it is weird to have `function expression` as default option since it is an uncommon style.

ljharb: I think it's just about as common as function declarations; the thing that's very common is using `function`.

Stefanwullems: @ljharb  I think there still should be an option for both function-expression as well as function declaration because at the end of the day they are different and people should still be able to prefer one over the other. Next to that I think function declarations should be the default because the react docs use those.

ljharb: I'm not objecting to there being an option for one or the other; I'm saying that the most common thing is what *does* need an option: which is "function expression or function declaration".

Stefanwullems: Wouldn't it be better to be able to specify your preference for unnamed components? Like curried components for instance. I think it would be nice to be able to choose between arrow-expression and function-expression in that case.

ljharb: That might work for "expression", but "declaration" means it wouldn't be valid in every context.

Either way, choosing an autofixing preference is something separate; the important thing is that the rule can be configured not to warn on good things, and to warn on bad things (for whatever definition of good and bad you prefer)

Stefanwullems: True, components with names can be arrow functions, function declarations and function expressions. Unnamed functions can only be arrow functions and function expressions. 

What I am trying to propose is that we make two things configurable that are completely separated of each other. The first one is how named components should be defined, the second one is how unnamed components should be defined. 

With named components you have the option of choosing between arrow functions, function declarations and function expressions and with unnamed components you can only choose between arrow functions and function expressions.

Stefanwullems:
I will start writing the documentation once I have some certainty that it now works the way you are okay with @ljharb

lib/rules/function-component-definition.js
```diff
@@ -0,0 +1,231 @@
...
+const docsUrl = require('../util/docsUrl');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+const NAMED_FUNCTION_TEMPLATES = {
+  'function-declaration': 'function $name$typeParams($params) $body',
+  'arrow-function': 'const $name$typeAnnotation =$typeParams($params) => $body',
+  'function-expression': 'const $name$typeAnnotation = function$typeParams($params) $body'
```
ljharb: tbh these should probably be `var`, so it's maximally compatible but the autofixes for `prefer-const` and `no-var` can kick in?

docs/rules/function-component-definition.md
```diff
@@ -0,0 +1,228 @@
...
+function getComponent() {
+  return (props) => {
+    return <div>{props.content}</div>
+  }
+}
+```
+
+## Rule Options
+
+This rule takes an options object as a second parameter where the preferred function type for components can be specified. The first property of the options object is `"namedComponents"` which can be `"function-declaration"`, `"function-expression"` or `"arrow-function"`. The second property is `"unnamedComponents"` that can be either `"function-expression"` or `"arrow-function"`
```
ljharb: What are the exact defaults?

docs/rules/function-component-definition.md
```diff
@@ -0,0 +1,228 @@
...
+  return <div/>
+}
+
+const Component: React.FC<Props> = (props) => {
+  return <div/>
+}
+
+// function components with one unconstrained type parameter cannot be autofixed to arrow functions because the syntax conflicts with jsx
+// [2, { "namedComponents": "arrow-function" }]
+function Component<T>(props: Props<T>) {
```
ljharb: why can't this be autofixed? there can't be jsx in these positions.

Stefanwullems: It's probably a bug in typescript but in tsx files this cannot be parsed correctly.

ljharb:
@Stefanwullems tests are failing.

docs/rules/function-component-definition.md
```diff
@@ -0,0 +1,229 @@
...
+function getComponent() {
+  return (props) => {
+    return <div>{props.content}</div>
+  }
+}
+```
+
+## Rule Options
+
+This rule takes an options object as a second parameter where the preferred function type for components can be specified. The first property of the options object is `"namedComponents"` which can be `"function-declaration"`, `"function-expression"` or `"arrow-function"` and has `'function-declaration'` as its default. The second property is `"unnamedComponents"` that can be either `"function-expression"` or `"arrow-function"` and has `'function-expression'` as its default.
```
ljharb:
```suggestion
This rule takes an options object as a second parameter where the preferred function type for components can be specified. The first property of the options object is `"namedComponents"` which can be `"function-declaration"`, `"function-expression"`, or `"arrow-function"` and has `'function-declaration'` as its default. The second property is `"unnamedComponents"` that can be either `"function-expression"` or `"arrow-function"`, and has `'function-expression'` as its default.
```

docs/rules/function-component-definition.md
```diff
@@ -0,0 +1,229 @@
...
+
+```js
+...
+"react/function-component-definition": [<enabled>, {
+  "namedComponents": "function-declaration"|"function-expression"|"arrow-function",
```
ljharb:
```suggestion
  "namedComponents": "function-declaration" | "function-expression" | "arrow-function",
```

docs/rules/function-component-definition.md
```diff
@@ -0,0 +1,229 @@
...
+  "unnamedComponents": "function-expression"|"arrow-function"
```
ljharb:
```suggestion
  "unnamedComponents": "function-expression" | "arrow-function"
```

docs/rules/function-component-definition.md
```diff
@@ -0,0 +1,229 @@
...
+
+// [2, { "unnamedComponents": "arrow-function" }]
+function getComponent() {
+  return function <T>(props: Props<T>) => {
+    return <div/>
+  }
+}
+```
+
+If the single type parameter either have a varraint
```
ljharb:
```suggestion
If the single type parameter either have a variant
```
this sentence reads oddly, however

lib/rules/function-component-definition.js
```diff
@@ -0,0 +1,184 @@
...
+    return node.id.name;
+  }
+
+  if (node.type === 'ArrowFunctionExpression' || node.type === 'FunctionExpression') {
+    return hasName(node) && node.parent.id.name;
+  }
+}
+
+function getParams(node, source) {
+  if (!node.params.length) return null;
```
ljharb:
```suggestion
  if (node.params.length === 0) return null;
```

docs/rules/function-component-definition.md
Stefanwullems: Ohh this was supposed to be constraint. That’s what you get if you blindly replace const for var.

tests/lib/rules/function-component-definition.js
```diff
@@ -0,0 +1,634 @@
...
+      '  return function(props): ReactNode {',
+      '    return <div><Component {...props}/></div>',
+      '  }',
+      '}'
+    ].join('\n'),
+    errors: [{message: 'Function component is not a function expression'}],
+    options: [{unnamedComponents: 'function-expression'}],
+    parser: parsers.TYPESCRIPT_ESLINT
+  }
+  ]
```
golopot: Some missing test cases:
```js
export function Foo() {...};
export default function Foo() {...};
export default { Foo() {...} };
exports.Foo = function Foo() {...};
```

Stefanwullems: [] `export function Foo() {...};`
[] `export default function Foo() {...};`
[] `export default { Foo() {...} };`
[] `exports.Foo = function Foo() {...};`

Stefanwullems:
@ljharb If everything is allright, let's get this merged??

ljharb:
@Stefanwullems this needs to be rebased on latest master, ideally squashed down to one commit, and all conflict markers removed (look for `<<<<<`, `>>>>>`, and `=====`).

Stefanwullems:
@ljharb I have no idea how to squash so it would be perfect if you or someone else can take is off my hands before I explode something.

ljharb:
I’m talking about git rebase; i can certainly try to do it for you but you have more context on the change than i do.

Stefanwullems:
> I’m talking about git rebase; i can certainly try to do it for you but you have more context on the change than i do.

@ljharb I think I rebased it but I have no idea if I did it right.

Stefanwullems:
@ljharb I think it's ready

PR #2418, opened 2019-09-24
update grammar for no-access-state-in-setstate.md

neaumusic:


docs/rules/no-access-state-in-setstate.md
```diff
@@ -11,12 +11,12 @@ function increment() {
 }
 ```
 
-If these two `setState` operations is grouped together in a batch it will
-look be something like the following, given that value is 1:
+If two `setState` operations are grouped together in a batch, they
```
ljharb:
```suggestion
If two `setState` operations are grouped together in a batch; they
```

neaumusic: I don't think `;` is right here, I believe `,` or `:` are the correct punctuation. Semicolon is for when two topics can be separated by either a period or a comma, but in this case it's dependent on the precursory clause. Colon is generally for lists, but I think would be fine too.

ljharb: you're right; i'm not sure what i was thinking.

PR #2419, opened 2019-09-25
[New] Add `jsx-no-script-url` to prevent usage of `javascript:` URLs

sergei-startsev:
**In React 16.9** any URLs starting with `javascript:` [scheme](https://wiki.whatwg.org/wiki/URL_schemes#javascript:_URLs) log a warning. React considers the pattern as a dangerous attack surface, see [details](https://reactjs.org/blog/2019/08/08/react-v16.9.0.html#deprecating-javascript-urls) and **in a future major release** it will throw an error if it encounters a `javascript:` URL.

The rule prevents usage of `javascript:` URLs in JSX.

### DIFF
--- a/docs/rules/jsx-no-script-url.md
+++ b/docs/rules/jsx-no-script-url.md
@@ -1,0 +1,57 @@
+# Prevent usage of `javascript:` URLs (react/jsx-no-script-url)
+
+**In React 16.9** any URLs starting with `javascript:` [scheme](https://wiki.whatwg.org/wiki/URL_schemes#javascript:_URLs) log a warning.
+React considers the pattern as a dangerous attack surface, see [details](https://reactjs.org/blog/2019/08/08/react-v16.9.0.html#deprecating-javascript-urls).
+**In a future major release**, React will throw an error if it encounters a `javascript:` URL.
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```jsx
+<a href="javascript:"></a>
+<a href="javascript:void(0)"></a>
+<a href="j\n\n\na\rv\tascript:"></a>
+```
+
+The following patterns are **not** considered warnings:
+
+```jsx
+<Foo href="javascript:"></Foo>
+<a href={"javascript:"}></a>
+```
+
+## Rule Options
+```json
+{
+  "react/jsx-no-script-url": [
+    "error",
+    [
+      {
+        "name": "Link",
+        "props": ["to"]
+      },
+      {
+        "name": "Foo",
+        "props": ["href", "to"]
+      }
+    ]
+  ]
+}
+```
+
+Allows you to indicate a specific list of properties used by a custom component to be checked.
+
+### name
+Component name.
+
+### props
+List of properties that should be validated.
+
+The following patterns are considered warnings with the options listed above:
+
+```jsx
+<Link to="javascript:void(0)"></Link>
+<Foo href="javascript:void(0)"></Foo>
+<Foo to="javascript:void(0)"></Foo>
+```
--- a/lib/rules/jsx-no-script-url.js
+++ b/lib/rules/jsx-no-script-url.js
@@ -1,0 +1,91 @@
+/**
+ * @fileoverview Prevent usage of `javascript:` URLs
+ * @author Sergei Startsev
+ */
+
+'use strict';
+
+const docsUrl = require('../util/docsUrl');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+// https://github.com/facebook/react/blob/d0ebde77f6d1232cefc0da184d731943d78e86f2/packages/react-dom/src/shared/sanitizeURL.js#L30
+/* eslint-disable-next-line max-len, no-control-regex */
+const isJavaScriptProtocol = /^[\u0000-\u001F ]*j[\r\n\t]*a[\r\n\t]*v[\r\n\t]*a[\r\n\t]*s[\r\n\t]*c[\r\n\t]*r[\r\n\t]*i[\r\n\t]*p[\r\n\t]*t[\r\n\t]*:/i;
+
+function hasJavaScriptProtocol(attr) {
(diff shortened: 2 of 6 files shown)


docs/rules/jsx-no-script-url.md
```diff
@@ -0,0 +1,22 @@
...
+```jsx
+<a href="javascript:"></a>
+<a href="javascript:void(0)"></a>
+<a href="j\n\n\na\rv\tascript:"></a>
+```
+
+The following patterns are **not** considered warnings:
+
+```jsx
+<Foo href="javascript:"></Foo>
```
ljharb: what about a `Link` component? we should provide a way to indicate that a specific list of custom components is like an "a", and also to override the name of the "href" prop on those elements.

sergei-startsev: @ljharb Exports can be renamed during import, so `Link` component can be imported as `Foo`:
```js
import { Link as Foo } from "react-router-dom";
```
How are you going to handle it?
Moreover if you pass `javascript:void(0)` as `to` property to  `Link`  component from `react-router-dom`, it doesn't set the same value as `href` attribute:
```jsx
<Link to="javascript:void(0)">Link</Link>

// it adds '/' and doesn't run js
<a href="/javascript:void(0)">Link</a>
```

ljharb: It'd be by name; if you rename it to `Link2` it wouldn't error, to `Link` it would.

That's certainly one example of a `Link` component, but not one I've ever seen since blindly prepending `/` would break some kinds of URLs.

sergei-startsev: @ljharb The PR has been updated, take a look when you get a chance.

sergei-startsev: @ljharb are there any updates here?

ljharb: i'll rereview, but i'm waiting to hear from @gaearon before merging this.

docs/rules/jsx-no-script-url.md
```diff
@@ -0,0 +1,57 @@
...
+<a href={"javascript:"}></a>
+```
+
+## Rule Options
+```json
+{
+  "react/jsx-no-script-url": [
+    "error",
+    [
```
ljharb: It's a shame that we can't share a global "settings" option for this with eslint-plugin-jsx-a11y :-/

lib/rules/jsx-no-script-url.js
```diff
@@ -0,0 +1,89 @@
...
+          },
+          props: {
+            type: 'array',
+            items: {
+              type: 'string',
+              uniqueItems: true
+            }
+          }
+        },
+        additionalProperties: false
```
ljharb: this schema should require one of "name" or "props"

sergei-startsev: Both are actually required, there's no sense to have one without the other. Updated PR.

docs/rules/jsx-no-script-url.md
sergei-startsev: @ljharb We have over two weeks since the last commit. Is there any chance to be merged?

sergei-startsev: @ljharb we have over month since the last discussion. Are there any concerns to merge it? The rule is disabled by default and I don't see how it can affect anybody, but it can help to prevent React errors. I think @gaearon had enough time to take a look at this.

sergei-startsev: @ljharb Any chance to merge?

lib/rules/jsx-no-script-url.js
```diff
@@ -0,0 +1,90 @@
...
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Forbid `javascript:` URLs',
+      category: 'Best Practices',
+      recommended: false,
+      url: docsUrl('jsx-no-script-url')
+    },
+    schema: [{
+      type: 'array',
```
ljharb:
```suggestion
      type: 'array',
      uniqueItems: true,
```

lib/rules/jsx-no-script-url.js
```diff
@@ -0,0 +1,90 @@
...
+        additionalProperties: false
+      }
+    }]
+  },
+
+  create(context) {
+    const configuration = context.options[0] || [];
+    const elements = configuration.map(i => i.name);
+
+    function shouldVerifyElement(node) {
```
ljharb: this function should be hoisted to module level

lib/rules/jsx-no-script-url.js
```diff
@@ -0,0 +1,90 @@
...
+
+      return node.name && props.indexOf(name) !== -1;
+    }
+
+    return {
+      JSXAttribute(node) {
+        const parent = node.parent;
+        if (shouldVerifyElement(parent) && shouldVerifyProp(node) && hasJavaScriptProtocol(node)) {
+          context.report(node, 'A future version of React will block javascript: URLs as a security precaution. ' +
+            'Use event handlers instead if you can. If you need to generate unsafe HTML try using dangerouslySetInnerHTML instead.');
```
ljharb:
```suggestion
            'Use event handlers instead if you can. If you need to generate unsafe HTML, try using dangerouslySetInnerHTML instead.');
```

lib/rules/jsx-no-script-url.js
sergei-startsev: Moved `shouldVerifyElement` as well as `shouldVerifyProp`.

lib/rules/jsx-no-script-url.js
sergei-startsev: Adjusted

lib/rules/jsx-no-script-url.js
sergei-startsev: Added

erunion:
Should this be added to the recommended config?

ljharb:
Not at this time, adding things to the recommended config is a breaking change.

If we ever are forced to do a semver-major, then we'll re-evaluate the recommended config.

PR #2429, opened 2019-10-02
Make jsx-sort-default-props fixable

### LINKED ISSUE #2347: [feature] react/jsx-sort-default-props is not fixable ?
I'm curious about the fact that the rule [`react/jsx-sort-props`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-sort-props.md) is fixable but not [`react/jsx-sort-default-props`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-sort-default-props.md). Is there a reason ? Is it possible to make it fixable ?

emroussel:
There didn't seem to be a reason for not fixing `jsx-sort-default-props` automatically.

Most of this code is copy and pasted from [`sort-prop-types`](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/sort-prop-types.js). It might be good to abstract it somewhere so that it's not duplicated, but I wasn't sure where that should be. If you agree, feel free to point me to where these type of reusable functions usually live, and I'll move them there.

Should fix #2347.

### DIFF
--- a/lib/util/propTypesSort.js
+++ b/lib/util/propTypesSort.js
@@ -1,0 +1,164 @@
+/**
+ * @fileoverview Common propTypes sorting functionality.
+ */
+
+'use strict';
+
+const astUtil = require('./ast');
+
+/**
+ * Returns the value name of a node.
+ *
+ * @param {ASTNode} node the node to check.
+ * @returns {String} The name of the node.
+ */
+function getValueName(node) {
+  return node.type === 'Property' && node.value.property && node.value.property.name;
+}
+
+/**
+ * Checks if the prop is required or not.
+ *
+ * @param {ASTNode} node the prop to check.
+ * @returns {Boolean} true if the prop is required.
+ */
+function isRequiredProp(node) {
+  return getValueName(node) === 'isRequired';
+}
+
+/**
+ * Checks if the proptype is a callback by checking if it starts with 'on'.
+ *
+ * @param {String} propName the name of the proptype to check.
+ * @returns {Boolean} true if the proptype is a callback.
+ */
+function isCallbackPropName(propName) {
+  return /^on[A-Z]/.test(propName);
+}
+
+/**
+ * Checks if the prop is PropTypes.shape.
+ *
+ * @param {ASTNode} node the prop to check.
+ * @returns {Boolean} true if the prop is PropTypes.shape.
+ */
+function isShapeProp(node) {
+  return Boolean(
+    node && node.callee && node.callee.property && node.callee.property.name === 'shape'
+  );
+}
+
+/**
+ * Returns the properties of a PropTypes.shape.
+ *
+ * @param {ASTNode} node the prop to check.
+ * @returns {Array} the properties of the PropTypes.shape node.
+ */
+function getShapeProperties(node) {
+  return node.arguments && node.arguments[0] && node.arguments[0].properties;
+}
+
+/**
+ * Compares two elements.
+ *
+ * @param {ASTNode} a the first element to compare.
+ * @param {ASTNode} b the second element to compare.
+ * @param {Context} context The context of the two nodes.
+ * @param {Boolean=} ignoreCase whether or not to ignore case when comparing the two elements.
+ * @param {Boolean=} requiredFirst whether or not to sort required elements first.
+ * @param {Boolean=} callbacksLast whether or not to sort callbacks after everyting else.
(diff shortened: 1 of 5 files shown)


ljharb:
Thanks, this is great! https://github.com/yannickcr/eslint-plugin-react/tree/master/lib/util seems like a good place to put it.

emroussel:
@ljharb done 👍

lib/util/propTypesSort.js
```diff
@@ -0,0 +1,176 @@
...
+      return -1;
+    }
+  }
+
+  if (ignoreCase) {
+    aKey = toLowerCase(aKey);
+    bKey = toLowerCase(bKey);
+  }
+
+  if (aKey < bKey) {
```
ljharb:
```suggestion
  return aKey.localeCompare(bKey);
```

lib/util/propTypesSort.js
```diff
@@ -0,0 +1,176 @@
...
+      }
+      return acc;
+    }, [[]]);
+
+    nodeGroups.forEach((nodes) => {
+      const sortedAttributes = nodes
+        .slice()
+        .sort((a, b) => sorter(a, b, context, ignoreCase, requiredFirst, callbacksLast));
+
+      for (let i = nodes.length - 1; i >= 0; i--) {
```
ljharb: this seems like it could be done with `nodes.reduceRight`?

lib/util/propTypesSort.js
emroussel: I might be missing something, but I don't think `String.localeCompare` supports sorting upper case letters before lower case ones (like this is doing). The `caseFirst` option will do that, but only for the same letter (i.e. `'a'` and `'A'`, but won't work for `'a'` and `'B'`).

I did use it for the `ignoreCase` condition above though, since that will work. Let me know if you have any other thoughts.

ljharb: You're right, but neither will `<` - `localeCompare` is identical to the code you have here, except it's more correct for some non-english characters.

emroussel: Hmm, this is copy and pasted from the fix function for `sort-prop-types` and if I change it to use `localeCompare` the tests fail.

It also looks like `'a' < 'B'` is false (so would return `1` in this code), and `'a'.localeCompare('B')` returns -1, so they aren't equivalent.

ljharb: oh - then perhaps you want `bKey.localeCompare(aKey)`? (i always mix up the order)

emroussel: That works for the case above, but isn't what we want: `'a' < 'b'` is true (so would return `-1` in the code above), but `'b'.localeCompare('a')` returns `1`.

From looking online, it looks like `localeCompare` cannot sort with all uppercase letters first, so it's really not the same as `<` or `>` comparison when dealing with capital letters. Let me know if I've missed something though.

ljharb: hmm, ok - let's leave it as is for now then, and i'll look into it later :-)

emroussel: Sounds good, thanks! After doing a bit of research it looks like `localeCompare` indeed [cannot do case sensitive comparison](https://stackoverflow.com/questions/34861386/force-localecompare-to-be-case-sensitive#answer-34862951).

There's also a [compare function on the `Intl.Collator` object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Collator/compare) that is also meant to deal with strings of different locales, but it has a similar API to `localeCompare` and doesn't do case sensitive comparison either.

I wasn't able to find any other way than `<` and `>` since that uses the ASCII sort order (uppercase before lowercase in English), but not sure what that would mean for other languages. Let me know if I can do anything else to help, I'll keep digging for a little bit and let you know if I find anything useful.

ljharb: That SO link suggests a second argument to localeCompare would help, though

emroussel: The answer is a bit confusing, but it only does if you compare the same letter (`'a'` vs `'A'`), not for different ones. Around the middle of the answer, it says
> This test shows that switching from upper to lower does only affect the same letter, and does not sort uppercase-alphabet before lowercase or visa versa.

and 

> There might be a locale that sorts uppercase/lowercase differently, but I'm not aware of any...

Issue #2377, opened 2019-08-08
react/jsx-no-literals: Add config to allow a set of strings


benhollander:
I propose adding a config option to allow a specified list of strings to pass the `react/jsx-no-literals` rule. I'm finding myself writing code similar to the below pretty often.

## Semi-contrived examples

```javascript

// This would throw an error because of "&nbsp;-&nbsp;"
// Can't do {'&nbsp;-&nbsp;'} because it will be rendered as text instead of an HTML entity
export const DisplayRangeComponent = (props) => (
  <div>
    {/* eslint-disable react/jsx-no-literals */}
    {props.min}&nbsp;-&nbsp;{props.max}
    {/* eslint-enable react/jsx-no-literals */}
  </div>
);

// This would throw because of '*'
export const AddAsterisk = ({ children }) => (
  {/* eslint-disable react/jsx-no-literals */}
  <span>{children}*</span>
  {/* eslint-enable react/jsx-no-literals */}
);
```

It could be convenient (at least in my use-case, which doesn't seem like it would be all that uncommon) to pass in a config like 
```
rules: { 'react/jsx-no-literals': [ 'error', { ignoreStrings: [ '&nbsp;', '*' ] } ] }
```

ljharb:
Is there a reason you can't do: `{this.props.min}{'&nbsp;-&nbsp;'}{this.props.max}`?

benhollander:
> Is there a reason you can't do: `{this.props.min}{'&nbsp;-&nbsp;'}{this.props.max}`?

That will cause the `&nbsp;` to be displayed in the browser rather than be escaped.

I made a codepen to maybe illustrate the problem more clearly: https://codepen.io/bhollander/pen/zgmGyj

![image](https://user-images.githubusercontent.com/1233977/62807241-82a44d80-baba-11e9-92ff-24cbf3496e87.png)

ljharb:
ah right, then `&nbsp;{'-'}&nbsp;`?

benhollander:
That would still fail the rule :sweat_smile:

ljharb:
Gotcha.

In that case, allowing an array of strings to ignore seems reasonable.

PR #2436, opened 2019-10-02
[Fix] `jsx-no-literals`: trim whitespace for `allowedStrings` check

cainlevy:
The `allowedStrings` option for `jsx-no-literals` is checked against an untrimmed string containing potential spaces and newlines. This is at odds with the [error message shown](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/jsx-no-literals.js#L53) and substantially increases the difficulty of managing the `allowedStrings` list.

I propose trimming strings before checking inclusion in `allowedStrings`.

related:
* #2377 
* https://github.com/yannickcr/eslint-plugin-react/pull/2380#issuecomment-537597565

### DIFF
--- a/lib/rules/jsx-no-literals.js
+++ b/lib/rules/jsx-no-literals.js
@@ -40,10 +40,15 @@ module.exports = {
   },
 
   create(context) {
-    const isNoStrings = context.options[0] ? context.options[0].noStrings : false;
-    const allowedStrings = context.options[0] ? new Set(context.options[0].allowedStrings) : false;
+    function trimIfString(val) {
+      return typeof val === 'string' ? val.trim() : val;
+    }
 
-    const message = isNoStrings ?
+    const defaults = {noStrings: false, allowedStrings: []};
+    const config = Object.assign({}, defaults, context.options[0] || {});
+    config.allowedStrings = new Set(config.allowedStrings.map(trimIfString));
+
+    const message = config.noStrings ?
       'Strings not allowed in JSX files' :
       'Missing JSX expression container around literal string';
 
@@ -63,7 +68,7 @@ module.exports = {
     }
 
     function getValidation(node) {
-      if (allowedStrings && allowedStrings.has(node.value)) {
+      if (config.allowedStrings.has(trimIfString(node.value))) {
         return false;
       }
       const parent = getParentIgnoringBinaryExpressions(node);
@@ -71,7 +76,7 @@ module.exports = {
           typeof node.value === 'string' &&
           parent.type.indexOf('JSX') !== -1 &&
           parent.type !== 'JSXAttribute';
-      if (isNoStrings) {
+      if (config.noStrings) {
         return standard;
       }
       return standard && parent.type !== 'JSXExpressionContainer';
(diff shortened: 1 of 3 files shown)


tests/lib/rules/jsx-no-literals.js
```diff
@@ -226,7 +240,7 @@ ruleTester.run('jsx-no-literals', rule, {
           }
         }
       `,
-      options: [{noStrings: true, allowedStrings: ['foo: ', '*']}]
+      options: [{noStrings: true, allowedStrings: ['foo:', '*']}]
```
cainlevy: this is either a backwards incompatibility or a bugfix, depending on perspective. my vote is the latter based on the conflict with error messages.

ljharb: would passing in `foo: ` (with the space) no longer work?

Could/should we trim all the allowedStrings before checking?

cainlevy: ah, of course! running `allowedStrings` through trim would be a simple way to make this backwards compatible.

cainlevy: done

lib/rules/jsx-no-literals.js
```diff
@@ -40,8 +40,14 @@ module.exports = {
...
 
   create(context) {
+    function trimIfString(val) {
+      return typeof val === 'string' ? val.trim() : val;
+    }
+
     const isNoStrings = context.options[0] ? context.options[0].noStrings : false;
-    const allowedStrings = context.options[0] ? new Set(context.options[0].allowedStrings) : false;
+    const allowedStrings = context.options[0] ?
+      new Set((context.options[0].allowedStrings || []).map(trimIfString)) :
```
ljharb: why the ternary? just making this unconditionally be a Set simplifies the check below

tests/lib/rules/jsx-no-literals.js
```diff
@@ -218,6 +218,20 @@ ruleTester.run('jsx-no-literals', rule, {
...
+          render() {
+            return (
+              <div>
+                &nbsp;
+              </div>
+            );
+          }
+        }
+      `,
+      options: [{noStrings: true, allowedStrings: ['&nbsp;']}]
```
ljharb: could we add a few more tests, one with leading and trailing spaces on `allowedStrings`, and a different one with them on the rendered text?

cainlevy: apologies, i'm not sure how to action this request. could you say it another way?

for context, we happen to have a pre-existing test with spaces in `allowedStrings` that would fail if `allowedStrings` were not trimmed. here are all the combinations of spaces and not spaces with test coverage:

| tested | spaces in `allowedStrings` | spaces in rendered text |
| --- | --- | ---
| ✅  | ✅ | ✅  |
| ✅  | 🚫 | ✅ |
| ✅  | 🚫 | 🚫 |
| 🚫 | ✅ | 🚫 |

is that the test gap you'd like me to fix?

ljharb: Yes, exactly right :-)

ljharb: I'd also like a test with noStrings false, and an allowedStrings value :-)

cainlevy: ✅ I've made `noStrings: false` explicit in the existing test case

cainlevy:
possible to get this in a release soon?

thanks for all the work you put into maintaining this plugin! 🙇

PR #2438, opened 2019-10-02
Add section about overriding for trusted links

aschriner:


docs/rules/jsx-no-target-blank.md
```diff
@@ -74,6 +74,9 @@ var Hello = <Link target="_blank" to="/absolute/path/in/the/host"></Link>
 var Hello = <Link />
 ```
 
+## When To Override It
+For internal links, or links to a "safe" host, you may want to keep the HTTP Referer header for analytics purposes.
```
ljharb: can you elaborate on what you mean by "safe"?

aschriner: I updated with additional text.  To elaborate further on the reason for this change: this lint rule and the linked page make it sound like one should *never* use target="_blank" without noreferer and noopener.  However, as I understand, if you control the site to which you are linking, there is no vulnerability. And there are good reasons to keep the referer header (e.g. interal analytics/tracking).

Issue #2430, opened 2019-10-02
react/jsx-no-bind: Dealing with the state hook


kdex:
Let's assume that we have a functional (yet stateful) component like,

```js
const Button = () => {
	const [text, setText] = useState("Text A");
	const onClick = () => setText("Text B");
	return (
		<button onClick={ onClick }>{ text }</button>
	);
};
```

With `react/jsx-no-bind`, this use of `onClick` is forbidden, as it will create a new binding to `onClick` on every render.

We can't really move it out of the function either, like so:

```js
const onClick = setText => setText("Text B");
const Button = () => {
	const [text, setText] = useState("Text A");
	return (
		<button onClick={ onClick }>{ text }</button>
	);
};
```

Now, `onClick` has no chance to receive the correct `setText` parameter, as we can't control this without `.bind` or creating a new binding for `onClick` on every render by other means.

This makes me wonder if, despite its usefulness, React's state hook is just wasteful with memory to begin with, or whether I've overlooked something.

Is there a way to avoid this situation with the state hook?

kdex:
I'll be giving an answer to my own question. This is achievable by doing event handling manually. That is, use both React's `useEffect` as well as its `useRef` hook and declare the `Button` like so:
```js
const Button = () => {
	const reference = useRef();
	const [text, setText] = useState("Text A");
	useEffect(() => {
		const onClick = () => setText("Text B");
		reference.current.addEventListener("click", onClick);
		return () => {
			reference.current.removeEventListener("click", onClick);
		};
	}, []);
	return (
		<button ref={ reference }>{ text }</button>
	);
};
```

Unfortunately though, the code has become less readable as this involves quite a bit of boilerplate.

I'm leaving this issue open just so that, if the maintainers know of a cleaner way to do this, they can add another answer before this is closed. (Otherwise, just close it.)

kdex:
An even better solution that removes all boilerplate can be achieved by memoizing the function, which can be done using the [useCallback](https://reactjs.org/docs/hooks-reference.html#usecallback) hook.

draperunner:
@kdex Does `jsx-no-bind` stop complaining if you're using `useCallback`? If that's the case I think the [docs for this rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md) should be updated with hooks examples. It should probably be updated with some mention about function components anyhow.

kdex:
[...]

PR #2443, opened 2019-10-03
[docs] `jsx-no-bind`: add section about React Hooks

kdex:
This PR is a response to @ljharb's and @draperunner's call for documentation in #2430.

If there's still room for improvement, feel free to review. I'll squash any edits down to a single commit in case you want to rebase this.

### DIFF
--- a/docs/rules/jsx-no-bind.md
+++ b/docs/rules/jsx-no-bind.md
@@ -37,7 +37,7 @@ When `true` the following are **not** considered warnings:
 ```jsx
 <div onClick={this._handleClick.bind(this) />
 <span onClick={() => console.log("Hello!")} />
-<button onClick={function() { alert("1337") }} />
+<button type="button" onClick={function() { alert("1337") }} />
 ```
 
 ### `ignoreRefs`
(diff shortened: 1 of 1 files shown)


docs/rules/jsx-no-bind.md
```diff
@@ -151,6 +151,35 @@ class Foo extends React.Component {
...
+```
+
+Otherwise, the idiomatic way to avoid redefining callbacks on every render would be to memoize them using the [`useCallback`](https://reactjs.org/docs/hooks-reference.html#usecallback) hook:
+
+```jsx
+const Button = () => {
+  const [text, setText] = useState("Before click");
+  const onClick = useCallback(() => {
+    setText("After click");
+  }, [text]); // Array of dependencies for which the memoization should update
```
ljharb: since `setText` is what's called, that's the dependency here
```suggestion
  }, [setText]); // Array of dependencies for which the memoization should update
```

docs/rules/jsx-no-bind.md
```diff
@@ -151,6 +151,35 @@ class Foo extends React.Component {
...
+  return (
+    <button onClick={onClick}>{text}</button>
```
ljharb:
```suggestion
    <button type="button" onClick={onClick}>{text}</button>
```
no button should ever lack a `type` :-)

docs/rules/jsx-no-bind.md
```diff
@@ -151,6 +151,35 @@ class Foo extends React.Component {
...
+
+Functional components are often used alongside hooks, and the most trivial case would occur if your callback is completely independent from your state. In this case, the solution is as simple as moving the callback out of your component:
+
+```jsx
+const onClick = () => {
+  console.log("Independent callback");
+};
+const Button = () => {
+  return (
+    <button onClick={onClick}>Label</button>
```
ljharb:
```suggestion
    <button type="button" onClick={onClick}>Label</button>
```

docs/rules/jsx-no-bind.md
kdex: Hah, good catch. I knew I should have run this. :)

docs/rules/jsx-no-bind.md
kdex: Agreed. I was thinking about putting a `type` there, but decided not to, as it had also been omitted in line 40 of the same file, where it says:

```jsx
```

I hope it's okay if I add a `type` attribute there as well for consistency.

PR #2446, opened 2019-10-05
[Fix] Support Flow type spread

moroine:
This PR aims to support type spread like:

```typescript
// @flow

type DefaultProps = {|
  foo: number,
|}

type Props = {
  bar: string,
};

function MyComponent(props: Props) {}

MyComponent.defaultProps = {
  foo: 42,
};
```

This PR also fixes https://github.com/yannickcr/eslint-plugin-react/issues/2138

### DIFF
--- a/lib/util/propTypes.js
+++ b/lib/util/propTypes.js
@@ -32,12 +32,18 @@ function isSuperTypeParameterPropsDeclaration(node) {
  * @param {Object[]} properties Array of properties to iterate.
  * @param {Function} fn Function to call on each property, receives property key
     and property value. (key, value) => void
+  * @param {Function} [handleSpreadFn] Function to call on each ObjectTypeSpreadProperty, receives the
+    argument
  */
-function iterateProperties(context, properties, fn) {
+function iterateProperties(context, properties, fn, handleSpreadFn) {
   if (properties && properties.length && typeof fn === 'function') {
     for (let i = 0, j = properties.length; i < j; i++) {
       const node = properties[i];
       const key = getKeyValue(context, node);
+
+      if (node.type === 'ObjectTypeSpreadProperty' && typeof handleSpreadFn === 'function') {
+        handleSpreadFn(node.argument);
+      }
 
       const value = node.value;
       fn(key, value, node);
@@ -121,7 +127,8 @@ module.exports = function propTypesInstructions(context, components, utils) {
     },
 
     ObjectTypeAnnotation(annotation, parentName, seen) {
-      let containsObjectTypeSpread = false;
+      let containsUnresolvedObjectTypeSpread = false;
+      let containsSpread = false;
       const containsIndexers = Boolean(annotation.indexers && annotation.indexers.length);
       const shapeTypeDefinition = {
         type: 'shape',
@@ -129,9 +136,7 @@ module.exports = function propTypesInstructions(context, components, utils) {
       };
       iterateProperties(context, annotation.properties, (childKey, childValue, propNode) => {
         const fullName = [parentName, childKey].join('.');
-        if (!childKey && !childValue) {
-          containsObjectTypeSpread = true;
-        } else {
+        if (childKey || childValue) {
           const types = buildTypeAnnotationDeclarationTypes(childValue, fullName, seen);
           types.fullName = fullName;
           types.name = childKey;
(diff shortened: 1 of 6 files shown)


lib/util/propTypes.js
```diff
@@ -121,29 +127,36 @@ module.exports = function propTypesInstructions(context, components, utils) {
...
+        if (!types.children) {
+          containsUnresolvedObjectTypeSpread = true;
+        } else {
+          Object.assign(shapeTypeDefinition, types.children);
+        }
       });
 
       // Mark if this shape has spread or an indexer. We will know to consider all props from this shape as having propTypes,
       // but still have the ability to detect unused children of this shape.
-      shapeTypeDefinition.containsSpread = containsObjectTypeSpread;
```
ljharb: this might be a breaking change for anyone using the utility from outside. can we keep both?

moroine: @ljharb updated

moroine:
What's your opinion about trying to resolve types from external files, or add an option to warn when type can't be resolved

ljharb:
@moroine i'm not a flow user, so i'm not sure. what do you think?

moroine:
@ljharb I think it can be a good idea to warn first, as I discover that limitation only when reading the code.

tests/lib/rules/prop-types.js
```diff
@@ -2335,6 +2352,30 @@ ruleTester.run('prop-types', rule, {
...
+
+        type Props = {| ...UsedProps, ...UnusedProps |};
+
+        function MyComponent({ usedProp }: Props) {
+          return <div>{usedProp}</div>;
+        }
+      `,
+      parser: parsers.BABEL_ESLINT,
+      errors: [{
+        message: "'notOne' is missing in props validation",
```
ghiculescu: possibly a dumb question, but why is this erroring? `notOne` doesn't appear anywhere else in this test?

levenecav:
After this update, it stopped working for me. Before that, everything was fine.

`type PropsState = {|
    brokenExchangeConnection: boolean
|};

type PropsActions = {|
    updateUserParams: (string, StreamingUserParamsItem) => StreamingUpdateUserParamsAction
|};

type PropsParent = {|
    onTenorChange: (SettingsStreamerAccountsPair, StreamingUserParamsItem) => void
|};

type Props = {|
    ...PropsState,
    ...PropsActions,
    ...PropsParent
|};`

`51:31  error  'brokenExchangeConnection' PropType is defined but prop is never used  react/no-unused-prop-types
57:23  error  'updateUserParams' PropType is defined but prop is never used          react/no-unused-prop-types
70:20  error  'onTenorChange' PropType is defined but prop is never used             react/no-unused-prop-types`

moroine:
@levenecav could we have the code of the component?

levenecav:
He is big, but the essence of this:

```

import ...
import ...
import ...

type PropsState = {|
    brokenExchangeConnection: boolean
|};

type PropsActions = {|
    updateUserParams: (string, StreamingUserParamsItem) => StreamingUpdateUserParamsAction
|};

type PropsParent = {|
    onTenorChange: (SettingsStreamerAccountsPair, StreamingUserParamsItem) => void
|};

type Props = {|
    ...PropsState,
    ...PropsActions,
    ...PropsParent
|};

class Instrument extends PureComponent<Props> {
    render(): Node {
        return (
            <div
                data-marker={currency}
                className={styles.root}>
                {this.renderBody()}
            </div>
        );
    }

    renderBody(): Node {
        const { brokenExchangeConnection } = this.props;
        
        return (
            <div className={className}>
                <InstrumentErrorMessage
                    brokenExchangeConnection={brokenExchangeConnection}
                    onUpdateUserParams={this.handleUpdateUserParams}
                    onTenorChange={this.handleTenorChange} />
            </div>
        );
    }

    handleUpdateUserParams = () => {
        this.props.updateUserParams('id', { id: 'id', name: 'name' });
    };

    handleTenorChange = () => {
        this.props.onTenorChange('USD/RUB', { id: 'id', name: 'name' });
    };
}

export default SortableElement(connect(
    (state: StoreState, props: Props): PropsState => {
        const { brokenExchangeConnection } = state;

        return { brokenExchangeConnection };
    },
    ({
        updateUserParams: Actions.streaming.updateUserParams
    }: PropsActions)
)(Instrument));

**Comment by levenecav** [comment]
I use flow for types.

**Comment by moroine** [comment]
If you use the following props do you have the same error? Because this PR is only about converting spread to the following props. Here I suspect the rule not finding usage of props in other methods

```
type Props = {|
    brokenExchangeConnection: boolean,
    updateUserParams: (string, StreamingUserParamsItem) => StreamingUpdateUserParamsAction,
    onTenorChange: (SettingsStreamerAccountsPair, StreamingUserParamsItem) => void,
|};
```

File: tests/lib/rules/prop-types.js (hunk 2 above)
Comment (moroine) [reply tests/lib/rules/prop-types.js]: @ghiculescu it is at line 4791

**Comment by levenecav** [comment]
I have exactly the same problem as in # 2138.
As far as I understand, this PR should solve this problem.

**Comment by moroine** [comment]
I'm sure your issue is caused by this PR. But this should be because now spreading is recognized as being 

```
type Props = {|
    brokenExchangeConnection: boolean,
    updateUserParams: (string, StreamingUserParamsItem) => StreamingUpdateUserParamsAction,
    onTenorChange: (SettingsStreamerAccountsPair, StreamingUserParamsItem) => void,
|};
```

Could you check using Props without spreading with v7.16.0?

I suspect the error is due not used Props in the following function:

```
(state: StoreState, props: Props): PropsState => {
        const { brokenExchangeConnection } = state;

        return { brokenExchangeConnection };
    }
```

ljharb:
@levenecav Please file a new issue.

levenecav:
@moroine 
Yes, you're right, I combined all the props and the problem remains.
It turned out that the problem arises because getDerivedStateFromProps has a call to a function that is on the same level as the component class.
If this function is put into import, then everything is fine.
I will file a new issue, thanks.

ljharb:
Passing the props object around is an antipattern; always de structure what you need immediately.

tests/lib/rules/prop-types.js
joeyparis: @moroine I'm not sure I understand. Why is a separate test causing this test to fail? Aren't they each independent tests?

PR #2452, opened 2019-10-07
Add contextType to lifecycle group in sort-comp

webkonstantin:


lib/rules/sort-comp.js
```diff
@@ -24,6 +24,7 @@ const defaultConfig = {
     lifecycle: [
       'displayName',
       'propTypes',
+      'contextType',
```
ljharb: adding this to the default is a breaking change.

PR #2459, opened 2019-10-11
[fix] Allow SCU to be decalred as arrow function

niki-timofe:


tests/lib/rules/require-optimization.js
```diff
@@ -131,10 +131,19 @@ ruleTester.run('react-require-optimization', rule, {
...
-    errors: [{
-      message: MESSAGE
-    }]
+    parser: parsers.BABEL_ESLINT
+  }, {
+    code: `
+    import React from "react";
+    class YourComponent extends React.Component {
+      handleClick = () => {}
+      shouldComponentUpdate = () => true;
```
ljharb: why is this desirable? lifecycle methods should never be implemented in this way, only as proper intance methods.

niki-timofe: Ok, I've figured out, that it's not desirable.

niki-timofe:
Makes no sense as described in https://github.com/facebook/react/issues/10810#issuecomment-332067094

PR #2463, opened 2019-10-13
[Docs] `props-no-multi-spaces`: suggest using core rule instead

golopot:


ljharb:
This is not accurate. The core rules do not provide sufficient jsx indentation control; the proper thing to do is disable the eslint core rule for all jsx, as the airbnb config does, and rely on this plugin to handle all jsx indentation.

golopot:
I believe the capability of the core rule `indent` is in par with `jsx-indent` and `jsx-indent-props`, except for the option `{indentLogicalExpressions: true}`. Can you give provide some examples where the core rule is insufficient? Configuring indent to ignore jsx is very unergonomic.

ljharb:
Not off the top of my head, but the airbnb config ignores jsx here: https://github.com/airbnb/javascript/blob/ab72ab9e90f403e90590f8815e0563248af10470/packages/eslint-config-airbnb-base/rules/style.js#L146 and sets all the react rules here: https://github.com/airbnb/javascript/blob/ab72ab9e90f403e90590f8815e0563248af10470/packages/eslint-config-airbnb/rules/react.js

If in fact *every single permutation* of our rules are possible with core, then this is fine to merge, but until then, the core indent rule should not be used with jsx whatsoever.

golopot:
The core rule does not offer compatibility with _every single permutation_ of our indent rules,  so I will drop that part. I amended this pr to only deal with `props-no-multi-spaces`.

PR #2469, opened 2019-10-20
[eslint] enable eslint-plugin-eslint-plugin

golopot:
Most of the changes are deleting duplicate test cases.


billyjanitsch:
@ljharb shouldn't this have been added as a dev dependency, rather than as a dependency?

ljharb:
yes, it should. thanks for the catch (#2510)

PR #2496, opened 2019-11-15
[New] `jsx-no-literals`: add `noAttributeStrings` option

### LINKED ISSUE #2486: support jsx-no-literals on specific attributes
For `react/jsx-no-literals` the default likely should not allow string literals on "title" and "alt" attributes, but allow configuration for component patterns that pass user strings in other ways.

tanmoyopenroot:
#### PR checklist

- [x] Addresses an existing issue: fixes #2486
- [x] enhancement
- [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Check for `JSXExpressionContainer` in `JSXAttribute`

##### Is there anything you'd like reviewers to focus on?
This will result in `break-in changes` because by default we are checking for  `JSXExpressionContainer` in `JSXAttribute`

### DIFF
--- a/tests/lib/rules/jsx-no-literals.js
+++ b/tests/lib/rules/jsx-no-literals.js
@@ -39,6 +39,10 @@ function invalidProp(str) {
   return `Invalid prop value: “${str}”`;
 }
 
+function attributeMessage(str) {
+  return `Strings not allowed in attributes: “${str}”`;
+}
+
 const ruleTester = new RuleTester({parserOptions});
 ruleTester.run('jsx-no-literals', rule, {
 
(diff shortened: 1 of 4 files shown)


tests/lib/rules/jsx-no-literals.js
```diff
@@ -128,37 +128,42 @@ ruleTester.run('jsx-no-literals', rule, {
       parser: parsers.BABEL_ESLINT
     }, {
       code: `
-        <Foo bar='test'>
+        <Foo bar={'test'}>
```
ljharb: changing any "valid" tests makes this a breaking change. i think this needs to be behind an option that's disabled by default, so that this PR does not touch any "valid" tests, only adds ones (and invalid ones too).

ljharb (approved):
Completed via #2782.

PR #2505, opened 2019-11-23
Disable --fix for sort-prop-types

webOS101:
This addresses #1940 by disabling the fix option for `sort-prop-types` until a better fix to copy the comments is available.

### DIFF
--- a/lib/rules/sort-prop-types.js
+++ b/lib/rules/sort-prop-types.js
@@ -8,7 +8,7 @@ const variableUtil = require('../util/variable');
 const propsUtil = require('../util/props');
 const docsUrl = require('../util/docsUrl');
 const propWrapperUtil = require('../util/propWrapper');
-const propTypesSortUtil = require('../util/propTypesSort');
+// const propTypesSortUtil = require('../util/propTypesSort');
 
 // ------------------------------------------------------------------------------
 // Rule Definition
@@ -23,7 +23,7 @@ module.exports = {
       url: docsUrl('sort-prop-types')
     },
 
-    fixable: 'code',
+    // fixable: 'code',
 
     schema: [{
       type: 'object',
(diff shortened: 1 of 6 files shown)


webOS101:
I'd hoped to keep this PR simple, but it required removing quite a bit of code.  Also, there are some failing Travis tests that seem unrelated to this PR.  Finally, not sure why, but I can't seem to execute tests locally.

lib/rules/sort-prop-types.js
```diff
@@ -98,18 +95,6 @@ module.exports = {
         return;
       }
 
-      function fix(fixer) {
```
ljharb: could we keep this as commented code, and do something like `const fix = undefined;` or something, so that all the rest of the code passing `fix` doesn't have to change?

tests/lib/rules/sort-prop-types.js
```diff
@@ -486,18 +486,7 @@ ruleTester.run('sort-prop-types', rule, {
       line: 4,
       column: 5,
       type: 'Property'
-    }],
-    output: [
```
ljharb: similarly i wonder if we could comment out instead of remove all these `output`s - since they're all correct for when we re-enable the fixer.

lib/rules/sort-prop-types.js
webOS101: I thought it might be easier to make this easy to revert to restore the missing code when desired.  I tend to not like to keep large swaths of commented out code around, but I can do that if desired.

webOS101:
I can add the failing test case, commented out.  As mentioned, I was keeping this easy to revert.  I can either add the failing test here or in another PR, or I if you do want me to go the comment-out route, I can easily add the test in.  I jus remembered that `sort-default-props` also uses the same code and probably suffers the same problem, I didn't test it though.

webOS101:
@ljharb I went ahead and changed the removals to comments and added failing (but commented out) tests for sort with comments.  I did not yet disable `--fix` for `jsx-sort-default-props` but think I should probably do that as well in this same PR.  Agree?

lib/rules/sort-prop-types.js
ljharb: generally i'd agree, but in this case i think it'd be preferable.

ljharb:
Yes, I agree. Thanks!

webOS101:
@ljharb Disabled fix for `jsx-sort-default-props`.  I _think_ this is more-or-less ready to go.

ljharb:
Looks like there's some tests failing on master :-/

PR #2513, opened 2019-12-05
[Fix] raise an error when a method name is in string form

shijistar:
I got a compiling error when a component method name is in string form, but not an identifier. The sample code is,
```javascript
class MyComponent {
    "special-method-name"() {
        // ...
    }
}
```
Normally method names should be valid identifiers, but for some reason, I need one of my methods in a special form. However any strings should be valid method name and both babel and Chrome run without any problem.

The difference between `Identifier` and `String` method is that the `node.key.type` is `Identifier`/`Literal`. So I made a special process here.

### DIFF
--- a/lib/rules/no-typos.js
+++ b/lib/rules/no-typos.js
@@ -144,7 +144,11 @@ module.exports = {
 
     function reportErrorIfLifecycleMethodCasingTypo(node) {
       LIFECYCLE_METHODS.forEach((method) => {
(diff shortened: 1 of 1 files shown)


ljharb:
Duplicate of #2514. it seems.

shijistar:
> Duplicate of #2514. it seems.

Yes, it's duplicated, and this PR should be ignored.

PR #2514, opened 2019-12-06
[Fix] Compilation error when method name is string instead of identifier

shijistar:
I got a compilation error when a component method name is string instead of identifier. The sample code is,
```javascript
class MyComponent {
    "special-method-name"() {
        // ...
    }
}
```
Normally method names should be valid identifiers. However for some reason, I need one of my methods must be in a special form. `Eslint` raises an error, but babel goes well and the code works well in chrome too.

The difference between `Identifier` and `String` method is that the `node.key.type` is `Identifier`/`Literal`. If method name is `Identifier`, the `node.key.name` is the identifier name, however if method name is `String`, we should get the method name by `node.key.value`.

### DIFF
--- a/lib/rules/no-typos.js
+++ b/lib/rules/no-typos.js
@@ -144,7 +144,11 @@ module.exports = {
 
     function reportErrorIfLifecycleMethodCasingTypo(node) {
       LIFECYCLE_METHODS.forEach((method) => {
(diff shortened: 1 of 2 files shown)


shijistar:
> Thanks! Can you please provide a regression test?

Regression test added.

PR #2523, opened 2019-12-13
Fix `curly-brace-presence` edge cases

rafbgarcia:
#### Formats texts that contain line breaks

Fixes https://github.com/yannickcr/eslint-plugin-react/issues/2524

The example below was not being handled.

```jsx
<Component>
  text
</Component>
```

#### Handle HTML entities

Fixes https://github.com/yannickcr/eslint-plugin-react/issues/2522

- Ignores HTML entities
- Correctly handle their output

### DIFF
--- a/lib/rules/jsx-curly-brace-presence.js
+++ b/lib/rules/jsx-curly-brace-presence.js
@@ -62,6 +62,7 @@ module.exports = {
   },
 
   create(context) {
+    const HTML_ENTITY_REGEX = () => /&[A-Za-z\d#]+;/g;
     const ruleOptions = context.options[0];
     const userConfig = typeof ruleOptions === 'string' ?
       {props: ruleOptions, children: ruleOptions} :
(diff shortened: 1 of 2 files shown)


rafbgarcia:
@ljharb what do you think of this?

About the HTML entities, I did what I think would be the best solution for that. If we did `{"&nbsp;"}` my code would break and I would not use that lint.

Please let me know if you the fixes on separate PRs.

rafbgarcia:
@ljharb kind ping as, again, this bug prevents using this lint which is one of the best. I appreciate your help, thank you!

lib/rules/jsx-curly-brace-presence.js
```diff
@@ -62,6 +62,7 @@ module.exports = {
   },
 
   create(context) {
+    const HTML_ENTITY_REGEX = /&[A-Za-z\d#]+;/g;
```
ljharb: global regexes are stateful when used with `test`; we should make this a function that always returns a new one, if it's global.

lib/rules/jsx-curly-brace-presence.js
```diff
@@ -111,6 +116,42 @@ module.exports = {
...
+      if (!containsLineTerminators(rawText)) {
+        return `{${JSON.stringify(rawText)}}`;
+      }
+
+      return rawText.split('\n').map((line) => {
+        if (line.trim() === '') {
+          return line;
+        }
+        const firstCharIndex = line.search(/[^\s]/);
+        const leftWhitespace = line.substring(0, firstCharIndex);
```
ljharb: `slice` should always be preferred over `substring` or `substr`

PR #2532, opened 2019-12-24
feat(require-default-props): add option to ignore functional components

RedTn:
Addresses the discussion from https://github.com/yannickcr/eslint-plugin-react/issues/1009
Specifically from the recommendation at: https://github.com/yannickcr/eslint-plugin-react/issues/1009#issuecomment-516485145 to 'warn less'.

This PR adds an option to `require-default-props` to ignore functional components, as react will soon deprecate defaultProps on functional components https://github.com/reactjs/rfcs/pull/107.

### DIFF
--- a/docs/rules/require-default-props.md
+++ b/docs/rules/require-default-props.md
@@ -188,12 +188,13 @@ NotAComponent.propTypes = {
 
 ```js
 ...
-"react/require-default-props": [<enabled>, { forbidDefaultForRequired: <boolean> }]
+"react/require-default-props": [<enabled>, { forbidDefaultForRequired: <boolean>, ignoreFunctionalComponents: <boolean> }]
 ...
 ```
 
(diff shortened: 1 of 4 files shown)


docs/rules/require-default-props.md
```diff
@@ -269,6 +270,67 @@ MyStatelessComponent.propTypes = {
 };
 ```
 
+### `ignoreFunctionalComponents`
+
+Ignores this rule for all functional components.
```
ljharb:
```suggestion
When set to `true`, ignores this rule for all functional components.
```

RedTn:
> Let's also add some tests for `export function Foo() {}` and `export const Foo = () => {}` and `export default function Foo() {}`, and let's make sure all these tests run on all the parsers.

to which rules specifically? these `exports` tests, all of my new tests, or all tests in this file? i assumed the first option

ljharb:
The first option, yes :-)

PR #2533, opened 2019-12-24
[Fix]: Does not validate for LogicalExpression

### LINKED ISSUE #2526: react/prop-types does not validate missing propTypes for functions in version 7.17.0
Seems like this is still an issue in version 7.17.0

Missing prop types do not raise errors in a function component with a return of:

```
return (
    ordering.length > 0 && (
      <Paginator items={ordering} pageSize={10} />
    )
)
```

But do with a function component with a return of:

```
  if (ordering.length > 0) {
    return false;
  }

  return (
    <Paginator items={ordering} pageSize={10} />
  )
```

and:

```
  return ordering.length > 0 ? (
    <Paginator items={ordering} pageSize={10} />
  ) : null;
```
_Originally posted by @johnarnfield in https://github.com/yannickcr/eslint-plugin-react/issues/2134#issuecomment-566035325_

toshi-toma:
Fixes #2526.

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -46,6 +46,30 @@ function mergeUsedPropTypes(propsList, newPropsList) {
   return propsList.concat(propsToAdd);
 }
 
+function isReturnsConditionalJSX(node, property, strict) {
+  const returnsConditionalJSXConsequent = node[property] &&
+    node[property].type === 'ConditionalExpression' &&
+    jsxUtil.isJSX(node[property].consequent);
+  const returnsConditionalJSXAlternate = node[property] &&
+    node[property].type === 'ConditionalExpression' &&
+    jsxUtil.isJSX(node[property].alternate);
+  return strict ?
+    (returnsConditionalJSXConsequent && returnsConditionalJSXAlternate) :
+    (returnsConditionalJSXConsequent || returnsConditionalJSXAlternate);
+}
+
+function isReturnsLogicalJSX(node, property, strict) {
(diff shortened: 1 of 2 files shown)


lib/util/Components.js
```diff
@@ -373,22 +373,40 @@ function componentRule(rule, context) {
...
         node[property].type === 'ConditionalExpression' &&
         jsxUtil.isJSX(node[property].consequent);
-      const returnsConditionalJSXAlternate = node[property] &&
+        const returnsConditionalJSXAlternate = node[property] &&
         node[property].type === 'ConditionalExpression' &&
         jsxUtil.isJSX(node[property].alternate);
-      const returnsConditionalJSX = strict ?
-        (returnsConditionalJSXConsequent && returnsConditionalJSXAlternate) :
-        (returnsConditionalJSXConsequent || returnsConditionalJSXAlternate);
+        return strict ?
```
toshi-toma: I think `strict` property not used.
I should probably remove `strict` property from `isReturningJSX`.

lib/util/Components.js
```diff
@@ -373,22 +373,40 @@ function componentRule(rule, context) {
         return false;
       }
 
-      const returnsConditionalJSXConsequent = node[property] &&
+      const isReturnsConditionalJSX = () => {
```
ljharb: i'll hoist these functions out so they don't need to be defined every time.

toshi-toma:
Thanks!!

PR #2536, opened 2019-12-30
[Fix]: false positives for `no-unused-...` when using TS type assertions

### LINKED ISSUE #2517: no-unused-prop-types returns false positive when destructuring props cast to a different type
# Sample Code

```tsx
const propTypes = {
  foo: PropTypes.number.isRequired,
};

type Props = PropTypes.InferProps<typeof propTypes>;

const Foo: React.FC<Props> = (props) => {
  const {
    foo,
  } = props as any; // works if `as any` assertion is removed

  return (
    <>{foo}</>
  );
};
```

# Issue

The linter throws an error: `'foo' PropType is defined but prop is never used` even tho the prop is used.

# Expected

The linter should not care about props being asserted to be of a different type. If an assertion is made it's most likely for a good reason and the actual prop is still referenced inside the component (no matter its type).

kdmadej:
Fixes #2517:
* adds an ast util for unwrapping TS `as` assertion.
* fixes false positives for `no-unused-prop-types`
* fixes false positives for `no-unused-state`

I tried to reuse as much of the existing solution as possible by handling cases where the `as` expression might occur.

I'm wondering tho if it would be better to just process the code with anything that would remove TS-specific syntax (type annotations, `as` assertions, etc.) and then parse the result so the entire logic could be reused. Just thinking out loud 😅

### DIFF
--- a/tests/lib/rules/no-unused-prop-types.js
+++ b/tests/lib/rules/no-unused-prop-types.js
@@ -3125,8 +3125,7 @@ ruleTester.run('no-unused-prop-types', rule, {
         }
       `,
       parser: parsers.BABEL_ESLINT
-    },
-    {
+    }, {
       code: `
         export default class Foo extends React.Component {
           render() {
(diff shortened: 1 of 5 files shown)


tests/lib/rules/no-unused-prop-types.js
```diff
@@ -3136,6 +3135,73 @@ ruleTester.run('no-unused-prop-types', rule, {
...
+            thisPropsAliasProp,
+            thisDestructPropsAliasDestructProp,
+            thisDestructPropsAliasProp,
+            thisDestructPropsDestructProp,
+            thisPropsDestructProp,
+            thisPropsProp,
+          };
+
+          componentDidUpdate(prevProps) {
+            (prevProps as any).prevProp as any;
```
ljharb: can we use `unknown` instead of `any` in all these tests?

kdmadej: Sure thing. Not sure it changes anything in this case but I'm more then fine with spreading the `unknown` 😄

tests/lib/rules/no-unused-prop-types.js
```diff
@@ -3136,6 +3135,73 @@ ruleTester.run('no-unused-prop-types', rule, {
 
         Foo.defaultProps = Object.assign({});
       `
+    }, {
+      code: `
+        const Foo = (props) => {
+          const { foo } = props as unknown;
+          (props as unknown).bar as unknown;
```
ljharb: could we also add this same component as a failing test, with only one of these two propTypes present?

tests/lib/rules/no-unused-prop-types.js
```diff
@@ -3136,6 +3135,73 @@ ruleTester.run('no-unused-prop-types', rule, {
...
+        Foo.propTypes = {
+          foo,
+          bar,
+        };
+      `,
+      parser: parsers.TYPESCRIPT_ESLINT
+    }, {
+      code: `
+        class Foo extends React.Component {
+          static propTypes = {
```
ljharb: same here; it'd be great to get this test also as a failing one, with some propTypes present and some not.

PR #2541, opened 2020-01-13
[Docs] Auto-generate list of rules in README

ybiquitous:
This adds a script to auto-generate the list of rules and embeds special tags to README.

Usage:

    $ npm run generate-list-of-rules

Check whether the auto-generated changes are committed:

    $ npm run generate-list-of-rules:check

See also #1742

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,9 @@
     "pretest": "npm run lint",
     "test": "npm run unit-test",
     "type-check": "tsc",
-    "unit-test": "istanbul cover node_modules/mocha/bin/_mocha tests/lib/**/*.js tests/util/**/*.js tests/index.js"
+    "unit-test": "istanbul cover node_modules/mocha/bin/_mocha tests/lib/**/*.js tests/util/**/*.js tests/index.js",
+    "generate-list-of-rules": "md-magic --path README.md",
+    "generate-list-of-rules:check": "npm run generate-list-of-rules && git diff --exit-code README.md"
   },
   "files": [
     "LICENSE",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -11,7 +11,9 @@
     "pretest": "npm run lint",
     "test": "npm run unit-test",
     "type-check": "tsc",
-    "unit-test": "istanbul cover node_modules/mocha/bin/_mocha tests/lib/**/*.js tests/util/**/*.js tests/index.js"
+    "unit-test": "istanbul cover node_modules/mocha/bin/_mocha tests/lib/**/*.js tests/util/**/*.js tests/index.js",
+    "generate-list-of-rules": "./scripts/generate-list-of-rules.js README.md",
```
ljharb: this would be great as a published package, so we can use it in eslint-plugin-react and eslint-plugin-jsx-a11y as well :-D (but we should wait until this PR lands to do that)

scripts/generate-list-of-rules.js
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+const targetFile = process.argv[2];
+const targetFileContent = fs.readFileSync(targetFile, 'utf8');
+
+const NOTICE = '<!-- THIS SECTION IS AUTO-GENERATED. DO NOT EDIT. -->';
+const TAG_BASIC_RULES = '<!-- BASIC_RULES_SECTION -->';
+const TAG_JSX_RULES = '<!-- JSX_RULES_SECTION -->';
+
+const newContent = targetFileContent.replace(
```
ljharb: perhaps it would be worth using something like https://www.npmjs.com/package/markdown-magic rather than messing with regexes?

ybiquitous: Looks good! I will try it on this PR later, thanks! 👍

ybiquitous:
I've synced this PR with #1742, so closing this.

scripts/generate-list-of-rules.js
ybiquitous: I will try it with #1742.

ljharb:
To be clear, i need both to stay open until both are merged :-)

ybiquitous:
Sorry, I see 👍

PR #2542, opened 2020-01-14
[Fix] `jsx-indent`: Does not check indents for JSXText

### LINKED ISSUE #1136: react/jsx-indent not working when using string literal as a child
eslint 3.19.0, eslint-plugin-react 6.10.3

setting:
```
rules
  react/jsx-indent: [ error, 2 ]
```

minimal test using a string literal which causes no error:
```
(
  <div>
  1
  <div>Car</div>
  </div>
)
```

expected this to be correct:
```
(
  <div>
    1
    <div>Car</div>
  </div>
)
```

though the error is correctly thrown when you are using an expression:
```
(
  <div>
  {`1`}
  <div>Car</div>
  </div>
)
```

```
22:3  error  Expected indentation of 4 space characters but found 2  react/jsx-indent
23:3  error  Expected indentation of 4 space characters but found 2  react/jsx-indent
```

### LINKED ISSUE #2467: react/jsx-one-expression-per-line
Similar to #1775 ...

The bug with the `--fix` part of this rule is still there, and the PR did not fix it.

**Rules**
```JSON
"react/jsx-one-expression-per-line": "error",
"react/jsx-indent": [
     "error",
      4
],
```

**Original Code**
```jsx
return (
    <div>text</div>
);
```

**Actual (after fix)**
```jsx
return (
    <div>
text
    </div>
);
```

**Expected**
```jsx
return (
    <div>
        text
    </div>
);
```

### LINKED ISSUE #2484: jsx + eslint --fix not properly spacing text
My application is a node app that server side renders React components to strings.  Whenever I run `eslint --fix`, it fixes spacing correctly except for plain text in jsx.  For example, after running eslint --fix here is what some of my jsx looks like:
```
    myAccount: {
        option1: <>
Shop Smarter,
            <br />
Earn Rewards
        </>,
        option2: 'No Interest Financing',
        option3: 'No Credit Needed',
        option4: 'Shop Smart',
        option5: <>
You’re still
            <br />
pre-approved!
        </>
    },
```
html entities are spaced just fine while text always has no indentation.  This happens almost everywhere in my repo and even my other repos.  Please help.

Node version: 10.16.3 or 12.13.0
yarn version: 1.17.3 or 1.19.1
package.json:
```
    "@babel/cli": "^7.6.4",
    "@babel/core": "^7.6.4",
    "@babel/plugin-transform-react-jsx": "^7.3.0",
    "@babel/preset-env": "^7.6.3",
    "babel-eslint": "^10.0.2",
    "babel-preset-minify": "^0.5.1",

    "eslint": "^6.6.0",
    "eslint-plugin-react": "^7.15.1",

    "react": "^16.11.0",
    "react-dom": "^16.11.0",
```

Additional eslint config:
```

    "env": {
        "es6": true,
        "jest": true,
        "node": true
    },
    "extends": "airbnb",
    "parser": "babel-eslint",
    "parserOptions": {
        "ecmaFeatures": {
            "jsx": true
        },
        "ecmaVersion": 6,
        "sourceType": "module"
    },
    "plugins": [
        "react",
        "react-hooks",
        "jsx-a11y",
        "import"
    ],
    "rules": {
        "max-len": ["error", 120, 4, {
            ignoreUrls: true,
            ignoreComments: false,
            ignoreRegExpLiterals: true,
            ignoreStrings: true,
            ignoreTemplateLiterals: true,
        }],
        "comma-dangle": 0,
        "no-underscore-dangle": 0,
        "import/prefer-default-export": 0,
        "indent": ["error", 4, {
            "SwitchCase": 1
        }],
        "react-hooks/rules-of-hooks": "error",
        "react-hooks/exhaustive-deps": "warn",
        "react/jsx-filename-extension": 0,
        "react/jsx-indent": ["error", 4],
        "react/jsx-indent-props": ["error", 4],
        "jsx-a11y/no-static-element-interactions": 0,

        "array-callback-return": 0,
        "consistent-return": 0,
        "import/no-extraneous-dependencies": 0,
        "max-len": 0,
        "no-console": 0,
        "no-plusplus": 0,
        "react/jsx-closing-tag-location": 0,
        "react/jsx-curly-brace-presence": 0,
        "react/jsx-uses-react": 2,
        "react/jsx-uses-vars": 2,
        "react/prop-types": 0,
        "react/no-danger": 0
    }
```

toshi-toma:
Fixes #2467, Fixes #2484, Fixes #1136 
This PR fixes the problem of does not check `jsx-indent` for `JSXText`.

`JSXText` Node is different from other formats.
e.g. The following code.  `JSXText`'s node.value returns `\n    text\n    text\n    `.
```js
const f = () => (
    <div>
        text
        text
    </div>
);
```

<details><summary>changes</summary>

1. add handler for JSXText Node
https://github.com/yannickcr/eslint-plugin-react/blob/0076cb09abfed7a3777102359efb5f39dee2e01e/lib/rules/jsx-indent.js#L406-L412

2. check JSXText Indent
https://github.com/yannickcr/eslint-plugin-react/blob/0076cb09abfed7a3777102359efb5f39dee2e01e/lib/rules/jsx-indent.js#L304-L341

3. add the fix for JSXText
https://github.com/yannickcr/eslint-plugin-react/blob/0076cb09abfed7a3777102359efb5f39dee2e01e/lib/rules/jsx-indent.js#L100-L104

4. add test cases
tests/lib/rules/jsx-indent.js
</details>

### DIFF
--- a/lib/rules/jsx-indent.js
+++ b/lib/rules/jsx-indent.js
@@ -30,6 +30,8 @@
 
 'use strict';
 
+const matchAll = require('string.prototype.matchall');
+
 const astUtil = require('../util/ast');
 const docsUrl = require('../util/docsUrl');
 
@@ -97,6 +99,11 @@ module.exports = {
     function getFixerFunction(node, needed) {
       return function fix(fixer) {
         const indent = Array(needed + 1).join(indentChar);
+        if (node.type === 'JSXText' || node.type === 'Literal') {
+          const regExp = /\n[\t ]*(\S)/g;
+          const fixedText = node.raw.replace(regExp, (match, p1) => `\n${indent}${p1}`);
+          return fixer.replaceText(node, fixedText);
+        }
         return fixer.replaceTextRange(
           [node.range[0] - node.loc.start.column, node.range[0]],
           indent
@@ -290,6 +297,29 @@ module.exports = {
       }
     }
 
+    /**
+     * Check indent for Literal Node or JSXText Node
+     * @param {ASTNode} node The node to check
+     * @param {Number} indent needed indent
+     */
+    function checkLiteralNodeIndent(node, indent) {
+      const value = node.value;
+      const regExp = indentType === 'space' ? /\n( *)[\t ]*\S/g : /\n(\t*)[\t ]*\S/g;
+      const nodeIndentsPerLine = Array.from(
+        matchAll(value, regExp),
+        match => (match[1] ? match[1].length : 0)
+      );
+      const hasFirstInLineNode = nodeIndentsPerLine.length > 0;
+      if (
+        hasFirstInLineNode &&
+        !nodeIndentsPerLine.every(actualIndent => actualIndent === indent)
+      ) {
+        nodeIndentsPerLine.forEach((nodeIndent) => {
+          report(node, indent, nodeIndent);
+        });
+      }
+    }
+
     function handleOpeningElement(node) {
       const sourceCode = context.getSourceCode();
       let prevToken = sourceCode.getTokenBefore(node);
@@ -340,6 +370,14 @@ module.exports = {
       checkNodesIndent(firstInLine, indent);
     }
 
+    function handleLiteral(node) {
+      if (!node.parent) {
+        return;
+      }
+      const parentNodeIndent = getNodeIndent(node.parent);
+      checkLiteralNodeIndent(node, parentNodeIndent + indentSize);
+    }
+
     return {
       JSXOpeningElement: handleOpeningElement,
       JSXOpeningFragment: handleOpeningElement,
@@ -352,7 +390,9 @@ module.exports = {
         }
         const parentNodeIndent = getNodeIndent(node.parent);
         checkNodesIndent(node, parentNodeIndent + indentSize);
-      }
+      },
+      Literal: handleLiteral,
+      JSXText: handleLiteral
     };
   }
 };
--- a/tests/lib/rules/jsx-indent.js
+++ b/tests/lib/rules/jsx-indent.js
@@ -300,22 +300,21 @@ ruleTester.run('jsx-indent', rule, {
     ].join('\n'),
     parser: parsers.BABEL_ESLINT
   }, {
-    // Literals indentation is not touched
     code: [
       '<div>',
-      'bar <div>',
-      '   bar',
-      '   bar {foo}',
-      'bar </div>',
+      '    bar <div>',
+      '        bar',
+      '        bar {foo}',
+      '        bar </div>',
       '</div>'
     ].join('\n')
   }, {
     code: [
       '<>',
-      'bar <>',
-      '   bar',
-      '   bar {foo}',
-      'bar </>',
+      '    bar <>',
+      '        bar',
+      '        bar {foo}',
+      '        bar </>',
       '</>'
     ].join('\n'),
     parser: parsers.BABEL_ESLINT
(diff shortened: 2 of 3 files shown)


lib/rules/jsx-indent.js
```diff
@@ -290,6 +295,51 @@ module.exports = {
       }
     }
 
+    /**
+     * Instead String.prototype.mathAll
+     * @param {String} str
+     * @param {RegExp} regex
+     * @returns {Array}
+     */
+    function matchAll(str, regex) {
```
ljharb: there's no need to hardcode this when https://npmjs.com/string.prototype.matchall exists; please use that instead.

lib/rules/jsx-indent.js
```diff
@@ -290,6 +295,51 @@ module.exports = {
...
+     * Check indent for JSXText
+     * @param {ASTNode} node The node to check
+     * @param {Number} indent needed indent
+     */
+    function checkJSXTextNodeIndent(node, indent) {
+      const value = node.value;
+      const regExp = indentType === 'space' ? /\n( *)[\t ]*\S/g : /\n(\t*)[\t ]*\S/g;
+      const nodeIndentsPerLine = [];
+      const matches = matchAll(value, regExp);
+      matches.forEach((match) => {
```
ljharb:
```suggestion
      const nodeIntentsPerLine = matches.map((match) => {
```

lib/rules/jsx-indent.js
```diff
@@ -290,6 +295,51 @@ module.exports = {
...
+        if (match[1]) {
+          nodeIndentsPerLine.push(match[1].length);
```
ljharb: this whole body can then be replaced with:
```js
return match[1] ? match[1].length : 0;
```

toshi-toma:
@ljharb 
Thank you for the review!! 
I fixed.

but, CI is failed...
The cause is  `JSXText` only available Node to `ESLint v5` or more.
https://github.com/eslint/eslint/pull/10152

What do you think....?

ljharb:
@toshi-toma is there no way to identify text inside JSX in eslint < 5? It seems pretty likely that there is something, it's just got a different node type.

golopot:
In older eslint (and current babel-eslint), the node type of text inside JSX is `Literal`.

tests/lib/rules/jsx-indent.js
```diff
@@ -1883,5 +1926,46 @@ const Component = () => (
...
+  }, {
+    code: [
+      '<div>',
+      '\t\ttext',
+      '</div>'
+    ].join('\n'),
+    options: ['tab'],
+    errors: [
+      {message: 'Expected indentation of 1 tab character but found 2.'}
+    ]
```
golopot: A test with `babel-eslint` is needed here.

toshi-toma: Thanks!! I agree. added a test with `babel-eslint`.

toshi-toma:
@ljharb 
I fixed to work for older eslint. and fixed some tests.
Now, CI is failing in Node v4...

lib/rules/jsx-indent.js
```diff
@@ -290,6 +297,29 @@ module.exports = {
...
+     * Check indent for Literal Node or JSXText Node
+     * @param {ASTNode} node The node to check
+     * @param {Number} indent needed indent
+     */
+    function checkLiteralNodeIndent(node, indent) {
+      const value = node.value;
+      const regExp = indentType === 'space' ? /\n( *)[\t ]*\S/g : /\n(\t*)[\t ]*\S/g;
+      const nodeIndentsPerLine = Array.from(
+        matchAll(value, regExp),
+        ([, firstMatch]) => (firstMatch ? firstMatch.length : 0)
```
ljharb: oops, array destructuring doesn't work in node 4
```suggestion
        (match) => (match[1] ? match[1].length : 0)
```

ljharb:
This caused #2561, which was fixed in 3da6eb36327f77adfdd856296c3d663bffe288ee - namely, the issue was that the literal check assumed strings.

PR #2546, opened 2020-01-18
Document that rules dealing with prop types recognize static types, too

silvenon:
In a separate commit I cleaned up examples to be a bit more modern, preferring function components, so they are easier to read, then documented that this rule isn't only for `PropTypes`, but that it accepts static types, too.

### DIFF
--- a/docs/rules/prop-types.md
+++ b/docs/rules/prop-types.md
@@ -1,37 +1,66 @@
 # Prevent missing props validation in a React component definition (react/prop-types)
 
-PropTypes improve the reusability of your component by validating the received data.
+Defining types for component props improves reusability of your components by
+validating received data. It can warn other developers if they make a mistake while reusing the component with improper data type.
 
-It can warn other developers if they make a mistake while reusing the component with improper data type.
+> **Note**: You can provide types in runtime types using [PropTypes] and/or
+statically using [TypeScript] or [Flow]. This rule will validate your prop types
+regardless of how you define them.
 
 ## Rule Details
 
 The following patterns are considered warnings:
 
 ```jsx
-var Hello = createReactClass({
-  render: function() {
-    return <div>Hello {this.props.name}</div>;
-  }
-});
+function Hello({ name }) {
+  return <div>Hello {name}</div>;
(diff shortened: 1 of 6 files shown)


silvenon:
I realized it would be a good idea to document the same for other rules like `require-default-props`, but I'll first wait to see if you agree with this direction.

docs/rules/prop-types.md
```diff
@@ -1,37 +1,66 @@
 # Prevent missing props validation in a React component definition (react/prop-types)
 
-PropTypes improve the reusability of your component by validating the received data.
+Defining types for component props improves reusability of your components by
+validating received data. It can warn other developers if they make a mistake while reusing the component with improper data type.
 
-It can warn other developers if they make a mistake while reusing the component with improper data type.
+You can provide types either in runtime types using [PropTypes] or statically
```
ljharb: let's reword this so it doesn't implicitly discourage using both - even when using static types, using PropTypes is recommended, because they can check many more things than static types can.

silvenon: Ok, I decided to simply use "and/or" instead of "or", that sounds fine to me. Let me know if you want something better.

docs/rules/prop-types.md
```diff
@@ -1,37 +1,66 @@
...
+
+In TypeScript:
+
+```tsx
+interface Props = {
+  age: number
+}
+function Hello({ name }: Props) {
   return <div>Hello {name}</div>;
+  // 'name' type is missing in props validation
```
golopot: I tested this and it seems that currently rule `prop-types` does not report on props with typescript type annotation, can you confirm or refute this?

Off topic: In my opinion rule `prop-types` should not report on props with ts/flow type annotation, since type-checker already excelled in that and that eslint-plugin-react also reporting the same thing just adds noise.

ljharb: I only think you're right if the propType exactly matches the type annotation - and since propTypes are more powerful than static types in many ways, this will often not be the case.

silvenon:
Sorry for never finishing this PR 😕 I hope you managed to fix problematic parts.

Issue #2530, opened 2019-12-21
react/jsx-sort-props + {ignoreCase: false} now ignores case when sorting props


hawkrives:
I've found an unexpected(?) change that stems from #2391.

```jsx
<Hello name="John" Number="2" />
``` 

With `{ignoreCase: false}`, which according to the docs [should take case into account](https://github.com/yannickcr/eslint-plugin-react/blob/8671f054191ad5e5cedf91c7c76999bea9b7bf7b/docs/rules/jsx-sort-props.md#ignorecase) when sorting the props, the above example now expects `name` to be sorted before `Number`.

I know _why_ it sorts the way it does: #2391 switched to using `String#localeCompare`, which does…

```sh
> 'name'.localeCompare('Number')
-1
```

Looking at the `lib/rules/jsx-sort-props.js` code, I'm not sure what changes would be required to accommodate both this issue and #2381, unfortunately.

So, the crux of this issue is: is this an unexpected change, or should I alter my expectations and update the documentation to reflect this behavior?

ljharb:
This definitely seems like an unexpected change.

ljharb:
Specifically, we should probably only use localeCompare when ignoreCase is true?

jcarty:
@ljharb any news on this one?

ljharb:
No, "help wanted" means it's awaiting a PR.

tanmoyopenroot:
We can't depend on `localeCompare` as it is not reliable.
Example:

<img width="245" alt="Screenshot 2020-01-28 at 12 30 03 PM" src="https://user-images.githubusercontent.com/25791025/73242187-3a90f880-41ca-11ea-96f7-8c8a7d3b6032.png">

Therefore, we need to replace `localeCompare` from [line](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/jsx-sort-props.js#L75) and [line](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/jsx-sort-props.js#L345).

tanmoyopenroot:
Hi @hawkrives ,
I have added a PR, Can you check the test cases?

ljharb:
@tanmoyopenroot we *must* depend on localeCompare whenever possible since english isn't the only language in the world.

tanmoyopenroot:
@ljharb  But each character in a string in any language will be compared in `Unicode order` for comparison operator.
In `Unicode` everything is ordered alphabetically.

[...]

PR #2556, opened 2020-01-28
[Fix] `jsx-sort-props`: only use localeCompare when case is ignored

### LINKED ISSUE #2381: react/jsx-sort-props gets confused when props are sorted with --fix but one has camelCase

tanmoyopenroot:
#### PR checklist

- [x] Addresses an existing issue: fixes #2381 #2530 
- [x] bugfix
- [x] Includes tests
- [ ] Documentation update

#### Overview of change:
remove the use  of `localCompare`

### DIFF
--- a/lib/rules/jsx-sort-props.js
+++ b/lib/rules/jsx-sort-props.js
@@ -71,8 +71,12 @@ function contextCompare(a, b, options) {
   if (options.ignoreCase) {
     aProp = aProp.toLowerCase();
     bProp = bProp.toLowerCase();
+    return aProp.localeCompare(bProp);
   }
-  return aProp.localeCompare(bProp);
+  if (aProp === bProp) {
+    return 0;
+  }
+  return aProp < bProp ? -1 : 1;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


lib/rules/jsx-sort-props.js
```diff
@@ -72,7 +72,7 @@ function contextCompare(a, b, options) {
     aProp = aProp.toLowerCase();
     bProp = bProp.toLowerCase();
   }
-  return aProp.localeCompare(bProp);
```
ljharb: 1. this needs to return `0` when `aProp === bProp`.
1. i would still prefer to use localeCompare when case is not ignored.

tanmoyopenroot: I will try to fix it with `localeCompare `.

PR #2557, opened 2020-01-29
[New] `jsx-pascal-case`: Support unicode characters

### LINKED ISSUE #1654: jsx-pascal-case warns on valid casing when using non-English characters
This rule only works when strictly using the 26 english characters hard coded in the regex.  

const PASCAL_CASE_REGEX = /^([A-Z0-9]|[A-Z0-9]+[a-z0-9]+(?:[A-Z0-9]+[a-z0-9]*)*)$/;

In order to support all cased languages, how do you feel about modifying the casing check for example using a technique similar to this (to detect letters):

```
function isLetter(c) {
  return c.toLowerCase() != c.toUpperCase();
}
```

Svish:
Fixes #1654 

Uses the `\p` Unicode character property matcher, and the tests run fine. 👍

Note
---

I pulled in the `xregexp` package to definitely support `\p`. The tests _does_ seem to work without it, at least it Works On My Machine™, but `eslint`, or something at least, doesn't seem to like it? 🤔

image

### DIFF
--- a/lib/rules/jsx-pascal-case.js
+++ b/lib/rules/jsx-pascal-case.js
@@ -6,6 +6,7 @@
 'use strict';
 
 const elementType = require('jsx-ast-utils/elementType');
+const XRegExp = require('xregexp');
 const docsUrl = require('../util/docsUrl');
 const jsxUtil = require('../util/jsx');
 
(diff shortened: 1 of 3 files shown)


Svish:
Can't seem to find any log or anything for why the travis build fails. But I do know there were failing tests in the project even before I started, so that could be one reason. 🤔

lib/rules/jsx-pascal-case.js
```diff
@@ -6,15 +6,16 @@
...
 const jsxUtil = require('../util/jsx');
 
 // ------------------------------------------------------------------------------
 // Constants
 // ------------------------------------------------------------------------------
 
-const PASCAL_CASE_REGEX = /^(.*[.])*([A-Z]|[A-Z]+[a-z0-9]+(?:[A-Z0-9]+[a-z0-9]*)*)$/;
-const ALL_CAPS_TAG_REGEX = /^[A-Z0-9]+([A-Z0-9_]*[A-Z0-9]+)?$/;
+const PASCAL_CASE_REGEX = XRegExp('^(.*[.])*([\\p{Lu}]|[\\p{Lu}]+[\\p{Ll}0-9]+(?:[\\p{Lu}0-9]+[\\p{Ll}0-9]*)*)$', 'u');
+const ALL_CAPS_TAG_REGEX = XRegExp('^[\\p{Lu}0-9]+([\\p{Lu}0-9_]*[\\p{Lu}0-9]+)?$', 'u');
```
golopot: Test is failing because you used RegExp `u` flag but it is not available in node 4. https://node.green/#ES2015-syntax-RegExp--y--and--u--flags

ljharb: @Svish we'll need to come up with another solution that doesn't rely on unicode mode in regexes

Svish: So, support unicode characters in a way that doesn't rely on unicode?  Guess I'm out and will have to just disable this rule then...

But really, why not just drop support for Node 4 and 5? They're like 4+ years old now. If someone for some reason needs to stick with Node 4 or 5, they can always stick with an `eslint-plugin-react` version pre Unicode as well?

ljharb: Hmm, you're using XRegExp tho, which shouldn't require native unicode support.

Their age isn't relevant, people still use them; and it's not worth a breaking change for it.

However, it *would* be fine if the unicode recognition only worked in versions of node that support the `u` flag - we could detect the support (`function hasU() { try { new RegExp('o', 'u'); return true; } catch (e) { return false; } }`) and treat the tests differently.

Svish: @ljharb I don't know how to do that though, so think someone else needs to take over here. Sorry. 😕

ljharb: Done.

PR #2578, opened 2020-02-22
[eslint] change `operator-linebreak` option for ternary operators

golopot:
Current configuration enforces ternary expressions to be written like:
```
a ?
b :
c
```
This is the opposite of the popular style. It is probably a mistake introduced during the time when this project was being converted to Airbnb style.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -23,7 +23,6 @@
           "ignoreTemplateLiterals": true,
           "ignoreComments": true,
         }],
-        "operator-linebreak": [2, "after"],
         "consistent-return": 0,
 
         "prefer-destructuring": [2, { "array": false, "object": false }, { "enforceForRenamedProperties": false }],
--- a/lib/rules/boolean-prop-naming.js
+++ b/lib/rules/boolean-prop-naming.js
@@ -117,9 +117,9 @@ module.exports = {
      */
     function flowCheck(prop) {
       return (
-        prop.type === 'ObjectTypeProperty' &&
-        prop.value.type === 'BooleanTypeAnnotation' &&
-        rule.test(getPropName(prop)) === false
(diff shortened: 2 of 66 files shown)


.eslintrc
```diff
@@ -23,7 +23,7 @@
           "ignoreTemplateLiterals": true,
           "ignoreComments": true,
         }],
-        "operator-linebreak": [2, "after"],
+        "operator-linebreak": [2, "after", {"overrides": {"?": "before", ":": "before" }}],
```
ljharb: if this line is removed, isn’t this what the Airbnb config already mandates?

golopot: Airbnb config uses "before".

https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/style.js#L422

ljharb: oh right. how much would change if we switched to that? Alternately, could we use before here, and use explicit overrides for the ones we want set to after, to match the Airbnb config?

golopot: There will be 840 lines of codes changed if that said option were to be switched to Airbnb style.
https://github.com/golopot/eslint-plugin-react/commit/267854438506153adb49a4ee72ab74306a3c0271
I would like to see the option being kept as is, due to my personal preference for "after" style, and for the sake of avoiding git diffs in history.

ljharb: I prefer the "before" style personally; but even if we wanted to continue setting operators like `&&` to "after", that seems better to me (ie, to be overriding airbnb's "before" setting for specific operators, instead of replacing them)

golopot: I suppose the next action is you come up with an configuration and I will accept it. I believe you.

ljharb:
I'm going ahead with the "before" style by removing the override.

PR #2583, opened 2020-03-01
[tests] delete test cases that became parsing error after acorn-jsx@5.2.0

golopot:
CI is failing after the release of acorn-jsx@5.2.0, 4 days ago, which is a dependency of espree. https://github.com/acornjs/acorn-jsx/releases/tag/5.2.0. That release introduced a breaking change that forbids `>` and `}` in JSXText. Therefore codes like `<p>></p>` became a parsing error.

This PR removed test cases that became parsing error after acorn-jsx@5.2.0.

An expensive alternative is to setup tests for both acorn-jsx < 5.2.0 and > 5.2.0. But that is expensive.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "object.values": "^1.1.1",
     "prop-types": "^15.7.2",
     "resolve": "^1.15.1",
+    "semver": "^6.3.0",
     "string.prototype.matchall": "^4.0.2",
     "xregexp": "^4.3.0"
   },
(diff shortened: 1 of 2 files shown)


ljharb:
Has an issue been filed upstream for the breaking change in a minor release?

ljharb:
Looks like they unfortunately made an intentional breaking change in a minor version.

ljharb:
I've gone ahead and updated the tests so they're kept when acorn-jsx is < 5.2, and skipped when it's absent or when it's >= 5.2. That way they can still run on eslint 4, for example: https://travis-ci.org/yannickcr/eslint-plugin-react/builds/657558888

PR #2585, opened 2020-03-03
[New] Create a named-import rule for React

el-angel:


alexzherdev:
Note that this can be achieved with the core [`no-restricted-syntax`](https://eslint.org/docs/rules/no-restricted-syntax) rule by using the following selector: `"ImportDeclaration[source.value=\"react\"] :matches(ImportSpecifier, ImportNamespaceSpecifier)"`

docs/rules/no-named-import.md
```diff
@@ -0,0 +1,20 @@
...
+```jsx
+import React, { Component } from 'react';
+import { Component } from 'react';
+```
+
+The following patterns are **not** considered warnings:
+
+```jsx
+import React from 'react';
+import * as React from 'react';
```
ljharb: this should always be a warning, and is always incorrect. TypeScript's module system is broken unless you enable synthetic imports and esModuleInterop, at which point this syntax isn't needed.

docs/rules/no-named-import.md
```diff
@@ -0,0 +1,20 @@
+# Prevent named imports from react (react/no-named-import)
+
+Prevent the use of named imports from React, so it's visible in code that it's React and not custom code. This rule
```
ljharb: What about hooks, like `useEffect` etc?

ljharb (changes requested):
I think if we wanted to do this, it'd need to be able to support both patterns - forcing names, as well as disallowing them.

docs/rules/no-named-import.md
el-angel: I will add an options object which allows the developer to whitelist imports

docs/rules/no-named-import.md
el-angel: I've added a new rule called `no-namespace-import`. I didn't think that rule should be within this rule.

README.md
```diff
@@ -122,6 +122,8 @@ Enable the rules that you would like to use.
 * [react/no-find-dom-node](docs/rules/no-find-dom-node.md): Prevent usage of findDOMNode
 * [react/no-is-mounted](docs/rules/no-is-mounted.md): Prevent usage of isMounted
 * [react/no-multi-comp](docs/rules/no-multi-comp.md): Prevent multiple component definition per file
+* [react/no-named-import](docs/rules/no-named-import.md): No named imports on React
+* [react/no-namespace-import](docs/rules/no-namespace-import.md): No namespace import on React (fixable)
```
ljharb: this is already a generic rule, and should not just apply to react: https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-namespace.md

docs/rules/no-named-import.md
```diff
@@ -0,0 +1,60 @@
...
+This rule has two different properties that can be used (not at the same time): `whitelist` and `blacklist`. Both are an array of strings. By default this rule whitelists nothing, so every named import will error.
+
+Example configurations:
+
+```javascript
+// Blacklist
+{
+  "rules": {
+    "react/no-named-import": ["error", {
+      "blacklist": ["Component", "useEffect"]
```
ljharb: "whitelist" and "blacklist" are not inclusive terms; please change it to something like "only" vs "except"?

ljharb (changes requested):

README.md
el-angel: Ah ofcourse, I'll remove it.

docs/rules/no-named-import.md
el-angel: I've changed it to `allow` and `forbid`

docs/rules/no-named-import.md
el-angel: I've added an option (`forceImport`, which could be true/false), where the linter will report when `React.useState` is used for example and should be imported. This way the rule can force certain imports which were used (`React.useState`) but not imported (`import React from 'react'`).

el-angel:
@ljharb Is there any feedback you have for me regarding this rule? And do you have a clue on what goes wrong on Node V4?

el-angel:
@ljharb I have tried to check on how to fix the node v4 pipeline, but I could use some help if that's possible. I would like to get this PR through

lib/rules/no-named-import.js
```diff
@@ -0,0 +1,123 @@
...
+      dontImport: "Don't import {{name}} from React. Use React.{{name}} to be consistent.",
+      shouldImport: 'Import {{name}} from React'
+    }
+  },
+  create(context) {
+    // ignore non-modules
+    if (context.parserOptions.sourceType !== 'module') {
+      return {};
+    }
+    const {options} = context;
```
golopot, line 58: Destructuring syntax is not supported in Node 4. For debugging without istanbul messing up codes you can run `$(npm bin)/mocha tests/lib/rules/no-named-import.js`.

el-angel: Thanks! It's fixed :)

docs/rules/no-named-import.md
```diff
@@ -0,0 +1,86 @@
...
+Example configurations:
+
+```javascript
+// forbid
+{
+  "rules": {
+    "react/no-named-import": ["error", {
+      "forbid": ["Component", "useEffect"],
+      // Optional
+      forceImport: true,
```
ljharb, line 21: i'm a bit skeptical about this one; when this is `true`, then imports are forced, but when it's set to `false`, are properties forced?

el-angel: `
// { allow: ['useState'], forceImport: true }
import React from 'react';
// will error, because it will force to import useState
const [loading, setLoading] = React.useState(false);
`

`
// { allow: ['useState'], forceImport: false }
import React from 'react';
// will not error, because it will allow the user to use React.useState
const [loading, setLoading] = React.useState(false);
`

`
// { allow: ['useState'], forceImport: false }
// will error on the Component import, because that's not allowed
const [loading, setLoading] = React.useState(false);
`

ljharb: ok, so then it seems like the entire rule has two "modes" - import, or property. If that's the case, then that shouldn't be in the options bag, it should be the first enum argument in a 2-argument schema (the object comes second). Does that make sense?

el-angel: I get what you wrote, only that's not the current implementation, because the `forceImport` property doesn't force a mode (property/import), let me clarify using these examples:

```
// allow ['useState']
// forceImport: false (= default value)

import React from 'react';

// passes, because forceImport === false
const [loading, setLoading] = React.useState(false);

// passes, because allow['useState']
import React, { useState } from 'react';

// errors, because Component is not allowed to import

// passes, because Component is not included in allow
class MyComponent extends React.Component {}
``` 

Another example:
```
// allow['useState']
// forceImport: true

import React from 'react';
// error, because forceImport: true, useState should be imported
const [loading, setLoading] = React.useState(false);

// Like this:
import React, { useState } from 'react';
const [loading, setLoading] = useState(false);
```

with the forbid option:
```
// forbid: ['Component']
// forceImport: true

import React from 'react';
// error, forceImport: true, so all except Component should be imported
const [loading, setLoading] = React.useState(false);

// forbid: ['Component']
// forceImport: false
import React from 'react';

// passes, because forceImport: false. User is allowed to access property and use it
const [loading, setLoading] = React.useState(false);

// errors, because Component is forbidden to import
```

ljharb: It seems like you envision `forceImport false` to mean "ignore" where `forceImport true` means "force imports" - where i'm looking for "force properties". iow, it seems like this shouldn't be a boolean `forceImport` but should instead be an enum `style` or something with 3 values?

docs/rules/no-named-import.md
```diff
@@ -7,80 +7,81 @@ also creates consistency inside a codebase because some people use named imports
 
 ### Options
 
-This rule has two different properties that can be used (not at the same time): `allow` and `forbid`. Both are an array of strings. By default this rule allows nothing, so every named import will error.
```
ljharb: removing "allow" and "forbid" is a breaking change, and isn't something i'll find acceptable.

ljharb:
(disregard my previous review; i was looking at the "view changes" diff and not the entire diff)

docs/rules/no-named-import.md
```diff
@@ -0,0 +1,87 @@
...
+also creates consistency inside a codebase because some people use named imports and some don't.
+
+## Rule Details
+
+### Options
+
+This rule has two different modes: 'import' and 'property' which can be set using the first argument. The second arguments accepts an array of strings, which are the allowed modules to be imported or accessed via property.
+
+```
+"react/no-named-import": [<enabled>, 'import' | 'property', []]
```
ljharb, line 13: what if someone wants to simultaneously require one thing be imported, and one thing be a property? For example, in my codebases I require hooks to be named imports, but I never permit `Component` to be anything except `React.Component`.

One possibility I'd expect here is a string argument, but the next option should be an object, whose keys are the names, and whose values are either "import" or "property".

Another would be an object with "import" and "property" properties, whose values are an array of names.

The advantage of the object key form is that it's trivial to ensure that each name only has one "mode" set; the advantage of the object array form is that it's much less boilerplate to write.

Perhaps there's a third alternative I'm not considering?

ljharb (changes requested):

docs/rules/no-named-import.md
el-angel: @ljharb 
I see your point and thought about this too. We have three possibilities as I see it:

```
// current implementation
const options = ["error", "import", [/* <names> */]];
```

```
const options = ["error", {
  "import": ["Component"],
  "property": ["useEffect", "useState"],
}];
```

```
const options = ["error", {
  "useEffect": "import",
  "Component": "property",
}]
```

The current implementation automatically forces a mode for the non-specified names, which should not be the case after reading your comment. 

After giving it some thought I think the best way is to go for the 2nd implementation. It's also consistent with other rules that for example have a `forbid` option which is an array of strings.

What do you think?

ljharb: I'd expect, in both 2 and 3 in your comment, that the object would be a *third* argument - that the second argument would be a string so as to set the mode for all non-specified names, and the object would override some of those.

The reason I'm leaning towards the third is because it ensures that "foo" won't appear in both `import` and `property` - otherwise we have to do a runtime check and throw an error when there's overlap.

el-angel: @ljharb I've implemented the third option. Looking forward to your feedback

ljharb (commented):
This seems great! It also seems like it should be autofixable. Is that something you're willing to add?

el-angel:
@ljharb absolutely! I'll work on it

stefcameron:
This looks like exactly the rule I'm looking for! React 16.14.0 was just announced, and along with it, a new codemod to [remove unused react imports](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#removing-unused-react-imports) -- if you move to the new JSX Transform available in 16.14.0.

But the codemod also has the option of moving all statements like `React.useState()` to `useState()` with an `import { useState } from 'react'`, and I've run the codemod to move to these destructured imports.

Now I'd like to enforce that in the code base, because, as the article says at the very bottom,

> In addition to cleaning up unused imports, this will also help you prepare for a future major version of React (not React 17) which will support ES Modules and not have a default export.

Note, "not have a default export".

Based on the examples in the README for this new rule, I think it would be perfect to make sure we keep consistently using destructured imports in our codebase with the `import` mode.

el-angel:
@ljharb is there anything regarding this PR that needs improvement? It would be great if this rule could make a next (major) release

ljharb:
Sorry for the delay, I'll take a look. New rules are semver-minor.

docs/rules/no-named-import.md
```diff
@@ -0,0 +1,80 @@
...
+
+```
+
+The following patterns are **not** considered warnings:
+
+```jsx
+import React, { useState } from 'react';
+
+// ['property', {}]
+import React from 'react'; const [loading,setLoading] = React.useState(false);
```
ljharb, line 69:
```suggestion
import React from 'react'; const [loading, setLoading] = React.useState(false);
```

docs/rules/no-named-import.md
```diff
@@ -0,0 +1,80 @@
...
+
+// ['property', {'useState': 'import' }]
+import { useState } from 'react';
+
+// ['import', {}]
+import React, { useState } from 'react'; const [loading,setLoading] = useState(false);
```
ljharb, line 75:
```suggestion
import React, { useState } from 'react';

 const [loading,setLoading] = useState(false);
```

docs/rules/no-named-import.md
```diff
@@ -0,0 +1,80 @@
...
+
+// ['import', {'useState': 'property' }]
+const [loading,setLoading] = React.useState(false);
```
ljharb, line 78:
```suggestion
const [loading, setLoading] = React.useState(false);
```

lib/rules/no-named-import.js
```diff
@@ -0,0 +1,244 @@
...
+
+      return str;
+    }
+
+    function fixImportDeclaration() {
+      const node = importDeclaration;
+      const tokens = sourceCode.getTokens(node);
+
+      const importToken = tokens[0];
+      const fromToken = tokens.filter((token) => token.value === 'from')[0];
```
ljharb, line 87:
```suggestion
      const fromToken = tokens.find((token) => token.value === 'from');
```

lib/rules/no-named-import.js
```diff
@@ -0,0 +1,244 @@
...
+      return (fixer) => fixer.replaceTextRange(range, getImportDeclarationString());
+    }
+
+    function shouldUpdateImportStatement() {
+      const currentSpecifiers = importDeclaration.specifiers
+        .filter((specifier) => specifier.type !== 'ImportDefaultSpecifier')
+        .map((specifier) => specifier.imported.name);
+
+      return specifiers.some((spec) => currentSpecifiers.indexOf(spec) === -1)
+      || currentSpecifiers.some((spec) => isNameOfType(spec, 'property'));
```
ljharb, line 103:
```suggestion
      return specifiers.some((spec) => currentSpecifiers.indexOf(spec) === -1)
        || currentSpecifiers.some((spec) => isNameOfType(spec, 'property'));
```

lib/rules/no-named-import.js
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    function getFixer(name, range, newType) {
+      let newText = '';
+
+      if (newType === 'property') {
+        newText += 'React.';
+      }
+
+      newText += name;
```
ljharb, line 113:
```suggestion
      const newText = `${newType === 'property' ? 'React.' : ''}${name}`;
```

lib/rules/no-named-import.js
```diff
@@ -0,0 +1,244 @@
...
+        importDeclaration = node;
+        node.specifiers.forEach((specifier) => {
+          if (specifier.type === 'ImportDefaultSpecifier') {
+            const variables = context.getDeclaredVariables(specifier);
+
+            const variablesWithRefs = variables.filter((_variable) => _variable.references.length);
+
+            if (!variablesWithRefs.length) { return; }
+
+            variablesWithRefs[0].references.forEach((reference) => {
```
ljharb, line 219:
```suggestion
            const firstVariableWithRefs = variables.find((_variable) => _variable.references.length > 0);

            if (!firstVariableWithRefs) { return; }

            firstVariableWithRefs.references.forEach((reference) => {
```

tests/lib/rules/no-named-import.js
```diff
@@ -0,0 +1,371 @@
...
+    {
+      code: "import React from 'react';"
+    },
+    {
+      code: "import React, { useState } from 'react';"
+    },
+    {
+      code: [
+        'import React from \'react\';',
+        'const [loading,setLoading] = React.useState(false);'
```
ljharb, line 39:
```suggestion
        'const [loading, setLoading] = React.useState(false);'
```

tests/lib/rules/no-named-import.js
```diff
@@ -0,0 +1,371 @@
...
+      ].join('\n'),
+      options: ['property']
+    },
+    {
+      code: "import React, { useState } from 'react'; const [loading,setLoading] = useState(false);",
```
ljharb, line 44:
```suggestion
      code: "import React, { useState } from 'react';
      const [loading, setLoading] = useState(false);",
```

tests/lib/rules/no-named-import.js
```diff
@@ -0,0 +1,371 @@
...
+      options: ['import']
+    },
+    {
+      code: `
+        import React, { useEffect, Component } from 'react';
+        const [loading,setLoading] = React.useState(false);
```
ljharb, line 50:
```suggestion
        const [loading, setLoading] = React.useState(false);
```

tests/lib/rules/no-named-import.js
```diff
@@ -0,0 +1,371 @@
...
+      `,
+      options: ['import', {
+        useEffect: 'import',
+        useState: 'property'
+      }]
```
ljharb, line 55:
```suggestion
      options: [
        'import',
        {
          useEffect: 'import',
          useState: 'property'
        }
      ]
```

tests/lib/rules/no-named-import.js
```diff
@@ -0,0 +1,371 @@
...
+    },
+    {
+      code: `
+        import { useEffect } from 'react';
+        const [loading,setLoading] = React.useState(false);
```
ljharb, line 60:
```suggestion
        const [loading, setLoading] = React.useState(false);
```

tests/lib/rules/no-named-import.js
```diff
@@ -0,0 +1,371 @@
...
+        const a = React.Component;
+      `,
+      options: ['property', {
+        useEffect: 'import',
+        useState: 'property'
+      }]
```
ljharb, line 66:
```suggestion
      options: [
        'property',
        {
          useEffect: 'import',
          useState: 'property'
        }
      ]
```

tests/lib/rules/no-named-import.js
```diff
@@ -0,0 +1,371 @@
...
+          messageId: 'usePropertyAccessor',
+          data: {name: 'useState'}
+        }
+      ]
+    },
+    {
+      code: `
+        import React from 'react';
+        const [value, setValue] = React.useState('');
+        React.useEffect(() => {},[]);
```
ljharb, line 133:
```suggestion
        React.useEffect(() => {}, []);
```

tests/lib/rules/no-named-import.js
```diff
@@ -0,0 +1,371 @@
...
+      `,
+      output: `
+        import React, { useEffect } from 'react';
+        const [value, setValue] = React.useState('');
+        useEffect(() => {},[]);
```
ljharb, line 138:
```suggestion
        useEffect(() => {}, []);
```

tests/lib/rules/no-named-import.js
```diff
@@ -0,0 +1,371 @@
...
+          messageId: 'useNamedImport',
+          data: {name: 'useEffect'}
+        }
+      ]
+    },
+    {
+      code: `
+        import React from 'react';
+        const [value, setValue] = React.useState('');
+        React.useEffect(() => {},[]);
```
ljharb, line 155:
```suggestion
        React.useEffect(() => {}, []);
```

tests/lib/rules/no-named-import.js
```diff
@@ -0,0 +1,371 @@
...
+      `,
+      output: `
+        import React, { useEffect } from 'react';
+        const [value, setValue] = React.useState('');
+        useEffect(() => {},[]);
```
ljharb, line 160:
```suggestion
        useEffect(() => {}, []);
```

ljharb (changes requested):

ljharb:
Perhaps we should rename this rule to "named-import"?

el-angel:
> Perhaps we should rename this rule to "named-import"?

@ljharb  I agree, changed it :-)

docs/rules/named-import.md
```diff
@@ -0,0 +1,83 @@
+# Prevent named imports from react (react/named-import)
```
ljharb, line 1:
```suggestion
# Prevent named imports from React (react/named-import)
```

docs/rules/named-import.md
```diff
@@ -0,0 +1,83 @@
...
+
+Enforce the use of named imports from React, so it's visible in code that it's React and not custom code. This rule also creates consistency inside a codebase because it enforces or prohibits the use of specified named imports.
```
ljharb, line 3:
```suggestion
Prohibit the use of named imports from React, so it's visible in code that it's React and not custom code; or enforce the use of named imports from React, to avoid needing the `React` default import. This rule also creates consistency inside a codebase because it enforces or prohibits the use of specified named imports.
```

docs/rules/named-import.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Rule Details
+
+### Options
+
+This rule has two different modes: 'import' and 'property' which can be set using the first argument. The second arguments accepts an object where the keys are equal to React exports and their value overrides the default mode
```
ljharb, line 9:
```suggestion
This rule has two different modes: 'import' and 'property', which can be set using the first argument. The second arguments accepts an object where the keys are equal to React named exports, and their value overrides the default mode.
```

docs/rules/named-import.md
```diff
@@ -0,0 +1,83 @@
...
+
+```
+"react/named-import": [<enabled>, 'import' | 'property', {}]
+```
+
+By default the rule is set to `import` with an empty object, which means that all modules should be imported
```
ljharb, line 15:
```suggestion
By default the rule is set to `import` with an empty object, which means that all of React's named exports will have the rule applied.
```

docs/rules/named-import.md
```diff
@@ -0,0 +1,83 @@
...
+
+// property
+{
+  "rules": {
+    "react/named-import": ["error", 'property', {
+      'Component': 'import',
+      'useState': 'import'
+    }]
+  }
+}
```
ljharb, line 40:
```suggestion
  "rules": {
    "react/named-import": ["error", "import", {
      "Component": "property",
      "useState": "property"
    }]
  }
}

// property
{
  "rules": {
    "react/named-import": ["error", "property", {
      "Component": "import",
      "useState": "import"
    }]
  }
}
```

lib/rules/named-import.js
```diff
@@ -0,0 +1,238 @@
...
+
+    /** @type {Context} */
+    const sourceCode = context.getSourceCode();
+
+    function isNameOfType(name, type) {
+      return (overrides[name] || mode) === type;
+    }
+
+    function getImportDeclarationString() {
+      let str = 'React';
```
ljharb, line 71: we probably need a way to configure whether the user is using the jsx runtime or not.

i think we'll need to add a global plugin setting, that this rule (and a few others) can look to. When set, we'd only provide the `React` import if something referenced it, not automatically.

lib/rules/named-import.js
```diff
@@ -0,0 +1,238 @@
...
+            fix: fixImportDeclaration()
+          });
+        }
+
+        importDeclaration = null;
+        specifiers = [];
+        propertySpecifiers = [];
+      },
+      TSTypeReference: (node) => typeSelector(node.typeName),
+      'ImportDeclaration[source.value=\'react\']'(node) {
```
ljharb, line 203: this needs to use the pragma setting, not hardcode "react"

lib/rules/named-import.js
```diff
@@ -0,0 +1,238 @@
...
+        importDeclaration = node;
+        node.specifiers.forEach((specifier) => {
+          if (specifier.type === 'ImportDefaultSpecifier') {
+            const variables = context.getDeclaredVariables(specifier);
+
+            const firstVariableWithRefs = variables.find((_variable) => _variable.references.length);
```
ljharb, line 209: let's avoid variable names with a leading `_` :-)

ljharb (changes requested):

lib/rules/named-import.js
el-angel: Can you point me in the right direction to make this work. I'm not sure how to implement this

lib/rules/named-import.js
el-angel: `import React from "react"`

the first "React" would be the pragma right? This hardcoded string stands for the package name, which is always "react". Or am I mistaken?

ljharb: oh hm, that's true. Looks like we currently just hardcode the react package name.

lib/rules/named-import.js
ljharb: By global settings, I mean the "settings" stuff referenced here: https://github.com/yannickcr/eslint-plugin-react#configuration

If we add a boolean `jsx-runtime` setting, when it's false, they're using the React 16 jsx compilation, and React must always be in scope. When it's true, they're using the React 17 jsx compilation, and React need not be in scope.

ljharb (approved):
LGTM once the typescript error is resolved

ljharb:
@el-angel this is almost ready to land if we can resolve the tsc error.

el-angel:
@ljharb pipeline failes because it says rebase fails, but I don't see any merge conflicts? Can you help me out?

ljharb:
@el-angel looks like you didn’t pull my latest changes when updating - could you reset your local branch to https://github.com/yannickcr/eslint-plugin-react/commit/a156b282586a093c134a0dc3994d24afaddf5725, add your fix on top, and force push that up?

ljharb:
nvm, i'll take care of it :-)

ljharb:
the only thing I'm still thinking about is https://github.com/yannickcr/eslint-plugin-react/pull/2585#discussion_r608809722.

In other words, there seems to be a few choices here:
1. assume that anyone using the jsx runtime will extend `react/jsx-runtime`, and in that config, set the rule to "named only".
1. do not assume this, and find a way to *detect* use of the jsx runtime, and either allow the rule to be configured to "auto" or error when the setting conflicts with the usage of the runtime (when the jsx runtime is enabled, "named only" must be selected).
1. do neither, and just force people to "know" they need to set this rule to "named only" when using the jsx runtime.

Gipsy9:
I want to make group ROCKY123GIPSY@GMAIL.COM

jonkoops:
@ljharb what do you think would be needed to get this landed? It seems that the React maintainers are aligned that named imports will be the way going forward (see https://github.com/facebook/react/issues/11503), so it would be good for users to start getting informed so they can switch before this becomes a migration issue further down the road.

I'd be more than willing to volunteer my time to get this landed in a new PR, if we can settle on a direction for the sensible defaults. My vote would be to include this rule by default for `react/jsx-runtime`, perhaps as a warning (depending on whether you consider introducing this as an error as a breaking change).

Another thing to consider is that React 19 has made the automatic JSX transform mandatory, which means `testReactVersion(context, '>= 19.0.0')` is now a reliable signal we could start using.

PR #2587, opened 2020-03-07
[fix] `displayName`: avoid a crash when using React.memo

golopot:
This pr adds a null check, to avoid a crash that happens when using `React.memo` plus a return statement without an argument. Like
```js
import React from 'react';
const Hello = React.memo(function Hello() {
  return;
});
```


golopot:
Thanks!

Issue #2218, opened 2019-03-28
Warning: React version not specified when used with monorepo


jagretz:
I looked through the code for [version.js](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/util/version.js) and the [version test](https://github.com/yannickcr/eslint-plugin-react/blob/master/tests/util/version.js) so I believe this is a valid issue.

I am running eslint from a monorepo I've set up with lerna. One of the packages in the monorepo using react so I've only installed react-specific eslint plugins in that package. ie `eslint-plugin-react`, `eslint-plugin-react-hooks` etc

Using eslints [overrides feature](https://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy) and running an `eslint` command from the _root_ (`./`) works picking up and overriding the root configuration when it finds a `eslintrc` in a sub package, however, when I've set my react version to `detect` from the react projects `.eslintrc` it throws this warning.

```bash
Warning: React version was set to "detect" in eslint-plugin-react settings, but the "react" package is not installed. Assuming latest React version for linting.
```

`react` _is_ installed and included in _that_ packages `/react-foo/package.json`. But in order to prevent this warning I would need to install `react` at the root package of the monorepo which isn't correct or desired.

Here is an example project hierarchy with the `.eslintrc` config overrides:

```bash
./
├── package.json            # react is not installed in this package
├── .eslintrc               # root lint config
├── packages/
|   └── react-foo/
|   |   ├── package.json    # react installed
|   |   ├── .eslintrc       # react-specific lint overrides
|   |   └── src/
|   |       └──  ...
|   └── not-react-foo/
|       ├── package.json
|       └── src/
|           └── ...
└── ...
```

I believe this is due to this tools `version.js` using `process.cwd()` to check for react at the (cwd) directory where the command was ran, ie the root (`./.eslintrc`), instead of directory from where the eslint configuration for eslint-plugin-react is specified, ie `./packages/react-foo/.eslintrc`.

ljharb:
It needs to be a sibling to the root .eslintrc, so yes, i'd say it is correct to have react located in a root node_modules. Typically in a monorepo, all deps are hoisted to the root and symlinked down - why is this not the case here?

ljharb:
(related to #2214)

ljharb:
hmm, that's a good point tho - maybe the detection should start looking from the location of the file being linted, not from the cwd? That seems a reasonable change.

[...]

PR #2609, opened 2020-03-29
WIP: react version detection in a monorepo

migueloller:
WIP pull request to track progress on #2218 .

### DIFF
--- a/lib/util/version.js
+++ b/lib/util/version.js
@@ -14,9 +14,9 @@ function resetWarningFlag() {
   warnedForMissingVersion = false;
 }
 
-function detectReactVersion() {
+function detectReactVersionFromContext(context) {
   try {
-    const reactPath = resolve.sync('react', {basedir: process.cwd()});
+    const reactPath = resolve.sync('react', {basedir: context.getCwd()});
     const react = require(reactPath); // eslint-disable-line global-require, import/no-dynamic-require
     return react.version;
   } catch (e) {
@@ -38,7 +38,7 @@ function getReactVersionFromContext(context) {
   if (context.settings && context.settings.react && context.settings.react.version) {
     let settingsVersion = context.settings.react.version;
     if (settingsVersion === 'detect') {
-      settingsVersion = detectReactVersion();
+      settingsVersion = detectReactVersionFromContext(context);
     }
     if (typeof settingsVersion !== 'string') {
       error('Warning: React version specified in eslint-plugin-react-settings must be a string; ' +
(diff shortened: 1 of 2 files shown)


ljharb:
Seems like a good first start, but this would need tests.

Can you convert it to a draft PR, until it's ready?

migueloller:
@ljharb, yup! I was just trying to see how involved this change was going to be. Because of the current test set up I'm probably going to have to spend some time figuring out the least invasive way to implement the tests.

I'll try to look into this again during the weekend.

ljharb:
@migueloller any interest in completing this PR?

migueloller:
@ljharb, in addition to the small tweak I did to the existing tests, were there specific test cases you were thinking we should cover?

ljharb:
My hope would be, inside `tests/files`, a recreation of a monorepo setup where `detect` would have returned the wrong result without your change.

migueloller:
@ljharb, got it. That sounds reasonable! I will do my best to get something set up sometime soon. I apologize I haven't put in more work on this PR, I've been busy 😬

ljharb:
@migueloller are you no longer interested in this feature?

ljharb:
The passage of time doesn't have any impact on relevance in open source :-)

My hope is that everybody using this plugin can use "detect", so if there are monorepo users unable to use it, then it's definitely worthwhile to fix it.

migueloller:
Sounds good!

Thanks for the nudge 😄

migueloller:
Well, as I was doing the rebase I noticed that `mater` now does consider `context` in the `detectReactVersion` function.

https://github.com/jsx-eslint/eslint-plugin-react/blob/master/lib/util/version.js#L48-L72

Here's the PR that made the change: https://github.com/jsx-eslint/eslint-plugin-react/commit/eeb92731bd922af4493cf938f0547a021e5cdbcf

Seems like that's satisfactory?

migueloller:
Ok, sorry for the delay in the response but I wanted to make sure I understood well how my changes and the current implementation in `master` differ before I could confirm it solves the problem. I just got around to that.

In my implementation I was using `context.getCwd` to determine the current working directory passed to the linter instance as documented [here](https://eslint.org/docs/latest/extend/custom-rules#the-context-object).

The current implementation is using `context.getFilename` instead, which seem like a better solution since it will find the relevant React version relative to the source file.

I did mention that using this API would be ideal: https://github.com/jsx-eslint/eslint-plugin-react/issues/2218#issuecomment-605589405. @jcrosetto ultimately got to it a year later: https://github.com/jsx-eslint/eslint-plugin-react/pull/2943

So, the project is indeed solved and I think we're good to close this.

Thanks again @ljharb and thanks @jcrosetto for taking the lead here and getting it fixed using `context.getFilename`.

Issue #2594, opened 2020-03-16
BUG "jsx-key"  should warning when iterate element with optional chaining syntax


JonathanLee-LX:
Iterate jsx element with optional chaining syntax, `react/jsx-key` rule don't show  warnings.

Just like this

![new-react-jsx](https://user-images.githubusercontent.com/20569396/76747410-0b533c80-67b4-11ea-8af9-751ca2531cf8.png)

JonathanLee-LX:
Maybe this example will be more accurate
![no-react-key](https://user-images.githubusercontent.com/20569396/76783243-79b3f100-67ec-11ea-9217-4c0100477022.png)

ljharb:
(in general, please post code samples as text; images of text are not accessible, can't be copy-pasted, and difficult to read in many circumstances)

That second example would never happen in real code, of course, but yes, I agree that in that case it should show a warning.

JonathanLee-LX:
Thank you first, yes, in the second example, the initial value of `arr` is normally set to `[]` instead of `undefined`, so this rule is generally not a problem.

JonathanLee-LX:
I cloned this repository, and try to add a test case like below
`    {
      code: '[1, 2, 3]?.map(x => <App />)',
      parser: parsers.TYPESCRIPT_ESLINT,
      errors: [{message: 'Missing "key" prop for element in array.'}]
    }
`
This test case will fail, if add `key` in `App`,  this case will pass。
So，`TYPESCRIPT_ESLINT` parser has already implemented this use case.

ljharb:
perhaps - is that the parser you're using? only the typescript parser, and babel-eslint, currently support optional chaining,

JonathanLee-LX:
I'm using `@typescript-eslint/parser`, the version is `^2.7.0`.

ljharb:
TYPESCRIPT_ESLINT is the old parser; you'll need `'@TYPESCRIPT_ESLINT'` for your failing test case. Want to make a PR with it?

JonathanLee-LX:
Yes, I want, but I can't make this use case pass the test with `@typescript-eslint/parser`.

ljharb:
That’s fine, a PR with a failing test is still much easier for me to work with than issue comments :-)

PR #2610, opened 2020-03-29
[Fix] `jsx-key`: support optional chaining

JonathanLee-LX:
For [#2594](https://github.com/yannickcr/eslint-plugin-react/issues/2594)

### DIFF
--- a/tests/lib/rules/jsx-key.js
+++ b/tests/lib/rules/jsx-key.js
@@ -50,7 +50,7 @@ ruleTester.run('jsx-key', rule, {
     {code: 'foo(() => <></>);', parser: parsers.BABEL_ESLINT},
     {code: '<></>;', parser: parsers.BABEL_ESLINT}
   ],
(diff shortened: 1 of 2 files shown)


tests/lib/rules/jsx-key.js
```diff
@@ -68,6 +68,9 @@ ruleTester.run('jsx-key', rule, {
   }, {
     code: '[1, 2 ,3].map(x => { return <App /> });',
     errors: [{message: 'Missing "key" prop for element in iterator'}]
+  }, {
+    code: '[1, 2, 3]?.map(x => <App />)',
+    errors: [{message: 'Missing "key" prop for element in iterator'}]
```
ljharb, line 73: this will have to be restricted to the babel and `@typescript` parsers, until eslint supports it.

JonathanLee-LX: Yeah, that's right

PR #2615, opened 2020-04-03
Allow a custom message with forbid props

mtamhankar1:
Hi team! When considering using the forbid-component-props and forbid-dom-props rules, an early topic of discussion was whether or not it would be possible to provide a custom error message. This gives the user a chance to better explain _why_ the prop is forbidden, and possible offer an alternative, or best practice, etc.

This PR aims to add custom messaging functionality to both rules of these rules. Let me know if such an enhancement is something you would consider adding and I'm happy to make improvements on it and take any suggestions! Thanks!

### DIFF
--- a/docs/rules/forbid-component-props.md
+++ b/docs/rules/forbid-component-props.md
@@ -42,15 +42,16 @@ The following patterns are **not** considered warnings:
 ### `forbid`
 
 An array specifying the names of props that are forbidden. The default value of this option is `['className', 'style']`.
-Each array element can either be a string with the property name or object specifying the property name and a component whitelist:
+Each array element can either be a string with the property name or object specifying the property name, an optional
+custom message, and a component whitelist:
 
 ```js
 {
   "propName": "someProp",
-  "allowedFor": [SomeComponent, AnotherComponent]
+  "allowedFor": [SomeComponent, AnotherComponent],
+  "message": "Avoid using someProp"
 }
 ```
-
 
 ### Related rules
 
--- a/docs/rules/forbid-dom-props.md
+++ b/docs/rules/forbid-dom-props.md
@@ -36,14 +36,22 @@ The following patterns are **not** considered warnings:
 
 ```js
 ...
-"react/forbid-dom-props": [<enabled>, { "forbid": [<string>] }]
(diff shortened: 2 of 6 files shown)


docs/rules/forbid-component-props.md
```diff
@@ -35,7 +35,7 @@ The following patterns are **not** considered warnings:
 
 ```js
 ...
-"react/forbid-component-props": [<enabled>, { "forbid": [<string>|<object>] }]
+"react/forbid-component-props": [<enabled>, { "forbid": [<string>|<object>], "message": <string> }]
```
ljharb, line 38: this seems like it would allow only one message for each "forbid" value; it seems like it. might make more sense to add a `message` property on the object form of a `forbid` value?

docs/rules/forbid-dom-props.md
```diff
@@ -36,14 +36,17 @@ The following patterns are **not** considered warnings:
 
 ```js
 ...
-"react/forbid-dom-props": [<enabled>, { "forbid": [<string>] }]
+"react/forbid-dom-props": [<enabled>, { "forbid": [<string>], "message": <string> }]
```
ljharb, line 39: similarly here - it seems like instead of an overarching message property, it'd be better to allow each forbid value to be a string *or* an object, with the object form allowing a custom message?

docs/rules/forbid-component-props.md
mtamhankar1: @ljharb Thanks for taking a look. I've updated this to allow a separate message for each forbid option. Let me know what you think!

mtamhankar1:
@ljharb Any chance you'll have time to take another look at this soon? Thanks!

lib/rules/forbid-dom-props.js
```diff
@@ -43,11 +55,17 @@ module.exports = {
...
+      const options = {
+        message: typeof value === 'string' ? null : value.message
+      };
+      return [propName, options];
+    }));
 
-      const forbid = configuration.forbid || DEFAULTS;
-      return forbid.indexOf(prop) >= 0;
+    function isForbidden(prop) {
+      return forbid.get(prop) !== undefined;
```
ljharb, line 68:
```suggestion
      return forbid.has(prop);
```
?

PR #2619, opened 2020-04-08
Add support --fix-type

### LINKED ISSUE #2618: Autofix --fix-type layout
I want to use eslint's autofix to fix layouts only.
Like this: `eslint --fix --fix-type layout`
(--fix-type: https://eslint.org/docs/user-guide/command-line-interface#fix-type )
To work that one plugin's rules must contains the type field.
( the type field: https://eslint.org/docs/developer-guide/working-with-rules#rule-basics )
Are there any reasons why this field cannot be added?

svetley1:
1) Add the type field in rule's meta
2) Fix the fixable field

PR resolve #2618


svetley1:
What do you think sorting rules can be attributed to the layout type?

lib/rules/jsx-wrap-multilines.js
```diff
@@ -32,6 +32,7 @@ const PARENS_NEW_LINES = 'Parentheses around JSX should be on separate lines';
 
 module.exports = {
   meta: {
+    type: 'suggestion',
```
golopot, line 35:
```suggestion
    type: 'layout',
```
This rule is similar to [arrow-parens](https://github.com/eslint/eslint/blob/master/lib/rules/arrow-parens.js ) and [no-extra-parens](https://github.com/eslint/eslint/blob/master/lib/rules/no-extra-parens.js). Both of them are considered 'layout'.

lib/rules/forbid-component-props.js
```diff
@@ -19,6 +19,7 @@ const DEFAULTS = ['className', 'style'];
 
 module.exports = {
   meta: {
+    type: 'suggestion',
```
ljharb, line 22: i think this is a "problem" if it's configured to forbid a prop.

lib/rules/forbid-dom-props.js
```diff
@@ -19,6 +19,7 @@ const DEFAULTS = [];
 
 module.exports = {
   meta: {
+    type: 'suggestion',
```
ljharb, line 22:
```suggestion
    type: 'problem',
```

lib/rules/forbid-elements.js
```diff
@@ -14,6 +14,7 @@ const docsUrl = require('../util/docsUrl');
 
 module.exports = {
   meta: {
+    type: 'suggestion',
```
ljharb, line 17:
```suggestion
    type: 'problem',
```

lib/rules/jsx-child-element-spacing.js
```diff
@@ -39,13 +39,14 @@ const INLINE_ELEMENTS = new Set([
...
   meta: {
+    type: 'layout',
     docs: {
       description: 'Ensures inline tags are not rendered without spaces between them',
       category: 'Stylistic Issues',
       recommended: false,
       url: docsUrl('jsx-child-element-spacing')
     },
-    fixable: null,
+    // fixable: null,
```
ljharb, line 49: why is this commented out?

lib/rules/jsx-no-comment-textnodes.js
```diff
@@ -31,6 +31,7 @@ function checkText(node, context) {
 
 module.exports = {
   meta: {
+    type: 'suggestion',
```
ljharb, line 34:
```suggestion
    type: 'problem',
```

lib/rules/jsx-sort-default-props.js
```diff
@@ -16,6 +16,7 @@ const propWrapperUtil = require('../util/propWrapper');
 
 module.exports = {
   meta: {
+    type: 'suggestion',
```
ljharb, line 19: would sorting rules not be "layout"?
```suggestion
    type: 'layout',
```

lib/rules/jsx-sort-props.js
```diff
@@ -203,6 +203,7 @@ function validateReservedFirstConfig(context, reservedFirst) {
 
 module.exports = {
   meta: {
+    type: 'suggestion',
```
ljharb, line 206:
```suggestion
    type: 'layout',
```

lib/rules/jsx-uses-react.js
```diff
@@ -14,6 +14,7 @@ const docsUrl = require('../util/docsUrl');
 
 module.exports = {
   meta: {
+    type: 'suggestion',
```
ljharb, line 17: i'm not sure if this one should have a fix type; it needs to be run *always* to be sure that variables are properly marked as used.

lib/rules/jsx-uses-vars.js
```diff
@@ -13,6 +13,7 @@ const docsUrl = require('../util/docsUrl');
 
 module.exports = {
   meta: {
+    type: 'suggestion',
```
ljharb, line 16: same with this one

lib/rules/no-array-index-key.js
```diff
@@ -16,9 +16,10 @@ const pragma = require('../util/pragma');
 
 module.exports = {
   meta: {
+    type: 'problem',
     docs: {
       description: 'Prevent usage of Array index in keys',
-      category: 'Best Practices',
+      category: 'Possible Errors',
```
ljharb, line 22:
```suggestion
      category: 'Best Practices',
```
it's not an error to use array indexes, it just defeats the point of `key`.

lib/rules/no-typos.js
```diff
@@ -33,6 +33,7 @@ const LIFECYCLE_METHODS = [
 
 module.exports = {
   meta: {
+    type: 'suggestion',
```
ljharb, line 36: this feels more like a "problem"

lib/rules/sort-comp.js
```diff
@@ -82,6 +82,7 @@ function getMethodsOrder(userConfig) {
 
 module.exports = {
   meta: {
+    type: 'suggestion',
```
ljharb, line 85:
```suggestion
    type: 'layout',
```

lib/rules/sort-prop-types.js
```diff
@@ -16,6 +16,7 @@ const propWrapperUtil = require('../util/propWrapper');
 
 module.exports = {
   meta: {
+    type: 'suggestion',
```
ljharb, line 19:
```suggestion
    type: 'layout',
```

lib/rules/static-property-placement.js
```diff
@@ -49,13 +49,14 @@ const schemaProperties = fromEntries(classProperties.map(property => [property,
...
   meta: {
+    type: 'suggestion',
     docs: {
       description: 'Defines where React component static properties should be positioned.',
       category: 'Stylistic Issues',
       recommended: false,
       url: docsUrl('static-property-placement')
     },
-    fixable: null, // or 'code' or 'whitespace'
+    // fixable: null, // or 'code' or 'whitespace'
```
ljharb, line 59: why is this commented out?

lib/rules/void-dom-elements-no-children.js
```diff
@@ -50,6 +50,7 @@ function errorMessage(elementName) {
 
 module.exports = {
   meta: {
+    type: 'suggestion',
```
ljharb, line 53:
```suggestion
    type: 'problem',
```
this is always a bug

lib/rules/forbid-component-props.js
svetley1: I was based on the rule category. If the rule category is "Best Practices" then I specified the "suggestion" type.
If we refer the rule to the “problem” type, then we need to change the category. 
Is not it?

lib/rules/jsx-child-element-spacing.js
svetley1: The string "fixable: null" does not make sense. But maybe someone needs it, so I did not delete it.

lib/rules/jsx-uses-react.js
svetley1: That's the meta type. I think all the rules should have it.

lib/rules/no-array-index-key.js
svetley1: This may result in items not being rendered.(https://reactjs.org/docs/reconciliation.html#recursing-on-children)

lib/rules/no-typos.js
svetley1: But `category: 'Stylistic Issues',`

lib/rules/static-property-placement.js
svetley1: The string "fixable: null" does not make sense. But maybe someone needs it, so I did not delete it.

lib/rules/void-dom-elements-no-children.js
svetley1: Change category to "Possible errors"?

lib/rules/static-property-placement.js
ljharb: ah, fair. in this case, i'd say it can just be deleted then.
```suggestion
```

lib/rules/no-typos.js
ljharb: In the case of react components, i think that category is incorrect.

lib/rules/jsx-child-element-spacing.js
ljharb:
```suggestion
```

lib/rules/jsx-sort-default-props.js
ljharb: hmm, fair point.

lib/rules/forbid-component-props.js
ljharb: I'm not sure that's always true. This rule is a "best practice" rule, but violating an explicitly  best practice is a problem.

lib/rules/jsx-uses-react.js
ljharb: I'm saying that this rule shouldn't, because it shouldn't be filtered out of any category. In other words, if i say "only run problems", and this rule doesn't run, things won't work properly.

lib/rules/no-array-index-key.js
ljharb: > This can work well if the items are never reordered, but reorders will be slow.

Can you provide an example of how it could cause that problem?

lib/rules/jsx-uses-react.js
svetley1: If we do not specify a type, then --fix with --fix-type will ignore this rule. I am not sure that we can specify array of types.
What are your suggestions?

lib/rules/no-array-index-key.js
svetley1: Here: https://jsbin.com/wohima/edit?output

lib/rules/no-typos.js
svetley1, line 36:
```suggestion
    type: 'problem',
```

lib/rules/no-typos.js
```diff
@@ -33,6 +33,7 @@ const LIFECYCLE_METHODS = [
...
     docs: {
       description: 'Prevent common typos',
       category: 'Stylistic Issues',
```
svetley1, line 39:
```suggestion
      category: 'Possible Errors',
```

lib/rules/forbid-component-props.js
svetley1, line 22:
```suggestion
    type: 'problem',
```

lib/rules/no-array-index-key.js
svetley1: Here: https://jsbin.com/wohima/edit?output

lib/rules/jsx-uses-react.js
svetley1: If we do not specify a type, then --fix with --fix-type will ignore this rule. I am not sure that we can specify array of types.
What are your suggestions?

lib/rules/no-array-index-key.js
ljharb: Thanks, that's pretty clear (altho adding an item to the beginning of the list seems like terrible UX).

lib/rules/jsx-uses-react.js
ljharb: hmm - maybe we could file a bug upstream with eslint to allow supporting an array of types, or, to include rules with unspecified types?

ljharb:
Seems like the final issue is the rules that exist to mark things as "used" - I'm pretty concerned that any `fix-type` usage *without* these rules enabled will break people's code.

lib/rules/jsx-uses-react.js
svetley1: The first idea sounds good, but it's a long time. And this bug can be rejected.
And I'm not sure if this is really necessary.

svetley1: Maybe we specify the type as "error" and that's it?

ljharb:
Unfortunately this PR's author has deleted their account, so the PR can't be landed.

sryze:
Would it be OK if someone opened a new PR and just copied the changes from this one? It looks like we can still access the diff

ljharb:
@sryze normally no, but since this one is unrecoverable, sure. please keep the commit authorship, and also please make sure it's rebased :-)

PR #2635, opened 2020-05-07
[New] support eslint v7

ljharb:


toshi-toma:
@ljharb 
Would you mind if I fix tests and `node.start`?

ljharb:
@toshi-toma i restored the test you'd removed; we shouldn't need to remove tests to get this shipped.

That test is the one failing CI, legitimately.

toshi-toma:
@ljharb 

Oh. I understand.

I made a mistake that added the test into valid.
https://github.com/yannickcr/eslint-plugin-react/pull/2635/files#diff-f8e412f530183ea4c0e4bbfd3c3f2b0eR576-R587

We should remove the test from valid. then, should pass the failing test.

toshi-toma:
There is an issue -> #1673

ljharb:
ah, thanks, i understand - the test was incorrectly in "invalid" (#1673 would fix that and make it properly invalid), but the parse error in the test was making it be silently ignored.

PR #2636, opened 2020-05-09
Remove dependency on xregexp

yacinehmito:
Reimplemented the logic of extended regular expressions with functions in the `jsx-pascal-case` rule in order to remove the dependency on `xregexp`. This reduced the size of `node_modules` on my machine from 150MB to 140MB.

----

The dependency on `xregexp` was introduced in #2557 to support the full Unicode range. More specifically, it allows for using the `\p{Lu}` and `\p{Ll}` character classes, which are defined as such:

> \p{Ll} or \p{Lowercase_Letter}: a lowercase letter that has an uppercase variant.
> \p{Lu} or \p{Uppercase_Letter}: an uppercase letter that has a lowercase variant.

([Source](http://www.regular-expressions.info/unicode.html))

However, this logic can be implemented in JS by the respective expressions:
- `char === char.toLowerCase() && char.toUpperCase() !== char.toLowerCase()`
- `char === char.toUpperCase() && char.toUpperCase() !== char.toLowerCase()`

I have used this to replace the regular expressions with simple loops.

----

While working on this pull request, I might have found the origin of a bug. Discussion in this issue: https://github.com/yannickcr/eslint-plugin-react/issues/1334

The code is not the most readable; I tried to strike the right balance between expressivity and performance. I also dumped everything in the rule file.
I am happy to perform any changes that would improve code quality or code organisation following the PR review.

### DIFF
--- a/lib/rules/jsx-pascal-case.js
+++ b/lib/rules/jsx-pascal-case.js
@@ -6,19 +6,58 @@
 'use strict';
 
 const elementType = require('jsx-ast-utils/elementType');
-const XRegExp = require('xregexp');
 const docsUrl = require('../util/docsUrl');
 const jsxUtil = require('../util/jsx');
 
-// ------------------------------------------------------------------------------
-// Constants
-// ------------------------------------------------------------------------------
+function testDigit(char) {
+  const charCode = char.charCodeAt(0);
+  return charCode >= 48 && charCode <= 57;
+}
 
-// eslint-disable-next-line no-new
-const hasU = (function hasU() { try { new RegExp('o', 'u'); return true; } catch (e) { return false; } }());
+function testUpperCase(char) {
+  const upperCase = char.toUpperCase();
+  return char === upperCase && upperCase !== char.toLowerCase();
+}
 
-const PASCAL_CASE_REGEX = XRegExp('^(.*[.])*([\\p{Lu}]|[\\p{Lu}]+[\\p{Ll}0-9]+(?:[\\p{Lu}0-9]+[\\p{Ll}0-9]*)*)$', hasU ? 'u' : '');
-const ALL_CAPS_TAG_REGEX = XRegExp('^[\\p{Lu}0-9]+([\\p{Lu}0-9_]*[\\p{Lu}0-9]+)?$', hasU ? 'u' : '');
+function testLowerCase(char) {
+  const lowerCase = char.toLowerCase();
+  return char === lowerCase && lowerCase !== char.toUpperCase();
+}
+
+function testPascalCase(name) {
+  if (!testUpperCase(name.charAt(0))) {
+    return false;
+  }
+  let atLeastOneLowerCase = false;
+  for (let i = 1; i < name.length; i += 1) {
+    const char = name.charAt(i);
+    if (!(char.toLowerCase() !== char.toUpperCase() || testDigit(char))) {
+      return false;
+    }
+    if (!atLeastOneLowerCase) {
+      atLeastOneLowerCase = testLowerCase(char);
+    }
+  }
+  return atLeastOneLowerCase;
+}
+
+function testAllCaps(name) {
+  const firstChar = name.charAt(0);
+  if (!(testUpperCase(firstChar) || testDigit(firstChar))) {
+    return false;
+  }
+  for (let i = 1; i < name.length - 1; i += 1) {
+    const char = name.charAt(i);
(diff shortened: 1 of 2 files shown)


yacinehmito:
The job is failing on the CI for Node 4 and I don't understand why. https://travis-ci.org/github/yannickcr/eslint-plugin-react/jobs/684908128
🤷

ljharb:
It may be that node 4's regexes don't have the proper native unicode support, which is why xregexp is needed.

yacinehmito:
Yes, and even more than that because even recent RegExp implementations don't support matching on lowercase or uppercase classes.

I didn't want to imply that xregexp was pointless; I wanted to highlight that we can replicate the functionality we get from it using simple JS methods.
It's more code to write and maintain but I believe that side of the tradeoff to be the good one.

lib/rules/jsx-pascal-case.js
```diff
@@ -6,19 +6,60 @@
...
 
-const PASCAL_CASE_REGEX = XRegExp('^(.*[.])*([\\p{Lu}]|[\\p{Lu}]+[\\p{Ll}0-9]+(?:[\\p{Lu}0-9]+[\\p{Ll}0-9]*)*)$', hasU ? 'u' : '');
-const ALL_CAPS_TAG_REGEX = XRegExp('^[\\p{Lu}0-9]+([\\p{Lu}0-9_]*[\\p{Lu}0-9]+)?$', hasU ? 'u' : '');
+function testPascalCase(name) {
+  if (!testUpperCase(name.charAt(0))) {
+    return false;
+  }
+  const {length} = name;
+  let atLeastOneLowerCase = false;
+  for (let i = 1; i < length; i += 1) {
```
ljharb, line 33: caching the length in a for loop in JS is slower, don't try to outsmart the engine.
```suggestion
  let atLeastOneLowerCase = false;
  for (let i = 1; i < name.length; i += 1) {
```

ljharb: should perhaps this loop be a `some` instead of a loop?

lib/rules/jsx-pascal-case.js
```diff
@@ -6,19 +6,60 @@
...
+  return atLeastOneLowerCase;
+}
+
+function testAllCaps(name) {
+  const {length} = name;
+  const firstChar = name.charAt(0);
+  if (!(testUpperCase(firstChar) || testDigit(firstChar))) {
+    return false;
+  }
+  for (let i = 1; i < length - 1; i += 1) {
```
ljharb, line 51:
```suggestion
  const firstChar = name.charAt(0);
  if (!(testUpperCase(firstChar) || testDigit(firstChar))) {
    return false;
  }
  for (let i = 1; i < name.length - 1; i += 1) {
```

lib/rules/jsx-pascal-case.js
yacinehmito: If speed is a concern, I don't know that `some` will help.
However it's not something I care a lot about; I would find `some` more readable.

I'll write it with `some` and you'll tell me if I should keep it or revert.

yacinehmito: I just attempted it and:
- We don't start at `0`. The logic in the loop would make it work even if we were because there is the right overlap between what we test on the first character and what we test on the rest, but that would be confusing to read.
- String doesn't have the `some` method. I would need to do a `.split()` first, which would probably come at a cost.

I'll still do it if you want me to, though.

ljharb: Nah, fair point, a loop is fine.

yacinehmito:
> As for the CI issue, it doesn't seem to be RegExp related; there is a syntax error on an innocuous expression.

You suggestion of removing the caching of `length` solved it at the same time. It was because I was relying on destructuring, and this isn't supported in Node 4.

yacinehmito:
Thank you very much for the prompt review by the way! 🤗

yacinehmito:
> In general, I believe that more dependencies is always better; I trust xregexp to handle edge cases more than some hand-rolled JS functions in one repo.

I share the sentiment. I looked for a package that tests casing and I sadly didn't find any. 😞

yacinehmito:
What is needed to make this PR mergeable?

yacinehmito:
@ljharb Polite bump 🤗

> What is needed to make this PR mergeable?

dadleyy:
FWIW this also removes a dependency on `core-js` which [seems to be in a concerning state](https://github.com/zloirock/core-js/issues/767) currently.

PR #2643, opened 2020-05-17
Relax JSX pragma regexp

### LINKED ISSUE #2642: JSX pragma matching too strict
The current JSX pragma detector will only detect pragmas included at the beginning of a docstring comment. However, pragmas can appear in regular block comments and at any position in them.

Additionally, the current requirement to use docstring comments causes issues with other eslint plugins that lint docstrings, as `@jsx` is not a valid JSDoc identifier.

The solution is to simply relax the regular expression that checks for this a bit.

gfmio:
Fixes #2642.

This PR relaxes the JSX pragma detection slightly.

Previously, only pragmas at the beginning of docstring comments were recognised. But pragmas can occur anywhere in any block comment (not just docstring comments). This PR fixes the regexp that recognises pragmas.

### DIFF
--- a/lib/util/pragma.js
+++ b/lib/util/pragma.js
@@ -5,7 +5,7 @@
 
 'use strict';
 
-const JSX_ANNOTATION_REGEX = /^\*\s*@jsx\s+([^\s]+)/;
+const JSX_ANNOTATION_REGEX = /@jsx\s+([^\s]+)/;
 // Does not check for reserved keywords or unicode characters
 const JS_IDENTIFIER_REGEX = /^[_$a-zA-Z][_$a-zA-Z0-9]*$/;
 
@@ -42,7 +42,7 @@ function getFromContext(context) {
   if (pragmaNode) {
     const matches = JSX_ANNOTATION_REGEX.exec(pragmaNode.value);
     pragma = matches[1].split('.')[0];
-  // .eslintrc shared settings (http://eslint.org/docs/user-guide/configuring#adding-shared-settings)
+    // .eslintrc shared settings (http://eslint.org/docs/user-guide/configuring#adding-shared-settings)
   } else if (context.settings.react && context.settings.react.pragma) {
     pragma = context.settings.react.pragma;
   }
--- a/tests/util/pragma.js
+++ b/tests/util/pragma.js
@@ -1,0 +1,64 @@
+'use strict';
+
+const assert = require('assert');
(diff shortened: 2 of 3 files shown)


gfmio:
I actually just realised that the previous pragma matcher has a bug anyway, as it would recognise line comments of the form `//* @jsx myPragma` as valid even though they're not.

gfmio:
@ljharb I've added the tests

lib/util/pragma.js
```diff
@@ -5,19 +5,20 @@
...
   let pragma = 'createReactClass';
   // .eslintrc shared settings (http://eslint.org/docs/user-guide/configuring#adding-shared-settings)
   if (context.settings.react && context.settings.react.createClass) {
     pragma = context.settings.react.createClass;
   }
   if (!JS_IDENTIFIER_REGEX.test(pragma)) {
-    throw new Error(`createClass pragma ${pragma} is not a valid function name`);
+    throw new Error(
+      `createClass pragma ${pragma} is not a valid function name`
+    );
```
ljharb, line 21: this looks like `prettier` nonsense; let's revert it please
```suggestion
    throw new Error(`createClass pragma ${pragma} is not a valid function name`);
```

lib/util/pragma.js
```diff
@@ -38,12 +39,14 @@ function getFromContext(context) {
   let pragma = 'React';
 
   const sourceCode = context.getSourceCode();
-  const pragmaNode = sourceCode.getAllComments().find((node) => JSX_ANNOTATION_REGEX.test(node.value));
+  const pragmaNode = sourceCode
+    .getAllComments()
+    .find((node) => JSX_ANNOTATION_REGEX.test(node.value));
```
ljharb, line 44:
```suggestion
  const pragmaNode = sourceCode.getAllComments().find((node) => JSX_ANNOTATION_REGEX.test(node.value));
```

tests/util/pragma.js
```diff
@@ -0,0 +1,64 @@
+'use strict';
+
+const assert = require('assert');
+const SourceCode = require('eslint').SourceCode;
+const espree = require('espree');
```
ljharb, line 5: is there any way to do this with eslint itself, rather than adding a dev dep on espree?

gfmio: Not as far as I could tell, unfortunately. I tried using eslint first, but eslint does not expose the API to generate the AST object needed to create the SourceCode object.

PR #2661, opened 2020-06-07
[Fix] no-unused-prop-types with typescript eslint parser

eltonio450:
Hi there,

This is a first draft to solve https://github.com/yannickcr/eslint-plugin-react/issues/2569

I would be thankful if someone involved in this rule creation could give feedback about edge cases to be handled (spread operator, etc.). I am afraid to have forgotten some...

Thank you !

### DIFF
--- a/lib/util/propTypes.js
+++ b/lib/util/propTypes.js
@@ -4,6 +4,8 @@
 
 'use strict';
 
+const flatMap = require('array.prototype.flatmap');
+
 const annotations = require('./annotations');
 const propsUtil = require('./props');
 const variableUtil = require('./variable');
@@ -281,6 +283,47 @@ module.exports = function propTypesInstructions(context, components, utils) {
     return ignorePropsValidation;
   }
 
+  function declarePropTypesForTSTypeAnnotation(propTypes, declaredPropTypes) {
+    let foundDeclaredPropertiesList = [];
+    if (propTypes.typeAnnotation.type === 'TSTypeReference') {
+      const typeName = propTypes.typeAnnotation.typeName.name;
+      if (!typeName) {
+        return true;
+      }
+      // the game here is to find the type declaration in the code
+      const candidateTypes = context.getSourceCode().ast.body.filter((item) => item.type === 'VariableDeclaration' && item.kind === 'type');
+      const declarations = flatMap(candidateTypes, (type) => type.declarations);
+
+      // we tried to find either an interface or a type with the TypeReference name
+      const typeDeclaration = declarations.find((dec) => dec.id.name === typeName);
+      const interfaceDeclaration = context.getSourceCode().ast.body.find((item) => item.type === 'TSInterfaceDeclaration' && item.id.name === typeName);
+
+      if (typeDeclaration) {
+        foundDeclaredPropertiesList = typeDeclaration.init.members;
+      } else if (interfaceDeclaration) {
+        foundDeclaredPropertiesList = interfaceDeclaration.body.body;
+      } else {
+        // type not found, for example can be an exported type, etc. Can issue a warning in the future.
+        return true;
+      }
+    } else if (propTypes.typeAnnotation.type === 'TSTypeLiteral') {
+      foundDeclaredPropertiesList = propTypes.typeAnnotation.members;
+    } else {
+      // weird cases such as TSTypeFunction
+      return true;
+    }
+
+    foundDeclaredPropertiesList.forEach((tsPropertySignature) => {
+      declaredPropTypes[tsPropertySignature.key.name] = {
+        fullName: tsPropertySignature.key.name,
+        name: tsPropertySignature.key.name,
+        node: tsPropertySignature,
+        isRequired: !tsPropertySignature.optional
+      };
+    });
+    return false;
+  }
+
   /**
    * Marks all props found inside IntersectionTypeAnnotation as declared.
    * Since InterSectionTypeAnnotations can be nested, this handles recursively.
@@ -546,6 +589,9 @@ module.exports = function propTypesInstructions(context, components, utils) {
         } else {
           ignorePropsValidation = true;
         }
+        break;
+      case 'TSTypeAnnotation':
+        ignorePropsValidation = declarePropTypesForTSTypeAnnotation(propTypes, declaredPropTypes);
         break;
       case null:
         break;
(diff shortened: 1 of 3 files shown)


lib/util/propTypes.js
```diff
@@ -281,6 +281,44 @@ module.exports = function propTypesInstructions(context, components, utils) {
...
+    let members = [];
+
+    if (propTypes.typeAnnotation.type === 'TSTypeReference') {
+      const typeName = propTypes.typeAnnotation.typeName.name;
+      if (!typeName) {
+        return true;
+      }
+      // the game here is to find the type declaration in the code
+      const candidateTypes = context.getSourceCode().ast.body.filter((item) => item.type === 'VariableDeclaration' && item.kind === 'type');
+      const declarations = candidateTypes.map((type) => type.declarations).reduce((acc, val) => acc.concat(val), []);
```
ljharb, line 294:
```suggestion
      const declarations = flatMap(candidateTypes, (type) => type.declarations);
```
where `flatMap` is `array.prototype.flatmap`?

lib/util/propTypes.js
```diff
@@ -281,6 +281,44 @@ module.exports = function propTypesInstructions(context, components, utils) {
...
+
+      if (!type) {
+        return true;
+      }
+
+      members = type.init.members;
+    } else if (propTypes.typeAnnotation.type === 'TSTypeLiteral') {
+      members = propTypes.typeAnnotation.members;
+    } else {
+      return true; // we don't handle cases where the type is declared directly. Can be improved in the future, with TSTypeLiteral for example
```
ljharb, line 305: What would be examples of this?

lib/util/propTypes.js
```diff
@@ -281,6 +281,44 @@ module.exports = function propTypesInstructions(context, components, utils) {
...
+    }
+
+    members.forEach((member) => {
+      declaredPropTypes[member.key.name] = {
+        fullName: member.key.name,
+        name: member.key.name,
+        node: member
+        // node: propNode
```
ljharb, line 313:
```suggestion
```

lib/util/propTypes.js
eltonio450: not sure, but weird cases such as type TSFunctionType

```
{
      // neither TSTypeReference or TSTypeLiteral, we do nothing. Weird case
      code: [
        'const Hello = (props: () => any) => {',
        '    return <div>{props.firstname}</div>;',
        '}'
      ].join('\n'),
      parser: parsers.TYPESCRIPT_ESLINT
    }
```
on the other hand, member is empty, so declaredPropsTypes ends up at {}

What is better ? ignoring props validation or returning an empty declaredPropsTypes ?

eltonio450: oops, sorry, this is the response for the comment below :)

lib/util/propTypes.js
eltonio450: see above

lib/util/propTypes.js
eltonio450: for the flatMap, it needs to be polyfilled for node < 11, and I think it's not the case here. 

It does not work on my computer with node 10 for example

ljharb: Yes, we already have `array.prototype.flatmap` in our deps, so it works down to IE 6 via the import :-)

lib/util/propTypes.js
ljharb: hmm. so the choice is between:
1. possibly give a false positive, which would force the user to change their code how?
1. silently accept something that should be warned on
?

lib/util/propTypes.js
eltonio450: ok cool! trying that asap :)

lib/util/propTypes.js
eltonio450: I think both cases are weird, but I have a slight preference for 1 :
1. setting ignore at true : what happens in this case ? does it emit a warning ? Maybe in the future we could signal the user that the props are not checked because of improper declaration/compatibility issue with the rule ? It could be an option, like "props validation has been ignored for this component" no ?
2. declaring props as {} (ignore = false, members = []): in this case, it fails silently since there is no unused props in {}. 

I think 1 gives a bit more perspective for this rule, as 2 just swallows the error and the next maintainer will have to dig in the code to understand why declared props are {} (as I had to do to fix the issue here)

what do you think :) ?

eltonio450:
ok I pushed a new version with : 
- flatMap
- handling interfaces props declaration (and a test)
- better variable names

eltonio450:
ok, I don't get how to make flatmap work :(

lib/util/propTypes.js
ljharb: seems like the first option is better

ljharb:
@eltonio450 i can fix the branch, but since you made the PR from `ovrsea` and not from `eltonio450`, i can't. can you add me to https://github.com/ovrsea/eslint-plugin-react so i can force push to it?

eltonio450:
done :) thank you :)

lib/util/propTypes.js
eltonio450: ok i leave it as it is now then 😃

lib/util/propTypes.js
```diff
@@ -281,6 +283,49 @@ module.exports = function propTypesInstructions(context, components, utils) {
...
+      const interfaceDeclaration = context.getSourceCode().ast.body.find((item) => item.type === 'TSInterfaceDeclaration' && item.id.name === typeName);
+
+
+      if (typeDeclaration) {
+        foundDeclaredPropertiesList = typeDeclaration.init.members;
+      } else if (interfaceDeclaration) {
+        foundDeclaredPropertiesList = interfaceDeclaration.body.body;
+      } else {
+        // type not found, for example can be an exported type, etc. Can issue a warning in the future.
+        return true;
```
ljharb, line 308: it would be great to add test cases to cover this, so we're aware when it changes in the future

lib/util/propTypes.js
```diff
@@ -281,6 +283,49 @@ module.exports = function propTypesInstructions(context, components, utils) {
...
+      }
+    } else if (propTypes.typeAnnotation.type === 'TSTypeLiteral') {
+      foundDeclaredPropertiesList = propTypes.typeAnnotation.members;
+    } else {
+      // weird cases such as TSTypeFunction
+      return true;
```
ljharb, line 314: it'd be great to include test cases for this, as well (or at least, an example of "weird" in the code)

eltonio450:
thank you !

ljharb:
no no, thank *you*!

PR #2664, opened 2020-06-08
Added support (without options) for spread arguments with keys

machineghost:
This is just a first step to try and solve: https://github.com/yannickcr/eslint-plugin-react/issues/1753

This is ONLY a request for code review: please do not merge.  I just wanted to get feedback on what I have so far before I add in options.  Also I wanted to get your thoughts on whether we should have two options or just one.

It seems like we definitely want to guard this new flexibility with some sort of "allow spread keys" option, but then do we want just one, or do we want to allow *explicit* spread keys:

    <Foo {...{ key }} />

separately from *implicit* ones:

    <Foo {...someObjectThatPresumablyHasAKey} />

Hypothetically speaking  I could see a dev wanting to allow the former but not the latter ... but maybe that's YAGNI?

### DIFF
--- a/tests/lib/rules/jsx-key.js
+++ b/tests/lib/rules/jsx-key.js
@@ -39,6 +39,8 @@ ruleTester.run('jsx-key', rule, {
     {code: 'fn()'},
     {code: '[1, 2, 3].map(function () {})'},
     {code: '<App />;'},
+    {code: '[1, 2, 3].map(x => <App {...{ key }} />);'},
+    {code: '[1, 2, 3].map(x => <App {...objectWithKey} />);'},
     {code: '[<App key={0} />, <App key={1} />];'},
     {code: '[1, 2, 3].map(function(x) { return <App key={x} /> });'},
     {code: '[1, 2, 3].map(x => <App key={x} />);'},
--- a/tests/lib/rules/jsx-key.js
+++ b/tests/lib/rules/jsx-key.js
@@ -39,8 +39,14 @@ ruleTester.run('jsx-key', rule, {
     {code: 'fn()'},
     {code: '[1, 2, 3].map(function () {})'},
     {code: '<App />;'},
-    {code: '[1, 2, 3].map(x => <App {...{ key }} />);'},
-    {code: '[1, 2, 3].map(x => <App {...objectWithKey} />);'},
+    {
+      code: '[1, 2, 3].map(x => <App {...{ key }} />);',
+      options: [{allowSpreadKeys: true}]
+    },
+    {
+      code: '[1, 2, 3].map(x => <App {...objectWithKey} />);',
+      options: [{allowSpreadKeys: true}]
+    },
     {code: '[<App key={0} />, <App key={1} />];'},
     {code: '[1, 2, 3].map(function(x) { return <App key={x} /> });'},
     {code: '[1, 2, 3].map(x => <App key={x} />);'},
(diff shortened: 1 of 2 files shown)


tests/lib/rules/jsx-key.js
```diff
@@ -39,6 +39,8 @@ ruleTester.run('jsx-key', rule, {
     {code: 'fn()'},
     {code: '[1, 2, 3].map(function () {})'},
     {code: '<App />;'},
+    {code: '[1, 2, 3].map(x => <App {...{ key }} />);'},
+    {code: '[1, 2, 3].map(x => <App {...objectWithKey} />);'},
```
ljharb, line 43: this should fail, i think, unless we want to try to track down `objectWithKey` and ensure it is an object literal with a hardcoded `key` property.

tests/lib/rules/jsx-key.js
```diff
@@ -39,6 +39,8 @@ ruleTester.run('jsx-key', rule, {
     {code: 'fn()'},
     {code: '[1, 2, 3].map(function () {})'},
     {code: '<App />;'},
+    {code: '[1, 2, 3].map(x => <App {...{ key }} />);'},
```
ljharb, line 42:
```suggestion
    {code: '[1, 2, 3].map(x => <App {...{ key }} />);'},
    {code: '[1, 2, 3].map(x => <App {...{ key: x.id }} />);'},
    {code: '[1, 2, 3].map(x => <App {...{ "key": x.id }} />);'},
    {code: '[1, 2, 3].map(x => <App {...{ 'key': x.id }} />);'},
    {code: '[1, 2, 3].map(x => <App {...{ ['key']: x.id }} />);'},
    {code: '[1, 2, 3].map(x => <App {...{ ["key"]: x.id }} />);'},
    {code: '[1, 2, 3].map(x => <App {...{ [`key`]: x.id }} />);'},
```

ljharb (commented):
I'll wait to review the rule changes until you've marked the PR as ready for review - but the direction seems fine to me.

tests/lib/rules/jsx-key.js
machineghost: I'll have to confess my ignorance of ESLint here: I didn't realize that latter option was even possible.

I thought we'd either have to blanket allow or deny such spread objects .... and I'd argue there *is* value in letting a dev say "I use ESLint as a tool to help me catch problems, but I have no problem remembering to put keys in my objects, so I'd prefer ES Lint not bother me with pointless error messages every time I spread an object" :)

But of course, if it's possible to make ESLint do something that would make any dev happy, and only warn about the legitimately problematic cases, that'd be best.  Can you point me to any rules that do something similar so I could get an idea how?

machineghost:
Since you never responded about the spread stuff I just made it an option, but that can easily change.

tests/lib/rules/jsx-key.js
ljharb: When the object is defined in the same file, as a literal, eslint provides the tools to read it as if it was defined inline.

Also, it's often better to wrongly block something than to wrongly allow it.

I'm not sure off the top of my head which rules do similar things; i'd look into our propType detection code.

machineghost:
From the lack of response (one way or the other) here, is it safe to assume that I will need to fork the library if I want this change published (and just in general if I want to see other improvements made to it)?

ljharb:
@machineghost sorry for the delay; what are you still waiting for a response on? I do have a reply [here](https://github.com/yannickcr/eslint-plugin-react/pull/2664#discussion_r446567651) that you haven't replied to yet?

PR #2670, opened 2020-06-13
[New] add `jsx-no-random-key` rule

pustovalov:
This rule prevent usage of random value for `key` prop

https://reactjs.org/docs/reconciliation.html#tradeoffs

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,7 @@ Enable the rules that you would like to use.
 * [react/jsx-no-comment-textnodes](docs/rules/jsx-no-comment-textnodes.md): Comments inside children section of tag should be placed inside braces
 * [react/jsx-no-duplicate-props](docs/rules/jsx-no-duplicate-props.md): Enforce no duplicate props
 * [react/jsx-no-literals](docs/rules/jsx-no-literals.md): Prevent using string literals in React component definition
+* [react/jsx-no-random-key](docs/rules/jsx-no-random-key.md): Prevent usage of random value for `key` prop
 * [react/jsx-no-script-url](docs/rules/jsx-no-script-url.md): Forbid `javascript:` URLs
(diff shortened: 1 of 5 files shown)


ljharb:
I'm pretty confused about this one. That it's a function call in no way implies it's a "random value" - it's quite impossible to statically know if a function call produces a random value, or one that's static.

I don't think proving key determinism is something that is possible to do with a linter; it requires runtime tests.

pustovalov:
I agree that this rule cannot determine exactly what value will be for `key`, but many people use these libraries in such a way:

https://github.com/search?p=1&q=key%3D%7Bnanoid%28%29%7D&type=Code
https://github.com/search?q=key%3D%7Bshortid.generate%28%29%7D&type=Code
https://github.com/search?q=key%3D%7Bcuid%28%29%7D&type=Code
https://github.com/search?q=key%3D%7Buuidv1%28%29%7D&type=Code
https://github.com/search?p=3&q=key%3D%7Buuidv3%28%29%7D&type=Code
https://github.com/search?p=2&q=key%3D%7Buuidv4%28%29%7D&type=Code
https://github.com/search?p=4&q=key%3D%7BMath.random%28&type=Code



some libraries even describe that you can’t use them for key:
https://github.com/ai/nanoid/blob/master/README.md#react

I thought that this rule would be useful when I saw this use in my team.

this rule can be simplified:
https://github.com/pustovalov/eslint-plugin-react/commit/c47f7d3476c3b5e3d66821fe7c38225638e392b3

ljharb:
Sure, but "it's a function call" isn't why that's a problem, it's because of *what those functions do*, and that's not something that's lintable.

pustovalov:
what about this one: restrict certain function calls for key property, is this better or still doesn't make sense?

PR #2676, opened 2020-06-16
Forbid objectof 2662

ajkovar:
In this case the type of the callee was 'Identifier' (vs 'MemberExpression'), which seems to makes sense because it is not being called as a property on PropTypes object.  I looked back at the history to see if maybe that check was added for a particular reason but it seems like it has been there from the beginning:

https://github.com/yannickcr/eslint-plugin-react/commit/00f716a990c247e844a5dba85220d3af7d6e3693

I can't think of any reason why it would be needed but if someone who knows better than me does, I'd be happy to hear it.

### DIFF
--- a/lib/rules/forbid-prop-types.js
+++ b/lib/rules/forbid-prop-types.js
@@ -59,6 +59,15 @@ module.exports = {
       return forbid.indexOf(type) >= 0;
     }
 
+    function reportIfForbidden(type, declaration, target) {
+      if (isForbidden(type)) {
+        context.report({
+          node: declaration,
+          message: `Prop type \`${target}\` is forbidden`
+        });
+      }
+    }
+
     function shouldCheckContextTypes(node) {
       if (checkContextTypes && propsUtil.isContextTypesDeclaration(node)) {
         return true;
@@ -93,10 +102,10 @@ module.exports = {
         ) {
           value = value.object;
         }
-        if (
-          value.type === 'CallExpression'
-          && value.callee.type === 'MemberExpression'
-        ) {
+        if (value.type === 'CallExpression') {
+          value.arguments.forEach((arg) => {
+            reportIfForbidden(arg.name, declaration, target);
+          });
           value = value.callee;
         }
         if (value.property) {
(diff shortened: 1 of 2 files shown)


ljharb:
@ajkovar so i notice you modified my failing test; i rebased this PR so that it's 1 commit, and we have both tests. The one that's failing still needs fixing, though.

ajkovar:
Sure.  Thank you for reviewing.  I think I had assumed there was a typo.  Let me take a look and see.

ajkovar:
Ok.  I see now.  Investigating.  Thanks for adding the additional test.

ajkovar:
Alright.  So I fixed those original test cases.  I think it might be necessary to tweak the algorithm a bit though to make it recursive or something of that nature.  That I think would require a larger refactor so I held off for now.  I added another test case to demonstrate what I mean.  There might be an easier shortcut.  Would be interested to hear your thoughts.

ajkovar:
Ok.  I have an update for you guys.  After reviewing the code further, my thoughts about recursion seem to be unnecessary since es-lint is already recursively descending through the AST.   I added another selector to handle calls to the 'shape' function specifically.  I think this should be sufficient but if anyone can think of why not, I would love more about it.  All tests in forbid-prop-types now pass (although tests seem to be broken on master). 

In this process, this seemed to fix another test that was passing but in a broken way (with a comment).  This *does not* fix  [#1673](https://github.com/yannickcr/eslint-plugin-react/issues/1673).  It doesn't seem to me that the test was accurately covering what was being requested in that ticket.  That test *should* have failed but for a different reason ('PropTypes.object' is forbidden).

lib/rules/forbid-prop-types.js
```diff
@@ -161,6 +165,14 @@ module.exports = {
         checkNode(node.parent.right);
       },
 
+      CallExpression(node) {
+        if ((node.callee.name === 'shape'
+             || astUtil.getPropertyName(node.callee) === 'shape')
+           && node.arguments.length > 0) {
```
ljharb, line 171:
```suggestion
        if (
          node.arguments.length > 0
          && (node.callee.name === 'shape' || astUtil.getPropertyName(node.callee) === 'shape')
         ) {
```

PR #2677, opened 2020-06-23
Doc disabled example

enBonnet:


ljharb:
Per similar discussion here: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/715#issuecomment-648340645, closing.

PR #2678, opened 2020-06-25
[Tests] `no-typos`:  move a valid case into valid section

toshi-toma:
It seems CI is failing.

```
  4345 passing (12s)
  1 failing
  1) no-typos
       invalid
         
     import 'react';
     class Component extends React.Component {};
   :
     AssertionError [ERR_ASSERTION]: Invalid cases must have at least one error
      at testInvalidTemplate (node_modules/eslint/lib/rule-tester/rule-tester.js:648:24)
      at Context.RuleTester.it (node_modules/eslint/lib/rule-tester/rule-tester.js:883:25)
      at processImmediate (internal/timers.js:443:21)
```

I moved a valid case into a valid section.

### DIFF
--- a/tests/lib/rules/no-typos.js
+++ b/tests/lib/rules/no-typos.js
@@ -8,6 +8,8 @@
 // Requirements
 // -----------------------------------------------------------------------------
 
+const semver = require('semver');
+const eslintPkg = require('eslint/package.json');
 const RuleTester = require('eslint').RuleTester;
 const rule = require('../../../lib/rules/no-typos');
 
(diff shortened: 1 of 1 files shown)


toshi-toma:
@ljharb 

Oh, Thanks. I understand.
But This case does not report in `no-typoes` rule. So, We can't fill in the errors field. ESLint will throw an error (AssertionError [ERR_ASSERTION]: Invalid cases must have at least one error)

```
import 'react';
class Component extends React.Component {};
```

Do you have any good ideas?

ljharb:
`import 'react'` should be the typo.

toshi-toma:
`no-typos` seems does not report ImportDeclaration. only PropTypes, lifecycle methods, and class properties...

Do you think should I add the report for ImportDeclaration??
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-typos.md

ljharb:
Hmm - looks like the test fails in eslint v7.3 but passes in eslint v7.2.

ljharb:
I'll just pin travis-ci to v7.2 for now, but we should fix the real bug.

ljharb:
(maybe i have to pin to 7.1, since 7.2 breaks in older nodes)

PR #2679, opened 2020-06-29
[Fix]`no-unused-prop-types`:handle optional chaining

hank121314:
Handle optional chaining in usedPropTypes.

Before:
```
const Hello = props => {
  const firstname = props?.name;
  return <div />;
};

Hello.propTypes = {
  name: PropTypes.string.isRequired, <<< 'name' PropType is defined but prop is never used eslint(react/no-unused-prop-types)
};
```
After this pr the error should be resolved.


ljharb:
I'll have the fix released later today.

ljharb:
v7.20.2 is released.

PR #2692, opened 2020-06-30
[New] : Avoid unsafe global window use

### LINKED ISSUE #2057: Rule proposal: ssr-friendly
In order to be SSR compliant, you can only access browser globals such as `window`, `document` and so on, inside of `componentDidMount`, `componentDidUpdate` and `componentWillUnmount`

screen shot 2018-11-26 at 11 05 04 am

A rule could be created to avoid the use of unguarded browser globals in other lifecycle methods.

Example of a guard:

```js
if (typeof window !== 'undefined') {
 // I can use browser globals in here
}
```

jzabala:
This rule will prevent the unsafe use on `window` resulting from rendering React apps in environments where the global is not available.

I'll be grateful for getting all your comments and suggestions before summiting the final version.

### Tested Conditions

- [x] used in a guard
- [x] used in componentDidMount, componentDidUpdate and componentWillUnmount
- [x] used in useEffect, useLayoutEffect and useCallback
- [x] used in an event handler

### Considerations

The issue talks about disabling the use of various globals variables aside from `window`. I think an easier solution would be to use the rule [no-restricted-globals](https://eslint.org/docs/rules/no-restricted-globals) to force the use of `window` in the variables that we want.

Another solution would be to have the option of setting what globals will be disabled directly in this plugin with `window` as default. And to change the name of the rule to something like: `no-unsafe-globals-use`.

Let me know your opinion about this or if you have other ideas 

Closes #2057

### DIFF
--- a/lib/rules/no-unsafe-window-use.js
+++ b/lib/rules/no-unsafe-window-use.js
@@ -1,0 +1,148 @@
+/**
+ * @fileoverview Rule to avoid unsafe global window use
+ * @author Johnny Zabala
+ */
+
+'use strict';
+
+const docsUrl = require('../util/docsUrl');
+
+// ------------------------------------------------------------------------------
+// Helpers
+// ------------------------------------------------------------------------------
+
+function isMethod(node) {
+  return node.type === 'Property'
+    || node.type === 'MethodDefinition'
+    || node.type === 'ClassProperty';
+}
+
+function isAllowedLifecycle(node) {
+  return ['componentDidMount', 'componentDidUpdate', 'componentWillUnmount'].includes(node.key.name);
+}
+
+function isAllowedHook(node) {
+  return (node.type === 'CallExpression')
+  && ['useEffect', 'useLayoutEffect'].includes(node.callee.name);
+}
+
+// guard: typeof window !== 'undefined'
+function isGuard(node) {
+  return node.type === 'BinaryExpression'
+    && ['!==', '!='].includes(node.operator)
+    && node.left.type === 'UnaryExpression'
+    && node.left.operator === 'typeof'
+    && node.right.type === 'Literal'
+    && node.right.value === 'undefined';
+}
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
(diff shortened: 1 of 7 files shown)


lib/rules/no-unsafe-window-use.js
```diff
@@ -0,0 +1,148 @@
...
+    || node.type === 'ClassProperty';
+}
+
+function isAllowedLifecycle(node) {
+  return ['componentDidMount', 'componentDidUpdate', 'componentWillUnmount'].includes(node.key.name);
+}
+
+function isAllowedHook(node) {
+  return (node.type === 'CallExpression')
+  && ['useEffect', 'useLayoutEffect'].includes(node.callee.name);
```
ljharb, line 26: `.includes` won't work on node 4; use the `array-includes` package. or, for such short lists, just use `||` and `===` comparisons :-)

ljharb (commented):
This looks like a good start :-)

lib/rules/no-unsafe-window-use.js
jzabala: Done!

jzabala:
> This looks like a good start :-)

Thanks for your initial review @ljharb. 

I think it is ready now 🙂

osdiab:
i'm interested in this rule or something like it getting published - anything that needs to be done to help it along?

osdiab:
FYI currently this errors incorrectly:

```
typeof window === "undefined" ? undefined : window.location.href
```

Also shadowing `window` still causes errors, like so:

```
function getWindow() {
  return typeof window !== "undefined" ? window : undefined;
}

// another file
const window = getWindow(); // this line errors but pretty sure it's actually fine
console.log(window?.location.href); // this errors too
```

though just giving it a different name works fine.

osdiab:
also would probably make sense to check for `document`, but yeah maybe that makes more sense for `no-restricted-globals`

ljharb (commented):
Seems like this also needs handling for hooks in SFCs, to ensure window and whatnot is only used in effects?

I'm not stoked on the use of a class and `this`, or all the loops, in the implementation, but that's something we can worry about later.

PR #2696, opened 2020-07-06
[Fix]:`prop-types`: handle component returning null

### LINKED ISSUE #2705: 'missing in props validation' when returning with a ternary operator
It took me a bit to figure out why I was getting a 'missing in props validation' error in my project. Turns out, returning with a ternary operator was causing the linter to produce this error. Refactoring a bit fixes it. Can anyone else reproduce this?

```javascript
// @flow
import React from 'react'

type Props = {
  x: string
}

const val = true

// ************* The following produces:
// 'x' is missing in props validation - eslintreact/prop-types
export const problem = ({ x }: Props) => {
  return val ? (
    <div>
      {x}
    </div>
  ) : (
    <div>Nothing</div>
  )
}

// ************ This does not cause an issue
export const notProblem = ({ x }: Props) => {
  const div = val ? (
    <div>
      {x}
    </div>
  ) : (
    <div>Nothing</div>
  )
  return div
}

// ************* Nor does this
export const alsoNotProblem = ({ x }: Props) => {
  if (val) {
    return (
      <div>
        {x}
      </div>
    )
  } else {
    return (
      <div>Nothing</div>
    )
  }
}
```

My eslint config is:
```json
{
"eslintConfig": {
    "env": {
      "browser": true,
      "jest": true,
      "es6": true
    },
    "extends": [
      "plugin:react/recommended",
      "plugin:flowtype/recommended",
      "standard",
      "react-app"
    ],
    "globals": {
      "Atomics": "readonly",
      "SharedArrayBuffer": "readonly"
    },
    "parserOptions": {
      "ecmaFeatures": {
        "jsx": true
      },
      "ecmaVersion": 2018,
      "sourceType": "module"
    },
    "plugins": [
      "react",
      "flowtype",
      "simple-import-sort"
    ],
    "settings": {
      "react": {
        "version": "detect"
      },
      "import/resolver": {
        "node": {
          "moduleDirectory": [
            "node_modules",
            "src/"
          ]
        }
      },
      "editor.codeActionsOnSave": {
        "source.fixAll.eslint": true
      }
    },
    "rules": {
      "simple-import-sort/sort": ["error",
        {
          "groups": [
            ["^react$", "^\\u0000", "^[^A-Z].*"],
            ["^[A-Z].*"],
            ["^\\."]
          ]
        }]
    }
  }
}
```

Package versions:
```
    "eslint": "^7.4.0",
    "eslint-config-react-app": "^5.2.1",
    "eslint-config-standard": "^14.1.1",
    "eslint-plugin-flowtype": "^5.2.0",
    "eslint-plugin-import": "^2.22.0",
    "eslint-plugin-node": "^11.1.0",
    "eslint-plugin-promise": "^4.2.1",
    "eslint-plugin-react": "^7.20.3",
    "eslint-plugin-simple-import-sort": "^5.0.3",
    "eslint-plugin-standard": "^4.0.1",
```

hank121314:
This pr is related to https://github.com/yannickcr/eslint-plugin-react/issues/2654#issuecomment-652868302
Previous isJSXFunctionComponent did not handle nullable component correctly.
So I replaced it with utils.isReturningJSXOrNull.
Sorry for the issue.

Fixes #2705.

### DIFF
--- a/tests/lib/rules/prop-types.js
+++ b/tests/lib/rules/prop-types.js
@@ -2524,7 +2524,52 @@ ruleTester.run('prop-types', rule, {
       MyComponent.propTypes = {
         hello: PropTypes.string.isRequired,
       };
-    `
+    `,
+    {
+      code: `
+      interface Props {
+        value?: string;
+      }
+
+      // without the | null, all ok, with it, it is broken
+      function Test ({ value }: Props): React.ReactElement<Props> | null {
+        if (!value) {
+          return null;
+        }
+
+        return <div>{value}</div>;
+      }`,
+      parser: parsers.TYPESCRIPT_ESLINT
+    },
+    {
+      code: `
+      interface Props {
+        value?: string;
+      }
+
+      // without the | null, all ok, with it, it is broken
+      function Test ({ value }: Props): React.ReactElement<Props> | null {
+        if (!value) {
+          return <div>{value}</div>;;
+        }
+
+        return null;
+      }`,
+      parser: parsers.TYPESCRIPT_ESLINT
+    },
+    {
+      code: `
+      interface Props {
+        value?: string;
+      }
+      const Hello = (props: Props) => {
+        if (props.value) {
+          return <div></div>;
+        }
+        return null;
+      }`,
+      parser: parsers.TYPESCRIPT_ESLINT
+    }
   ],
 
   invalid: [
@@ -5100,6 +5145,21 @@ ruleTester.run('prop-types', rule, {
       `,
       errors: [{
         message: '\'foo.baz\' is missing in props validation'
+      }]
+    },
+    {
+      code: `
+      interface Props {
+      }
+      const Hello = (props: Props) => {
+        if (props.value) {
+          return <div></div>;
+        }
+        return null;
+      }`,
+      parser: parsers.TYPESCRIPT_ESLINT,
+      errors: [{
+        message: '\'value\' is missing in props validation'
       }]
     }
   ]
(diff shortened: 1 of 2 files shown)


tests/lib/rules/prop-types.js
```diff
@@ -2493,8 +2493,7 @@ ruleTester.run('prop-types', rule, {
     {
       code: `
         interface Props {
-          'aria-label': string // 'undefined' PropType is defined but prop is never used eslint(react/no-unused-prop-types)
-          // 'undefined' PropType is defined but prop is never used eslint(react-redux/no-unused-prop-types)
+          'aria-label': string;
         }
```
ljharb, line 2497: why does this test case have to change? i realize they're just comments but it's nice when test diffs are only green :-)

tests/lib/rules/no-unused-prop-types.js
```diff
@@ -3404,7 +3404,7 @@ ruleTester.run('no-unused-prop-types', rule, {
         ': Props): JSX.Element {',
         'const handleVerifySubmit = ({',
         '  otp,',
-        ' }) => {',
+        ' }: {otp:string}) => {',
```
ljharb, line 3407: why does this need to change?

hank121314: I am embarrassed to say that this is a typo. 😭 
This test case is trying to test the issue from https://github.com/yannickcr/eslint-plugin-react/issues/2689.
I added it back in this pr.
I will change the pr title.

tests/lib/rules/prop-types.js
hank121314: OK! I will add it back!

ljharb:
@hank121314 turns out this caused #2728.

hank121314:
Hi @ljharb !
I think this pr did not cause that issue.
https://github.com/yannickcr/eslint-plugin-react/pull/2699 is more related.

PR #2699, opened 2020-07-06
[Fix]: Functional components first letter capitalization check

### LINKED ISSUE #2352: eslint showing error when using variable from parent scope

* **ESLint Version:** 6.0.1
* **Node Version:** v10.15.3
* **npm Version:** 6.4.1

**What parser (default, Babel-ESLint, etc.) are you using?**
default 

<details>

```js

```

</details>

```js
import React from 'react';

interface Props {
    readonly history: number[];
}

export const MyHistory = (props: Props) => {
    const renderContent = () => {
        if (props.history.length > 0) {
            return <div> My history here</div>
        }
        return <div> No historical events</div>
    };

    return renderContent();
};
```

```bash
npm run eslint --fix --ext .tsx --ext .ts src
```

eslint show no errors

```
  9:19  error  'history' is missing in props validation         react/prop-types
  9:27  error  'history.length' is missing in props validation  react/prop-types
```

Additional info: issue first created on `eslint` repo https://github.com/eslint/eslint/issues/12004

### LINKED ISSUE #2495: Understanding "PropType is defined but prop is not never used" + Formik
Hello all!

I am making a Formik form and ESLint was complaining about some props missing validation. Then, I added propTypes and defaults and I got the error "PropType is defined but prop is never used".

I believe this error is due to the fact that when I place MyComponent.propTypes blabla I was assigning props to my functional components. However, ESLint complains about the props passed to the render prop function. An example of such props are _errors_ and _touched_ from the function ValidationSchemaExample on https://jaredpalmer.com/formik/docs/guides/validation

In my personal code I had something like this:
```
  const renderResetPasswordRequestForm = ({ errors, touched, handleSubmit } ) => {
    return (
      <form onSubmit={handleSubmit}>
      </form>
    );
  };
```

Which was triggering the error, however, with the below code it does not happen:

```
const renderResetPasswordRequestForm = formikProps => {
    const { errors, touched, handleSubmit } = formikProps;
    return (
      <form onSubmit={handleSubmit}>
      </form>
    );
  };
```
So the key question:
**Anyone understand why if I destructure the parameters it complains and if I do so in the body of the function ESLint does not complain?**

Thank you in advance and regards.

### LINKED ISSUE #2554: False positive for function-component-definition when function returns null
I hava a function which `react/function-component-definition` is falsely reporting as a function component. This only seems to occur when null is returned at the end of a function.

My rule config is:
```javascript
'react/function-component-definition': ['error', {
  namedComponents: 'function-declaration',
  unnamedComponents: 'arrow-function',
}],
```

function that triggers a false positive: 
```javascript
const selectAvatarByUserId = (state, id) => {
  const user = selectUserById(state, id)

  if (user) {
    return user.attributes.image || `//api.adorable.io/avatars/${user.id}`
  }

  return null
}
```

Removing `return null` or returning any other value fixes the error. I was able to get around this by reversing my if statement, but I'm still reporting this as a potential bug.

### LINKED ISSUE #2607: [BUG] react/prop-types - false positive in functional components
When I use props inside a function of a functional component. The eslint PropTypes validation fails. Which is wrong.

My eslintrc.json:

```
{
  "settings": {
    "react": {
      "createClass": "createReactClass",
      "pragma": "React",
      "version": "16"
    }
  },
  "parser": "babel-eslint",
  "parserOptions": {
    "ecmaVersion": 6,
    "sourceType": "module",
    "ecmaFeatures": {
      "jsx": true
    }
  },
  "extends": [
    "plugin:react/recommended"
  ],
  "plugins": [
    "babel"
  ],
  "rules": {
    "react/prop-types": ["error"]
  }
}
```

**Example with bug**

```
import React from 'react';

const MyComponent = (props) => {
    function render() {
        return <test>{props.text}</test>;
    }

    return render();
};

MyComponent.propTypes = {
    text: PropTypes.string.isRequired,
};

export default MyComponent;

```

Error messge: ` 6:29  error  'text' is missing in props validation  react/prop-types`

---

If I don't use a function inside my functional component. The eslint rule validates my code correctly.

**Working example:**

```
import React from 'react';

const MyComponent = (props) => {
    return <test>{props.text}</test>;
};

MyComponent.propTypes = {
    text: PropTypes.string.isRequired,
};

export default MyComponent;
```

jzabala:
As stated by @ljharb in https://github.com/yannickcr/eslint-plugin-react/issues/2554#issuecomment-596631942, functions starting with a lowercase letter are not components.

This PR adds a check for the capitalization of the first letter in functional components detection.

With this change all the reported cases in the issue work properly.

Fixes #2554. Fixes #2495. Fixes #2607. Fixes #2352

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -70,6 +70,14 @@ function isReturnsLogicalJSX(node, property, strict) {
     : (returnsLogicalJSXLeft || returnsLogicalJSXRight);
 }
 
+function isFirstLetterCapitalized(word) {
+  if (!word) {
+    return false;
+  }
+  const firstLetter = word.charAt(0);
+  return firstLetter.toUpperCase() === firstLetter;
+}
+
 const Lists = new WeakMap();
 
 /**
@@ -624,13 +632,21 @@ function componentRule(rule, context) {
      * @returns {ASTNode | undefined}
      */
     getStatelessComponent(node) {
-      if (node.type === 'FunctionDeclaration') {
-        if (utils.isReturningJSXOrNull(node)) {
-          return node;
-        }
+      if (
+        node.type === 'FunctionDeclaration'
+        && isFirstLetterCapitalized(node.id.name)
+        && utils.isReturningJSXOrNull(node)
+      ) {
+        return node;
       }
 
       if (node.type === 'FunctionExpression' || node.type === 'ArrowFunctionExpression') {
+        if (node.parent.type === 'VariableDeclarator' && utils.isReturningJSXOrNull(node)) {
+          if (isFirstLetterCapitalized(node.parent.id.name)) {
+            return node;
+          }
+          return undefined;
+        }
         if (utils.isInAllowedPositionForComponent(node) && utils.isReturningJSXOrNull(node)) {
           return node;
         }
--- a/tests/lib/rules/function-component-definition.js
+++ b/tests/lib/rules/function-component-definition.js
@@ -69,6 +69,32 @@ ruleTester.run('function-component-definition', rule, {
     code: 'var Foo = React.memo(function Foo() { return <p/> })',
     options: [{namedComponents: 'function-declaration'}]
   }, {
+    // shouldn't trigger this rule since functions stating with a lowercase
+    // letter are not considered components
+    code: `
+    const selectAvatarByUserId = (state, id) => {
+      const user = selectUserById(state, id)
+      return null
+    }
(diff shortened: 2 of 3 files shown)


jzabala:
> Thanks! Would you mind adding some test cases to various rules?

Thanks for you review @ljharb. I added some tests based on the issues this PR solves.

ljharb:
This may have caused #2728.

PR #2701, opened 2020-07-07
[New]`jsx-key`: add check for unique key

### LINKED ISSUE #2614: Rule idea: `jsx-unique-key`
I'd like to propose a rule that triggers when duplicate `key` properties are defined in on the same JSX nesting level or in arrays containing JSX at the top-level element.

React depends on unique keys and will render very unexpected results when duplicate keys are present which shoult to my knowledge almost never be intentional.

#### Examples where the rule would trigger an error:

```jsx
const div = (
  <div>
    <span key="notunique"/>
    <span key="notunique"/>
  </div>
);
```
``` jsx
const spans = [
  <span key="notunique"/>,
  <span key="notunique"/>,
];

### PR DESCRIPTION
**Pull request by jzabala**
Add `key` uniqueness check in array literal and sibling JSX elements.

Closes #2614

### DIFF
--- a/docs/rules/jsx-key.md
+++ b/docs/rules/jsx-key.md
@@ -3,6 +3,8 @@
 Warn if an element that likely requires a `key` prop--namely, one present in an
 array literal or an arrow function expression.
 
+Also validates `key` uniqueness in array literals and JSX sibling elements.
+
 ## Rule Details
 
 The following patterns are considered warnings:
@@ -11,6 +13,13 @@ The following patterns are considered warnings:
 [<Hello />, <Hello />, <Hello />];
 
 data.map(x => <Hello>{x}</Hello>);
+
+[<Hello key="one" />, <Hello key="one" />];
+
+<div>
+  <Hello key="hello" />
+  <Hello key="hello" />
+</div>
 
 <Hello {...{ key: id, id, caption }} />
 ```
@@ -25,19 +34,24 @@ The following patterns are **not** considered warnings:
 data.map((x, i) => <Hello key={i}>{x}</Hello>);
 
 <Hello key={id} {...{ id, caption }} />
+
+<div>
+  <Hello key="hi" />
+  <Hello key="hello" />
+</div>
 ```
 
 ## Rule Options
 
 ```js
 ...
-"react/jsx-key": [<enabled>, { "checkFragmentShorthand": <boolean> }]
+"react/jsx-key": [<enabled>, { "checkFragmentShorthand": <boolean>, "checkUniquePropKey": <boolean> }]
 ...
 ```
 
 ### `checkFragmentShorthand` (default: `false`)
 
-When `true` the rule will check if usage of the [shorthand fragment syntax][short_syntax] requires a key. This option was added to avoid a breaking change and will be the default in the next major version. 
+When `true` the rule will check if usage of the [shorthand fragment syntax][short_syntax] requires a key. This option was added to avoid a breaking change and will be the default in the next major version.
 
 The following patterns are considered warnings:
 
@@ -46,6 +60,10 @@ The following patterns are considered warnings:
 
 data.map(x => <>{x}</>);
 ```
+
+### `checkUniquePropKey` (default: `true`)
+
+When `false` the validation for unique `key` in array literal and sibling JSX elements is disabled.
 
 ## When not to use
 
--- a/lib/rules/jsx-key.js
+++ b/lib/rules/jsx-key.js
@@ -6,6 +6,8 @@
 'use strict';
 
 const hasProp = require('jsx-ast-utils/hasProp');
+const getPropValue = require('jsx-ast-utils/getPropValue');
+const getProp = require('jsx-ast-utils/getProp');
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: docs/rules/jsx-key.md [hunk 1]
```diff
@@ -47,6 +61,10 @@ The following patterns are considered warnings:
 data.map(x => <>{x}</>);
 ```
 
+### `checkUniquePropKey` (default: `true`)
+
+When `false` the validation for unique `key` in array literal and sibling JSX elements is disabled.
```
ljharb, line 66: to avoid a breaking change, this also needs to default to false, just like checkFragmentShorthand.

lib/rules/jsx-key.js
```diff
@@ -40,8 +50,10 @@ module.exports = {
   create(context) {
     const options = Object.assign({}, defaultOptions, context.options[0]);
     const checkFragmentShorthand = options.checkFragmentShorthand;
+    const checkUniquePropKey = options.checkUniquePropKey;
     const reactPragma = pragmaUtil.getFromContext(context);
     const fragmentPragma = pragmaUtil.getFragmentFromContext(context);
+    const processedNodes = new Set();
```
ljharb, line 56:
```suggestion
    const processedNodes = new WeakSet();
```

lib/rules/jsx-key.js
```diff
@@ -61,18 +73,61 @@ module.exports = {
       return body.filter((item) => item.type === 'ReturnStatement')[0];
     }
 
-    return {
-      JSXElement(node) {
-        if (hasProp(node.openingElement.attributes, 'key')) {
-          return;
-        }
+    function nodeProcessed(node) {
```
ljharb, line 76: this reads like it's a predicate, but it also has the side effect of marking the node as processed. maybe it needs a better name?

tests/lib/rules/jsx-key.js
```diff
@@ -47,46 +46,117 @@ ruleTester.run('jsx-key', rule, {
     {code: 'var App = () => <div />;'},
     {code: '[1, 2, 3].map(function(x) { return; });'},
     {code: 'foo(() => <div />);'},
-    {code: 'foo(() => <></>);', parser: parsers.BABEL_ESLINT},
-    {code: '<></>;', parser: parsers.BABEL_ESLINT}
```
ljharb, line 51: to ensure that this isn't a breaking change, please don't alter any existing test cases.

tests/lib/rules/jsx-key.js
```diff
@@ -47,46 +46,117 @@ ruleTester.run('jsx-key', rule, {
...
-  }, {
-    code: '[1, 2 ,3].map(function(x) { return <App /> });',
-    errors: [{message: 'Missing "key" prop for element in iterator'}]
-  }, {
-    code: '[1, 2 ,3].map(x => <App />);',
-    errors: [{message: 'Missing "key" prop for element in iterator'}]
-  }, {
-    code: '[1, 2 ,3].map(x => { return <App /> });',
-    errors: [{message: 'Missing "key" prop for element in iterator'}]
-  }, {
```
ljharb, line 71: also no need to reformat lines like these; that just clutters the diff.

docs/rules/jsx-key.md
ljharb: (i can see an argument for it not being enabled by default, but i don't think it's worth the risk)

tests/lib/rules/jsx-key.js
jzabala: I understand your argument but I would like to make a case for the change. 

I think moving tests like this is low risk, as you can see they are still listed in a `tests` object and they are being passed to the test runners. Also, the majority of tests are now being run with the 4 test runners improving the test suite.

If you prefer I can revert the change but I think the benefits outweigh the risk. What do you think?

tests/lib/rules/jsx-key.js
jzabala: Yeah, sorry about that. The current indentation make it easier for me to comment and uncomment then while testing and I forgot the change them back.

lib/rules/jsx-key.js
jzabala: Yeah, you are right. 🤔 What do you think about `checkAndMarkNodeProcessed`? I am not crazy about this one but at least express better the behavior.

tests/lib/rules/jsx-key.js
ljharb: Fair enough.

ljharb:
Please mark this PR as ready for review once it is ready.

tido64:
This rule is now being triggered for all elements inside a `React.Fragment`. Is that intentional? Our CI failed with 7.29: https://github.com/microsoft/rnx-kit/runs/5333921004?check_suite_focus=true

The code that it trips on: https://github.com/microsoft/rnx-kit/blob/main/packages/test-app/src/App.native.tsx#L39

ljharb:
This was done in dbc8e8c35a39e9351ce73d3132e00af2d0113b80 / f51ec4551b3fc85d971fc4d65414684bfabe3974.

PR #2707, opened 2020-07-10
[Fix]: only detect functions that return JSX somewhere as components

### LINKED ISSUE #2010: Non sfc function shows "react/no-this-in-sfc"
Hi, i have the following file which does not have a stateless functional component, it does not involve react at all but is showing me the error `react/no-this-in-sfc` when i call `this.connection`.

```javascript
import { ValidatedMethod } from "meteor/mdg:validated-method";
import { SimpleSchema } from "meteor/aldeed:simple-schema";

export const prepareLogin = new ValidatedMethod({
    name: "user.prepare",
    validate: new SimpleSchema({
        // ...
    }).validator(),
    run({ remember }) {
        if (Meteor.isServer) {
            const connectionId = this.connection.id; // react/no-this-in-sfc
            return Methods.prepareLogin(connectionId, remember);
        }
        return null;
    },
});
```

jzabala:
Registering all functions that return JSX + null as components has bring a lot of false positives for normal functions that just happen to return `null` somewhere.

In this PR I propose a new approach to detecting components by finding all return statements in a function and only marking the function as a component if one of them returns JSX.

Fixes #2010

### DIFF
--- a/tests/lib/rules/no-this-in-sfc.js
+++ b/tests/lib/rules/no-this-in-sfc.js
@@ -151,6 +151,42 @@ ruleTester.run('no-this-in-sfc', rule, {
       };
     };`,
     parser: parsers.BABEL_ESLINT
+  }, {
+    code: `
+      export const prepareLogin = new ValidatedMethod({
+        name: "user.prepare",
+        validate: new SimpleSchema({
+        }).validator(),
+        run({ remember }) {
+            if (Meteor.isServer) {
+                const connectionId = this.connection.id;
+                return Methods.prepareLogin(connectionId, remember);
+            }
+            return null;
+        },
+      });
+    `
+  }, {
+    // functions returning null are not necessarily components.
+    code: `
+    class Foo {
+      bar() {
+        function Bar(){
+          return () => {
+            this.something();
+            return null;
+          }
+        }
+      }
+    }`
+  }, {
+    code: `
+    function Foo(props) {
+      if (this.props.foo) {
+        something();
+      }
+      return null;
+    }`
   }],
   invalid: [{
     code: `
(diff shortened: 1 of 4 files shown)


ljharb:
In React 16, returning an array, a string, or a number (and I think a boolean) are all valid components.

tests/lib/rules/no-this-in-sfc.js
```diff
@@ -193,15 +229,6 @@ ruleTester.run('no-this-in-sfc', rule, {
...
     }`,
     errors: [{message: ERROR_MESSAGE}]
-  }, {
-    code: `
-    function Foo(props) {
-      if (this.props.foo) {
-        something();
-      }
-      return null;
-    }`,
```
ljharb, line 203: this is a valid component, and must be treated as one.

jzabala: But what exactly makes this function a component? this could perfectly pass as a function [constructor](https://css-tricks.com/understanding-javascript-constructors/).

jzabala:
> In React 16, returning an array, a string, or a number (and I think a boolean) are all valid components.

Yeah, but adding those types in my opinion will bring even more false positives that the `null` case. 

By the way the only meaningful thing that this PR those is to remove the null check and collect all return statements from functions. The `isReturningJSX` algorithm is the same. That is why almost no test was broken.

tests/lib/rules/no-this-in-sfc.js
ljharb: That's the challenge with component detection.

Legacy constructors tho are fine to be false positives - nobody should be writing those anymore, especially not in parts an app using React (and thus linted by this plugin). Constructors should be written with `class`.

The false positives that we need to focus on avoiding are render props, or normal non-component functions to return jsx for refactoring purposes inside a component.

In other words, "returns jsx in any branch" means it's definitely a component; and "returns something a component *can't* in any branch" means it's definitely not; there's ambiguity in between.

jzabala: > The false positives that we need to focus on avoiding are render props, or normal non-component functions to return jsx for refactoring purposes inside a component.

In that case I think I'll close this PR since other than the uppercase `F` there is not really a way of knowing if that function is really a component or other thing.

jzabala:
Closed based in https://github.com/yannickcr/eslint-plugin-react/pull/2707#discussion_r453122044

tests/lib/rules/no-this-in-sfc.js
ljharb: If it returns a class or a function inline, it's not a component, but sure.

PR #2711, opened 2020-07-12
[Fix] `no-access-state-in-setstate`: add check for class component

### LINKED ISSUE #2464: [no-access-state-in-setstate] inspects non-react classes
The `react/no-access-state-in-setstate` rule can find errors in non-react classes, which doesn't make a lot of sense.

In my specific use case I'm using [overstated](https://github.com/fabiospampinato/overstated) for state management, which provides a similar API to React's own API for updating the state, which trips this rule.

It's impossible in general to statically detect if a class is a react component or not, e.g. `class Foo extends Abstract {}`, who knows if that's a react component or not.

Therefor I propose adding an option for blacklisting specific classes, so that I can tell the rule that "Abstract" in my example above is not a react class.

jzabala:
Fixes #2464

### DIFF
--- a/tests/lib/rules/no-access-state-in-setstate.js
+++ b/tests/lib/rules/no-access-state-in-setstate.js
@@ -10,8 +10,8 @@
 // ------------------------------------------------------------------------------
 
 const RuleTester = require('eslint').RuleTester;
-const rule = require('../../../lib/rules/no-access-state-in-setstate');
 const parsers = require('../../helpers/parsers');
+const rule = require('../../../lib/rules/no-access-state-in-setstate');
 
 const parserOptions = {
   ecmaVersion: 2018,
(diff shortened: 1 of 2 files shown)


tests/lib/rules/no-access-state-in-setstate.js
```diff
@@ -19,11 +20,17 @@ const parserOptions = {
   }
 };
 
+const settings = {
+  react: {
+    createClass: 'createClass'
+  }
+};
```
ljharb, line 27: what does adding this here help? none of your new test cases use createClass.

jzabala: @ljharb 

None of the ones I added use it but a lot of tests in this rule use `createClass`. 

The default pragma is `createReactClass`, and since I added component detection to this rule I need to do one of two things: update all tests using `createClass`  to the default pragma `createReactClass` or what I did here: to set the default pragma to `createClass`.

ljharb: Thanks, makes perfect sense.

PR #2718, opened 2020-07-19
[New] version: detect-no-warn

### LINKED ISSUE #2717: react version "detect-no-warn"
I want to use a single Eslint configuration for multiple projects. However, some of the projects do not use react. When I set the react version settings to detect, I get a warning that react is not installed in those repositories.
```json
  "settings": {
    "react": {
      "version": "detect"
    }
  }
```

I wish there was an option to set the version to `"version": "detect-no-warn"`, so I can use a single Eslint preset for all of the projects no matter if they are using react or not.

aminya:
Fixes #2717 

This adds an option to set the version to `"version": "detect-no-warn"`, which is the same as "detect", but does not give a warning. 

This is useful if someone wants to use a single Eslint preset for multiple projects no matter if they are using react or not (common in organizations).

```json
  "settings": {
    "react": {
      "version": "detect-no-warn"
    }
  }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ You should also specify settings that will be shared across all the plugin rules
                            // You can also use `16.0`, `16.3`, etc, if you want to override the detected value.
                            // default to latest and warns if missing
                            // It will default to "detect" in the future
+                           // You can also set it to `"detect-no-warn"`, which is the same as "detect", but does not give a warning
       "flowVersion": "0.53" // Flow version
     },
     "propWrapperFunctions": [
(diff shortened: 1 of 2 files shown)


aminya:
> It's not possible (or imo sensible) to have a single eslint config that works for both react and non-react projects.

Why is it not possible to do so?

ljharb:
For one, because if it's a non-react project, it's likely to be using a different component framework, many of which use jsx - so this plugin's linting rules will give you lots of false positives.

See https://npmjs.com/eslint-config-airbnb and https://npmjs.com/eslint-config-airbnb-base; that's the approach I'd suggest.

PR #2730, opened 2020-07-27
[Fix] `prop-types`: handle anonymous functions

### LINKED ISSUE #2728: TypeError: Cannot read property 'name' of null
Version 7.20.4 seems to crash with the following error in a non-tsx file.

```
yarn run v1.22.4
$ prettier packages/core/src/**/*.less --check && eslint packages --ext .js,.ts,.tsx --cache
Checking formatting...
All matched files use Prettier code style!

ESLint: 7.5.0

TypeError: Cannot read property 'name' of null
Occurred while linting /home/runner/work/exam-engine/exam-engine/packages/cli/src/commands/create-mex.ts:23
    at Object.getStatelessComponent (/home/runner/work/exam-engine/exam-engine/node_modules/eslint-plugin-react/lib/util/Components.js:692:45)
    at Object.getParentStatelessComponent (/home/runner/work/exam-engine/exam-engine/node_modules/eslint-plugin-react/lib/util/Components.js:728:42)
    at Object.getParentComponent (/home/runner/work/exam-engine/exam-engine/node_modules/eslint-plugin-react/lib/util/Components.js:623:18)
    at Object.ArrowFunctionExpression (/home/runner/work/exam-engine/exam-engine/node_modules/eslint-plugin-react/lib/util/Components.js:916:31)
    at updatedRuleInstructions.<computed> (/home/runner/work/exam-engine/exam-engine/node_modules/eslint-plugin-react/lib/util/Components.js:970:43)
    at /home/runner/work/exam-engine/exam-engine/node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Object.emit (/home/runner/work/exam-engine/exam-engine/node_modules/eslint/lib/linter/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (/home/runner/work/exam-engine/exam-engine/node_modules/eslint/lib/linter/node-event-generator.js:254:26)
    at NodeEventGenerator.applySelectors (/home/runner/work/exam-engine/exam-engine/node_modules/eslint/lib/linter/node-event-generator.js:283:22)
```

I haven't been able to produce a reduced test case yet.

odinho:
Fixes #2728.

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -689,7 +689,7 @@ function componentRule(rule, context) {
     getStatelessComponent(node) {
       if (
         node.type === 'FunctionDeclaration'
-        && isFirstLetterCapitalized(node.id.name)
+        && (!node.id || isFirstLetterCapitalized(node.id.name))
         && utils.isReturningJSXOrNull(node)
       ) {
         return node;
--- a/tests/lib/rules/prop-types.js
+++ b/tests/lib/rules/prop-types.js
@@ -2569,6 +2569,11 @@ ruleTester.run('prop-types', rule, {
         return null;
       }`,
       parser: parsers.TYPESCRIPT_ESLINT
+    },
+    {
+      code: `
+      export default function() {}
+      `
     }
   ],
 
@@ -5158,6 +5163,16 @@ ruleTester.run('prop-types', rule, {
         return null;
       }`,
       parser: parsers.TYPESCRIPT_ESLINT,
+      errors: [{
+        message: '\'value\' is missing in props validation'
+      }]
+    },
+    {
+      code: `
+        export default function ({ value = 'World' }) {
+          return <h1>Hello {value}</h1>
+        }
+      `,
       errors: [{
         message: '\'value\' is missing in props validation'
       }]


sluukkonen:
I'm no expert in eslint, but is the check actually correct? I think an anonymous function could also be considered to be a stateless component.

I was thinking of something like 

```javascript
      if (
        node.type === 'FunctionDeclaration'
        && (node.id == null || isFirstLetterCapitalized(node.id.name))
        && utils.isReturningJSXOrNull(node)
      ) {
```

odinho:
> I'm no expert in eslint, but is the check actually correct? I think an anonymous function could also be considered to be a stateless component.
> 
> I was thinking of something like
> 
> ```js
>       if (
>         node.type === 'FunctionDeclaration'
>         && (node.id == null || isFirstLetterCapitalized(node.id.name))
>         && utils.isReturningJSXOrNull(node)
>       ) {
> ```

As I read it, the code underneath will handle that case. It checks the variable that the anonymous function is being stored into.

ljharb:
Thanks! This does need a regression test first.

ljharb:
@odinho Since you made this PR from a fork that you don't own (whereby instead of odinho) I can't merge this. Can you please add me to https://github.com/whereby/eslint-plugin-react so I can force push to the PR branch?

odinho:
> @odinho Since you made this PR from a fork that you don't own (whereby instead of odinho) I can't merge this yet. Can you please add me to https://github.com/whereby/eslint-plugin-react so I can force push to the PR branch?

Oup. Why do I keep forgetting that this restriction exists. I've added you now :facepalm:

jzabala:
This indeed will solve #2728 but then components defined this way won't be recognized:

```js
export default function () {
  return <h1>Hello World</div>
}
```

I think a better check would be:

```js
if (
        node.type === 'FunctionDeclaration'
        && (!node.id || isFirstLetterCapitalized(node.id.name)) // skip first letter check if the node.id doesn't exists
        && utils.isReturningJSXOrNull(node)
) {
```

ljharb:
@jzabala thanks, i've added a test case and your suggested fix.

PR #2736, opened 2020-07-31
[Fix] jsx-no-literals with allowStrings doesn't work in props

### LINKED ISSUE #2720: jsx-no-literals with allowStrings doesn't work in props
the following setup seems not to work:

```js
<button type="button" ></button>
```

with the rule 
```js
    'react/jsx-no-literals': [
      'warn',
      {
        noStrings: true,
        allowedStrings: ['button', 'submit'],
      },
    ],
```

produces the following error: `Invalid prop value: “type="button"”  react/jsx-no-literals`

Context: I'm trying to find a balance between enforcing no literals but also `react/button-has-type`

karolina-benitez:
Checks if props string is included in `allowedStrings` option

Fixes #2720.

### DIFF
--- a/lib/rules/jsx-no-literals.js
+++ b/lib/rules/jsx-no-literals.js
@@ -120,7 +120,7 @@ module.exports = {
       },
 
       JSXAttribute(node) {
-        const isNodeValueString = node.value && node.value && node.value.type === 'Literal' && typeof node.value.value === 'string';
+        const isNodeValueString = node && node.value && node.value.type === 'Literal' && typeof node.value.value === 'string' && !config.allowedStrings.has(node.value.value);
 
         if (config.noStrings && !config.ignoreProps && isNodeValueString) {
           const customMessage = 'Invalid prop value';
--- a/tests/lib/rules/jsx-no-literals.js
+++ b/tests/lib/rules/jsx-no-literals.js
@@ -42,8 +42,63 @@ function invalidProp(str) {
 const ruleTester = new RuleTester({parserOptions});
 ruleTester.run('jsx-no-literals', rule, {
 
-  valid: [
+  valid: [].concat(
     {
+      code: `
+        class Comp1 extends Component {
+          render() {
+            return (
+              <div>
+                <button type="button"></button>
+              </div>
+            );
+          }
+        }
+      `,
+      options: [{noStrings: true, allowedStrings: ['button', 'submit']}]
+    }, {
+      code: `
+        class Comp1 extends Component {
+          render() {
+            return (
+              <div>
+                <button type="button"></button>
+              </div>
+            );
+          }
+        }
+      `,
+      options: [{noStrings: true, allowedStrings: ['button', 'submit']}],
+      parser: parsers.BABEL_ESLINT
+    }, parsers.TS([{
+      code: `
+        class Comp1 extends Component {
+          render() {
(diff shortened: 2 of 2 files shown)


tests/lib/rules/jsx-no-literals.js
```diff
@@ -44,6 +44,20 @@ ruleTester.run('jsx-no-literals', rule, {
 
   valid: [
     {
+      code: `
+        class Comp1 extends Component {
+          render() {
+            return (
+              <div>
+                <button type="button" ></button>
```
ljharb, line 52:
```suggestion
                <button type="button"></button>
```

lib/rules/jsx-no-literals.js
```diff
@@ -120,7 +120,12 @@ module.exports = {
       },
 
       JSXAttribute(node) {
-        const isNodeValueString = node.value && node.value && node.value.type === 'Literal' && typeof node.value.value === 'string';
+        let isNodeValueString;
+        if (context && context.options[0] && context.options[0].allowedStrings !== undefined) {
+          isNodeValueString = node.value && node.value && node.value.type === 'Literal' && typeof node.value.value === 'string' && context.options[0].allowedStrings.indexOf(node.value) !== -1;
```
ljharb, line 125:
```suggestion
          isNodeValueString = node && node.value && node.value.type === 'Literal' && typeof node.value.value === 'string' && context.options[0].allowedStrings.indexOf(node.value) !== -1;
```

lib/rules/jsx-no-literals.js
```diff
@@ -120,7 +120,12 @@ module.exports = {
       },
 
       JSXAttribute(node) {
-        const isNodeValueString = node.value && node.value && node.value.type === 'Literal' && typeof node.value.value === 'string';
+        let isNodeValueString;
+        if (context && context.options[0] && context.options[0].allowedStrings !== undefined) {
```
ljharb, line 124: why is this check needed? `config.allowedStrings` should already be an array containing either the default, or the provided value - what's the need to determine what the user actually passed?

tests/lib/rules/jsx-no-literals.js
```diff
@@ -44,6 +44,20 @@ ruleTester.run('jsx-no-literals', rule, {
...
+              </div>
+            );
+          }
+        }
+      `,
+      options: [{noStrings: true, allowedStrings: ['button', 'submit']}],
+      parser: parsers.BABEL_ESLINT
```
ljharb, line 59: let's also add the same test case for each of the other parsers (default, and both typescript-eslints)

lib/rules/jsx-no-literals.js
karolina-benitez: Thanks for catching that. Fixed & pushed changes

tests/lib/rules/jsx-no-literals.js
karolina-benitez: Can you please clarify?

ljharb: This test case works for babel-eslint; but the same test case should pass on the default parser, and on the old typescript-eslint, and on the new typescript eslint parser.

karolina-benitez: Thanks, added!

PR #2739, opened 2020-08-03
fix(callee): check to ensure callee object name is `PropTypes`

codecog:
An issue was raised where using an alternative library with a similar syntax (in this case Yup) resulted in the rule `forbid-prop-types` stating that `Yup.array()` was an invalid proptype. Adding another check to ensure the object name of callee begins with `PropTypes` ties the rule to check only properties off of that name.


lib/rules/forbid-prop-types.js
```diff
@@ -170,6 +170,7 @@ module.exports = {
       CallExpression(node) {
         if (
           node.arguments.length > 0
+          && node.callee.object.name === 'PropTypes'
```
ljharb, line 173: this shouldn't check for `PropTypes` specifically - it should ensure that it was either imported/required from `prop-types`, or, that it came from `React.PropTypes`.

codecog: Understood, will update.

jzabala: I think it should be `[pragma].PropTypes`.

ljharb: yes, thanks, good call.

codecog: Thanks for some direction. The check now uses the `pragmaUtil` instead of checking against the name inside callee

lib/rules/forbid-prop-types.js
```diff
@@ -170,6 +172,7 @@ module.exports = {
       CallExpression(node) {
         if (
           node.arguments.length > 0
+          && `${pragma}.PropTypes`
```
ljharb, line 175: should this be comparing this string to anything? otherwise it's just a noop and could be deleted.

ljharb:
@nonewcode are you still interested in completing this PR?

PR #2740, opened 2020-08-03
[New] `prop-types`: add support for `PropTypes.exact` 

### LINKED ISSUE #2590: prop-types does not recognize PropTypes.exact
`PropTypes.exact` should work just like `PropTypes.shape`. But this plugin does not recognize `PropTypes.exact`, so no error is detected.

```js
function Zoo(props) {
  return (
    <>
      {prop.foo.c} // Expects a report here.
    </>
  );
}

Zoo.propTypes = {
  foo: PropTypes.exact({
    a: PropTypes.number,
    b: PropTypes.number,
  }),
};
```

jzabala:
Closes #2590

### DIFF
--- a/lib/util/propTypes.js
+++ b/lib/util/propTypes.js
@@ -422,13 +422,14 @@ module.exports = function propTypesInstructions(context, components, utils) {
       const callName = value.callee.property.name;
       const argument = value.arguments[0];
       switch (callName) {
-        case 'shape': {
+        case 'shape':
+        case 'exact': {
           if (argument.type !== 'ObjectExpression') {
             // Invalid proptype or cannot analyse statically
             return {};
           }
           const shapeTypeDefinition = {
-            type: 'shape',
+            type: callName,
             children: {}
           };
           iterateProperties(context, argument.properties, (childKey, childValue, propNode) => {
--- a/tests/lib/rules/prop-types.js
+++ b/tests/lib/rules/prop-types.js
@@ -5827,6 +5827,70 @@ ruleTester.run('prop-types', rule, {
         errors: [{
           message: '\'dateCreated\' is missing in props validation'
         }]
+      },
+      {
+        code: `
+          function Zoo(props) {
+            return (
+              <>
+                {props.foo.c}
+              </>
+            );
+          }
+
+          Zoo.propTypes = {
+            foo: PropTypes.exact({
+              a: PropTypes.number,
+              b: PropTypes.number,
+            }),
+          };
+        `,
+        errors: [{
+          message: "'foo.c' is missing in props validation"
+        }]
+      },
+      {
+        code: `
+          function Zoo(props) {
+            return (
+              <>
+                {props.foo.c}
+              </>
+            );
+          }
+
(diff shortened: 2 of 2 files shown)


lib/util/propTypes.js
```diff
@@ -422,13 +422,14 @@ module.exports = function propTypesInstructions(context, components, utils) {
       const callName = value.callee.property.name;
       const argument = value.arguments[0];
       switch (callName) {
-        case 'shape': {
+        case 'shape':
+        case 'exact': {
```
ljharb, line 426: i don't think this should be included here - i think it should be a default member of `propWrapperFunctions`. Otherwise, how can you turn it off?

Also, when the installed `prop-types` package *doesn't* have `exact`, it should *not* be supported.

jzabala: But why we would turn it off? `exact` has been part of the library since [2017][exact] from what I see.

About validating that `prop-types` is installed, that would make us more correct but is not like `React.PropTypes.exact` is going to work anyway. In that case the user would get a `ReferenceError` on `exact`, I think.

[exact]: https://github.com/facebook/prop-types/pull/41

ljharb: Because someone might be using an older version of the library. There's lots of precedent; this plugin dynamically looks up the keys on `require('prop-types')` in order to determine what's available.

`React.PropTypes.exact` might work if a) the pragma isn't "React", b) they'd done `React.PropTypes = require('prop-types');`.

jzabala: > Because someone might be using an older version of the library. There's lots of precedent; this plugin dynamically looks up the keys on require('prop-types') in order to determine what's available.

Yeah, but the rule `no-types` works the user having the library installed or not since this project already has it as a dependency:

https://github.com/yannickcr/eslint-plugin-react/blob/d8741de74da0fb7a0cb730f0fea14de05e4faa9b/package.json#L39

Doing as suggested in comment https://github.com/yannickcr/eslint-plugin-react/issues/2590#issuecomment-597757907 would result in `exact` always being included since the library would always resolve to the internal version of `prop-types`.

> React.PropTypes.exact might work if a) the pragma isn't "React", b) they'd done React.PropTypes = require('prop-types');

Added two more tests for this cases.

ljharb: Hmm, that is a fair point on the prop-types dep.

jzabala: But we can still do it right.

The library that depends on `prop-types` is `no-typos`.

https://github.com/yannickcr/eslint-plugin-react/blob/d8741de74da0fb7a0cb730f0fea14de05e4faa9b/lib/rules/no-typos.js#L7

So, to correctly feature detect `exact` what do you think if I change `prop-types` to a dev dependency, check for the existent of `prop-types` before requiring it and fix `no-typos` to not validate `prop-types` names if the lib is not present.

ljharb: The intention in no-typos is that when new validators are added, they're suggested by default, without needing a change in this library.

PR #2741, opened 2020-08-04
[Fix] Improve error message for no-find-dom-node

ecraig12345:
The error message for `no-find-dom-node` wasn't very informative, which makes it less likely that people will respect the rule instead of just disabling it. This PR adds a more informative message.

### DIFF
--- a/lib/rules/no-find-dom-node.js
+++ b/lib/rules/no-find-dom-node.js
@@ -40,7 +40,7 @@ module.exports = {
 
         context.report({
           node: callee,
-          message: 'Do not use findDOMNode'
+          message: 'Do not use findDOMNode. It doesn’t work with function components and is deprecated in StrictMode. See https://reactjs.org/docs/react-dom.html#finddomnode'
         });
       }
     };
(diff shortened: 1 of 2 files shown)


lib/rules/no-find-dom-node.js
```diff
@@ -40,7 +40,7 @@ module.exports = {
 
         context.report({
           node: callee,
-          message: 'Do not use findDOMNode'
+          message: 'Do not use findDOMNode. It doesn\'t work with function components and is deprecated in StrictMode. See https://reactjs.org/docs/react-dom.html#finddomnode'
```
ljharb, line 43:
```suggestion
          message: 'Do not use findDOMNode. It doesn‘t work with function components and is deprecated in StrictMode. See https://reactjs.org/docs/react-dom.html#finddomnode'
```
straight quotes are always typographically incorrect in prose :-)

ecraig12345: I believe you meant to use the right single quotation mark ’ rather than the left one.

ljharb: lol yes, thanks :-)

PR #2744, opened 2020-08-06
[Fix] `jsx-no-useless-fragment`: accept fragments with call expressions

### LINKED ISSUE #2584: react/jsx-no-useless-fragment false positive
There's a false positive on the rule with forwarded children. Use cases are pretty rare but still exist. Without a fragment around the `children`, code breaks at runtime

```
const HollowComponent = ({ children }) => {
  // some logic here
  return <>{children}</>; // doesn't pass the rule
}
```

example: https://codesandbox.io/s/icy-lake-ri3uo

Temporary fix is to disable rule for this line but it would be nicer if this rule let this kind of Fragment use possible.

hasparus:
:wave: 

Solves part of #2584.

### DIFF
--- a/tests/lib/rules/jsx-no-useless-fragment.js
+++ b/tests/lib/rules/jsx-no-useless-fragment.js
@@ -63,11 +63,21 @@ ruleTester.run('jsx-no-useless-fragment', rule, {
     {
       code: '<Fooo content={<>eeee ee eeeeeee eeeeeeee</>} />',
       parser: parsers.BABEL_ESLINT
+    },
+    {
+      code: '<>{foos.map(foo => foo)}</>',
+      parser: parsers.BABEL_ESLINT
     }
   ],
   invalid: [
     {
       code: '<></>',
+      output: null,
+      errors: [{messageId: 'NeedsMoreChidren', type: 'JSXFragment'}],
+      parser: parsers.BABEL_ESLINT
+    },
+    {
+      code: '<>{}</>',
       output: null,
       errors: [{messageId: 'NeedsMoreChidren', type: 'JSXFragment'}],
       parser: parsers.BABEL_ESLINT
--- a/lib/rules/jsx-no-useless-fragment.js
+++ b/lib/rules/jsx-no-useless-fragment.js
@@ -66,6 +66,17 @@ function isKeyedElement(node) {
     && node.openingElement.attributes.some(jsxUtil.isJSXAttributeKey);
 }
 
+/**
+ * @param {ASTNode} node
+ * @returns {boolean}
+ */
+function containsCallExpression(node) {
+  return node
+    && node.type === 'JSXExpressionContainer'
+    && node.expression
+    && node.expression.type === 'CallExpression';
+}
+
 module.exports = {
   meta: {
     type: 'suggestion',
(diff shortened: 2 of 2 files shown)


tests/lib/rules/jsx-no-useless-fragment.js
```diff
@@ -63,6 +63,10 @@ ruleTester.run('jsx-no-useless-fragment', rule, {
     {
       code: '<Fooo content={<>eeee ee eeeeeee eeeeeeee</>} />',
       parser: parsers.BABEL_ESLINT
+    },
+    {
+      code: '<>{foos.map(foo => foo)}</>',
```
ljharb, line 68: this seems like a useless fragment to me tho? since `foos.map` produces an array, and an array is a valid render value.

tests/lib/rules/jsx-no-useless-fragment.js
```diff
@@ -72,6 +76,12 @@ ruleTester.run('jsx-no-useless-fragment', rule, {
       errors: [{messageId: 'NeedsMoreChidren', type: 'JSXFragment'}],
       parser: parsers.BABEL_ESLINT
     },
+    {
+      code: '<>{}</>',
```
ljharb, line 80: this seems like it's just a syntax error, and wouldn't parse in the first place?

tests/lib/rules/jsx-no-useless-fragment.js
hasparus: **In case if `foos` is an array**, the fragment wraps an array (a React node) into a React element. It's not useless if it's intentional (e.g. a function where you pass this treats arrays different than single elements -- requires a single child or sth like that).

https://github.com/yannickcr/eslint-plugin-react/issues/2584#issuecomment-627002431


**But actually, we can't assume what `map` does**, and even if `map` suggests an array, we treat all call expressions the same.
It may return `undefined` and then we have a crash at runtime if render it without wrapping in the fragment.

```js
const foos = { map: () => undefined };
```

tests/lib/rules/jsx-no-useless-fragment.js
hasparus: It's totally legal. CodeSandbox here:

https://codesandbox.io/s/jsx-no-useless-fragment-qw6cj

In this case the `{}` is useless, so fixing it into `<></>` could be nice.

ljharb: wow, ok then :-)

tests/lib/rules/jsx-no-useless-fragment.js
ljharb: What crash at runtime? A component that returns `[undefined]` is a valid component.

You're right that it wouldn't be useless if it's being passed somewhere, but since the introduction of fragments it seems like anything that accepts a fragment should also accept an array?

hasparus: > What crash at runtime? A component that returns [undefined] is a valid component.

Yes, but it's not valid if it returns `undefined`. There is no array here.

This doesn't have to be `map`, but even when the function is called "map" we have no guarantee that this is `Array.prototype.map`. A function may return undefined.

Vishwas-75:
How to disable this jsx-no-useless-frragments

ljharb:
@Vishwas-75 same way to disable any rule; set it to "off" in your eslint config, or with an inline eslint override comment. eslint's docs are a good resource.

PR #2746, opened 2020-08-08
[New] `jsx-filename-extension` add  "allow" option with "as-needed"

### LINKED ISSUE #2742: jsx-filename-extension allow always or as needed
`react/jsx-filename-extension` reports files using JSX which aren’t using the `.jsx` extension. However, it doesn’t report files that don’t use JSX syntax, regardless of the extension.

I would like to add an option `allow`, which can be set to `always` or `as-needed`. When this is set to `always` (default), the rule will only report JSX syntax in files not using the `.jsx` extension, as it does now. When this option is set to `as-needed`, this rule will also report files that end with the `.jsx` extension, but don’t use JSX syntax.

remcohaszing:
This option allows users to disallow the filename extension in files not using JSX syntax.

Closes #2742

### DIFF
--- a/docs/rules/jsx-filename-extension.md
+++ b/docs/rules/jsx-filename-extension.md
@@ -20,7 +20,21 @@ function MyComponent() {
 }
 ```
 
+Beware this rule **only** reports JSX syntax, **not** other non-standard syntax such as experimental features or type annotations.
+
 ## Rule Options
+
+### `allow` (default: `"always"`)
+
+When to allow a JSX filename extension. By default all files may have a JSX extension. Set this to `as-needed` to only allow JSX file extensions in files that contain JSX syntax.
+
+```js
+"rules": {
+  "react/jsx-filename-extension": [1, { "allow": "as-needed" }]
+}
+```
+
+### `extensions` (default: `[".jsx"]`)
 
 The set of allowed extensions is configurable. By default '.jsx' is allowed. If you wanted to allow both '.jsx' and '.js', the configuration would be:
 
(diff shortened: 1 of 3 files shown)


tests/lib/rules/jsx-filename-extension.js
```diff
@@ -74,6 +82,11 @@ ruleTester.run('jsx-filename-extension', rule, {
       filename: 'MyComponent.js',
       code: withJSXElement,
       errors: [{message: 'JSX not allowed in files with extension \'.js\''}]
+    }, {
+      filename: 'MyComponent.jsx',
+      code: withoutJSX,
+      options: [{allow: 'as-needed'}],
+      errors: [{message: 'Only files containing JSX may use the extension \'.jsx\''}]
```
ljharb, line 89: let's also add a test case for as-needed with `withJSXElement` and a `.js` file?

remcohaszing: Done

docs/rules/jsx-filename-extension.md
```diff
@@ -22,6 +22,18 @@ function MyComponent() {
 
 ## Rule Options
 
+### `allow` (default: `"always"`)
+
+When to allow a JSX filename extension. By default all files may have a JSX extension. Set this to `as-needed` to only allow JSX file extensions in files that contain JSX syntax.
```
ljharb, line 27: As commented in the linked issue, I'm not convinced this is a safe option to add. `.js` files are only supposed to be JS - jsx syntax isn't the only thing that would make a file not be standard JS. For example, if it used any pre-stage-3 features, like decorators or something.

At the very least, the docs here should add another paragraph explaining the reasons why this might be contentious.

remcohaszing: Done.

PR #2748, opened 2020-08-08
button-has-type: support trivial ternary expressions

Hypnosphi:
This adds support for trivial ternary expressions to `button-has-type` rule:

```jsx
<button type={condition ? 'button' : 'submit'} />
```

See https://github.com/yannickcr/eslint-plugin-react/issues/1555#issuecomment-362262866

### DIFF
--- a/tests/lib/rules/button-has-type.js
+++ b/tests/lib/rules/button-has-type.js
@@ -32,6 +32,12 @@ ruleTester.run('button-has-type', rule, {
     {code: '<button type="button"/>'},
     {code: '<button type="submit"/>'},
     {code: '<button type="reset"/>'},
+    {code: '<button type={"button"}/>'},
+    {code: '<button type={\'button\'}/>'},
+    {code: '<button type={`button`}/>'},
+    {code: '<button type={condition ? "button" : "submit"}/>'},
+    {code: '<button type={condition ? \'button\' : \'submit\'}/>'},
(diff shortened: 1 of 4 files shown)


tests/lib/rules/button-has-type.js
```diff
@@ -73,13 +76,13 @@ ruleTester.run('button-has-type', rule, {
...
       errors: [{
-        message: 'The button type attribute must be specified by a static string'
+        message: 'The button type attribute must be specified by a static string or a trivial ternary expression'
       }]
     },
     {
       code: '<button type={"foo"}/>',
       errors: [{
-        message: 'The button type attribute must be specified by a static string'
+        message: '"foo" is an invalid value for button type attribute'
```
ljharb, line 85: let's please add this test case also with single quotes, and also with backticks (it should support no-substitution-template-strings as well, and i believe that test would fail atm)

Hypnosphi: I don't think it should work with backticks. `` `foo` `` is actually a template string not a literal. Given that there's a fixed number of possible values, it doesn't seem practical to evaluate templates here

ljharb: I'm not sure why? Plenty of rules work with "no substitution" template literals, and i think this one should too. That's a much more common use case than a "trivial ternary".

PR #2750, opened 2020-08-09
[New] `no-unstable-nested-components`: Prevent creating unstable components inside components

### LINKED ISSUE #2749: Rule proposal: react/no-unstable-nested-components
I'd like to introduce a new rule for detecting React components which are declared inside another component or inside class component's render block. I was surprised to see this rule doesn't exist yet.

I've named this rule `react/no-unstable-nested-components` - if anyone can come up with a better name I'm open for it.

I'll set up the PR soon. Here's the implementation [AriPerkkio/eslint-plugin-react/tree/react/no-unstable-nested-components](https://github.com/AriPerkkio/eslint-plugin-react/tree/react/no-unstable-nested-components).

## Use case
Prevent developers from creating components inside another components.

### 1. Performance
```jsx
function Component() {
  function UnstableNestedComponent() {
    return <div />;
  }

  return (
    <div>
      <UnstableNestedComponent />
    </div>
  );
}
```

When `Component` re-renders (e.g. parent state updates) the `UnstableNestedComponent` is recreated. React reconciliation will compare the previous render's `UnstableNestedComponent` and the current one using [reference equality](https://github.com/facebook/react/blob/v16.13.1/packages/react-reconciler/src/ReactChildFiber.js#L407). Since the reference does not match, the element will be completely recreated - not to be confused with re-rendering. I guess this is a small performance optimization, especially in cases where this pattern is used on application root level.

Since all the children of the nested component are also recreated on each re-render, optimization provided by `React.PureComponent` is gone. The `PureComponent` is recreated on each re-render.
```jsx
class PureComponent extends React.PureComponent {
  render() {
    return <div />;
  };
};

function Component(props) {
  function Wrapper() {
    return <div>{props.children}</div>;
  };

  return (
    <div>
      <Wrapper />
    </div>
  )
};

...

<Component>
  <PureComponent />
</Component>
```

### 2. Prevent unnoticeable bugs

This is the most important use case of this rule. In addition to performance optimization this rule will prevent [possible bugs](https://codepen.io/ariperkkio/pen/vYLodLB).

In cases where the nested component is rendering given **stateful** children, each re-render will reset the children's state back to the initial state. Without this rule these cases are easily unnoticeable and cause weird behavior in apps.

```jsx
function Accordion(props) {
  const [isOpen, toggle] = React.useReducer(s => !s, false);

  return (
    <>
      <button onClick={toggle}>Toggle</button>
      <div style={{ opacity: isOpen ? 1 : 0 }}>
        {props.children}
      </div>
    </>
  );
};

function Menu(props) {
  function MenuWrapper() {
    return <div>{props.children}</div>;
  };

  return (
    <div>
      <h2>Menu</h2>
      <MenuWrapper />
    </div>
  );
};

export default function Navigation() {
  return (
    <nav>
      <button onClick={React.useReducer(() => [], [])[1]}>
        Force render
      </button>
      <Menu>
        <Accordion>About us</Accordion>
        <Accordion>Contact</Accordion>
      </Menu>
    </nav>
  );
};
```

In example above:

1. Click About us to open the "accordion"
2. Click force render to update parent state and to trigger re-render

- Expected: Accordion stays open
- Actual: Accordion closes. Component is recreated and initial state is set.

## Rule Details

```jsx
function Component() {
  function UnstableNestedComponent() {
    return <div />;
  }

  return (
    <div>
      <UnstableNestedComponent />
    </div>
  );
}
```

```jsx
class Component extends React.Component {
  render() {
    function UnstableNestedComponent() {
      return <div />;
    }

    return (
      <div>
        <UnstableNestedComponent />
      </div>
    );
  }
}
```

The following patterns are **not** considered warnings:

```jsx
function OutsideDefinedComponent(props) {
  return <div />;
}

function Component() {
  return (
    <div>
      <OutsideDefinedComponent />
    </div>
  );
}
```

```jsx
function Component() {
  const MemoizedNestedComponent = React.useCallback(() => <div />, []);

  return (
    <div>
      <MemoizedNestedComponent />
    </div>
  );
}
```

## Rule Options

```js
...
"react/no-unstable-nested-components": "off" | "warn" | "error"
...
```

AriPerkkio:
Fixes #2749.

This is my first time playing with AST. Feel free to guide me in any way.

### DIFF
--- a/lib/rules/no-unstable-nested-components.js
+++ b/lib/rules/no-unstable-nested-components.js
@@ -1,0 +1,468 @@
+/**
+ * @fileoverview Prevent creating unstable components inside components
+ * @author Ari Perkkiö
+ */
+
+'use strict';
+
+const Components = require('../util/Components');
+const docsUrl = require('../util/docsUrl');
+
+// ------------------------------------------------------------------------------
+// Constants
+// ------------------------------------------------------------------------------
+
+const ERROR_MESSAGE_WITHOUT_NAME = 'Declare this component outside parent component or memoize it.';
+const COMPONENT_AS_PROPS_INFO = ' If you want to allow component creation in props, set allowAsProps option to true.';
+const HOOK_REGEXP = /^use[A-Z0-9].*$/;
+
+// ------------------------------------------------------------------------------
+// Helpers
+// ------------------------------------------------------------------------------
+
+/**
+ * Generate error message with given parent component name
+ * @param {String} parentName Name of the parent component
+ * @returns {String} Error message with parent component name
+ */
+function generateErrorMessageWithParentName(parentName) {
+  return `Declare this component outside parent component "${parentName}" or memoize it.`;
+}
+
+/**
+ * Check whether given text starts with `render`. Comparison is case-sensitive.
+ * @param {String} text Text to validate
+ * @returns {Boolean}
+ */
+function startsWithRender(text) {
+  return (text || '').startsWith('render');
+}
+
+/**
+ * Get closest parent matching given matcher
+ * @param {ASTNode} node The AST node
+ * @param {Function} matcher Method used to match the parent
+ * @returns {ASTNode} The matching parent node, if any
+ */
+function getClosestMatchingParent(node, matcher) {
+  if (!node || !node.parent || node.parent.type === 'Program') {
+    return;
+  }
+
+  if (matcher(node.parent)) {
+    return node.parent;
+  }
+
+  return getClosestMatchingParent(node.parent, matcher);
+}
+
+/**
+ * Matcher used to check whether given node is a `createElement` call
+ * @param {ASTNode} node The AST node
+ * @returns {Boolean} True if node is a `createElement` call, false if not
+ */
+function isCreateElementMatcher(node) {
+  return (
+    node
+    && node.type === 'CallExpression'
+    && node.callee
+    && node.callee.property
+    && node.callee.property.name === 'createElement'
+  );
+}
+
+/**
+ * Matcher used to check whether given node is a `ObjectExpression`
+ * @param {ASTNode} node The AST node
+ * @returns {Boolean} True if node is a `ObjectExpression`, false if not
+ */
+function isObjectExpressionMatcher(node) {
+  return node && node.type === 'ObjectExpression';
+}
+
+/**
+ * Matcher used to check whether given node is a `JSXExpressionContainer`
+ * @param {ASTNode} node The AST node
+ * @returns {Boolean} True if node is a `JSXExpressionContainer`, false if not
+ */
+function isJSXExpressionContainerMatcher(node) {
+  return node && node.type === 'JSXExpressionContainer';
+}
+
+/**
+ * Matcher used to check whether given node is a `JSXAttribute` of `JSXExpressionContainer`
+ * @param {ASTNode} node The AST node
+ * @returns {Boolean} True if node is a `JSXAttribute` of `JSXExpressionContainer`, false if not
+ */
+function isJSXAttributeOfExpressionContainerMatcher(node) {
+  return (
+    node
+    && node.type === 'JSXAttribute'
+    && node.value
+    && node.value.type === 'JSXExpressionContainer'
+  );
+}
+
+/**
+ * Matcher used to check whether given node is a `CallExpression`
+ * @param {ASTNode} node The AST node
+ * @returns {Boolean} True if node is a `CallExpression`, false if not
+ */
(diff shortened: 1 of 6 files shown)


AriPerkkio:
Ready for review. @ljharb could you take a look at this some point?

lib/rules/no-unstable-nested-components.js
```diff
@@ -0,0 +1,434 @@
...
+const Components = require('../util/Components');
+const docsUrl = require('../util/docsUrl');
+
+// ------------------------------------------------------------------------------
+// Constants
+// ------------------------------------------------------------------------------
+
+const ERROR_MESSAGE_WITHOUT_NAME = 'Declare this component outside parent component or memoize it.';
+const COMPONENT_AS_PROPS_INFO = ' If you want to allow component creation in props, set allowAsProps option to true.';
+const RENDER_REGEXP = /^render/;
```
ljharb, line 17: instead of a regex, can we use `.startsWith` here? https://npmjs.com/string.prototype.startswith if it's not in supported node versions.

lib/rules/no-unstable-nested-components.js
```diff
@@ -0,0 +1,434 @@
...
+const HOOK_REGEXP = /^use[A-Z0-9].*$/;
```
ljharb, line 18: it'd be ideal to avoid hardcoding this, to avoid deviating from eslint-plugin-react-with-hooks. is there any way we could share hook detection code with that plugin?

ljharb (commented):
Sorry for the long delay. Overall, this looks great!

lib/rules/no-unstable-nested-components.js
AriPerkkio: `startsWith` can be used without polyfill. [According to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) it has been introduced in v4 which is the lowest version supported by `eslint-plugin-react`. 

I also verified this in container just to be sure:
```bash
> process.version
'v4.0.0'
> 'temp'.startsWith('te')
true
```

lib/rules/no-unstable-nested-components.js
AriPerkkio: At the moment, no. Currently they are not exporting [the hook detection methods](https://github.com/facebook/react/blob/6b28eb6175e5251b31507749b79c55c0e389b6ef/packages/eslint-plugin-react-hooks/src/RulesOfHooks.js#L12-L19) - I just decided to copy-paste the implementation. If they did export this, we could rely on that via dependencies/peerDependencies.

It's a shame hook related rules live in a separate package.

AriPerkkio:
As the component detection in various patterns is tricky I think this rule should be ran with `eslint-remote-tester` against 100-500 repositories before merging, once the implementation is reviewed and accepted. I could upload the results in another branch and check whether any false positives have been found.

ljharb:
Please mark this PR as ready for review when you're confident in the results, and we can land it.

AriPerkkio:
Final testing is now complete. I tested this against ~550 repositories. Fixed false positives of three rare minor cases. I'm confident to ship this now @ljharb. 

There are two false positive cases remaining which I'm unable to fix for now. These are caused by `component.get(node)` falsely marking them as components. These cases are found from test cases as placeholders:

```js
    /* TODO These minor cases are currently falsely marked due to component detection
    {
      code: `
      function ParentComponent() {
        const _renderHeader = () => <div />;
        return <div>{_renderHeader()}</div>;
      }
      `
    },
    {
      // https://github.com/emotion-js/emotion/blob/a89d4257b0246a1640a99f77838e5edad4ec4386/packages/jest/test/react-enzyme.test.js#L26-L28
      code: `
      const testCases = {
        basic: {
          render() {
            const Component = () => <div />;
            return <div />;
          }
        }
      }
      `
    },
    */
```

lib/rules/no-unstable-nested-components.js
```diff
@@ -0,0 +1,466 @@
...
+ * @param {ASTNode} node The AST node
+ * @returns {Boolean} True if component is declared inside a render prop, false if not
+ */
+function isComponentInRenderProp(node) {
+  if (
+    node
+    && node.parent
+    && node.parent.type === 'Property'
+    && node.parent.key
+    && startsWithRender(node.parent.key.name)) {
```
ljharb, line 169:
```suggestion
    node
    && node.parent
    && node.parent.type === 'Property'
    && node.parent.key
    && startsWithRender(node.parent.key.name)
  ) {
```

ljharb (approved):

PR #2759, opened 2020-08-15
[Fix] `display-name`: Get rid of false position on component detection

### LINKED ISSUE #2751: react/display-name false positive on ant-design ColumnsType

* **ESLint Version:** 7.5.0
* **ESLint Plugin React Version:** 7.20.3
* **Node Version:** 10.15.3
* **npm Version:** 6.4.1

**What parser (default, Babel-ESLint, etc.) are you using?**

<details>

```js
  extends: [
    'eslint:recommended',
    'plugin:@typescript-eslint/eslint-recommended',
    'plugin:@typescript-eslint/recommended',
    'plugin:react/recommended',
    'plugin:jsx-a11y/recommended',
    // Prettier plugin and recommended rules
    'prettier/@typescript-eslint',
    'plugin:prettier/recommended',
  ],
  rules: {
    // Include .prettierrc.js rules
    'prettier/prettier': ['error', {}, { usePrettierrc: true }],
    'react/prop-types': 'error',
    'react/react-in-jsx-scope': 'error',
    '@typescript-eslint/explicit-function-return-type': 'error',
    '@typescript-eslint/ban-ts-ignore': 'off',
    'jsx-a11y/label-has-associated-control': [
      'error',
      {
        labelComponents: [],
        labelAttributes: [],
        controlComponents: [],
        assert: 'either',
        depth: 25,
      },
    ],
    'jsx-a11y/anchor-is-valid': 'off',
    '@typescript-eslint/no-explicit-any': 'off',
    '@typescript-eslint/explicit-module-boundary-types': 'error',
    '@typescript-eslint/no-unused-vars': 'error',
  },
```

</details>

```js
...
import { ColumnsType } from "antd/lib/table";
import { Table } from "antd";

const profilesColumns: ColumnsType<any> = [
  {
    title: "ID",
    dataIndex: "id",
    key: "id"
  },
  {
    title: "URL",
    dataIndex: "url",
    key: "url",
    render: (url: string): React.ReactElement => (
      <a href={url} target="_blank" rel="noopener noreferrer">
        <p>lol</p>
      </a>
    )
  }
];

ReactDOM.render(
  <Table columns={profilesColumns} />,
  document.getElementById("root")
);
```

```bash
eslint --fix . --ext .ts,.tsx,.js,.jsx
```

For the Ant Design, `render` in `ColumnsType` is not a `React component` but function, hence the rule should not be applied in this scenario. No error should show up.

```bash
yarn run v1.17.3
$ eslint --fix . --ext .ts,.tsx,.js,.jsx

/home/user/my-project/components/ComponentTable.tsx
  32:13  error  Component definition is missing display name  react/display-name

✖ 1 problem (1 error, 0 warnings)

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

### LINKED ISSUE #2758: v7.20.6 has breaking changes for react/display-name
before `v7.20.6` the `react/display-name` rule was okay with this pattern:
```jsx
const renderer = a => listItem => (
	<div>{a} {listItem}</div>
);

// for context, that function might be used like this:
<List list={[1, 2, 3]} renderItem={renderer(a)} />
```

Since `v7.20.6` the `react/display-name` rule reports that "Component definition is missing display name" on line 1. 

There's nothing in the changelog for `v7.20.6` regarding the react/display-name rule so I suspect this is unintentional?

I'm also confused as to how we'd fix this because `renderer` is not a component, nor is the function it returns. So maybe this is a false positive?

iiison:
Fixes #2751. Closes #2758.

### DIFF
--- a/tests/lib/rules/display-name.js
+++ b/tests/lib/rules/display-name.js
@@ -504,6 +504,26 @@ ruleTester.run('display-name', rule, {
       class Link extends Component<LinkProps> {}
     `,
     parser: parsers.BABEL_ESLINT
+  }, {
+    code: `
+      const x = {
+        title: "URL",
+        dataIndex: "url",
+        key: "url",
+        render: url => (
+          <a href={url} target="_blank" rel="noopener noreferrer">
+            <p>lol</p>
+          </a>
+        )
+      }
+    `,
+    parser: parsers.BABEL_ESLINT
+  }, {
+    code: `
+      const renderer = a => function Component(listItem) {
+        return <div>{a} {listItem}</div>;
+      };
+    `
   }],
 
   invalid: [{
@@ -921,5 +941,12 @@ ruleTester.run('display-name', rule, {
       endLine: 11,
       endColumn: 11
     }]
+  }, {
+    code: `
+      const renderer = a => listItem => (
+        <div>{a} {listItem}</div>
+      );
+    `,
+    errors: [{message: 'Component definition is missing display name'}]
   }]
 });
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,11 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ### Added
 * add [`no-namespace`] rule ([#2640] @yacinehmito @ljharb)
 
+### Fixed
+* [`display-name`]: Get rid of false position on component detection ([#2759] @iiison)
+
 [#2640]: https://github.com/yannickcr/eslint-plugin-react/pull/2640
+[#2759]: https://github.com/yannickcr/eslint-plugin-react/pull/2759
 
 ## [7.25.3] - 2021.09.19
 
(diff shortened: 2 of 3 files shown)


tests/lib/rules/display-name.js
```diff
@@ -894,5 +908,21 @@ ruleTester.run('display-name', rule, {
     errors: [{
       message: 'Component definition is missing display name'
     }]
+  }, {
+    code: `
+      const x = {
+        title: "URL",
+        dataIndex: "url",
+        key: "url",
+        render: url => (
```
jzabala, line 917: I don't think that under any conditions this function should be considered a component. Maybe if the name were uppercase, since that is a pattern I have seen in some libraries, something like:

```jsx
<MyComponent.Button />
```

But not this way. What do you think @ljharb ?

iiison: I thought of the same thing after your last comment on the issue :D  

I am a little new to this codebase. could you please point me to the file/logic where we're classifying the code as a component?

ljharb: Yes, I agree. This shouldn't be a component.

ljharb:
Please unmark this as a draft when it's updated.

k-yle:
This also fixes #2758

koengsen:
Hey Guys, any news on this one?

PR #2761, opened 2020-08-20
[New] `jsx-handler-names`: add `checkInlineFunction` option

### LINKED ISSUE #2586: react/jsx-handler-names should not flag arrow functions
Given the following configuration:

```yml
react/jsx-handler-names:
  - error
  - eventHandlerPrefix: '(handle|on|set)'
    checkLocalVariables: true
```

The following code:

```tsx
<Button onClick={() => setVisible(!visible)}>
  {visible ? 'Hide' : 'Show'}
</Button>
```

fails linting with the following message:

```
Handler function for onClick prop key must begin with '(handle|on|set)'  react/jsx-handler-names
```

We believe `react/jsx-handler-names` should not trigger on arrow functions, because there is no way to address the handler naming violation without refactoring into a variable or similar (which, while probably valid, is another rule/problem).
Thoughts?

dididy:
Add `checkInlineFunction` option for `jsx-handler-names` rule

PR resolve #2586

### DIFF
--- a/docs/rules/jsx-handler-names.md
+++ b/docs/rules/jsx-handler-names.md
@@ -31,7 +31,8 @@ The following patterns are **not** considered warnings:
 "react/jsx-handler-names": [<enabled>, {
   "eventHandlerPrefix": <eventHandlerPrefix>,
   "eventHandlerPropPrefix": <eventHandlerPropPrefix>,
-  "checkLocalVariables": <boolean>
+  "checkLocalVariables": <boolean>,
+  "checkInlineFunction": <boolean>
 }]
 ...
 ```
@@ -39,6 +40,7 @@ The following patterns are **not** considered warnings:
 * `eventHandlerPrefix`: Prefix for component methods used as event handlers. Defaults to `handle`
 * `eventHandlerPropPrefix`: Prefix for props that are used as event handlers. Defaults to `on`
 * `checkLocalVariables`: Determines whether event handlers stored as local variables are checked. Defaults to `false`
+* `checkInlineFunction`: Determines whether event handlers set as inline functions are checked. Defaults to `false`
 
 ## When Not To Use It
 
--- a/lib/rules/jsx-handler-names.js
+++ b/lib/rules/jsx-handler-names.js
@@ -27,7 +27,8 @@ module.exports = {
           properties: {
             eventHandlerPrefix: {type: 'string'},
             eventHandlerPropPrefix: {type: 'string'},
-            checkLocalVariables: {type: 'boolean'}
+            checkLocalVariables: {type: 'boolean'},
+            checkInlineFunction: {type: 'boolean'}
           },
           additionalProperties: false
         }, {
@@ -38,7 +39,8 @@ module.exports = {
               type: 'boolean',
               enum: [false]
             },
-            checkLocalVariables: {type: 'boolean'}
+            checkLocalVariables: {type: 'boolean'},
+            checkInlineFunction: {type: 'boolean'}
           },
           additionalProperties: false
         }, {
@@ -49,13 +51,20 @@ module.exports = {
               enum: [false]
             },
             eventHandlerPropPrefix: {type: 'string'},
-            checkLocalVariables: {type: 'boolean'}
+            checkLocalVariables: {type: 'boolean'},
+            checkInlineFunction: {type: 'boolean'}
           },
           additionalProperties: false
         }, {
           type: 'object',
           properties: {
             checkLocalVariables: {type: 'boolean'}
+          },
+          additionalProperties: false
+        }, {
+          type: 'object',
+          properties: {
+            checkInlineFunction: {type: 'boolean'}
           },
           additionalProperties: false
         }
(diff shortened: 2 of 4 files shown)


ljharb:
If you end up submitting a new PR, please reopen and reuse this one, rather than creating a new one and leaving this one untouched. Thanks!

docs/rules/jsx-handler-names.md
```diff
@@ -31,14 +31,16 @@ The following patterns are **not** considered warnings:
...
+  "checkLocalVariables": <boolean>,
+  "checkInlineFunction": <boolean>
 }]
 ...
 ```
 
 * `eventHandlerPrefix`: Prefix for component methods used as event handlers. Defaults to `handle`
 * `eventHandlerPropPrefix`: Prefix for props that are used as event handlers. Defaults to `on`
 * `checkLocalVariables`: Determines whether event handlers stored as local variables are checked. Defaults to `false`
+* `checkInlineFunction`: Determines whether event hanlders used as inline function. Defaults to `false`
```
roshcagra, line 43: I think 'Determines whether event handlers set as inline functions are checked' would be better

lib/rules/jsx-handler-names.js
```diff
@@ -68,6 +77,16 @@ module.exports = {
       return prefix === false;
     }
 
+    function isInlineHandler(node) {
+      return node.value.expression.type === 'ArrowFunctionExpression';
+    }
+
+    function isLocalInlineHandler(node) {
```
roshcagra, line 84: I'm not exactly sure what this is returning, but it's not a boolean. Should this be called 'getInlineHandler' or something like that?

lib/rules/jsx-handler-names.js
```diff
@@ -86,14 +105,27 @@ module.exports = {
...
+            node,
+            message: 'Expected handler function should be inline function.'
+          });
+        }
+
         const propKey = typeof node.name === 'object' ? node.name.name : node.name;
-        const propValue = context.getSourceCode().getText(node.value.expression).replace(/^this\.|.*::/, '');
+        const propValue = isInlineHandler(node)
+          ? isLocalInlineHandler(node)
+          : context.getSourceCode().getText(node.value.expression).replace(/^this\.|.*::/, '');
```
roshcagra, line 128: I'm not familiar with how eslint rules work, so you should probably have another person review this just to be certain!

docs/rules/jsx-handler-names.md
dididy: Oh thank you very much! That sentence is proper to understand. I'll reflect it next commit!

lib/rules/jsx-handler-names.js
dididy: This function is the part to be examined for local variable if inline handler.

The AST in the case of inline handler and local variable is as follows. (node.value.expression.body.callee.name)
Screen Shot 2020-08-25 at 2 23 00 AM

If the inline handler is not a local variable, the AST is as follows. (node.value.expression.body.callee.property.name)
Screen Shot 2020-08-25 at 2 25 23 AM

In this case, `getPropValue` is a better name because it's not a boolean and it returning propValue. Would you agree with it?

lib/rules/jsx-handler-names.js
dididy: This situation is to get the name of handler function.

1. If handler function is an inline function, you have to care whether it's local variable or not. [AST structure](https://github.com/yannickcr/eslint-plugin-react/pull/2761/files#r475780562)
2. If handler function is not an inline function, the name of handler function can be imported into regex. [line 128](https://github.com/yannickcr/eslint-plugin-react/pull/2761/files#diff-ccf82a24ec6f364517e4d1f4aed7a0a8R128)

lib/rules/jsx-handler-names.js
roshcagra: Sounds good!

lib/rules/jsx-handler-names.js
roshcagra: Ah got it - ok this looks fine to me!

dididy:
@ljharb Could you review this PR?

lib/rules/jsx-handler-names.js
dididy: I remove getHandlerName function, because below code is more flexible.

`context.getSourceCode().getText(node.value.expression.body.callee).replace(/^this\.|.*::/, '')`

https://github.com/yannickcr/eslint-plugin-react/commit/ddcf12cf4b1e75b32b84807b3fc06d3a2c573d14#diff-ccf82a24ec6f364517e4d1f4aed7a0a8R114

lib/rules/jsx-handler-names.js
dididy: https://github.com/yannickcr/eslint-plugin-react/pull/2761#discussion_r476542714

roshcagra:
What's left to do on this PR?

roshcagra:
I'm trying this out on the new release and I don't think it identifies arrow functions without curly braces correctly - should I open a new issue?

ljharb:
Thanks, a new issue, or a PR with failing test cases, would be great.

PR #2763, opened 2020-08-22
[New] `jsx-no-constructed-context-values`: add rule for unstable context values

dylanOshima:
# Rule `JSX-no-constructed-context-values`
## Summary
Adds a new rule that checks if the value prop passed to a Context Provider will cause needless rerenders. A common example is inline object declaration such as:
```js
<Context.Provider value={{foo: 'bar'}}>
...
</Context.Provider>
```
which will cause the Context to rerender every time this code is run because the object is defined with a new reference. This is true for many other types of JS objects and variables.

## Rationale 
This can lead to performance hits since not only will this rerender all the children of a context, it will also update all consumers of the value. The search React does to search the tree can also be very expensive. Some other instances that this rule covers are: array declarations, function declarations (i.e. arrow functions), or new class object instantiates.

## Possible Fixes
For inline values this can be fixed by wrapping the in-line declaration with `useMemo` or `useCallback` as the lint rule suggests. I think this can depend on the usage and specific implementation and thus an auto-fix rule is not provided. For the example above this would look like:
```js
const contextVal = useMemo(() => {foo: 'bar'}, []);
<Context.Provider value={contextVal}>
...
</Context.Provider>
```

## Improvements
1. It currently doesn't provide any possible auto-fixes, although it provides two suggestions: `useCallback` for functions declarations/definitions, and `useMemo` for everything else. I was considering suggesting moving the reference outside of the component, but like I mentioned before I feel this can be case-by-case and I'm hesitant making this an auto-fix rule.
2. It doesn't handle variables being reassigned. It's noted in one of the tests and I would greatly appreciate help if there's an easy fix I'm not seeing. For example:
```js
function Component(foo) {
  let bar = 10;
  bar = {};
  return (<Context.Provider value={bar}></Context.Provider>);
}
// Does not fire the rule because it was initialized to a Literal
```
3. Not sure if this should be a recommended lint rule, I decided to be more conservative and chose not to.

## Contributions
Thanks to @captbaritone for noticing the issue and creating the [React hooks version](https://github.com/facebook/react/pull/19590#issuecomment-674041726) which this is largely based off of, and huge thanks to @bradzacher for suggesting I make this rule open source here 👍

### DIFF
--- a/docs/rules/jsx-no-constructed-context-values.md
+++ b/docs/rules/jsx-no-constructed-context-values.md
@@ -1,0 +1,37 @@
+# Prevent react contexts from taking non-stable values (react/jsx-no-constructed-context-values)
+
+This rule prevents non-stable values (i.e. object identities) from being used as a value for `Context.Provider`.
+
+## Rule Details
+
+One way to resolve this issue may be to wrap the value in a `useMemo()`. If it's a function then `useCallback()` can be used as well.
+
+If you _expect_ the context to be rerun on each render, then consider adding a comment/lint supression explaining why.
+
+## Examples
+
+Examples of **incorrect** code for this rule:
+
+```
+return (
+    <SomeContext.Provider value={{foo: 'bar'}}>
+        ...
+    </SomeContext.Provider>
+)
+```
+
+Examples of **correct** code for this rule:
+
+```
+const foo = useMemo(() => {foo: 'bar'}, []);
+return (
+    <SomeContext.Provider value={foo}>
+        ...
+    </SomeContext.Provider>
+)
+```
+
+## Legitimate Uses
+React Context, and all its child nodes and Consumers are rerendered whenever the value prop changes. Because each Javascript object carries its own *identity*, things like object expressions (`{foo: 'bar'}`) or function expressions get a new identity on every run through the component. This makes the context think it has gotten a new object and can cause needless rerenders and unintended consequences.
+
+This can be a pretty large performance hit because not only will it cause the context providers and consumers to rerender with all the elements in its subtree, the processing for the tree scan react does to render the provider and find consumers is also wasted.
--- a/lib/rules/jsx-no-constructed-context-values.js
+++ b/lib/rules/jsx-no-constructed-context-values.js
@@ -1,0 +1,215 @@
+/**
+ * @fileoverview Prevents jsx context provider values from taking values that
+ *               will cause needless rerenders.
+ * @author Dylan Oshima
+ */
+
+'use strict';
+
+const docsUrl = require('../util/docsUrl');
+
+// ------------------------------------------------------------------------------
+// Helpers
+// ------------------------------------------------------------------------------
+
+// Recursively checks if an element is a construction.
+// A construction is a variable that changes identity every render.
+function isConstruction(node, callScope) {
+  switch (node.type) {
+    case 'Literal':
+      if (node.regex != null) {
+        return {type: 'regular expression', node};
+      }
+      return null;
+    case 'Identifier': {
+      const variableScoping = callScope.set.get(node.name);
+
+      if (variableScoping == null || variableScoping.defs == null) {
+        // If it's not in scope, we don't care.
+        return null; // Handled
+      }
+
+      // Gets the last variable identity
+      const variableDefs = variableScoping.defs;
+      const def = variableDefs[variableDefs.length - 1];
+      if (def != null
+        && def.type !== 'Variable'
+        && def.type !== 'FunctionName'
+      ) {
+        // Parameter or an unusual pattern. Bail out.
+        return null; // Unhandled
+      }
+
+      if (def.node.type === 'FunctionDeclaration') {
+        return {type: 'function declaration', node: def.node, usage: node};
+      }
+
+      const init = def.node.init;
+      if (init == null) {
+        return null;
+      }
+
+      const initConstruction = isConstruction(init, callScope);
+      if (initConstruction == null) {
+        return null;
+      }
+
+      return {
+        type: initConstruction.type,
+        node: initConstruction.node,
+        usage: node
+      };
+    }
+    case 'ObjectExpression':
+      // Any object initialized inline will create a new identity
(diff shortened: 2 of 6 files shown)


captbaritone:
Looks like tests are failing on ESLint version 4

captbaritone:
Interesting! If destructing is not supported, I wonder if there's a lint rule we could add to this repo's eslint config that would disallow it.

That would certainly make it easier for the next person.

docs/rules/jsx-no-constructed-context-values.md
```diff
@@ -0,0 +1,37 @@
...
+const foo = useMemo(() => {foo: 'bar'}, []);
+return (
+    <SomeContext.Provider value={foo}>
+        ...
+    </SomeContext.Provider>
+)
+```
+
+## Legitimate Uses
+React Context, and all its child nodes and Consumers are rerendered whenever the value prop changes. Because of Javascript's pass by reference nature, things like object expressions (`{foo: 'bar'}`) or function expressions get a new reference on every run through the component. This can cause needless rerenders and unintended consequences.
```
ljharb, line 35: JavaScript is strictly, alway, and only ever pass by value (see https://web.archive.org/web/20161005155047/http://www.jon-carlos.com/2013/is-javascript-call-by-value-or-call-by-reference/).

The issue is that like everything else in javascript, objects are compared by *identity*, not contents, and if you create a new object on every render, that identity changes across renders.

lib/rules/jsx-no-constructed-context-values.js
```diff
@@ -0,0 +1,215 @@
...
+// ------------------------------------------------------------------------------
+// Helpers
+// ------------------------------------------------------------------------------
+
+// Recursively checks if an element is a construction.
+// A construction is a variable that changes reference every render.
+function isConstruction(node, callScope) {
+  switch (node.type) {
+    case 'Literal':
+      if (node.value instanceof RegExp) {
```
ljharb, line 20: instanceof is highly unreliable and should be avoided on principle; let's use https://www.npmjs.com/package/is-regex if this is needed.

tests/lib/rules/jsx-no-constructed-context-values.js
```diff
@@ -0,0 +1,408 @@
...
+        data: {
+          variableName: 'Wrapper',
+          type: 'JSX element',
+          nodeLine: '1',
+          usageLine: '1'
+        }
+      }]
+    },
+    {
+      // Invalid because RegEx returns a new object which has will be a new reference
```
ljharb, line 376: please update the terminology throughout the PR; "reference" is inaccurate.

docs/rules/jsx-no-constructed-context-values.md
dylanOshima: My mistake, thanks for clarifying!

lib/rules/jsx-no-constructed-context-values.js
dylanOshima: Before we go down that route another method I was using was just checking that there was a [regex property in the literal node](https://astexplorer.net/#/gist/ef24fe782783aedcdf565b24c2585f34/latest) but I had thought the `instanceof` might be safer.

dylanOshima:
> Is this rule only useful with React 16.3+'s Context? What about regular context?

This rule was built with React 16.3's contexts in mind and that's where the test coverage mainly lies. But its functionality is general in that it will detect any value that could change _identity_. Could you elaborate on "regular contexts"? I could try and add tests to see.

lib/rules/jsx-no-constructed-context-values.js
ljharb: I'm fine relying on the AST structure; instanceof isn't ever safe, unfortunately.

ljharb:
I mean like, pre-16.3 React context, like the `getChildContext` class component method in particular.

lib/rules/jsx-no-constructed-context-values.js
bradzacher: Could you please elaborate on why it's not safe in this context?

I would have thought that based on the context of a lint rule, we can make some assumptions that imply enough safety:
- The AST is provided by a parser (via ESLint), so it's not some user generated object passed to a function.
- The AST is assumed to conform [to the ESTree spec for literals](https://github.com/estree/estree/blob/master/es5.md#literal)
- The `value` prop is assumed to be one of a set of known of types (string, boolean, number, bigint, null, or RegExp)

Unless I'm wrong, the only case that it's not safe is if someone has purposely crafted a parser for eslint which provides an AST that is not safe?

bradzacher:
This rule won't work for pre <16.3 contexts because that mechanism is fundamentally different to the context used for 16.3+

I don't know too much about the legacy context API, but reading the old docs, I believe it only lets you pass objects down the chain, returned from the method.
I'd assume that API compares the objects by property rather than equality (or I'd hope so, considering its a lifecycle style method which always returns a new object).

ljharb:
Fair enough.

lib/rules/jsx-no-constructed-context-values.js
ljharb: It wouldn't work if a parser created the regex in another realm, like in the `vm` module, but you're right that in this specific use case it's probably fine. I'd still prefer to avoid it to keep up a good coding habit.

bradzacher: I didn't know about that API! That makes sense then.

dylanOshima:
Hi @ljharb just wanted to double check if there was still things that needed changing.

captbaritone:
Following up here to see if there's anything else needed before this can be merged.

ljharb (approved):
Rebased this; I think it looks good but I want to give it another look.

mqklin:
It's not released yet, right? I can see it in README, but eslint says that Definition for rule 'react/jsx-no-constructed-context-values' was not found

ljharb:
Yes - check the readme on the latest tag, not on master, to see what’s released.

PR #2764, opened 2020-08-23
jsx-no-script: fix <a href ... /> (no value)

wmertens:


ljharb:
This appears to have been fixed in 63d5a1b0, to fix #2871.

PR #2767, opened 2020-08-24
[Fix] `forbid-component-props`: Implemented support for "namespaced" components.

### LINKED ISSUE #2766: forbid-component-props: How to distinguish a component from a library?
Situation:

```tsx
import {Layout as AntdLayout} from 'antd';
...
            <AntdLayout.Content className={contentClassNames}>
...
```

Adding `AntdLayout` nor `AntdLayout.Content` into `allowedFor` isn't working. I don't want to add `Content`, because then linter won't report an error when our `Content` component is used with `className`.

Edit: I have attempted to improve it in https://github.com/yannickcr/eslint-plugin-react/pull/2767.

mnn:
I don't really know what I am doing. I have just looked at the structure of the `node`, added a test and updated code so the tests pass. 

Fixes #2766.

### DIFF
--- a/lib/rules/forbid-component-props.js
+++ b/lib/rules/forbid-component-props.js
@@ -78,8 +78,11 @@ module.exports = {
 
     return {
       JSXAttribute(node) {
-        const tag = node.parent.name.name;
-        if (tag && tag[0] !== tag[0].toUpperCase()) {
+        const parentName = node.parent.name;
+        // Extract a component name when using a "namespace", e.g. `<AntdLayout.Content />`.
+        const tag = parentName.name || `${parentName.object.name}.${parentName.property.name}`;
+        const componentName = parentName.name || parentName.property.name;
+        if (componentName && componentName[0] !== componentName[0].toUpperCase()) {
           // This is a DOM node, not a Component, so exit.
           return;
         }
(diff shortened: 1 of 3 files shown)


lib/rules/forbid-component-props.js
```diff
@@ -78,7 +78,12 @@ module.exports = {
 
     return {
       JSXAttribute(node) {
-        const tag = node.parent.name.name;
+        const parentName = node.parent.name;
+        let tag = parentName.name;
+        if (!tag && parentName.object.name !== 'this') {
```
ljharb, line 83: why not `this`? iow, `<this.Foo />` should be valid as well, no?

mnn: One test was failing, so I thought there is some reason I am not aware of behind forbidding it.

ljharb: Why would it be a different component every time? `this.Component` could just be an own property on the instance, the same every time.

I think it's reasonable to assume that `this` inside a class method is always the class instance.

ljharb: I've taken care of it; it's that the "first letter is capitalized" check wasn't yet taking into account the namespace and the dot.

PR #2769, opened 2020-08-26
[New] `jsx-max-props-per-line`: add multi/single line config options

### LINKED ISSUE #2735: Is there any way to multiline all JSX props if there are more than X props on element?
Hello, I want to achieve a behavior when JSX props will all be multiline if amount of props is more than defined value. Currently there is `react/jsx-max-props-per-line` that is the closest to what I need, but if I set its `maximum` to 2, it groups props in multiple lines grouped by 2. So this JSX:
```
<Component prop1="value 1" prop2="value 2" prop3="value 3" prop4="value 4" />
```

Will be formatted to this:
```
<Component
  prop1="value 1" prop2="value 2"
  prop3="value 3" prop4="value 4"
/>
```

While I want to achieve this:
```
<Component
  prop1="value 1"
  prop2="value 2"
  prop3="value 3"
  prop4="value 4"
/>
```

Is it possible to add a rule like `react/jsx-props-multiline` with attribute like `after` with amount of props after which it should be multiline?

iiison:
- Update rule to accept different configs for single and multiline tags.
- Update fixer for the same

Fixes #2735

### DIFF
--- a/lib/rules/jsx-max-props-per-line.js
+++ b/lib/rules/jsx-max-props-per-line.js
@@ -8,6 +8,13 @@
 const docsUrl = require('../util/docsUrl');
 const report = require('../util/report');
 
+function getPropName(context, propNode) {
+  if (propNode.type === 'JSXSpreadAttribute') {
+    return context.getSourceCode().getText(propNode.argument);
+  }
+  return propNode.name.name;
+}
+
 // ------------------------------------------------------------------------------
 // Rule Definition
 // ------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


tests/lib/rules/jsx-max-props-per-line.js
```diff
@@ -32,22 +48,23 @@ ruleTester.run('jsx-max-props-per-line', rule, {
     code: '<App foo />'
   }, {
     code: '<App foo bar />',
-    options: [{maximum: 2}]
+    options: [getConfigs(2)]
```
ljharb, line 51: i'm not convinced this is clearer than just repeating each configuration inline. It's much harder for me to evaluate the test cases this way. Can you update this to remove `getConfigs`?

lib/rules/jsx-max-props-per-line.js
```diff
@@ -24,8 +24,17 @@ module.exports = {
...
+          properties: {
+            single: {
+              type: 'integer',
+              minimum: 1
+            },
+            multi: {
+              type: 'integer',
+              minimum: 1
+            }
+          }
```
ljharb, line 37:
```suggestion
          },
          "additionalProperties": false
```
(if i'm remembering the syntax right)

lib/rules/jsx-max-props-per-line.js
```diff
@@ -24,8 +24,17 @@ module.exports = {
       type: 'object',
       properties: {
         maximum: {
-          type: 'integer',
-          minimum: 1
+          type: 'object',
+          properties: {
```
ljharb, line 28: this is a breaking change unless it uses `anyOf` or `oneOf` to allow both an integer *and* an object.

dherault:
Is anyone working on this? Can I take a shot at it?

iiison:
@dherault please go ahead. I'm a little busy with my office stuff.

ljharb:
@dherault please do not open a new PR - please post a link here to a branch/commit, and I'll update this PR.

SIL0RAK:
Duplicate of 3078

PR #2770, opened 2020-08-26
[Fix] react/prefer-read-only-props fails if using Flow's $ReadOnly<T>

karolina-benitez:


ljharb:
@karolina-benitez why close/delete?

karolina-benitez:
I was second guessing my understanding of the issue. Just to clarify, I need to adjust the rule to allow the object level annotation `$ReadOnly` to replace the need to add the individual annotation `+` to each attribute?

ljharb:
To be honest I'm not super clear, I'm not familiar with Flow :-/ my personal approach here would be to write the tests first, and then blunder through the implementation until i'd figured out how to fix it.

karolina-benitez:
Gotcha, from the Flow documentation:
``` 
props: $ReadOnly<{|
    dataTest?: string,
    className?: string,
  |}>,
``` is valid syntax but the current rule is returning an error because it's expecting each individual attribute to be annotated. Like so:
``` 
props: $ReadOnly<{|
    +dataTest?: string,
    +className?: string,
  |}>,
```
I think my initial understanding was correct and the first example should pass. I'll be adjusting my tests and will reopen the PR shortly

ljharb:
That seems right to me!

ljharb:
@karolina-benitez would you mind restoring the branch and reopening this PR, so i can keep it in sync with #2772?

ljharb:
To be specific, the desired workflow was to click the "restore branch" button on this PR, and then "reopen pull request". Pushing to the branch directly makes this PR unrecoverable.

karolina-benitez:
Oh no, my mistake. What should be my next steps?

ljharb:
If this PR in fact is unrecoverable (if there's no "restore branch" button) then I think at this point we'll just leave the other two PRs open; i'll keep them in sync, and this one will have to be left as-is.

karolina-benitez:
Yes it is unrecoverable, I don’t see a “restore branch” button. I appreciate your patience while I stumble around, I’m still trying to figure this all out

ljharb:
No worries, it happens.

Feel free to hit me up on any medium in the future if you want to pair on this sort of thing :-)

karolina-benitez:
That would be incredibly helpful, thank you!

PR #2771, opened 2020-08-26
function-component-definition: ignore properties

### LINKED ISSUE #2765: function-component-definition breaks code
When I auto-fix function-component-definition, it decides that the below is a function component (it's not really) and that it should use arrow notation:

```js
const RULES = [
	{
		serialize(el) {
			if (el.type !== 'tag') return
			const data = el.data.toJS ? el.data.toJS() : el.data
			const text = el.nodes[0].text

			return <Tag linked value={{...data, props: {label: text}}} />
		},
	},
]
```

It proceeds to auto-fix like this:
```js
const RULES = [
	{
		serialize(el) => {
[...]
```

which is invalid. It should have changed it to `serialize: el => {` instead

Another option is that the rule ignores object function notation.

stefan-wullems:
Fixes #2765

### DIFF
--- a/tests/lib/rules/function-component-definition.js
+++ b/tests/lib/rules/function-component-definition.js
@@ -166,6 +166,144 @@ ruleTester.run('function-component-definition', rule, {
     code: 'function Hello(props): ReactNode { return <p/> }',
     options: [{namedComponents: 'function-declaration'}],
     parser: parsers.TYPESCRIPT_ESLINT
+  },
+  // https://github.com/yannickcr/eslint-plugin-react/issues/2765
+  {
+    code: [
+      'const obj = {',
+      '  serialize: (el) => {',
+      '    return <p/>',
+      '  }',
+      '}'
+    ].join('\n'),
+    options: [{namedComponents: 'function-declaration'}]
+  }, {
+    code: [
+      'const obj = {',
+      '  serialize: (el) => {',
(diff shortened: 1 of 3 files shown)


tests/lib/rules/function-component-definition.js
```diff
@@ -166,6 +166,16 @@ ruleTester.run('function-component-definition', rule, {
...
+  }, {
+    // https://github.com/yannickcr/eslint-plugin-react/issues/2765
+    code: [
+      'const obj = {',
+      '  serialize(el) {',
+      '    return <p/>',
+      '  }',
+      '}'
+    ].join('\n'),
+    options: [{unnamedComponents: 'arrow-function'}]
```
ljharb, line 178: could we also add test cases with a bunch of other options, to be clear that no concise object method will ever be detected as a component?

wmertens:
Well, the test cases this PR uses should all have been ignored based on PascalCase alone. I don't see a test for it in lib/rules/function-component-definition.js either.

wmertens:
(this was in reply to the issue, sorry, on phone)

ljharb:
@wmertens yes, ignored because they're not PascalCase.

wmertens:
@ljharb no, they weren't ignored before this PR, this PR only adds being part of an object

PR #2779, opened 2020-09-01
[Fix] react/prefer-read-only-props fails if using Flow's $ReadOnly<T>

karolina-benitez:


ljharb:
@karolina-benitez is this a third duplicate of #2772 and #2770? The hope was to have *one* PR :-)

karolina-benitez:
Oof yea 😩 my apologies

Issue #2309, opened 2019-06-13
[react/destructuring-assignment] Shows weird behaviour when I declare a variable with the name context


roman-16:
In a functional component I get an error when I declare a variable with the name context.
Example code:

```javascript
const Test = props => {
  const context = useContext(someContext);

  return (
    <div {...props}>
      {context.test} // Here message: 'Must use destructuring context'
    </div>
  );
}
```

ljharb:
The error here should appear imo with any variable name coming from the useContext hook. Does the error disappear if you use a different variable name?

wa4-fearless-otter:
No, then it doesn't appear.

ljharb:
There seems like two bugs then:
1. the variable name `context` shouldn't be special
2. the return of `useContext` should be caught by this rule with every var name (altho this might need an option)

Zinyon:
I think the error message on that case is correct following this [rule,](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/destructuring-assignment.md) isn't it?
can you explain more why there is a bug ?

ljharb:
@Zinyon the rule is supposed to enforce _legacy_ context being destructured; it's not supposed to warn on `useContext`, or on any variable that happens to be named `context`.

Zinyon:
@ljharb Thank you, it makes sense! and I'll try fix this.

PR #2787, opened 2020-09-05
[New] 'destructuring-assigment' : add 'useContext' rule in SFC

Zinyon:
resolve issue #2309

1. delete rule about variable happens to be named ``context``

2. set rule on ``useContext`` with ``options: always`` and ``options: never``
- make ``contextSet`` to detect any renamed variable that is being assigned by ``useContext``

### DIFF
--- a/tests/lib/rules/destructuring-assignment.js
+++ b/tests/lib/rules/destructuring-assignment.js
@@ -176,6 +176,18 @@ ruleTester.run('destructuring-assignment', rule, {
     ].join('\n'),
     options: ['always', {ignoreClassFields: true}],
     parser: parsers.BABEL_ESLINT
+  }, {
+    code: `const MyComponent = (props) => {
+      const {foo} = useContext(aContext);
+      return <div>{foo}</div>
+    };`,
+    options: ['always']
+  }, {
+    code: `const MyComponent = (props) => {
+      const foo = useContext(aContext);
+      return <div>{foo.test}</div>
+    };`,
+    options: ['never']
   }],
 
   invalid: [{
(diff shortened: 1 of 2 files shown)


tests/lib/rules/destructuring-assignment.js
```diff
@@ -193,14 +205,6 @@ ruleTester.run('destructuring-assignment', rule, {
...
       {message: 'Must never use destructuring props assignment in SFC argument'}
     ]
-  }, {
-    code: `const MyComponent = (props, { color }) => (
-      <div id={props.id} className={props.className} />
-    );`,
-    options: ['never'],
-    errors: [
-      {message: 'Must never use destructuring context assignment in SFC argument'}
-    ]
```
ljharb, line 203: why is this test changed? this implies it's a breaking change.

ljharb (changes requested):

tests/lib/rules/destructuring-assignment.js
Zinyon: i think that case is legacy context assignment which is not used in current react version.
sorry for change without any comments.

ljharb: This plugin supports all react versions; nothing can be removed.

Zinyon: i'm sorry. actually i want to ask about that. my mistake. then should I edit and make new PR?

plus,  ``const destructuringContext = node.params && node.params[1] && node.params[1].type === 'ObjectPattern';``should be added in ``function handleStatelessComponent(node)``  right?

Zinyon: @ljharb 
i did revive test cases which i deleted and rules associated. please check again

PR #2790, opened 2020-09-09
Fix no-unknown-property: check attributes with any input case

julienw:
I think this is ready now. I checked my project with this rule and it didn't break anything :-)

I wasn't sure if this rule should also check completely unknown attributes. My view is that it should, but I didn't do it in this patch.

With this patch the rule will probably find new errors in existing code, so it should probably be versioned in a new minor or major version.

Tell me what you think!

(pretest fails but this seems to happen on master too?)

### DIFF
--- a/lib/rules/no-unknown-property.js
+++ b/lib/rules/no-unknown-property.js
@@ -147,7 +147,8 @@ function getDOMPropertyNames(context) {
 // ------------------------------------------------------------------------------
 
 /**
- * Checks if a node matches the JSX tag convention.
+ * Checks if a node matches the JSX tag convention. This also checks if a node
+ * is extended as a webcomponent using the attribute "is".
  * @param {Object} node - JSX element being tested.
  * @returns {boolean} Whether or not the node name match the JSX tag convention.
  */
@@ -194,7 +195,7 @@ function tagNameHasDot(node) {
  * Get the standard name of the attribute.
  * @param {String} name - Name of the attribute.
  * @param {String} context - eslint context
- * @returns {String} The standard name of the attribute.
+ * @returns {String | undefined} The standard name of the attribute, or undefined if no standard name was found.
  */
 function getStandardName(name, context) {
   if (DOM_ATTRIBUTE_NAMES[name]) {
@@ -203,13 +204,9 @@ function getStandardName(name, context) {
   if (SVGDOM_ATTRIBUTE_NAMES[name]) {
     return SVGDOM_ATTRIBUTE_NAMES[name];
   }
-  let i = -1;
   const names = getDOMPropertyNames(context);
-  const found = names.some((element, index) => {
-    i = index;
-    return element.toLowerCase() === name;
-  });
-  return found ? names[i] : null;
+  // Let's find a possible attribute match with a case-insensitive search.
+  return names.find((element) => element.toLowerCase() === name.toLowerCase());
 }
 
 // ------------------------------------------------------------------------------
@@ -259,6 +256,8 @@ module.exports = {
         }
 
         const tagName = getTagName(node);
+
+        // 1. Some attributes are allowed on some tags only.
         const allowedTags = ATTRIBUTE_TAGS_MAP[name];
         if (tagName && allowedTags && /[^A-Z]/.test(tagName.charAt(0)) && allowedTags.indexOf(tagName) === -1) {
           context.report({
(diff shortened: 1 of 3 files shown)


lib/rules/no-unknown-property.js
```diff
@@ -205,11 +206,11 @@ function getStandardName(name, context) {
   }
   let i = -1;
   const names = getDOMPropertyNames(context);
-  const found = names.some((element, index) => {
-    i = index;
-    return element.toLowerCase() === name;
-  });
-  return found ? names[i] : null;
+  // Let's find a possible attribute match with a case-insensitive search.
+  const foundName = names.find(
```
julienw, line 210: I checked that `find` is supported by node 4 before changing the code.

lib/rules/no-unknown-property.js
```diff
@@ -205,11 +206,11 @@ function getStandardName(name, context) {
...
+    (element, index) => element.toLowerCase() === name.toLowerCase()
+  );
+  return foundName === undefined ? null : foundName;
```
ljharb, line 213: any reason not to just `return foundName` here, and have the function return `String | undefined`?

ljharb (commented):
LGTM pending question

lib/rules/no-unknown-property.js
```diff
@@ -205,11 +206,11 @@ function getStandardName(name, context) {
   }
   let i = -1;
```
ljharb, line 207:
```suggestion
```
this is no longer used

lib/rules/no-unknown-property.js
```diff
@@ -205,11 +206,11 @@ function getStandardName(name, context) {
...
   const names = getDOMPropertyNames(context);
-  const found = names.some((element, index) => {
-    i = index;
-    return element.toLowerCase() === name;
-  });
-  return found ? names[i] : null;
+  // Let's find a possible attribute match with a case-insensitive search.
+  const foundName = names.find(
+    (element, index) => element.toLowerCase() === name.toLowerCase()
```
ljharb, line 211:
```suggestion
    (element) => element.toLowerCase() === name.toLowerCase()
```

lib/rules/no-unknown-property.js
julienw: There was 2 reasons:
* initially I wanted to keep it working just like before, without changing too much code -- I ended up changing more code than I wanted :)
* we have this habit in our project that we use `null` as an explicite "no value" whereas `undefined` is accidental.

In this context I agree this isn't so useful, where `getStandardName` is used at only one place. I can change this.

julienw:
Thanks for the review @ljharb, I think I fixed all your comments!

One question: I think this property should also check completely unknown attributes on the html tags (eg `<div foo="bar">`). Maybe this could be an option though... What do you think?

ljharb:
@julienw yes, ideally it should, but we'd need to ensure we can use the same mappings React itself uses, and i don't think they're exposed anywhere. I think we can do that in another PR :-)

lib/rules/no-unknown-property.js
```diff
@@ -194,7 +195,7 @@ function tagNameHasDot(node) {
  * Get the standard name of the attribute.
  * @param {String} name - Name of the attribute.
  * @param {String} context - eslint context
- * @returns {String} The standard name of the attribute.
+ * @returns {String | undefined} The standard name of the attribute, or null if no standard name was found.
```
ljharb, line 198:
```suggestion
 * @returns {String | undefined} The standard name of the attribute, or undefined if no standard name was found.
```

lib/rules/no-unknown-property.js
```diff
@@ -203,13 +204,12 @@ function getStandardName(name, context) {
...
-  const found = names.some((element, index) => {
-    i = index;
-    return element.toLowerCase() === name;
-  });
-  return found ? names[i] : null;
+  // Let's find a possible attribute match with a case-insensitive search.
+  const foundName = names.find(
+    (element) => element.toLowerCase() === name.toLowerCase()
+  );
+  return foundName; // This can be undefined if the value hasn't been found.
```
ljharb, line 212:
```suggestion
  return names.find(
    (element) => element.toLowerCase() === name.toLowerCase()
  );
```
the comment isn't really helpful since that's how `.find` already works :-)

ljharb (approved):
Thanks, I'll clean up these tweaks and land.

PR #2792, opened 2020-09-10
[Fix] "Expected no line gap between" false positive

### LINKED ISSUE #2774: "Expected no line gap between" false positive
I believe one of the last [commits](https://github.com/yannickcr/eslint-plugin-react/commit/5029bd1b588c3f72251e13de190f9df7bd1390d8) on master brings a new unexpected error. For example:

```typescript
<Tab.Screen
    name={`Dia ${index + 1}`}
    component={DayTab}
    // eslint-disable-next-line react/no-array-index-key
    key={index}
/>
```

`error  Expected no line gap between “component” and “key”  react/jsx-props-no-multi-spaces`

@iiison / @ljharb  is this okay? In these scenarios, how should we avoid the error?

karolina-benitez:
Fixes #2774.

### DIFF
--- a/lib/rules/jsx-props-no-multi-spaces.js
+++ b/lib/rules/jsx-props-no-multi-spaces.js
@@ -24,6 +24,8 @@ module.exports = {
   },
 
   create(context) {
+    const sourceCode = context.getSourceCode();
+
     function getPropName(propNode) {
       switch (propNode.type) {
         case 'JSXSpreadAttribute':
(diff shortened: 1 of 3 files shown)


lib/rules/jsx-props-no-multi-spaces.js
```diff
@@ -40,12 +42,17 @@ module.exports = {
...
+      const nodeLineDifference = currNodeStartLine - prevNodeEndLine;
 
-      if (currNodeStartLine - prevNodeEndLine > 1) {
-        context.report({
-          node,
-          message: `Expected no line gap between “${getPropName(prev)}” and “${getPropName(node)}”`
-        });
+      if (nodeLineDifference > 1) {
+        for (let i = 0; i < nodeLineDifference - 1; i++) {
+          if (!sourceCodeArray[prevNodeEndLine + i].includes('//')) {
```
ljharb, line 49: there could also be comments like `/* foo */` or also like `<!-- bar -->` - can we add some tests for those? I have a feeling there's an eslint helper method here that can help ensure there's just whitespace and comments, rather than relying on introspecting text.

ljharb (approved):
Thanks, looks great!

PR #2797, opened 2020-09-16
[Fix] `destructuring-assignment`: Handle destructuring of useContext in SFC

### LINKED ISSUE #2309: [react/destructuring-assignment] Shows weird behaviour when I declare a variable with the name context

Zinyon:
i reopen PR on resolve issue - #2309 cause i did a lot of commits while editing.

* add rule of handling destructuring of useContext in SFC
* delete detecting variable named 'context' in SFC

Fixes #2309. Closes #2787.

### DIFF
--- a/lib/rules/destructuring-assignment.js
+++ b/lib/rules/destructuring-assignment.js
@@ -8,6 +8,7 @@ const Components = require('../util/Components');
 const docsUrl = require('../util/docsUrl');
 const isAssignmentLHS = require('../util/ast').isAssignmentLHS;
 const report = require('../util/report');
+const testReactVersion = require('../util/version').testReactVersion;
 
 const DEFAULT_OPTION = 'always';
 
@@ -94,6 +95,8 @@ module.exports = {
     const destructureInSignature = (context.options[1] && context.options[1].destructureInSignature) || 'ignore';
     const sfcParams = createSFCParams();
 
+    // set to save renamed var of useContext
+    const contextSet = new Set();
     /**
      * @param {ASTNode} node We expect either an ArrowFunctionExpression,
      *   FunctionDeclaration, or FunctionExpression
@@ -128,7 +131,7 @@ module.exports = {
     function handleSFCUsage(node) {
       const propsName = sfcParams.propsName();
       const contextName = sfcParams.contextName();
-      // props.aProp || context.aProp
+      // props.aProp
       const isPropUsed = (
         (propsName && node.object.name === propsName)
           || (contextName && node.object.name === contextName)
@@ -140,6 +143,16 @@ module.exports = {
           data: {
             type: node.object.name,
           },
+        });
+      }
+
+      // const foo = useContext(aContext);
+      // foo.aProp
+      const isContextUsed = contextSet.has(node.object.name) && !isAssignmentLHS(node);
+      if (isContextUsed && configuration === 'always') {
+        context.report({
+          node,
+          message: `Must use destructuring ${node.object.name} assignment`,
         });
       }
     }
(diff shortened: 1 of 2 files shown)


lib/rules/destructuring-assignment.js
```diff
@@ -61,14 +62,24 @@ module.exports = {
     }
 
     function handleSFCUsage(node) {
-      // props.aProp || context.aProp
-      const isPropUsed = (node.object.name === 'props' || node.object.name === 'context') && !isAssignmentLHS(node);
+      // props.aProp
+      const isPropUsed = (node.object.name === 'props') && !isAssignmentLHS(node);
```
ljharb, line 66:
```suggestion
      const isPropUsed = node.object.name === 'props' && !isAssignmentLHS(node);
```

lib/rules/destructuring-assignment.js
```diff
@@ -125,13 +136,29 @@ module.exports = {
...
 
         const destructuring = (node.init && node.id && node.id.type === 'ObjectPattern');
+        const identifier = (node.init && node.id && node.id.type === 'Identifier');
         // let {foo} = props;
-        const destructuringSFC = destructuring && (node.init.name === 'props' || node.init.name === 'context');
+        const destructuringSFC = destructuring && node.init.name === 'props';
+        // let {foo} = useContext(aContext);
+        const destructuringUseContext = destructuring && node.init.callee && node.init.callee.name === 'useContext';
+        // let foo = useContext(aContext);
+        const assignUseContext = identifier && node.init.callee && node.init.callee.name === 'useContext';
```
ljharb, line 145: are we sure `useContext` is the react hook? I think we should make sure that nobody's using a custom hook or function named `useContext`, and that this `useContext` is imported/required from `react`.

ljharb (changes requested):

lib/rules/destructuring-assignment.js
Zinyon: https://reactjs.org/docs/hooks-reference.html#usecontext
here in official docs of react says ``useContext`` is one of basic hooks with ``useState``, ``useEffect``.
and we import ``useContext`` like ``import {useContext} from 'react'``;

I think if somebody's using function named ``useContext`` which overwrite basic hooks , I think we should make another rule about custom hooks which overwrite basic hooks.

ljharb: Someone might be using a version of React that doesn't have hooks.

Zinyon: then is there any possible way to check version of react? or we should not include this part.

ljharb: Yes! Check out `lib/util/version.js`'s `testReactVersion` method.

Zinyon: do I have to import ``testReactVersion`` method in ``lib/rules/destructuring-assignment'`` to check react version in rules? I have no idea.. could you help me with this?

ljharb: Yes, that's right

ljharb:
@Zinyon are you still interested in completing this PR?

ljharb:
The `isDestructuredFromPragmaImport` helper might be useful here.

Zinyon:
@ljharb sure, i will look around ``isDestructuredFromPragmaImport`` helper and complete PR soon!

ljharb:
I went ahead and added the react version dependence; i'll leave "make sure it's destructured from the pragma" as a followup.

ljharb (approved):

Issue #2795, opened 2020-09-15
TypeError: Cannot read property 'type' of undefined


Paperfeed:
eslint-plugin-react `v7.20.6` causes the following error:

```Oops! Something went wrong! :(

ESLint: 7.6.0   (or 7.9.0)

TypeError: Cannot read property 'type' of undefined
Occurred while linting /path-to-file/Component.tsx:101
    at Object.isRequiredPropType (/node_modules/eslint-plugin-react/lib/util/props.js:91:29)
    at /node_modules/eslint-plugin-react/lib/util/propTypes.js:669:54
    at iterateProperties (/node_modules/eslint-plugin-react/lib/util/propTypes.js:64:7)
    at /node_modules/eslint-plugin-react/lib/util/propTypes.js:664:23
    at Array.forEach (<anonymous>)
    at DeclarePropTypesForTSTypeAnnotation.convertReturnTypeToPropTypes (/node_modules/eslint-plugin-react/lib/util/propTypes.js:653:34)
    at DeclarePropTypesForTSTypeAnnotation.searchDeclarationByName (/node_modules/eslint-plugin-react/lib/util/propTypes.js:562:14)
    at DeclarePropTypesForTSTypeAnnotation.visitTSNode (/node_modules/eslint-plugin-react/lib/util/propTypes.js:517:14)
    at DeclarePropTypesForTSTypeAnnotation.traverseDeclaredInterfaceOrTypeAlias (/node_modules/eslint-plugin-react/lib/util/propTypes.js:610:14)
    at Array.forEach (<anonymous>)
```

At propTypes.js it goes into the switch case for 'ObjectExpression' 
```
switch (res.type) {
    case 'ObjectExpression':
        iterateProperties(context, res.properties, (key, value, propNode) => {
            const types = buildReactDeclarationTypes(value, key);
            types.fullName = key;
            types.name = key;
            types.node = propNode;
            types.isRequired = propsUtil.isRequiredPropType(value);
            this.declaredPropTypes[key] = types;
        });
        break;
    default:
}
```

with the following values:
```
┌───────────────┬──────────┬──────────┬──────────┬──────────┬──────────┬──────────┬──────────────────────┬─────────┬────────────────────┬─────────────────┬──────────┬──────────────────┬────────────────────────────────────┬──────────┬───────────┐
[...]

PR #2802, opened 2020-09-23
[Fix] `prop-types`: Should handle CallExpression in ReturnType

hank121314:
## Summary

This pr fixes: #2795.

This pr handle CallExpression which come from two places.
1. From ReturnType function body:
For example:
```
const mapDispatchToProps = (dispatch: ThunkDispatch<State, null, Action>) => 
  bindActionCreators<{props1: ()=>void,props2: ()=>string}>(
    { prop1: importedAction, prop2: anotherImportedAction },
    dispatch,
)
```

2. From RerturnType function return object spread properties:
For example:
```
const mapDispatchToProps = (dispatch: ThunkDispatch<State, null, Action>) => ({
  ...bindActionCreators<{prop1: ()=>void,prop2: ()=>string}>(
    { prop1: importedAction, prop2: anotherImportedAction },
    dispatch,
  ),
})
```

If user give us typeParameters of CallExpression, eslint will do prop-types check.
If user did not give us, eslint will ignore the prop-types check since we cannot get the return value.

### DIFF
--- a/tests/lib/rules/no-unused-prop-types.js
+++ b/tests/lib/rules/no-unused-prop-types.js
@@ -3645,7 +3645,7 @@ ruleTester.run('no-unused-prop-types', rule, {
         type User = {
           user: string;
         }
-        
+
         type Props = User;
 
         export default (props: Props) => {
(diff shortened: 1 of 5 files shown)


tests/lib/rules/no-unused-prop-types.js
```diff
@@ -6765,7 +6848,38 @@ ruleTester.run('no-unused-prop-types', rule, {
       }, {
         message: '\'username\' PropType is defined but prop is never used'
       }]
-    }])
+    }]),
```
ljharb, line 6851: the below TS tests need to be inside this call. I'll fix this when i rebase.

ljharb (approved):
Thanks!

tests/lib/rules/no-unused-prop-types.js
hank121314: OK! Thanks for your help.

Issue #2804, opened 2020-09-23
TypeError: Cannot read property 'type' of undefined in prop-types


Nick-bcm:
**error:**
```
TypeError: Cannot read property 'type' of undefined
Occurred while linting /InputField.js:1
    at /node_modules/eslint-plugin-react/lib/rules/prop-types.js:153:39
    at Array.forEach (<anonymous>)
    at /eslint-plugin-react/lib/rules/prop-types.js:152:26
    at Array.forEach (<anonymous>)
    at isDeclaredInDestructuredParam (/node_modules/eslint-plugin-react/lib/rules/prop-types.js:147:14)
    at isDeclaredInComponent (/node_modules/eslint-plugin-react/lib/rules/prop-types.js:186:18)
    at /node_modules/eslint-plugin-react/lib/rules/prop-types.js:201:13
    at Array.filter (<anonymous>)
    at reportUndeclaredPropTypes (/node_modules/eslint-plugin-react/lib/rules/prop-types.js:198:51)
    at /node_modules/eslint-plugin-react/lib/rules/prop-types.js:223:11
```

**InputField.js**
```
import React from 'react'

const InputField = ({ input, meta: { touched, error }, cssClasses = {}, ...restProps }) => {
  restProps.className = cssClasses.base

  if (cssClasses.custom) {
    restProps.className += ` ${cssClasses.custom}`
  }
  if (touched && error) {
    restProps.className += ` ${cssClasses.error}`
  }

  return(
    <input
      {...input}
      {...restProps}
    />
  )
}

export default InputField
```

**.eslintrc.js**
```
module.exports = {
  env: {
    es6: true,
    node: true,
  },
  extends: ['airbnb'],
  parser: 'babel-eslint',
  parserOptions: {
    ecmaFeatures: {
      jsx: true,
    },
    ecmaVersion: 12,
    sourceType: 'module',
  },
  plugins: ['react', 'import', 'jsx-a11y'],
  overrides: [
    {
      files: ['**/*.spec.js', '**/*.spec.jsx'],
      env: {
        jest: true,
      },
    },
  ],
  rules: {
    semi: ['error', 'never', { beforeStatementContinuationChars: 'always' }],
    'react/jsx-filename-extension': [1, { extensions: ['.js', '.jsx'] }],
    // "react/prop-types": 0, // when uncomment - no error.
  },
}
```

**versions:**
```
    "eslint": "^7.9.0",
[...]

PR #2805, opened 2020-09-24
[Fix] `prop-types`: handle RestElement in destructured param

hank121314:
## Summary

This pr fixes: #2804 .

Function parameter will have `RestElement` which do not have property value.
It will cause undefined property error.

PR #2807, opened 2020-09-27
[Fix]: fix `prop-types` Cannot read property 'type' of undefined error when destructured param

minwe:
## Summary

This pr fixes: #2804, #2805.

Function parameter will have `ExperimentalRestProperty`(JSX) and which do not have property value.
It will cause undefined property error.


ljharb (approved):
Thanks!

adnasa:
it is possible that this fixes https://github.com/yannickcr/eslint-plugin-react/issues/2809
as well. I just migrated to `7.21.3` and warning stopped emitting.

PR #2808, opened 2020-09-28
[Fix] `jsx-indent-props` : Apply indentation when operator is used in front of the upper line

### LINKED ISSUE #647: jsx-indent-props fails on lines following preceding operator
Example, with rule setting:

```
  react/jsx-indent-props:
    - error
    - 2
```

and source file:

```
import React from 'react'

const SpanElement = () => {
  const span = true
    ? <span
        attr='value'
      />
    : <span
        attr='otherValue'
      />

  return span
}

export default SpanElement
```

eslint gives output:

```
./SpanElement.jsx
  6:9   error  Expected indentation of 6 space characters but found 8  react/jsx-indent-props
  9:9   error  Expected indentation of 6 space characters but found 8  react/jsx-indent-props
```

I would expect it to understand that the indent is increased by the preceding operator on the line above.

Moong0122:
To apply indentation if there is an operator in front of the line above,
modified `lib/rules/jsx-indent-props`, and added test case in `tests/lib/rules/jsx-indent-props`.

PR fixes #647

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,10 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ### Fixed
 * [`no-unused-prop-types`]: Silence false positive on `never` type in TS ([#2815][] @pcorpet)
+* [`jsx-indent-props`]: Apply indentation when operator is used in front of the upper line ([#2808][] @Moong0122)
 
 [#2815]: https://github.com/yannickcr/eslint-plugin-react/pull/2815
+[#2808]: https://github.com/yannickcr/eslint-plugin-react/pull/2808
 
 ## [7.21.3] - 2020.10.02
 
--- a/lib/rules/jsx-indent-props.js
+++ b/lib/rules/jsx-indent-props.js
@@ -62,6 +62,10 @@ module.exports = {
     let indentType = 'space';
     /** @type {number|'first'} */
     let indentSize = 4;
+    const line = {
+      isUsingOperator: false,
+      currentOperator: false
+    };
 
     if (context.options.length) {
       if (context.options[0] === 'first') {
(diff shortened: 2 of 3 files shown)


ljharb:
@Moong0122 thanks! would you mind checking the "allow edits" box on the RHS of the PR?

Moong0122:
@ljharb I checked it, and was it correct?

ljharb:
@Moong0122 i don't see it being checked, no - it would tell me i can push commits to the branch if so.

ljharb (approved):

xfournet:
There is a regression with this update, as show in this example. At line "key=3" the linter is now expecting an indent of 6 instead of 4. If the first element of the array is commented (the one with the ternary expression) the expected indentation is correctly expected to 4 again.

````
import React from 'react';

const x = false;

const Test = () => ([
  (x
    ? <div key="1" />
    : <div key="2" />),
  <div
    key="3"
  />,
]);

export default Test;

````

Moong0122:
@xfournet 
First of all, I would like to say thank you for discovering the problem!
The cases I added didn't seem to solve the case you gave me.

@ljharb 
I want to add a case, modify the code, and deliver the PR again. Can I create a new issue?

ljharb:
@Moong0122 sure - no need for an issue tho, if you just want to open a PR directly?

Moong0122:
@ljharb 
In my opinion, some people may have similar errors, so I think it would be better to let them know that the problem is being solved through the issue :)

@xfournet 
I'm trying to revise the code again by creating a new issue, can I use the example you told me?

ljharb:
@Moong0122 hopefully we could solve and release it quickly enough that that wouldn't be a problem :-)

Moong0122:
@ljharb Thanks!! :-)

xfournet:
@Moong0122 sure you can freely use/adapt the example as you wish

bukowskiadam:
We're having this problem now. Using 7.21.3 causes no issues, but the latest one causes us a lot of errors. 👀 Thanks for fixing it in advance!

danrot:
Just a side question I discovered because of this BC break: I realized that the `indent` rule that comes with eslint itself also formatted my JSX code as I had it in mind. Does it make sense to keep using that rule then? :thinking:

ljharb:
@danrot yes; eslint's indent rule both doesn't format jsx in all the same ways our rule does. I suggest configuring `indent` to ignore JSX, as the [airbnb config](https://github.com/airbnb/javascript/blob/a24dc34a4a2748c99006a48e997aa0a06b1d4d94/packages/eslint-config-airbnb-base/rules/style.js#L150-L151) does.

PR #2812, opened 2020-10-01
improve es6 class detection

vankop:
Thanks for awesome eslint plugin!

Our team is using this plugin with Vue in tsx, but unfortunatelly not all rules that could be useful both for React and Vue work with Vue, because of ES6 class detection problem. Technically this could be useful for React as well.

[Example](https://github.com/wonderful-panda/vue-tsx-support#2-add-_tsx-field-to-tell-type-information-to-typescript) how component looks like.

Solution that I see is provide an option for superClass RegExp.

### DIFF
--- a/lib/util/pragma.js
+++ b/lib/util/pragma.js
@@ -53,7 +53,16 @@ function getFromContext(context) {
   return pragma;
 }
 
+function getComponentRegExpFromContext(context, pragma) {
+  if (context.settings.react && context.settings.react.component) {
+    return new RegExp(context.settings.react.component);
+  }
+
+  return new RegExp(`^(${pragma}\\.)?(Pure)?Component$`);
+}
+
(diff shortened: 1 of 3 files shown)


lib/util/pragma.js
```diff
@@ -53,7 +53,16 @@ function getFromContext(context) {
   return pragma;
 }
 
+function getComponentRegExpFromContext(context, pragma) {
+  if (context.settings.react && context.settings.react.component) {
+    return new RegExp(context.settings.react.component);
+  }
```
ljharb, line 59: i don't think this is a good solution. the appropriate way to pretend a class inherits from React.Component is https://github.com/yannickcr/eslint-plugin-react/issues/1220#issuecomment-303617161

ljharb (changes requested):

PR #2813, opened 2020-10-01
[Fix] `no-array-index-key`: catch toString and String() usage

RedTn:
Closes https://github.com/yannickcr/eslint-plugin-react/issues/2140


ljharb (commented):
This looks good, but it seems like *any* reference to the index should be warned on, not just `.toString` or `String()`. Can we have it do that?

RedTn:
> This looks good, but it seems like _any_ reference to the index should be warned on, not just `.toString` or `String()`. Can we have it do that?

This seems a bit tricky to write if I'm trying to explicitly hard code cases to look for. Is there an existing util in the codebase to do this, or something in ESLint? Do we take some sort of recursive algo approach in checking all the nodes arguments and callee.objects etc.?

ljharb:
I don't have a specific suggestion. I'd say that any identifier inside a jsx attribute that matches the index should be flagged?

ljharb (approved):
Let's go ahead and land this, and hopefully there will be further improvements in the future.

RedTn:
Sorry forgot about this. If we open another issue I can take a look again once I have some time. But happy to let others take it if they want to

PR #2814, opened 2020-10-01
[Refactor] cleanup no-mixed-operators warnings

RedTn:
Explicitly wrapping brackets around operators with higher precedence
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence


ljharb (approved):
Thanks!

PR #2826, opened 2020-10-13
[Fix] `jsx-indent-props` : Apply indentation when using operator

Moong0122:
Since there were cases where indentation was not properly applied after https://github.com/yannickcr/eslint-plugin-react/pull/2808, 
modified `lib/rules/jsx-indent-props`, and added test case in `tests/lib/rules/jsx-indent-props`.

### DIFF
--- a/lib/rules/jsx-indent-props.js
+++ b/lib/rules/jsx-indent-props.js
@@ -124,10 +124,14 @@ module.exports = {
 
       const indent = regExp.exec(src);
       const useOperator = /^([ ]|[\t])*[:]/.test(src) || /^([ ]|[\t])*[?]/.test(src);
+      const useBracket = /^([ ]|[\t])*[<]/.test(src);
+
       line.currentOperator = false;
       if (useOperator) {
         line.isUsingOperator = true;
         line.currentOperator = true;
+      } else if (useBracket) {
+        line.isUsingOperator = false;
       }
 
       return indent ? indent[0].length : 0;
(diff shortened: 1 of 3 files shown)


lib/rules/jsx-indent-props.js
```diff
@@ -124,10 +124,14 @@ module.exports = {
 
       const indent = regExp.exec(src);
       const useOperator = /^([ ]|[\t])*[:]/.test(src) || /^([ ]|[\t])*[?]/.test(src);
+      const useBracket = /^([ ]|[\t])*[<]/.test(src);
```
ljharb, line 127: it occurs to me that perhaps we could use eslint AST utilities rather than regexes to detect these things. have you explored that at all?

ljharb (approved):

lib/rules/jsx-indent-props.js
Moong0122: First, thanks for the quick feedback!
I haven't explored so far, can I ask where I can get more information about it?

ljharb: There's https://github.com/jsx-eslint/jsx-ast-utils, and https://eslint.org/docs/developer-guide/working-with-rules that might help.

I'll land this as-is for now, and we can improve it in a followup.

Moong0122: @ljharb 
Checking if there are other test cases with indentation error,
I will follow up with reference to what you have informed me.

ljharb: Sounds good!

PR #2835, opened 2020-10-19
[New]: `jsx-key`: added `checkKeyMustBeforeSpread` option for new jsx transform

### LINKED ISSUE #2830: Enhancement: added option checkKeyMustBeforeSpread of `react/jsx-key`
`jsx-runtime` landed. but there have different transform results when key before or after props spread

https://github.com/facebook/react/issues/20031

```jsx
<div key="key "{...props} />;

<div {...props} key="key"/>;

// will convert to 

import { createElement } from "react"
import { jsx } from "react/jsx-runtime"

jsx("div", props, "key")
createElement("div", { ...props, key: "key" }) // only key after props spread do the fallback 
```

to avoid the fallback to using `createElement`, 
need a rule to limit the key before spread.

morlay:
fix #2830 

related: https://github.com/facebook/react/issues/20031#issuecomment-710346866

### DIFF
--- a/lib/rules/jsx-key.js
+++ b/lib/rules/jsx-key.js
@@ -6,6 +6,7 @@
 'use strict';
 
 const hasProp = require('jsx-ast-utils/hasProp');
+const propName = require('jsx-ast-utils/propName');
 const docsUrl = require('../util/docsUrl');
 const pragmaUtil = require('../util/pragma');
 
@@ -14,7 +15,8 @@ const pragmaUtil = require('../util/pragma');
 // ------------------------------------------------------------------------------
 
 const defaultOptions = {
-  checkFragmentShorthand: false
+  checkFragmentShorthand: false,
+  checkKeyMustBeforeSpread: false
 };
 
 module.exports = {
@@ -31,6 +33,10 @@ module.exports = {
         checkFragmentShorthand: {
           type: 'boolean',
           default: defaultOptions.checkFragmentShorthand
+        },
+        checkKeyMustBeforeSpread: {
+          type: 'boolean',
+          default: defaultOptions.checkKeyMustBeforeSpread
         }
       },
       additionalProperties: false
@@ -40,6 +46,7 @@ module.exports = {
   create(context) {
     const options = Object.assign({}, defaultOptions, context.options[0]);
     const checkFragmentShorthand = options.checkFragmentShorthand;
+    const checkKeyMustBeforeSpread = options.checkKeyMustBeforeSpread;
     const reactPragma = pragmaUtil.getFromContext(context);
     const fragmentPragma = pragmaUtil.getFragmentFromContext(context);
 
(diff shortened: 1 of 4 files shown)


morlay:
@ljharb should be `default: true` ?

lib/rules/jsx-key.js
```diff
@@ -61,9 +68,34 @@ module.exports = {
...
+        }
+
+        if (attribute.type === 'JSXAttribute') {
+          if (propName(attribute) === 'key') {
+            keyAttributeIndex = i;
+          }
+        }
+      });
+
+      return spreadAttributeIndex !== -1 && keyAttributeIndex > spreadAttributeIndex;
```
ljharb, line 87:
```suggestion
      var hasFoundSpread = false;
      return attributes.some((attribute) => {
        if (attribute.type === 'JSXSpreadAttribute') {
          hasFoundSpread = true;
          return false;
        }
        if (attribute.type !== 'JSXAttribute') {
          return false;
        }
        return !hasFoundSpread || propName(attribute) === 'key';
      });
```
what about this alternative?

lib/rules/jsx-key.js
```diff
@@ -61,9 +68,34 @@ module.exports = {
...
+    }
+
     return {
       JSXElement(node) {
         if (hasProp(node.openingElement.attributes, 'key')) {
+          if (checkKeyMustBeforeSpread && isKeyAfterSpread(node.openingElement.attributes)) {
+            context.report({
+              node,
+              message: '"key" prop must before "...spread"'
```
ljharb, line 96:
```suggestion
              message: '`key` prop must before any `{...spread}, to avoid conflicting with React’s new JSX transform: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html`'
```

tests/lib/rules/jsx-key.js
```diff
@@ -48,7 +48,9 @@ ruleTester.run('jsx-key', rule, {
     {code: '[1, 2, 3].map(function(x) { return; });'},
     {code: 'foo(() => <div />);'},
     {code: 'foo(() => <></>);', parser: parsers.BABEL_ESLINT},
-    {code: '<></>;', parser: parsers.BABEL_ESLINT}
+    {code: '<></>;', parser: parsers.BABEL_ESLINT},
+    {code: '<App {...{}} />;', parser: parsers.BABEL_ESLINT},
+    {code: '<App key="keyBeforeSpread" {...{}} />;', parser: parsers.BABEL_ESLINT}
```
ljharb, line 53: let's include duplicate examples for an HTML element as well

tests/lib/rules/jsx-key.js
```diff
@@ -88,5 +90,11 @@ ruleTester.run('jsx-key', rule, {
     options: [{checkFragmentShorthand: true}],
     settings,
     errors: [{message: 'Missing "key" prop for element in array. Shorthand fragment syntax does not support providing keys. Use Act.Frag instead'}]
+  }, {
+    code: '[<App {...obj} key="keyAfterSpread" />];',
+    parser: parsers.BABEL_ESLINT,
+    options: [{checkKeyMustBeforeSpread: true}],
+    settings,
+    errors: [{message: '"key" prop must before "...spread"'}]
```
ljharb, line 98: let's include duplicate examples for an HTML element as well. also, duplicate examples for TypeScript, and for eslint's default parser.

ljharb (changes requested):
Thanks! The default can't be true, or that'd be a breaking change.

lib/rules/jsx-key.js
morlay: `hasFoundSpread && propName(attribute) === 'key'` may be clear

lib/rules/jsx-key.js
morlay: updated.

tests/lib/rules/jsx-key.js
morlay: sure

tests/lib/rules/jsx-key.js
morlay: updates.

lib/rules/jsx-key.js
ljharb: Sure, in that case you'd need to negate the overarching `.some`.

ljharb (approved):
Thanks, looks great! If there's further tweaking needed for the TS tests in older eslint/node, I'll handle that.

PR #2782, opened 2020-09-02
[New] `jsx-no-literals`: add `noAttributeStrings` option

TaLeaMonet:
Closes #2496. Fixes #2486.

PR #2836, opened 2020-10-19
Remove invalid prop

sirugh:
As far as I can tell, the `noAttributeStrings` property does not exist. Including it in my configuration caused my eslint to throw an error: 

> 	Configuration for rule "react/jsx-no-literals" is invalid:
	Value {"noStrings":true,"allowedStrings":[],"ignoreProps":false,"noAttributeStrings":true} should NOT have additional properties.


ljharb (changes requested):
It absolutely exists - see https://github.com/yannickcr/eslint-plugin-react/blob/153eac8d894a22ac2058dd4fbc9b491032176834/lib/rules/jsx-no-literals.js#L44-L46.

It was added in #2782, and released in v7.21.0. What version of eslint-plugin-react are you using?

sirugh:
Oops! Github has latest release at 7.20.0. I made a bad assumption. We can close this :)

ljharb:
Always check npm, not github :-)

PR #2837, opened 2020-10-20
Use correct type for children

FredericEspiau:
The correct type is `ReactNode` and not `React.Node`

Also added an example

### DIFF
--- a/docs/rules/prop-types.md
+++ b/docs/rules/prop-types.md
@@ -160,7 +160,7 @@ As it aptly noticed in
 to document `children` in the prop types.
 
 Generally, you should use `PropTypes.node` or static type `ReactNode` for
-`children`, like this:
+`children`, like this in TypeScript:
 
 ```tsx
 import React, { ReactNode } from "react";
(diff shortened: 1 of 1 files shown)


docs/rules/prop-types.md
```diff
@@ -159,8 +159,22 @@ As it aptly noticed in
 > Most components don't need `this.props.children`, so that makes it extra important
 to document `children` in the prop types.
 
-Generally, you should use `PropTypes.node` or static type `React.Node` for
-`children`. It accepts anything that can be rendered: numbers, strings, elements
```
ljharb, line 163: https://flow.org/en/docs/react/types/#toc-react-node is for Flow. `ReactNode` is for TypeScript.

We shouldn't remove the Flow instructions; let's keep both.

ljharb (changes requested):

docs/rules/prop-types.md
ljharb: This still needs the Flow instructions added.

ljharb:
@FredericEspiau are you no longer interested in correcting the instructions to account for both Flow and TS?

PR #2843, opened 2020-10-25
[Fix] `prop-types`: function in class that returns a component causes false warning in typescript

### LINKED ISSUE #2784: Function that returns a component causes false positive unused react/prop-types warning in typescript
The folllowing code causes a false positive for the react/proptypes, with 'className' missing in prop-types declaration. Since this is typescript, and ComponentProps includes className, I'd expect this to not happen.
```typescript
import * as React from "react";

type ComponentProps = {
  className?: string;
};

class Test {
  getAComponent(): React.FC<ComponentProps> {
    return function ComponentName(props: ComponentProps): React.ReactElement {
      return <div className={props.className}>Hello</div>;
    };
  }
} 

const test = new Test();
const Comp = test.getAComponent();

export default function App(): React.ReactElement {
  return (
    <div className="App">
      <Comp className="text-success" />
    </div>
  );
}
```

If the component is instead defined outside of the method that returns it, then the warning dissapears, like this (the rest of the example remains the same):

```typescript
const getAComponentNotInClass = function ComponentName(props: ComponentProps): React.ReactElement {
  return <div className={props.className}>Hello</div>;
};

class Test {
  getAComponent(): React.FC<ComponentProps> {
    return getAComponentNotInClass
  }
} 
```

SyMind:
Fixes #2784.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,9 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ### Fixed
 * [`display-name`]/component detection: avoid a crash on anonymous components ([#2840][] @ljharb)
+* [`prop-types`]: function in class that returns a component causes false warning in typescript ([#2843][] @SyMind)
 
+[#2843]: https://github.com/yannickcr/eslint-plugin-react/pull/2843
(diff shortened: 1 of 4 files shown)


ljharb (approved):
Thanks, seems reasonable

PR #2844, opened 2020-10-25
[Fix] `no-multi-comp`: improve component detection

### LINKED ISSUE #2842: Ignore nested components in no-multi-comp
I don’t think `react/no-multi-comp` should trigger on nested components or should have an option to disable it. A practical example:

```jsx
import { FormattedMessage } from 'react-intl';

export function App() {
  return (
    <FormattedMessage
      values={{
        link:
          // The next line triggers react/no-multi-comp
          (text: string) => <a>{text}</a>,
      }}
    />
  );
}
```

I suggest to add an option `ignoreNested` which ignores any components that aren’t declared as part of the `Program` node body.

rrhg:
Fixes #2842. 

As explained by @ljharb, the issue there isn’t that it’s nested - it’s that it’s treated as a component despite being used in an object literal with a key name that can’t be a custom component name.

Our component detection should be improved here.

### DIFF
--- a/tests/lib/rules/destructuring-assignment.js
+++ b/tests/lib/rules/destructuring-assignment.js
@@ -265,7 +265,7 @@ ruleTester.run('destructuring-assignment', rule, {
     errors: [{message: 'Must use destructuring props assignment'}]
   }, {
     code: `
-      function hof() {
+      function HOF() {
         return (props) => <p>{props.a}</p>;
       }
     `,
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -708,6 +708,14 @@ function componentRule(rule, context) {
         }
         if (utils.isInAllowedPositionForComponent(node) && utils.isReturningJSXOrNull(node)) {
           if (!node.id || isFirstLetterCapitalized(node.id.name)) {
+            if (
+              node.body
+              && node.body.openingElement
+              && node.body.openingElement.name
+              && !isFirstLetterCapitalized(node.body.openingElement.name.name)
+            ) {
+              return undefined;
+            }
             return node;
           }
           return undefined;
--- a/tests/lib/rules/no-multi-comp.js
+++ b/tests/lib/rules/no-multi-comp.js
@@ -31,6 +31,80 @@ ruleTester.run('no-multi-comp', rule, {
 
   valid: [{
     code: [
+      'import { FormattedMessage } from \'react-intl\';',
+      'export function App() {',
+      '  return (',
+      '    <FormattedMessage',
+      '      values={{',
+      '        link:',
+      '        (text) => <a>{text}</a>,',
+      '      }}',
+      '    />',
+      '  );',
+      '}'
+    ].join('\r')
+  }, {
+    code: [
+      'import { FormattedMessage } from \'react-intl\';',
+      'const App = () => {',
+      '  return (',
+      '    <FormattedMessage',
+      '      values={{',
+      '        link:',
+      '        (text) => <a>{text}</a>,',
+      '      }}',
+      '    />',
+      '  );',
+      '}'
+    ].join('\r')
+  }, {
+    code: [
+      'import { FormattedMessage } from \'react-intl\';',
+      'const App = function() {',
+      '  return (',
+      '    <FormattedMessage',
+      '      values={{',
+      '        link:',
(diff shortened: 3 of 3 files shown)


tests/lib/rules/destructuring-assignment.js
```diff
@@ -265,8 +265,8 @@ ruleTester.run('destructuring-assignment', rule, {
     errors: [{message: 'Must use destructuring props assignment'}]
   }, {
     code: `
-      function hof() {
-        return (props) => <p>{props.a}</p>;
```
ljharb, line 269: let's find a way to keep this test case as-is, perhaps just by renaming `hof` to `HOF`?

ljharb (changes requested):
Thanks! Let's add a bunch more test cases, and try to avoid changing existing ones.

tests/lib/rules/destructuring-assignment.js
rrhg: Thanks :+1:  Could you help figuring out what's wrong here ?
I reversed the changes as suggested but now is failing. It's suppose to throw an error message but it's not throwing.
Could be wrong, but I think that this unrelated test was failing for the wrong reason & my PR fix, just unmasked the bug.
**This appears to be a regular function that returns an anonymous React component.** I have research a lot and can't find an example of how that is used.
That's why previously thought it was a typo or bug.

rrhg: Wait ! You said something relevant in describing the original issue(#2842):
> The issue there isn’t that it’s nested - it’s that it’s treated as a component despite being used in an object literal with a key name that can’t be a custom component name.   
   
So if we take a look at the original issue(#2842) relevant code   
```
export function App() {
  return (
    <FormattedMessage
      values={{
        link:
          // The next line triggers react/no-multi-comp
          (text: string) => <a>{text}</a>,
      }}    /> ); }
```   

We see that is very similar to the [unrelated test](https://github.com/yannickcr/eslint-plugin-react/blob/edbbd7931902ec0b1cc1941df4155e39acc06f31/tests/lib/rules/destructuring-assignment.js#L267) that now is failing.   

So the questions is: Should this also be ignored ?:
```
      function hof() {
        return (props) => <p>{props.a}</p>;   
```
IF, they are similar & both should be ignored, then I should delete this code from [/tests/lib/rules/destructuring-assignment.js](https://github.com/yannickcr/eslint-plugin-react/blob/edbbd7931902ec0b1cc1941df4155e39acc06f31/tests/lib/rules/destructuring-assignment.js#L267) & add it to my valid code tests

ljharb: I think that `function hof() {` here should be ignored - however, if it's in a test case, it's likely an author error, and *should* be named "HOF" so it's a proper component that is *not* ignored.

ljharb (approved):

ljharb:
@rrhg i've rebased this; if tests pass i'll land it, but if not then please take a look :-)

tests/lib/rules/destructuring-assignment.js
rrhg: The only way I could find to keep this test was like this:
```
     function HOF(props) {
       return <p>{props.a}</p>;
     }
```
The reason, is that the way it was done before (returning an anonymous function), it was been ignored & the destructuring was not been checked. Remember that the purpose of this was testing invalid (not-destructuring) code but only on React components, not anonymous functions.

BTW, if you accept this, it needs to be rebased again. Since you rebased it, (& due to my inexperience), now it have a rebase conflict that I'm not able to resolve, sorry.

ljharb:
@rrhg i've rebased this again, and fixed the HOF test case. It's supposed to remain as-is - a higher-order function, that takes props and returns a component - and the linter should detect its return value as one (but not HOF itself)

ljharb:
@rrhg i don't understand, why did you close the PR and delete the branch?

rrhg:
Oh, thanks :pray: . Sorry about that, didn't want to waist your time. The thing is, my PR was about no-multi-comp, and it took me a couple of days to figure it out(honestly I was lucky). But apparently it uncovered this problem with destructuring which I been unable figure out :worried:

ljharb:
No problem, let's leave this open so if you do figure it out, or if someone else comes along who can, we can update the PR to fix it :-)

PR #2846, opened 2020-10-28
[New] `jsx-indent-props`: add `ignoreTernaryOperator` option

SebastianZimmer:


ljharb (changes requested):
Let's pick an option name where absence and `false` are the same.

How about, `ignoreTernaryOperator` that defaults to false?

SebastianZimmer:
> Let's pick an option name where absence and false are the same.
> 
> How about, ignoreTernaryOperator that defaults to false?

Good idea, I've changed it that way and adjusted documentation and tests.

ljharb (approved):

PR #2848, opened 2020-10-30
[New] add `no-object-type-as-default-prop` rule

cyan33:
Discussed in 

https://github.com/yannickcr/eslint-plugin-react/issues/2847

## Test

```
npm run lint
npm run test
```

shows no errors and warnings

---

- [x] Added it in README

### DIFF
--- a/docs/rules/no-object-type-as-default-prop.md
+++ b/docs/rules/no-object-type-as-default-prop.md
@@ -1,0 +1,70 @@
+# Disallow usage of referential-type variables as default param in functional component (react/no-object-type-as-default-prop)
+
+💼 This rule is enabled in the following [configs](https://github.com/jsx-eslint/eslint-plugin-react#shareable-configurations): `all`.
+
+Warns if in a functional component, an object type value (such as array/object literal/function/etc) is used as default prop, to prevent potential unnecessary rerenders, and performance regressions.
+
+## Rule Details
+
+Certain values (like arrays, objects, functions, etc) are compared by identity instead of by value. This means that, for example, whilst two empty arrays conceptually represent the same value - JavaScript semantics dictate that they are distinct and unequal as they represent two distinct values.
+
+When using object destructuring syntax you can set the default value for a given property if it does not exist. If you set the default value to one of the values that is compared by identity, it will mean that each time the destructure is evaluated the JS engine will create a new, distinct value in the destructured variable.
+
+In the context of a React functional component's props argument this means for each render, the property has a new, distinct value. When this value is passed to a hook as a dependency or passed into a child component as a property React will see this as a new value - meaning that a hook will be re-evaluated, or a memoized component will rerender.
+
+This obviously destroys any performance benefits you get from memoization. Additionally, in certain circumstances this can cause infinite rerender loops, which can often be hard to debug.
+
+It's worth noting that primitive literal values (`string`, `number`, `boolean`, `null`, and `undefined`) can be considered to be compared "by value", or alternatively, as always having the same identity (every `3` is the same exact `3`). Thus, it's safe for those to be inlined as a default value.
+
+To fix the violations, the easiest way is to use a referencing variable in module scope instead of using the literal values, e.g:
+
+```jsx
+const emptyArray = [];
+
+function Component({
+  items = emptyArray,
+}) {}
+```
+
+Examples of ***invalid*** code for this rule:
+
+```jsx
+function Component({
+  items = [],
+}) {}
+
+const Component = ({
+  items = {},
+}) => {}
+
+const Component = ({
+  items = () => {},
+}) => {}
+```
+
+Examples of ***valid*** code for this rule:
+
+```jsx
+const emptyArray = [];
+
+function Component({
+  items = emptyArray,
+}) {}
+
+const emptyObject = {};
+const Component = ({
+  items = emptyObject,
+}) => {}
+
+const noopFunc = () => {};
+const Component = ({
+  items = noopFunc,
+}) => {}
+
+// primitives are all compared by value, so are safe to be inlined
+function Component({
+  num = 3,
+  str = 'foo',
+  bool = true,
+}) {}
+```
--- a/lib/rules/no-object-type-as-default-prop.js
+++ b/lib/rules/no-object-type-as-default-prop.js
@@ -1,0 +1,108 @@
+/**
+ * @fileoverview Prevent usage of referential-type variables as default param in functional component
+ * @author Chang Yan
+ */
+
+'use strict';
+
+const values = require('object.values');
+
+const Components = require('../util/Components');
+const docsUrl = require('../util/docsUrl');
+const report = require('../util/report');
+
+const FORBIDDEN_TYPES_MAP = {
+  ArrowFunctionExpression: 'arrow function',
+  FunctionExpression: 'function expression',
+  ObjectExpression: 'object literal',
+  ArrayExpression: 'array literal',
+  ClassExpression: 'class expression',
+  NewExpression: 'construction expression',
+  JSXElement: 'JSX element',
+};
+
+const FORBIDDEN_TYPES = new Set(Object.keys(FORBIDDEN_TYPES_MAP));
+const MESSAGE_ID = 'forbiddenTypeDefaultParam';
+
+const messages = {
+  [MESSAGE_ID]: '{{propName}} has a/an {{forbiddenType}} as default prop. This could lead to potential infinite render loop in React. Use a variable reference instead of {{forbiddenType}}.',
+};
+function hasUsedObjectDestructuringSyntax(params) {
+  return (
+    params != null
+    && params.length === 1
+    && params[0].type === 'ObjectPattern'
+  );
+}
+
+function verifyDefaultPropsDestructuring(context, properties) {
+  // Loop through each of the default params
+  properties.filter((prop) => prop.type === 'Property').forEach((prop) => {
+    const propName = prop.key.name;
+    const propDefaultValue = prop.value;
+
+    if (propDefaultValue.type !== 'AssignmentPattern') {
+      return;
+    }
+
+    const propDefaultValueType = propDefaultValue.right.type;
+
+    if (
+      propDefaultValueType === 'Literal'
+      && propDefaultValue.right.regex != null
+    ) {
+      report(context, messages[MESSAGE_ID], MESSAGE_ID, {
+        node: propDefaultValue,
+        data: {
+          propName,
+          forbiddenType: 'regex literal',
+        },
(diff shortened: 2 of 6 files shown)


cyan33:
cc @ljharb

docs/rules/no-reference-type-as-default-prop.md
```diff
@@ -0,0 +1,68 @@
+# Prevent usage of referential-type variables as default param in functional component (react/no-reference-type-as-default-prop)
+
+Warns if in a functional component, a referential type value (such as array/object literal) is used as default prop, to prevent potential un-necessary re-renders, and performance regressions.
```
ljharb, line 3: i'm not sure if the name "referential type" is helpful - the hazard is with any object, a category which arrays and functions belong to, so `no-object-as-default-prop` might be clearer? (especially since any implication that JS is a pass-by-reference language would be incorrect and confusing)

docs/rules/no-reference-type-as-default-prop.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Rule Details
+
+Certain values (like arrays, objects, functions, etc) are compared by reference instead of by value. This means that, for example, whilst two empty arrays represent the same value - to the JS engine they are distinct and inequal as they represent references to different things in memory.
```
ljharb, line 7: I'd prefer to describe this as comparing by identity - and note that this also includes Symbols. "reference" is an overloaded term.

docs/rules/no-reference-type-as-default-prop.md
```diff
@@ -0,0 +1,68 @@
...
+
+When using object destructuring syntax you can set the default value for a given property if it does not exist. If you set the default value to one of the values that is compared by reference, it will mean that each time the destructure is executed the JS engine will create a new, non-equal value in the destructured variable.
+
+In the  context of a React functional component's props argument this means that each render the property has a new, distinct value. When this value is passed to a hook as a dependency or passed into a child component as a property react will see this as a new value - meaning that a hook will be re-evaluated, or a memoized component will rerender.
+
+This obviously destroys any performance benefits you get from memoisation. Additionally in certain circumstances this can cause infinite rerender loops, which can often be hard to debug.
```
ljharb, line 13:
```suggestion
This obviously destroys any performance benefits you get from memoization. Additionally in certain circumstances this can cause infinite rerender loops, which can often be hard to debug.
```

tests/lib/rules/no-reference-type-as-default-prop.js
```diff
@@ -0,0 +1,212 @@
...
+    }
+  `,
+    `
+    const Foo = () => {};
+  `,
+    `
+    const Foo = ({bar = 1}) => {};
+  `
+  ],
+  invalid: [
```
ljharb, line 66: let's make sure each of these test cases is duplicated for the default parser, babel-eslint, and the typescript parsers.

ljharb (changes requested):
LGTM minus the naming/prose concerns, the question about new Symbols, and the test parser comment

AriPerkkio:
This new rule looked useful so I quickly tested it against +150 repositories. It seems to be [crashing quite often](https://github.com/AriPerkkio/eslint-plugin-react/blob/test/no-reference-type-as-default-prop/eslint-remote-tester-results/baseweb.md) on `isReactComponentName` method. This test should cover at least one of those cases
```js
export default function() {
    return <div />;
}
```

I'm not sure if all crashes were caused by the same issue. You can find all results from [here](https://github.com/AriPerkkio/eslint-plugin-react/tree/test/no-reference-type-as-default-prop/eslint-remote-tester-results) and log containing all crashing cases from [here](https://github.com/AriPerkkio/eslint-plugin-react/blob/test/no-reference-type-as-default-prop/lint-log).

ljharb:
@cyan33 ping :-) there's a few things to address

docs/rules/no-reference-type-as-default-prop.md
cyan33: Sounds good. If we see `object` as a broad term, the indication does seem clearer with the proposed name `no-object-as-default-prop`.

cyan33:
Sorry it took me a bit to get back to this 🙂

Updated:

- Tweaking the name and docs of the rule
- Adding support to check for `Symbol('foo')`
- Add test case for anonymous function
  - Simply skip as it's hard to guarantee if it's a real React component just based on its content. I think some false positive is fine in this case since it's not encouraged to not provide a name to the component?

cyan33:
Ping @ljharb

docs/rules/no-object-type-as-default-prop.md
```diff
@@ -0,0 +1,68 @@
...
+
+When using object destructuring syntax you can set the default value for a given property if it does not exist. If you set the default value to one of the values that is compared by identity, it will mean that each time the destructure is executed the JS engine will create a new, non-equal value in the destructured variable.
+
+In the  context of a React functional component's props argument this means that each render the property has a new, distinct value. When this value is passed to a hook as a dependency or passed into a child component as a property react will see this as a new value - meaning that a hook will be re-evaluated, or a memoized component will rerender.
+
+This obviously destroys any performance benefits you get from memoisation. Additionally in certain circumstances this can cause infinite rerender loops, which can often be hard to debug.
+
+It's worth noting that primitive literal values (`string`, `number`, `boolean`, `null`, and `undefined`) are equal when two same value are being compared, so it's safe for those to be inlined as default prop.
```
ljharb, line 15:
```suggestion
It's worth noting that primitive literal values (`string`, `number`, `boolean`, `null`, and `undefined`) can be considered to be compared "by value", or alternatively, as always having the same identity (every `3` is the same exact `3`). Thus, it's safe for those to be inlined as a default value.
```

docs/rules/no-object-type-as-default-prop.md
```diff
@@ -0,0 +1,68 @@
...
+
+To fix the violations, the easiest way is to use a referencing variable instead of using the literal values, e.g:
```
ljharb, line 17:
```suggestion
To fix the violations, the easiest way is to use a referencing variable in module scope instead of using the literal values, e.g:
```

docs/rules/no-object-type-as-default-prop.md
```diff
@@ -0,0 +1,68 @@
...
+const Component = ({
+  items = emptyObject,
+}) => {}
+
+const noopFunc = () => {};
+const Component = ({
+  items = noopFunc,
+}) => {}
+
+// primitive literals are all compared by value, so are safe to be inlined
```
ljharb, line 62:
```suggestion
// primitives are all compared by value, so are safe to be inlined
```

lib/rules/no-object-type-as-default-prop.js
```diff
@@ -0,0 +1,150 @@
...
+
+const FORBIDDEN_TYPES = new Set(Object.keys(FORBIDDEN_TYPES_MAP));
+const MESSAGE_ID = 'forbiddenTypeDefaultParam';
+
+function isReactComponentName(node) {
+  if (node.id && node.id.type === 'Identifier' && node.id.name) {
+    const firstLetter = node.id.name[0];
+    if (firstLetter.toUpperCase() === firstLetter) {
+      return true;
+    }
```
ljharb, line 28:
```suggestion
    return firstLetter.toUpperCase() === firstLetter;
```

lib/rules/no-object-type-as-default-prop.js
```diff
@@ -0,0 +1,150 @@
...
+
+function hasUsedObjectDestructuringSyntax(params) {
+  if (
+    params == null
+    || params.length !== 1
+    || params[0].type !== 'ObjectPattern'
+  ) {
+    return false;
+  }
+  return true;
```
ljharb, line 52:
```suggestion
  return
    params != null
    && params.length === 1
    && params[0].type === 'ObjectPattern';
```

lib/rules/no-object-type-as-default-prop.js
```diff
@@ -0,0 +1,150 @@
...
+}
+
+function verifyDefaultPropsDestructuring(context, properties) {
+  // Loop through each of the default params
+  properties.forEach((prop) => {
+    if (prop.type !== 'Property') {
+      return;
+    }
+
```
ljharb, line 61:
```suggestion
  properties.filter((prop) => prop.type === 'Property').forEach((prop) => {
```

lib/rules/no-object-type-as-default-prop.js
```diff
@@ -0,0 +1,150 @@
...
+      context.report({
+        node: propDefaultValue,
+        messageId: MESSAGE_ID,
+        data: {
+          propName,
+          forbiddenType: 'regex literal'
+        }
+      });
+    } else if (propDefaultValueType === 'CallExpression'
+      && propDefaultValue.right.callee.type === 'Identifier'
```
ljharb, line 84:
```suggestion
    } else if (
      propDefaultValueType === 'CallExpression'
      && propDefaultValue.right.callee.type === 'Identifier'
```

ljharb (changes requested):
Overall, I think this looks good. After this pass, if my only feedback is docs prose, I'll tweak it myself prior to landing.

fengkx:
Is there anything more to be done before this rule can be release? It looks very useful.  cc @ljharb

ljharb:
@fengkx the review comment can be addressed either by @cyan33, or, by someone commenting a link to their branch (NOT another PR) that addresses them.

ljharb (changes requested):
This rule should use the `Components.detect` utility rather than trying to implement its own component detection.

ljharb:
@fengkx thanks! this PR has been freshly rebased, and I updated the test conventions as well. The only missing piece here is to use `Components.detect`. Please ping me when your branch link is updated and I'll be able to pull it in and rereview.

fengkx:
[This branch](https://github.com/fengkx/eslint-plugin-react/tree/pr-2848-continue?rgh-link-date=2022-10-11T12%3A37%3A00Z) switched the component detection to use `Components.detect`. Because the `Component.detect` function no only check the function has a leading upper case letter but also return value( to be jsx or null), I change all test cases to return a `null`.

I can't find any doc about the `Components.detect` function. Should we mention this kind of utils in CONTRIBUTING.md ?

ljharb:
Sure, it couldn't hurt to mention it in CONTRIBUTING.md.

Thanks, the added commit looks good; i'll pull it in shortly.

ljharb (approved):

fengkx:
Could you release a new version for this? @ljharb

ljharb:
When I’m prepared to field bug reports from the unreleased changes, i will.

PR #2852, opened 2020-11-08
[Docs] Add new example to no-unused-prop-types

### LINKED ISSUE #2849: `react/no-unused-prop-types` false positive if any one component doesn't use a field
In the following example both `name` and `placeholder` are triggering the `no-unused-prop-types` rule, despite both being used in separate components. I _think_ the rule is triggering for `name` because it is unused in `Comp1` and `placeholder` because it is unused in `Comp2`.

```javascript
import React from "react";

type Props = {
  name: string;
  placeholder: string;
};

export const Comp1 = (props: Props): React.ReactElement => {
  const { placeholder } = props;
  return <div placeholder={placeholder} />
}

export const Comp2 = (props: Props): React.ReactElement => {
  const { name } = props;
  return <div>{name}</div>;
}
```

I appreciate that this example is contrived, but _I think_ is the most basic reproduction case of a more complicated manifestation of the issue. The behaviour is kind of logical but not exactly intuitive.

thehereward:
Adding a new example to resolve #2849.
Also fixing a comment in the same file that got out of sync with its code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 * [`prop-types`]: default argument does not count as props-types declaration ([#2877][] @golopot)
 * [`jsx-props-no-multi-spaces`]: fix a false positive for beside comments ([#2878][] @golopot)
 
+### Docs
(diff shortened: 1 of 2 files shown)


ljharb (approved):
Thanks!

PR #2854, opened 2020-11-10
Added a new rule to allow linting of spread prop declarations in JSX to ensure alphabetical ordering

IRoninCoder:
Based on the question at SO [react functional component props sort alphabetically at the definition side](https://stackoverflow.com/questions/64750228/react-functional-component-props-sort-alphabetically-at-the-definition-side) the idea is to have the ability to lint for spread property names in order to ensure they are sorted alphabetically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,7 @@ Enable the rules that you would like to use.
 * [react/jsx-props-no-multi-spaces](docs/rules/jsx-props-no-multi-spaces.md): Disallow multiple spaces between inline JSX props (fixable)
 * [react/jsx-props-no-spreading](docs/rules/jsx-props-no-spreading.md): Prevent JSX prop spreading
 * [react/jsx-sort-default-props](docs/rules/jsx-sort-default-props.md): Enforce default props alphabetical sorting
+* [react/jsx-sort-functional-props](docs/rules/jsx-sort-functional-props.md): Enforce props of functional components to be sorted alphabetically
 * [react/jsx-sort-props](docs/rules/jsx-sort-props.md): Enforce props alphabetical sorting (fixable)
 * [react/jsx-space-before-closing](docs/rules/jsx-space-before-closing.md): Validate spacing before closing bracket in JSX (fixable)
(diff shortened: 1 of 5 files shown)


ljharb:
I'm a bit confused. eslint already has a core rule to enforce ordering of object keys, in destructuring as well as objects. Why would you want it enforced for spreaded props but not every other object literal?

IRoninCoder:
Are you referring to [sort-keys](https://eslint.org/docs/rules/sort-keys#require-object-keys-to-be-sorted-sort-keys)? if so then it does not capture JSX prop destructures:

`index.jsx`
```jsx
import React from "react";

let obj = {a: 1, c: 3, b: 2}; //this is captured by sort-keys

const Component = ({two, one}) => { //this is not captured by sort-keys, which is what this PR attempts to solve
  return <div>Yes!</div>;
};

export default Component;
```

I also think it's a matter of choice to have it everywhere or just for the prop spreads in JSX. We may prefer it everywhere, but surely not everyone does.

ljharb:
This seems like a reasonable enhancement to the core sort-keys rule. Can you first try filing an issue on eslint itself about it?

ljharb:
Please link that issue here when you do, i'd be interested to follow it.

IRoninCoder:
https://github.com/eslint/eslint/issues/14873

p-runge:
Since `ESLint's core stylistic rules are frozen` (as mentioned in the linked issue at eslint itself), it seems to make sense to reopen this issue and possibly implement the related feature, no?

taozhou-glean:
> Since `ESLint's core stylistic rules are frozen` (as mentioned in the linked issue at eslint itself), it seems to make sense to reopen this issue and possibly implement the related feature, no?

FYI: there is a plugin just for that: https://www.npmjs.com/package/eslint-plugin-sort-destructure-keys

igorpupkinable:
Any chance this MR or [plugin mentioned](https://www.npmjs.com/package/eslint-plugin-sort-destructure-keys) can be part of core JSX-ESLint? Just one less package to install and keep track of.

ljharb:
Adding an entirely new rule to this plugin just to largely duplicate `sort-keys`, and then to only apply to react props, seems like a lot of effort and complexity for minimal value. Open to being convinced tho!

PR #2855, opened 2020-11-11
[New] `jsx-no-target-blank`: Add `warnOnSpreadAttributes` option

### LINKED ISSUE #2827: jsx-target-no-blank should warn when using JSX spread attributes
At the moment, something like this would _not_ trigger a warning:

```
<a {...otherProps} target="_blank" />
```

Since the rule can't verify whether this will compile to have an `href` that points to an external url, wouldn't it be safer to warn the user? If they are sure that this will never compile to be an external link, they can explicitly disable the rule for this line.

michael-yx-wu:
Resolves #2827

I've changed the rule to warn on the `JSXOpeningElement` instead of a specific attribute. There is a technical reason and an architectural reason:
- when disabling `trustSpreadAttributes`, we could end up with multiple warnings for the same problem e.g. once on a `target="_blank"` attribute and once on `{...myUnsafeObject}` spread attribute
- it makes more sense (to me) that the rule warns on the element instead of a specific attribute since the problem is not necessarily due to any specific attribute, but the combination and order of the attributes on the element

Commit message
---
```
Add trustSpreadAttributes for jsx-no-target-blank

Defaults to true. When disabled, treats spread attributes as dangerous
unless explicitly overriden e.g. the following is safe:

<a {...dangerousObject} rel="noreferrer" target="_blank"></a>

This change also extends target="_blank" detection to include
conditional expressions whose alternate or consequent is the "_blank"
string (case-insensitive).
```

### DIFF
--- a/docs/rules/jsx-no-target-blank.md
+++ b/docs/rules/jsx-no-target-blank.md
@@ -1,15 +1,11 @@
 # Prevent usage of unsafe `target='_blank'` (react/jsx-no-target-blank)
 
-When creating a JSX element that has an `a` tag, it is often desired to have
-the link open in a new tab using the `target='_blank'` attribute. Using this
-attribute unaccompanied by `rel='noreferrer'`, however, is a severe
-security vulnerability ([see here for more details](https://html.spec.whatwg.org/multipage/links.html#link-type-noopener))
+When creating a JSX element that has an `a` tag, it is often desired to have the link open in a new tab using the `target='_blank'` attribute. Using this attribute unaccompanied by `rel='noreferrer'`, however, is a severe security vulnerability ([see here for more details](https://html.spec.whatwg.org/multipage/links.html#link-type-noopener))
 This rules requires that you accompany `target='_blank'` attributes with `rel='noreferrer'`.
 
 ## Rule Details
 
-This rule aims to prevent user generated links from creating security vulnerabilities by requiring
-`rel='noreferrer'` for external links, and optionally any dynamically generated links.
+This rule aims to prevent user generated links from creating security vulnerabilities by requiring `rel='noreferrer'` for external links, and optionally any dynamically generated links.
 
 ## Rule Options
 ```json
@@ -20,7 +16,8 @@ This rule aims to prevent user generated links from creating security vulnerabil
 
 * allow-referrer: optional boolean. If `true` does not require `noreferrer`. Defaults to `false`.
 * enabled: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
-* enforce: optional string, 'always' or 'never'
+* enforceDynamicLinks: optional string, 'always' or 'never'
+* warnOnSpreadAttributes: optional boolean. Defaults to `false`.
 
 ### `enforceDynamicLinks`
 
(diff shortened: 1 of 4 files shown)


michael-yx-wu:
cc @ljharb

docs/rules/jsx-no-target-blank.md
```diff
@@ -20,7 +16,8 @@ This rule aims to prevent user generated links from creating security vulnerabil
 
 * allow-referrer: optional boolean. If `true` does not require `noreferrer`. Defaults to `false`.
 * enabled: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
-* enforce: optional string, 'always' or 'never'
+* enforceDynamicLinks: optional string, 'always' or 'never'
+* trustSpreadAttributes: optional boolean. Defaults to `true`.
```
ljharb, line 20: i can't decide if this is fine, or if it'd be better to stick with the normal best practice of making "absence" and "false" be the same (and thus, inverting the name, eg `warnOnSpreadSyntax`)

ljharb (commented):
LGTM pending comment.

docs/rules/jsx-no-target-blank.md
michael-yx-wu: ah, i wasn't aware of that pattern! i'll change to `warnOnSpreadSyntax` to make it consistent

michael-yx-wu:
I've changed the option name to `warnOnSpreadAttributes` (defaults to `false`). Thanks for pointing me to the right pattern for these things. This is ready for review again!

lib/rules/jsx-no-target-blank.js
```diff
@@ -12,51 +12,75 @@ const linkComponentsUtil = require('../util/linkComponents');
...
+        if (expr.alternate.type === 'Literal' && expr.alternate.value && expr.alternate.value.toLowerCase() === '_blank') {
+          return true;
+        }
+        if (expr.consequent.type === 'Literal' && expr.consequent.value && expr.consequent.value.toLowerCase() === '_blank') {
+          return true;
+        }
+      }
+    }
+    return false;
+  })(node.attributes[targetIndex]);
```
ljharb, line 44: this IIFE is a bit weird here. can it be refactored?

lib/rules/jsx-no-target-blank.js
```diff
@@ -12,51 +12,75 @@ const linkComponentsUtil = require('../util/linkComponents');
...
+  const value = relAttribute.value
+    && ((
+      relAttribute.value.type === 'Literal'
+      && relAttribute.value.value
+    ) || (
+      relAttribute.value.type === 'JSXExpressionContainer'
+      && relAttribute.value.expression
+      && relAttribute.value.expression.value
+    ));
+  const tags = value && value.toLowerCase && value.toLowerCase().split(' ');
```
ljharb, line 82:
```suggestion
  const tags = value && typeof value === 'string' && value.toLowerCase().split(' ');
```

ljharb (changes requested):
LGTM overall

lib/rules/jsx-no-target-blank.js
michael-yx-wu: I had left this as is since it was carried over from the previous change, but I like this more explicit check as well.

lib/rules/jsx-no-target-blank.js
michael-yx-wu: Hm, yes maybe it's a bit too terse. I will break out the function, which might also make this a little faster since the rule won't have to recreate the function.

michael-yx-wu:
This is ready for review again.

ljharb (approved):

astorije:
Any chance this could be enabled by default or added to the recommended preset in the next major release? We're using the recommended preset, and that would save an override of the preset with this rule simply to enable this option 🙏

ljharb:
In the next major release, yes (along with many other changes), because it’d be a breaking change - but there’s no major releases planned any time soon.

PR #2862, opened 2020-11-25
[New] `jsx-no-target-blank`: add fixer

Nokel81:


lib/rules/jsx-no-target-blank.js
```diff
@@ -103,7 +104,47 @@ module.exports = {
...
             message: 'Using target="_blank" without rel="noreferrer" '
-              + 'is a security risk: see https://html.spec.whatwg.org/multipage/links.html#link-type-noopener'
+              + 'is a security risk: see https://html.spec.whatwg.org/multipage/links.html#link-type-noopener',
+            fix(fixer) {
+              const relAttribute = node.parent.attributes.find((attr) => attr.name.name === 'rel');
+              if (!relAttribute) {
+                return fixer.insertTextAfter(node, ' rel="noreferrer"');
+              }
+
+              if (!relAttribute.value) {
```
ljharb, line 114: can you add test cases for `rel=""`, `rel={false}`, `rel={0}`, etc?

lib/rules/jsx-no-target-blank.js
```diff
@@ -103,7 +104,47 @@ module.exports = {
...
+                const parts = relAttribute.value.value
+                  .split('noreferrer')
+                  .filter((part) => part);
+                return fixer.replaceText(relAttribute.value, `"${parts.join(' ')} noreferrer"`);
+              }
+
+              if (relAttribute.value.expression.type === 'Literal') {
+                const valueType = typeof relAttribute.value.expression.value;
+                if (
+                  ['boolean', 'number', 'symbol', 'bigint', 'undefined'].includes(valueType)
```
ljharb, line 132: what about `null`?

Nokel81: That is caught by the `|| <expr>` on line 133 (right below this one). Remember, `typeof null` results in `"object"`.

lib/rules/jsx-no-target-blank.js
Nokel81: sure

Nokel81: already existed for `""`. So I add for false, 0, and null

lib/rules/jsx-no-target-blank.js
ljharb, line 132: not a huge deal, but engines optimize the inline `typeof x === 'y'` pattern and do not optimize this pattern, so it'd likely be faster to do:
```suggestion
                if (
                  typeof relAttribute.value.expression.value === 'undefined'
                  || typeof relAttribute.value.expression.value === 'boolean'
                  || typeof relAttribute.value.expression.value === 'number'
                  || typeof relAttribute.value.expression.value === 'symbol'
                  || typeof relAttribute.value.expression.value === 'bigint'
```

lib/rules/jsx-no-target-blank.js
```diff
@@ -103,7 +104,47 @@ module.exports = {
...
+                  || relAttribute.value.expression.value === null
+                ) {
+                  return fixer.replaceText(relAttribute.value.expression, '"noreferrer"');
+                }
+
+                if (valueType === 'string') {
+                  const parts = relAttribute.value.expression.value
+                    .split('noreferrer')
+                    .filter((part) => part);
```
ljharb, line 141:
```suggestion
                    .filter(Boolean);
```

tests/lib/rules/jsx-no-target-blank.js
```diff
@@ -96,60 +96,97 @@ ruleTester.run('jsx-no-target-blank', rule, {
...
     code: '<a target="_blank" rel="" href="http://example.com"></a>',
+    output: '<a target="_blank" rel="noreferrer" href="http://example.com"></a>',
+    errors: defaultErrors
+  }, {
+    code: '<a target="_blank" rel={0} href="http://example.com"></a>',
+    output: '<a target="_blank" rel={"noreferrer"} href="http://example.com"></a>',
+    errors: defaultErrors
+  }, {
+    code: '<a target="_blank" rel={false} href="http://example.com"></a>',
+    output: '<a target="_blank" rel={"noreferrer"} href="http://example.com"></a>',
```
ljharb, line 111:
```suggestion
    output: '<a target="_blank" rel="noreferrer" href="http://example.com"></a>',
```
we shouldn't include the curly braces for a string literal autofix.

tests/lib/rules/jsx-no-target-blank.js
```diff
@@ -96,60 +96,97 @@ ruleTester.run('jsx-no-target-blank', rule, {
...
     code: '<a target="_blank" href="http://example.com"></a>',
+    output: '<a target="_blank" rel="noreferrer" href="http://example.com"></a>',
     errors: defaultErrors
   }, {
     code: '<a target="_blank" rel="" href="http://example.com"></a>',
+    output: '<a target="_blank" rel="noreferrer" href="http://example.com"></a>',
+    errors: defaultErrors
+  }, {
+    code: '<a target="_blank" rel={0} href="http://example.com"></a>',
+    output: '<a target="_blank" rel={"noreferrer"} href="http://example.com"></a>',
```
ljharb, line 107:
```suggestion
    output: '<a target="_blank" rel="noreferrer" href="http://example.com"></a>',
```

tests/lib/rules/jsx-no-target-blank.js
```diff
@@ -96,60 +96,97 @@ ruleTester.run('jsx-no-target-blank', rule, {
...
+    errors: defaultErrors
+  }, {
+    code: '<a target="_blank" rel={false} href="http://example.com"></a>',
+    output: '<a target="_blank" rel={"noreferrer"} href="http://example.com"></a>',
+    errors: defaultErrors
+  }, {
+    code: '<a target="_blank" rel={null} href="http://example.com"></a>',
+    output: '<a target="_blank" rel={"noreferrer"} href="http://example.com"></a>',
```
ljharb, line 115:
```suggestion
    output: '<a target="_blank" rel="noreferrer" href="http://example.com"></a>',
```

tests/lib/rules/jsx-no-target-blank.js
```diff
@@ -96,60 +96,97 @@ ruleTester.run('jsx-no-target-blank', rule, {
...
     code: '<a target="_blank" href="//example.com" rel={"noopenernoreferrer"}></a>',
+    output: '<a target="_blank" href="//example.com" rel={"noopener noreferrer"}></a>',
     errors: defaultErrors
   }, {
     code: '<a target={"_blank"} href={"//example.com"} rel={"noopenernoreferrer"}></a>',
+    output: '<a target={"_blank"} href={"//example.com"} rel={"noopener noreferrer"}></a>',
+    errors: defaultErrors
+  }, {
+    code: '<a target={"_blank"} href={"//example.com"} rel={"noopenernoreferrernoreferrernoreferrernoreferrernoreferrer"}></a>',
+    output: '<a target={"_blank"} href={"//example.com"} rel={"noopener noreferrer"}></a>',
```
ljharb, line 158: it's fine to keep it here since the original was both a string and had curly braces

ljharb (commented):
Almost there :-)

lib/rules/jsx-no-target-blank.js
Nokel81: Ah thanks, didn't know they normally optimize that. Will do.

ljharb: nobody does :-p it's super obscure knowledge

ljharb (approved):

ljharb:
@Nokel81 ok! i think this time i've got it down to 4 failures, not just in older eslint versions: namely, that the fixer *is* fixing the code in the spread attribute case. If CI ends up showing those same 4 failures, then fixing that would be the last step to landing this.

ljharb:
Please mark the PR as ready for review once either the tests are passing or you're stuck again (hopefully the former :-p )

Nokel81:
@ljharb All tests are passing.

ljharb (approved):

PR #2863, opened 2020-11-30
[New] add `no-invalid-html-attribute` rule

Nokel81:
This rule (and fixer) checks the "rel" value to make sure that it correctly relates to the tag that the "rel" property is on.

### DIFF
--- a/docs/rules/no-invalid-html-attribute.md
+++ b/docs/rules/no-invalid-html-attribute.md
@@ -1,0 +1,17 @@
+# Prevent usage of invalid attributes (react/no-invalid-html-attribute)
+
+Some HTML elements have a specific set of valid values for some attributes.
+For instance the elements: `a`, `area`, `link`, or `form` all have an attribute called `rel`.
+There is a fixed list of values that have any meaning for this attribute on these tags (see [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel)).
+To help with minimizing confusion while reading code, only the appropriate values should be on each attribute.
+
+## Rule Details
+
+This rule aims to remove invalid attribute values.
+
+## Rule Options
+The options is a list of attributes to check. Defaults to `["rel"]`.
+
+## When Not To Use It
+
+When you don't want to enforce attribute value correctness.
--- a/lib/rules/no-invalid-html-attribute.js
+++ b/lib/rules/no-invalid-html-attribute.js
@@ -1,0 +1,537 @@
+/**
+ * @fileoverview Check if tag attributes to have non-valid value
+ * @author Sebastian Malton
+ */
+
+'use strict';
+
+const matchAll = require('string.prototype.matchall');
+const docsUrl = require('../util/docsUrl');
+const report = require('../util/report');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+const rel = new Map([
+  ['alternate', new Set(['link', 'area', 'a'])],
+  ['author', new Set(['link', 'area', 'a'])],
+  ['bookmark', new Set(['area', 'a'])],
+  ['canonical', new Set(['link'])],
+  ['dns-prefetch', new Set(['link'])],
+  ['external', new Set(['area', 'a', 'form'])],
+  ['help', new Set(['link', 'area', 'a', 'form'])],
+  ['icon', new Set(['link'])],
+  ['license', new Set(['link', 'area', 'a', 'form'])],
+  ['manifest', new Set(['link'])],
+  ['modulepreload', new Set(['link'])],
+  ['next', new Set(['link', 'area', 'a', 'form'])],
+  ['nofollow', new Set(['area', 'a', 'form'])],
+  ['noopener', new Set(['area', 'a', 'form'])],
+  ['noreferrer', new Set(['area', 'a', 'form'])],
+  ['opener', new Set(['area', 'a', 'form'])],
+  ['pingback', new Set(['link'])],
+  ['preconnect', new Set(['link'])],
+  ['prefetch', new Set(['link'])],
+  ['preload', new Set(['link'])],
+  ['prerender', new Set(['link'])],
+  ['prev', new Set(['link', 'area', 'a', 'form'])],
+  ['search', new Set(['link', 'area', 'a', 'form'])],
+  ['stylesheet', new Set(['link'])],
+  ['tag', new Set(['area', 'a'])],
+]);
+
+/**
+ * Map between attributes and a mapping between valid values and a set of tags they are valid on
+ * @type {Map<string, Map<string, Set<string>>>}
+ */
+const VALID_VALUES = new Map([
+  ['rel', rel],
+]);
+
+/**
+ * The set of all possible HTML elements. Used for skipping custom types
+ * @type {Set<string>}
+ */
+const HTML_ELEMENTS = new Set([
+  'html',
+  'base',
+  'head',
+  'link',
+  'meta',
+  'style',
+  'title',
+  'body',
+  'address',
+  'article',
+  'aside',
+  'footer',
+  'header',
+  'h1',
+  'h2',
+  'h3',
+  'h4',
+  'h5',
+  'h6',
+  'main',
+  'nav',
+  'section',
+  'blockquote',
+  'dd',
+  'div',
+  'dl',
+  'dt',
+  'figcaption',
+  'figure',
+  'hr',
+  'li',
+  'ol',
+  'p',
+  'pre',
+  'ul',
+  'a',
+  'abbr',
+  'b',
+  'bdi',
+  'bdo',
+  'br',
+  'cite',
+  'code',
+  'data',
+  'dfn',
+  'em',
+  'i',
+  'kbd',
+  'mark',
+  'q',
+  'rp',
+  'rt',
+  'ruby',
+  's',
+  'samp',
+  'small',
+  'span',
+  'strong',
+  'sub',
+  'sup',
+  'time',
+  'u',
+  'var',
+  'wbr',
+  'area',
+  'audio',
+  'img',
+  'map',
+  'track',
+  'video',
+  'embed',
+  'iframe',
+  'object',
+  'param',
+  'picture',
+  'portal',
+  'source',
+  'svg',
+  'math',
+  'canvas',
+  'noscript',
+  'script',
+  'del',
+  'ins',
+  'caption',
+  'col',
+  'colgroup',
+  'table',
+  'tbody',
+  'td',
+  'tfoot',
+  'th',
+  'thead',
+  'tr',
+  'button',
+  'datalist',
+  'fieldset',
+  'form',
+  'input',
+  'label',
+  'legend',
+  'meter',
+  'optgroup',
+  'option',
+  'output',
+  'progress',
+  'select',
+  'textarea',
+  'details',
+  'dialog',
+  'menu',
+  'summary',
+  'slot',
+  'template',
+  'acronym',
+  'applet',
+  'basefont',
+  'bgsound',
+  'big',
+  'blink',
+  'center',
+  'content',
+  'dir',
+  'font',
+  'frame',
+  'frameset',
+  'hgroup',
+  'image',
+  'keygen',
+  'marquee',
+  'menuitem',
+  'nobr',
+  'noembed',
+  'noframes',
+  'plaintext',
+  'rb',
+  'rtc',
+  'shadow',
+  'spacer',
+  'strike',
+  'tt',
+  'xmp',
+]);
+
+/**
+* Map between attributes and set of tags that the attribute is valid on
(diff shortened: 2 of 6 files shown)


lib/rules/jsx-no-invalid-rel.js
```diff
@@ -0,0 +1,190 @@
...
+      realRelValues: '{{ value }} is never a valid "rel" attribute value.',
+      matchingRelValues: '"{{ value }}" is not a valid "rel" attribute value for <{{ tag }}>.'
+    }
+  },
+
+  create(context) {
+    return {
+      JSXAttribute(node) {
+        // ignore attributes that aren't "rel"
+        if (node.type !== 'JSXIdentifier' && node.name.name !== 'rel') {
```
golopot, line 135: `node.type` is always `JSXAttribute` here.

tests/lib/rules/jsx-no-invalid-rel.js
```diff
@@ -0,0 +1,464 @@
...
+    {code: '<form rel="noopener"></form>'},
+    {code: '<form rel="noreferrer"></form>'},
+    {code: '<form rel="opener"></form>'},
+    {code: '<form rel="prev"></form>'},
+    {code: '<form rel="search"></form>'},
+    {code: '<form rel={callFoo()}></form>'},
+    {code: '<a rel={{a: "noreferrer"}["a"]}></a>'},
+    {code: '<a rel={{a: "noreferrer"}["b"]}></a>'}
+  ],
+  invalid: [
```
golopot, line 91: missing tests for `<a rel="" />` and `<any rel />`

lib/rules/jsx-no-invalid-rel.js
```diff
@@ -0,0 +1,190 @@
...
+            fix(fixer) {
+              return fixer.remove(node);
+            }
+          });
+        }
+
+        if (node.value.type === 'Literal') {
+          return checkLiteralValueNode(context, node.value, parentNodeName);
+        }
+
```
golopot, line 163:
```suggestion
         if (node.value.type !== 'JSXExpressionContainer') {
          return;
        }
```
For readability it is nice to always assert node types.

tests/lib/rules/jsx-no-invalid-rel.js
```diff
@@ -0,0 +1,464 @@
...
+};
+
+// ------------------------------------------------------------------------------
+// Tests
+// ------------------------------------------------------------------------------
+
+const ruleTester = new RuleTester({parserOptions});
+
+ruleTester.run('jsx-no-invalid-rel', rule, {
+  valid: [
```
golopot, line 30: needs tests for non-html components like `<Foo rel="abc">`.

golopot (changes requested):

lib/rules/jsx-no-invalid-rel.js
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  return res;
+}
+
+function checkLiteralValueNode(context, node, parentNodeName) {
+  if (typeof node.value !== 'string') {
+    return context.report({
+      node,
+      messageId: 'onlyStrings',
```
golopot, line 67: tests are failing because eslint@4 does not support messageId and template messages.

docs/rules/jsx-no-invalid-rel.md
```diff
@@ -0,0 +1,14 @@
+# Prevent usage of invalid `rel` (react/jsx-no-invalid-rel)
+
+The JSX elements: `a`, `area`, `link`, or `form` all have a attribute called `rel`. There is is fixed list of values that have any meaning on these tags (see [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel)). To help with minimizing confusion while reading code, only the appropriate values should be on each attribute.
```
ljharb, line 3:
```suggestion
The JSX elements: `a`, `area`, `link`, or `form` all have an attribute called `rel`. There is is fixed list of values that have any meaning on these tags (see [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel)). To help with minimizing confusion while reading code, only the appropriate values should be on each attribute.
```

tests/lib/rules/jsx-no-invalid-rel.js
```diff
@@ -0,0 +1,464 @@
...
+    {
+      code: '<area rel="icon"></area>',
+      output: '<area rel=""></area>',
+      errors: [{
+        messageId: 'matchingRelValues'
+      }]
+    },
+    {
+      code: '<area rel="manifest"></area>',
+      output: '<area rel=""></area>',
```
ljharb, line 255: seems like it should just remove the `rel` value entirely?

ljharb (changes requested):
Plus all of @golopot's comments.

tests/lib/rules/jsx-no-invalid-rel.js
Nokel81: The eslint docs state that if any fixes are applied then a subsequent linting cycle is done. I believe that that would fix this.

ljharb: While that's true, it seems ideal to avoid forcing an unnecessary fix pass, especially since in editor integrations, only one pass is run. I'd prefer to optimize this.

docs/rules/jsx-no-invalid-rel.md
```diff
@@ -0,0 +1,16 @@
+# Prevent usage of invalid `rel` (react/jsx-no-invalid-rel)
+
+The JSX elements: `a`, `area`, `link`, or `form` all have an attribute called `rel`.
+There is is fixed list of values that have any meaning on these tags (see [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel)).
```
ljharb, line 4:
```suggestion
There is a fixed list of values that have any meaning for this attribute on these tags (see [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel)).
```

lib/rules/jsx-no-invalid-rel.js
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        const parentNodeName = node.parent.name.name;
+        if (!components.has(parentNodeName)) {
+          return context.report({
+            node,
+            message: 'The "rel" attribute only has meaning on `<link>`, `<a>`, `<area>`, and `<form>` tags.',
+            fix(fixer) {
+              return fixer.remove(node);
+            }
```
ljharb, line 144: this fixer function can be defined at module level and reused in a bunch of places.

ljharb (changes requested):
Are there other HTML element attributes that this type of rule might work on?

It might be nicer to make this rule a generic `no-invalid-html-attribute` rule (one that also works on `React.createElement`, and thus isn't jsx-specific), with an option for a list of attributes to check (that would default to `['rel']`). That way, we could add support for a new attribute in a semver-minor without needing to add an entirely new rule.

Thoughts?

Nokel81:
@ljharb Good idea about making this a more generic rule, though I don't know if I want to make it working with `React.createElement` in this PR.

Nokel81:
@ljharb I have changed the name of the rule and added the option as you said. While reworking the internals a bit (though I haven't tested the code with non-rel items)

docs/rules/no-invalid-html-attribute.md
```diff
@@ -0,0 +1,17 @@
+# Prevent usage of invalid `rel` (react/no-invalid-html-attribute)
```
ljharb, line 1: i think the docs here still need updating?

lib/rules/no-invalid-html-attribute.js
```diff
@@ -0,0 +1,219 @@
...
+  .set('opener', new Set(['area', 'a', 'form']))
+  .set('pingback', new Set(['link']))
+  .set('preconnect', new Set(['link']))
+  .set('prefetch', new Set(['link']))
+  .set('preload', new Set(['link']))
+  .set('prerender', new Set(['link']))
+  .set('prev', new Set(['link', 'area', 'a', 'form']))
+  .set('search', new Set(['link', 'area', 'a', 'form']))
+  .set('stylesheet', new Set(['link']))
+  .set('tag', new Set(['area', 'a']));
```
ljharb, line 45:
```suggestion
const rel = new Map([
  ['alternate', new Set(['link', 'area', 'a'])],
  ['author', new Set(['link', 'area', 'a'])],
  ['bookmark', new Set(['area', 'a'])],
  ['canonical', new Set(['link'])],
  ['dns-prefetch', new Set(['link'])],
  ['external', new Set(['area', 'a', 'form'])],
  ['help', new Set(['link', 'area', 'a', 'form'])],
  ['icon', new Set(['link'])],
  ['license', new Set(['link', 'area', 'a', 'form'])],
  ['manifest', new Set(['link'])],
  ['modulepreload', new Set(['link'])],
  ['next', new Set(['link', 'area', 'a', 'form'])],
  ['nofollow', new Set(['area', 'a', 'form'])],
  ['noopener', new Set(['area', 'a', 'form'])],
  ['noreferrer', new Set(['area', 'a', 'form'])],
  ['opener', new Set(['area', 'a', 'form'])],
  ['pingback', new Set(['link'])],
  ['preconnect', new Set(['link'])],
  ['prefetch', new Set(['link'])],
  ['preload', new Set(['link'])],
  ['prerender', new Set(['link'])],
  ['prev', new Set(['link', 'area', 'a', 'form'])],
  ['search', new Set(['link', 'area', 'a', 'form'])],
  ['stylesheet', new Set(['link'])],
  ['tag', new Set(['area', 'a'])],
]);
VALID_VALUES.set('rel', rel);
```

lib/rules/no-invalid-html-attribute.js
```diff
@@ -0,0 +1,219 @@
...
+const COMPONENT_ATTRIBUTE_MAP = new Map();
+COMPONENT_ATTRIBUTE_MAP.set('rel', new Set(['link', 'a', 'area', 'form']));
+
+function splitIntoRangedParts(node, regex) {
+  const res = [];
+  const valueRangeStart = node.range[0] + 1; // the plus one is for the initial quote
+  let match;
+
+  // eslint-disable-next-line no-cond-assign
+  while ((match = regex.exec(node.value)) !== null) {
```
ljharb, line 60: please use https://npmjs.com/string.prototype.matchall and `Array.from` for this rather than a while loop.

lib/rules/no-invalid-html-attribute.js
```diff
@@ -0,0 +1,219 @@
...
+
+  function fix(fixer) {
+    return fixer.remove(node);
+  }
+
+  const parentNodeName = node.parent.name.name;
+  if (!COMPONENT_ATTRIBUTE_MAP.has(attribute) || !COMPONENT_ATTRIBUTE_MAP.get(attribute).has(parentNodeName)) {
+    const tagNames = Array.from(COMPONENT_ATTRIBUTE_MAP.get(attribute).values())
+      .map((tagName) => `"<${tagName}>"`)
+      .join(', ');
```
ljharb, line 143:
```suggestion
    const tagNames = Array.from(
      COMPONENT_ATTRIBUTE_MAP.get(attribute).values(),
      (tagName) => `"<${tagName}>"`
    ).join(', ');
```

ljharb (changes requested):
Thanks, this is looking much better

Nokel81:
I know that there are some failing tests for some older versions of eslint. Will look at them soon.

ljharb:
No rush, please mark as ready for review once tests are passing.

ljharb (approved):
Thanks, this looks great! (sorry for the review delay)

This rule isn't named "jsx-", but it doesn't check `React.createElement`. I think it probably should, though. Is that something you'd mind adding?

Nokel81:
@ljharb I have added checking `React.createElement`.

ljharb (approved):
Are there other attributes besides "rel" we can add here?

Nokel81:
Maybe https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete but I think that could easily be done in a separate PR.

ljharb:
Rebased again; I tweaked the test. `<Foo rel="" />` must always be valid - this rule should only be checking HTML elements, never custom elements.

A few tests are still failing. @Nokel81?

Nokel81:
Well since you just changed the logic or tests (cannot tell). I was whitelisting elements whereas I think you want me to blacklist html elements that aren't on the list.

ljharb:
All I did is move the two "Foo" tests from invalid to valid.

ljharb:
I'm fine with an inclusion-list-based approach, but it should ignore custom elements entirely.

Nokel81:
@ljharb Sorry it took so long, this PR has been updated to ignore non-HTML tags.

ljharb:
@Nokel81 i enabled all the test cases and used the messageId pattern and rebased.

ljharb (approved):

Issue #2417, opened 2019-09-19
Feature request: enforce naming convention for useState()


mbrowne:
It would be great to have a rule to enforce the standard naming convention for `useState()`:

```
const [x, setX] = useState(...)
```

For example, the rule would consider this an error:

```
const [x, setFoo] = useState(...)
```

This would ensure that all useState variables followed the same convention, i.e. `${varName}` and `set${upperFirstVarName}`.

ljharb:
This seems like a better rule for eslint-plugin-react-hooks.

mbrowne:
@ljharb I read somewhere that eslint-plugin-react-hooks deliberately avoids including stylistic rules, and intends to focus only on behavioral things, i.e. "rules of hooks". But maybe they would be open to adding this stylistic rule if it were disabled by default...I could open an issue there and see what they say.

karlhorky:
@ljharb Would the `eslint-plugin-react` project be opposed to adding this feature?

Seems like the issue in `facebook/react` to add it to `eslint-plugin-react-hooks` got no response and was autoclosed by the bot...

ljharb:
I'd love to see it go into eslint-plugin-react-hooks if possible.

karlhorky:
Ok, @mbrowne opened a new PR at `facebook/react`, let's see if it gets any traction: https://github.com/facebook/react/issues/19883

karlhorky:
So gaearon confirmed in https://github.com/facebook/react/issues/19883#issuecomment-722059542:

> We won't be focusing on stylistic enforcement like this in the official plugin

@ljharb would a pull request be welcome in `eslint-plugin-react`?

ljharb:
Is there a way the “hook detection” code could be shared between the two? I don’t want this plugin to have to reinvent it, and then risk deviating from it in the future.

karlhorky:
Maybe depending on `eslint-plugin-react-hooks` and then somehow using the `isHook` and `isHookName` functions?

https://github.com/facebook/react/blob/30b47103d4354d9187dc0f1fb804855a5208ca9f/packages/eslint-plugin-react-hooks/src/RulesOfHooks.js#L12-L40

Unfortunately they are not exported, but depending on the willingness of @gaearon to expose them and the value of sharing this code, I guess the file could be patched using `patch-package` or similar...

ljharb:
Hopefully they can either be exported, or published as a separate package that both consume. @gaearon?

[...]

PR #2873, opened 2020-12-09
[New] : Enforce useState hook destructuring and variable naming

duncanbeevers:
Ensure symmetric value and setter variable names are destructured from React.setState hook calls

Towards #2417

@karlhorky feel free to riff on this, or let me know how this compares to your own proof-of-concept.

* Supports both `useState` and `React.useState`
* Tested with TypeScript parser and `useState<>` type annotation
* Robust against weird `useState` uses like destructuring too many items, or not destructuring at all
* Fixable when 2nd destructured arg does not matching `set<X>` naming convention

@ljharb I included this in the `Best Practices` category, though perhaps a different categorization would be more appropriate.

Let me know if the documentation needs anything else, like more examples, or whether the laissez-faire approach to weird assignment is too forgiving.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,7 @@ Enable the rules that you would like to use.
 * [react/forbid-foreign-prop-types](docs/rules/forbid-foreign-prop-types.md): Forbid using another component's propTypes
 * [react/forbid-prop-types](docs/rules/forbid-prop-types.md): Forbid certain propTypes
 * [react/function-component-definition](docs/rules/function-component-definition.md): Standardize the way function component get defined (fixable)
+* [react/hook-use-state](docs/rules/hook-use-state.md): Ensure symmetric naming of useState hook value and setter variables (fixable)
 * [react/no-access-state-in-setstate](docs/rules/no-access-state-in-setstate.md): Reports when this.state is accessed within setState
 * [react/no-adjacent-inline-elements](docs/rules/no-adjacent-inline-elements.md): Prevent adjacent inline elements not separated by whitespace.
 * [react/no-array-index-key](docs/rules/no-array-index-key.md): Prevent usage of Array index in keys
--- a/docs/rules/hook-use-state.md
+++ b/docs/rules/hook-use-state.md
@@ -1,0 +1,23 @@
+# Ensure destructuring and symmetric naming of useState hook value and setter variables (react/hook-use-state)
+
+**Fixable:** In some cases, this rule is automatically fixable using the `--fix` flag on the command line.
+
+## Rule Details
+
+This rule checks whether the value and setter variables destructured from a `React.useState()` call are named symmetrically.
+
+Examples of **incorrect** code for this rule:
+
+```js
+const useStateResult = React.useState();
+```
+
+```js
+const [color, updateColor] = React.useState();
+```
+
+Examples of **correct** code for this rule:
+
+```js
+const [color, setColor] = React.useState();
+```
--- a/index.js
+++ b/index.js
@@ -16,6 +16,7 @@ const allRules = {
   'forbid-foreign-prop-types': require('./lib/rules/forbid-foreign-prop-types'),
   'forbid-prop-types': require('./lib/rules/forbid-prop-types'),
   'function-component-definition': require('./lib/rules/function-component-definition'),
+  'hook-use-state': require('./lib/rules/hook-use-state'),
   'jsx-boolean-value': require('./lib/rules/jsx-boolean-value'),
   'jsx-child-element-spacing': require('./lib/rules/jsx-child-element-spacing'),
   'jsx-closing-bracket-location': require('./lib/rules/jsx-closing-bracket-location'),
--- a/lib/rules/hook-use-state.js
+++ b/lib/rules/hook-use-state.js
@@ -68,7 +68,7 @@ module.exports = {
 
         const valueVariableName = valueVariable
           ? valueVariable.name
-          : 'state';
+          : undefined;
 
         const setterVariableName = setterVariable
           ? setterVariable.name
@@ -89,12 +89,10 @@ module.exports = {
           context.report({
             node: node.parent.id,
             messageId: USE_STATE_ERROR_MESSAGE,
-            fix(fixer) {
-              return fixer.replaceTextRange(
-                [node.parent.id.range[0] + 1, node.parent.id.range[1] - 1],
-                `${valueVariableName}, ${expectedSetterVariableName}`
-              );
-            }
+            fix: valueVariableName ? (fixer) => fixer.replaceTextRange(
+              [node.parent.id.range[0], node.parent.id.range[1]],
+              `[${valueVariableName}, ${expectedSetterVariableName}]`
+            ) : undefined
           });
         }
       }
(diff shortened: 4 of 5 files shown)


karlhorky:
Awesome @duncanbeevers ! I think you've probably put more effort into this than I have up until now - haven't found the time to learn what I would need to in order to tackle this 😬

I think it's looking good 🙌

@ljharb what are your thoughts?

docs/rules/hook-use-state-names.md
```diff
@@ -0,0 +1,19 @@
...
+Examples of **incorrect** code for this rule:
+
+```js
+const [color, updateColor] = React.useState();
+```
+
+Examples of **correct** code for this rule:
+
+```js
+const [color, setColor] = React.useState();
```
ljharb, line 18: Would someone want to configure "set" to make "update" allowed? Should we allow that to be configurable?

lib/rules/hook-use-state-names.js
```diff
@@ -0,0 +1,114 @@
...
+    },
+    schema: [{
+      type: 'object',
+      properties: {
+        ignoreTranspilerName: {
+          type: 'boolean'
+        }
+      },
+      additionalProperties: false
+    }]
```
ljharb, line 35: copypasta?
```suggestion
    schema: [{
      type: 'object',
      additionalProperties: false
    }]
```

lib/rules/hook-use-state-names.js
```diff
@@ -0,0 +1,114 @@
...
+
+          const setterVariableName = setterVariable
+            ? setterVariable.name
+            : undefined;
+
+          if (valueVariableName) {
+            const expectedSetterVariableName = (
+              `set${
+                valueVariableName.charAt(0).toUpperCase()
+              }${valueVariableName.substr(1, valueVariableName.length)}`
```
ljharb, line 87: substr is deprecated and should never be used.
```suggestion
              }${valueVariableName.slice(1)}`
```

tests/lib/rules/hook-use-state-names.js
```diff
@@ -0,0 +1,85 @@
...
+      code: 'const [color1, setColor1] = useState()'
+    },
+    {
+      code: 'const [, setColor] = useState()'
+    },
+    {
+      code: 'const useStateResult = useState()'
+    },
+    {
+      code: 'const [, , unused] = useState()'
```
ljharb, line 51: this seems like it should be an error, but i guess not from this rule

tests/lib/rules/hook-use-state-names.js
```diff
@@ -0,0 +1,85 @@
...
+      code: 'const [color, setColor, unused] = useState()'
+    },
+    {
+      code: 'const [color1, setColor1] = useState()'
+    },
+    {
+      code: 'const [, setColor] = useState()'
+    },
+    {
+      code: 'const useStateResult = useState()'
```
ljharb, line 48: what about if someone does `const [state, setThatState] = useStateResult;` or `const state = useStateResult[0]; const setThatState = useStateResult[1];` on the next lines?

tests/lib/rules/hook-use-state-names.js
```diff
@@ -0,0 +1,85 @@
...
+      code: 'const [, , unused] = useState()'
+    },
+    {
+      code: 'const [color, setColor] = useState<string>()',
+      parser: parsers.TYPESCRIPT_ESLINT
+    },
+    {
+      code: 'const [color, setColor] = useState<string>(\'#ffffff\')',
+      parser: parsers.TYPESCRIPT_ESLINT
+    }
```
ljharb, line 60: let's also duplicate these and add `parsers['@TYPESCRIPT_ESLINT']` so we can handle the new parser too. grep for `parsers.TS(` to find examples of how to do this safely.

tests/lib/rules/hook-use-state-names.js
```diff
@@ -0,0 +1,85 @@
...
+      }],
+      output: 'const [color, setColor] = useState()'
+    },
+    {
+      code: 'const [color, setFlavor] = useState<string>()',
+      errors: [{
+        message: 'setState variable names are not symmetric, setter should be setColor'
+      }],
+      output: 'const [color, setColor] = useState<string>()',
+      parser: parsers.TYPESCRIPT_ESLINT
```
ljharb, line 82: duplicate test case with the new parser here too

ljharb (changes requested):

docs/rules/hook-use-state-names.md
karlhorky: Have you seen this a lot before? I haven't seen this pattern before.

I suppose allowing arbitrary configuration of the prefix would be ok though (eg. a `setterPrefix` option?) - would allow for flexibility for some teams.

tests/lib/rules/hook-use-state-names.js
duncanbeevers: This rule only looks for calls to `useState` which are directly destructured+assigned;

These intermediate-value forms aren't recognized by this rule.

docs/rules/hook-use-state-names.md
duncanbeevers: > Would someone want to configure "set" to make "update" allowed? Should we allow that to be configurable?

Just as the naming of hooks is already opinionated, `useXYZ`, all the React docs consistently use `[thing, setThing]`.
I think it's best to lean on this convention here.

tests/lib/rules/hook-use-state-names.js
duncanbeevers: Yeah, in my workflow this issue would be caught by the type system.
This test is intended to indicate the rule's non-opinion around a third argument (and to make sure I didn't throw an error on weird inputs)

tests/lib/rules/hook-use-state-names.js
duncanbeevers: I think I added these correctly.

duncanbeevers:
@ljharb The scope of this rule has increased, and its naming has been changed from `hook-use-state-names` to the "generic" `hook-use-state`.

As you mentioned in your first review, certain unusual usages of `setState` weren't addressed by this rule.
In addressing these issues the other comments raised in this PR, the prescriptive nature of this rule seemed clearer.

Towards that end of providing gentle, but clear guidance in the usage of this hook+rule, the naming has been changed to `hook-use-state`.

It is now a bit more prescriptive, and therefore a bit more helpful in its fix suggestions.

### Stuff that didn't change
* there is no configuration
* all previously-invalid cases remain invalid (invalid cases were focused solely on variable naming violations)

### Stuff that changed
* enforced destructuring of `useState` call to value+setter
* added fix for overly-arity-destructured useState calls

tests/lib/rules/hook-use-state-names.js
duncanbeevers: This is now an error.

tests/lib/rules/hook-use-state-names.js
duncanbeevers: This is now an error as well.

duncanbeevers:
Closing now that the core rule change has landed.
https://github.com/eslint/eslint/pull/14012

duncanbeevers:
Oops wrong PR. I'll re-open this one and remove the no-unused-expresssion PR.

mea culpa

PR #2876, opened 2020-12-14
Reduce install size by 30%

developit:
`Object.entries` is supported in Node 6+ and does not need to be polyfilled.

`Object.fromEntries` is supported in Node 12+, and its usage here is minimal enough that an inline shim suffices.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,8 +1,5 @@
 'use strict';
 
-const fromEntries = require('object.fromentries');
-const entries = require('object.entries');
-
 /* eslint-disable global-require */
 const allRules = {
   'boolean-prop-naming': require('./lib/rules/boolean-prop-naming'),
@@ -97,8 +94,14 @@ const allRules = {
 };
 /* eslint-enable */
 
+const fromEntries = Object.fromEntries || function fromEntries(entries) {
+  const obj = {};
+  for (const [key, value] of entries) obj[key] = value;
+  return obj;
+}
+
 function filterRules(rules, predicate) {
-  return fromEntries(entries(rules).filter((entry) => predicate(entry[1])));
+  return fromEntries(Object.entries(rules).filter((entry) => predicate(entry[1])));
 }
 
 function configureAsError(rules) {


index.js
```diff
@@ -97,8 +94,14 @@ const allRules = {
 };
 /* eslint-enable */
 
+const fromEntries = Object.fromEntries || function fromEntries(entries) {
+  const obj = {};
+  for (const [key, value] of entries) obj[key] = value;
+  return obj;
+}
```
ljharb, line 101: inlining a polyfill (one that's not matching the spec) is a nonstarter.

index.js
```diff
@@ -1,8 +1,5 @@
 'use strict';
 
-const fromEntries = require('object.fromentries');
-const entries = require('object.entries');
```
ljharb, line 4: this plugin supports node 4, which lacks Object.entries, so we still need the polyfill.

lib/util/Components.js
```diff
@@ -6,8 +6,6 @@
 'use strict';
 
 const doctrine = require('doctrine');
-const arrayIncludes = require('array-includes');
-const values = require('object.values');
```
ljharb, line 10: node 4 also lacks these functions.

ljharb (changes requested):
This plugin supports node 4, and inlining a polyfill is always a much much worse solution than using a dependency.

index.js
developit: Why not rewrite it to avoid the expensive polyfill then?

ljharb: The polyfill isn't expensive, it's the cheapest possible way to correctly implement the method.

developit:
Why would this plugin have any reason to support Node 4 when ESlint itself only supports Node 10+?

ljharb:
Because it also supports down to eslint 3, which supports node 4. The gain is backwards compatibility, and avoiding a semver-major bump.

developit:
Have you evaluated the performance and disk usage tradeoff this is inherently making, or is this a purely philosophical stance?

ljharb:
It's a dev tool; install speed impact is negligible and disk usage is irrelevant.

developit:
I'm not sure that viewpoint is broadly shared. Developers have hundreds of projects on a machine, and this one is just under 8mb. Anecdotally, eslint-plugin-react accounts for more than 1GB of disk usage on my machine.

ljharb:
I have almost 500 projects on my personal machine that are fully npm-installed, a handful of them using eslint-plugin-react, and the total size of the entire dir that holds them (including `.git` and `node_modules` and the actual code) is around 20GB - anecdata is tricky :-/

If the goal is to take up less disk space, I'm happy to discuss ways to do that on those polyfill repos - that would have far greater impact than removing them from one eslint plugin.

developit:
That seems like a reasonable plan - I don't have exact numbers on the % impact of the polyfill repos, but estimating based on packagephobia seems to indicate they're at least 50% of the disk usage of this plugin.

TrySound:
Here's a number for one included polyfill. 3.39mb. Would be good at least to not have duplicated monstrous packages.
Would be good to not pay for theoretical users of eslint 3 and nonde 4.
https://packagephobia.com/result?p=string.prototype.matchall@4.0.3
https://packagephobia.com/result?p=es-abstract@1.17.7

chyzwar:
Maybe releasing library major version every 10 major version of node.js would be "stable enough". @ljharb how long do you plan to support node v4?

onigoetz:
Are there stats somewhere on the actual users of ESLint 3, node.js 4 that need the latest `eslint-plugin-react` for React 18 ?

I understand that people could be using older versions of ESLint and node.js, but supporting runtimes that were released 7 years ago seems a bit overkill. Even more so that for people stuck on old versions of ESLint and node.js are also probably stuck on React 0.14

Releasing a new major version of `eslint-plugin-react` would still allow these people to use the previous major version for their use case.

ljharb:
@onigoetz react 18 works just fine on ancient versions of node, so that assumption doesn't hold. Also, the instant we shipped a bugfix or a new feature, the folks stuck on the previous major would be left in the lurch. That's a high cost, compared to the virtually zero cost of "a few extra packages".

@chyzwar i maintain over 300 packages that still support node 0.4. So, "how long"? Probably forever.

onigoetz:
True, maybe it would suck /if/ those people are actually using Node 4 and ESLint 3

On another hand, your package has 12'500'000 downloads a week, [according to packagephobia](https://packagephobia.com/result?p=eslint-plugin-react), one install is 3.77MB (Unzipped, so let's say half that when transfered)

Which means (3.77/2)*12500000 = 23 562 500 MB = 23.5625 TB

In other words, this PR is proposing to shave off 7.8 TB of data transfer per week.

Here is an article written by `chokidar`'s maintainer Paul Miller on how the newest version is massively smaller and the implications of that size : https://paulmillr.com/posts/chokidar-3-save-32tb-of-traffic/

PR #2877, opened 2020-12-15
[Fix] `prop-types`: default argument does not count as props-types declaration

### LINKED ISSUE #2867: react/prop-types: components with props that all have default values fails to validate
### Problem
When using the `react/prop-types` rule when a component has default values for **all** of its props the rule fails to validate correctly.

### Reproduction case
_Run it in [repl.it/@jvdl/eslint-plugin-react-default-value-prop-validation-bug](https://repl.it/@jvdl/eslint-plugin-react-default-value-prop-validation-bug#example.js) or clone [jvanderloo/eslint-plugin-react-default-value-prop-validation-bug](https://github.com/jvanderloo/eslint-plugin-react-default-value-prop-validation-bug) and run locally._

eslint config:
```json
  "rules": {
    "react/prop-types": "error"
  }
```

Example that **does** validate as expected:
```jsx
function Foo({ fooProp1 }) {
  return (<p>{fooProp1}</p>);
}
```
Example that **does not** validate as expected:
```jsx
function FooBar({ foobarProp1 = "foo" }) {
  return (<p>{foobarProp1}</p>);
}
```
### Expected results
(Assuming a file that has both of the above examples in it)
```
./example.js
   4:16  error  'fooProp1' is missing in props validation  react/prop-types
  19:19  error  'foobarProp1' is missing in props validation  react/prop-types
```
Note in the above expected results that *both* functions fail props validation.

### Actual results
```
./example.js
  4:16  error  'fooProp1' is missing in props validation  react/prop-types
```
Note that for the function `FooBar` the props validation passes, even though there is no explicit prop types declaration, only the implicit one by virtue of the types of the default value for the prop.

### Additional notes
It appears that this case only occurs when **all** props have a default value, if only some of the props have one then all props will be properly validated as not having a prop type.
 e.g. in the following example only one of the props has a default value
```jsx
function Bar({ barProp1 = "foo", barProp2 }) {
  return (<p>{barProp1} and {barProp2}</p>);
}
```
This shows the expected result of:
```
  10:16  error  'barProp1' is missing in props validation  react/prop-types
  10:34  error  'barProp2' is missing in props validation  react/prop-types
```

golopot:
Fixes #2867.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,10 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 * [`no-typos`]: avoid crash with computed method name ([#2870][] @ljharb, @AriPerkkio)
 * [`jsx-max-depth`]: avoid crash with childless jsx child ([#2869][] @ljharb, @AriPerkkio)
 * [`no-unknown-property`]: avoid crash with prop named with Object.prototype key ([#2879][] @ljharb, @AriPerkkio)
+* [`prop-types`]: default argument does not count as props-types declaration ([#2877][] @golopot)
 
 [#2879]: https://github.com/yannickcr/eslint-plugin-react/issues/2879
+[#2877]: https://github.com/yannickcr/eslint-plugin-react/pull/2877
 [#2875]: https://github.com/yannickcr/eslint-plugin-react/issues/2875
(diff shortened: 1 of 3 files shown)


ljharb (approved):
Thanks!

PR #2881, opened 2020-12-20
[Docs] `prop-types`: fix example

technote-space:
Fixed typo to make the interface usage correct.


ljharb (approved):
For future reference, PRs from forks that aren't "yours" can't be edited by maintainers; you may want to avoid doing that in the future.

PR #2883, opened 2020-12-22
[Fix] `no-unused-prop-types`: apply `skipShapeProps` to exact types

### LINKED ISSUE #2850: `no-unused-prop-types`: All fields reported as unused with PropTypes.arrayOf(PropTypes.exact)
It seems like when a list of objects is typed using `PropTypes.arrayOf(PropTypes.exact({ ... }))` all properties of the inner object are reported as not in use by `no-unused-prop-types` regardless of them being used or not. This is not the case with `PropTypes.arrayOf(PropTypes.shape({ ... }))`.

~~I would expect `PropTypes.arrayOf(PropTypes.exact)` and `PropTypes.arrayOf(PropTypes.shape)` to behave the same when checking for unused properties~~
I would expect the two below cases to have no errors as they are very similar. However, case A has no errors, while case B reports that `a.*.b` is not in use.

### Case A:
No errors reported

```js
const Comp = ({ a }) => <div>{a.map(({ b }) => <div>{b}</div>)}</div>;

Comp.propTypes = {
  a: PropTypes.arrayOf(PropTypes.shape({ b: PropTypes.string }))
};
```

### Case B:

Error: `'a.*.b' PropType is defined but prop is never used`

```js
const Comp = ({ a }) => <div>{a.map(({ b }) => <div>{b}</div>)}</div>;

Comp.propTypes = {
  a: PropTypes.arrayOf(PropTypes.exact({ b: PropTypes.string }))
};
```

### .eslintrc.json

```json
{
  "parser": "babel-eslint",
  "plugins": ["react"],
  "rules": {
    "react/no-unused-prop-types": 1
  }
}
```

### package.json

```json
{
  "dependencies": {
    "babel-eslint": "^10.1.0",
    "eslint": "^7.12.1",
    "eslint-plugin-react": "^7.21.5"
  }
}
```

golopot:
Fixes #2850.

The option `skipShapeProps` was only applied to `PropTypes.shape`, this pr make the option apply to `PropTypes.exact` as well.

Props usages like `props.a.map({b} => use(b))` are hard to handle. Currently `PropTypes.shape` has the same bug under the option `skipShapeProps: false`. People do not complain about it because most people use `skipShapeProps: true`.

The bug has two parts:
1. `skipShapeProps` should applies to `PropTypes.exact`.
2. Props usages like `props.a.map({b} => use(b))` should be handled.

This pr only fixes 1.

### DIFF
--- a/lib/rules/no-unused-prop-types.js
+++ b/lib/rules/no-unused-prop-types.js
@@ -70,6 +70,7 @@ module.exports = {
         const usedProp = usedPropTypes[i];
         if (
           prop.type === 'shape'
+          || prop.type === 'exact'
           || prop.name === '__ANY_KEY__'
           || usedProp.name === prop.name
         ) {
@@ -98,7 +99,7 @@ module.exports = {
           return;
         }
 
-        if (prop.type === 'shape' && configuration.skipShapeProps) {
+        if ((prop.type === 'shape' || prop.type === 'exact') && configuration.skipShapeProps) {
           return;
         }
 
--- a/tests/lib/rules/no-unused-prop-types.js
+++ b/tests/lib/rules/no-unused-prop-types.js
@@ -3213,6 +3213,25 @@ ruleTester.run('no-unused-prop-types', rule, {
         Foo.defaultProps = Object.assign({});
       `
     },
+    {
+      code: `
+        const Hello = ({a}) => (
+          <div>
+            {a.map(({b}) => (
+              <div>{b}</div>
+            ))}
+          </div>
+        );
+        Hello.propTypes = {
+          a: PropTypes.arrayOf(
+            PropTypes.exact({
+              b: PropTypes.string,
+            })
+          ),
+        };
+      `,
+      parser: parsers.BABEL_ESLINT
+    },
     parsers.TS([
       {
         code: `


lib/rules/no-unused-prop-types.js
```diff
@@ -65,11 +65,10 @@ module.exports = {
      * @returns {Boolean} True if the prop is used, false if not.
      */
     function isPropUsed(node, prop) {
-      const usedPropTypes = node.usedPropTypes || [];
-      for (let i = 0, l = usedPropTypes.length; i < l; i++) {
-        const usedProp = usedPropTypes[i];
+      for (const usedProp of node.usedPropTypes || []) {
```
ljharb, line 68: i pretty intentionally didn't use `for..of` here, since the array iteration protocol is slow and brittle. i don't think this change is needed for this fix; let's revert it
```suggestion
      const usedPropTypes = node.usedPropTypes || [];
      for (let i = 0, l = usedPropTypes.length; i < l; i++) {
        const usedProp = usedPropTypes[i];
```

ljharb (changes requested):
LGTM pending the one comment

ljharb (approved):

PR #2888, opened 2020-12-26
[New Rule] jsx-embed-condition

steelbrain:
This rule disallows use of `&&` inside JSX Embeds to avoid conditional numbers from being rendered.

Fixes https://github.com/yannickcr/eslint-plugin-react/issues/1979

### DIFF
--- a/lib/rules/jsx-embed-condition.js
+++ b/lib/rules/jsx-embed-condition.js
@@ -1,0 +1,45 @@
+/**
+ * @fileoverview Prevents usage of && condition in JSX Embeds.
+ * @author Anees Iqbal <i@steelbrain.me>
+ */
+
+'use strict';
+
+const docsUrl = require('../util/docsUrl');
+
+// -----------------------------------------------------------------------------
+// Rule Definition
+// -----------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Prevents usage of && condition in JSX embed',
+      category: 'Best Practices',
+      recommended: false,
+      url: docsUrl('jsx-embed-condition')
+    },
+
+    schema: []
+  },
+
+  create(context) {
+    return {
+      JSXExpressionContainer(node) {
+        if (
+          node.parent == null
+          || node.parent.type !== 'JSXElement'
+          || node.expression == null
+          || node.expression.type !== 'LogicalExpression'
+          || node.expression.operator === '??'
+        ) {
+          return;
+        }
+        context.report({
+          node,
+          message: 'Using && to condition JSX embeds is forbidden. Convert it to a ternary operation instead'
+        });
+      }
+    };
+  }
+};
--- a/tests/lib/rules/jsx-embed-condition.js
+++ b/tests/lib/rules/jsx-embed-condition.js
@@ -1,0 +1,72 @@
+/**
+ * @fileoverview Prevents usage of && condition in JSX Embeds.
+ * @author Anees Iqbal <i@steelbrain.me>
+ */
+
+'use strict';
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+const RuleTester = require('eslint').RuleTester;
+const version = require('eslint/package.json').version;
+const semver = require('semver');
+const rule = require('../../../lib/rules/jsx-embed-condition');
+const parsers = require('../../helpers/parsers');
+
+const parserOptions = {
+  ecmaVersion: 2018,
+  sourceType: 'module',
+  ecmaFeatures: {
+    jsx: true
+  }
+};
+
+// ------------------------------------------------------------------------------
+// Tests
+// ------------------------------------------------------------------------------
+
+const ruleTester = new RuleTester({parserOptions});
+ruleTester.run('jsx-embed-condition', rule, {
+  valid: [].concat({
+    code: '<App>Test</App>'
+  }, {
+    code: '<App test>Another</App>'
+  }, {
+    code: '<App foo={e => bar(e)}>Hello World</App>'
+  }, {
+    code: '<App>{x ? <div></div> : null}</App>'
+  }, {
+    code: '<App>{x ? <div>Hello</div> : null}</App>'
+  }, {
+    code: '<App>{x ? <div>{y ? <y /> : <z />}</div> : null}</App>'
+  }, {
+    code: '<App x={x && y}>{x ? <div>{y ? <y /> : <z />}</div> : null}</App>'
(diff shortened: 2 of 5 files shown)


lib/rules/jsx-embed-condition.js
```diff
@@ -0,0 +1,46 @@
...
+          || node.expression == null
+          || node.expression.type !== 'LogicalExpression'
+          || node.expression.operator === '??'
+        ) {
+          return;
+        }
+        context.report({
+          node,
+          message:
+            'Using && to condition JSX embeds is forbidden. Convert it to a ternary operation instead'
```
ljharb, line 41:
```suggestion
          message: 'Using && to condition JSX embeds is forbidden. Convert it to a ternary operation instead'
```
this looks suspiciously like terrible prettier-style indentation :-/ please don't run prettier on repos that don't have it configured.

tests/lib/rules/jsx-embed-condition.js
```diff
@@ -0,0 +1,64 @@
...
+  }, {
+    code: '<App>{x ? <div>{y ? <y /> : <z />}</div> : null}</App>'
+  }, {
+    code: '<App x={x && y}>{x ? <div>{y ? <y /> : <z />}</div> : null}</App>'
+  }
+  // {
+  //   code: '<App test>{x ?? <div />}</App>'
+  // }
+  // TODO: ^ Using above gives a Parsing error in ESLint.
+  // Is the parser out of date
```
ljharb, line 48: no, but you can specify the `parserOptions.ecmaVersion` to be `2020` in this test case.

ljharb (changes requested):
I'm still not convinced this is a rule worth adding. If it were added, it'd need autofix to avoid being too disruptive.

steelbrain:
> I'm still not convinced this is a rule worth adding

It is a convention recommended by [The official React docs](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator). A whole bunch of people have been stung by this bug, many didn't even know what caused it, see:

- https://github.com/facebook/react-native/issues/20764
- https://github.com/facebook/react-native/issues/23607
- https://github.com/facebook/react-native/issues/23735
- https://github.com/artsy/emission/pull/1706
- https://blog.echobind.com/a-common-conditional-rendering-bug-in-react-native-8bbbc400abd8

(and you'll find many more examples by searching for "Text strings must be rendered within a Text component" on Google)

Apart from crashing React Native hard, the bug is known to add random `0`s to the DOM in HTML renderer. I wrote this PR because my team got bit by this a few times.

> If it were added, it'd need autofix to avoid being too disruptive.

This Lint rule is opt-in only. I can commit to adding an autofix for it later next month. My work schedule is pretty tight before that.

> you can specify the parserOptions.ecmaVersion to be 2020 in this test case.

Doing so fixed the check on tests for latest Node.js runners, but broke them on the rest. I looked around the repository and couldn't find any tests with `ecmaVersion` newer than 2018. I would appreciate some help here! Thanks!

tests/lib/rules/jsx-embed-condition.js
ljharb: to clarify, i meant in this *one test case*, not for all test cases.

ljharb:
You can pass `parserOptions.ecmaVersion` directly in the one test case.

steelbrain:
> You can pass parserOptions.ecmaVersion directly in the one test case.

Attempted in https://github.com/yannickcr/eslint-plugin-react/pull/2888/commits/8abb5dcd900845d416f4469d7fc2fbf9197c4040 No dice. Any other ideas?

haroldwaste (approved):

ljharb (commented):
I've rebased and hopefully fixed the tests. I'm not sold on the name of the rule - "embeds" isn't a term I've ever heard before to describe jsx curly-brace expressions.

Specifically, this is a rule concerned with preventing accidental *rendering* of a zero. This can happen only in the "children" position in jsx, or in a `React.createElement` call.

So:
 - let's add support for `React.createElement`, both in the `children` position and in a literal `children` prop passed to that function
 - let's rename the rule. perhaps, since it's no longer just jsx, `no-conditional-children`?

ljharb:
Please mark the PR as ready for review once the next round of changes is done. Thanks!

duncanbeevers:
We have a similar rule in an internal codebase called `logical-and-expressions`; its existence in our jsx-ruleset makes it clear what its scope is.

Since this plugin's purview is a bit broader, perhaps a name like `jsx-logical-and-expressions` would capture the rule's scope+spirit.

duncanbeevers:
@steelbrain Here's the internal rule we use, which includes a simple fixer.

```js
module.exports = {
  meta: {
    docs: {
      url: 'SECRET URL'
    },
    fixable: 'code'
  },
  create(context) {

    return {
      JSXExpressionContainer(node) {
        if (
          node.expression.type === 'LogicalExpression' &&
          node.expression.operator === '&&' &&
          node.parent.type !== 'JSXAttribute'
        ) {
          context.report({
            node,
            message: 'JSX expression should not use logical-and.',
            fix: (fixer) => fixer.replaceText(
              node,
              `{${sourceCode.getText(node.expression.left)} ? ${sourceCode.getText(node.expression.right)} : null}`
            )
          });
        }
      }
    };
  }
};
```

steelbrain:
I do not have the bandwidth to work on it this week, I'll continue this work on the weekend. Thank you for your continued support!

Dakuan:
We've also been bit by this bug and would like to see this rule enforced in our code base

ljharb:
@steelbrain are you still planning to finish this?

holic:
@ljharb Happy to open up a new PR with the lint rule + fix + tests if ya'll are interested in this. I was gonna release it as a separate package otherwise.

steelbrain:
I do not have the bandwidth in the near future and this needs more love than I can give right now :)
Good luck @holic

ljharb:
Let's please keep this open so it can be reused.

ljharb:
@holic please post a branch link here, and i'll update the PR

holic:
Added a PR here: https://github.com/jsx-eslint/eslint-plugin-react/pull/3318

This ports over the code we used internally at my last company that I was intending to release as its own package. Let me know if you'd prefer I build off of this PR instead.

ljharb:
@holic i thought i was pretty clear i did *not* want a second PR opened; that pollutes the repo forever. Now I have to keep both PRs open and in sync until both are landed together.

steelbrain:
Closed by https://github.com/jsx-eslint/eslint-plugin-react/pull/3203

PR #2891, opened 2020-12-26
[Fix] `no-adjacent-inline-elements`: add url

thofmann:
The url for the `react/style-prop-object` rule is missing. This adds the correct url using the `docsUrl` util.


ljharb:
@thofmann for future reference, these kinds of things can be grouped as a single PR, rather than multiple.

PR #2893, opened 2020-12-30
[Fix] `static-property-placement`: do not report non-components

### LINKED ISSUE #2884: static-property-placement reports violations on non-React classes
Version 7.21.5

A `static-property-placement` violation is reported on a non-React class e.g.

```typescript
class Foo {
    static context: {};
}
```

Should it not check that the class extends `React.Component`?

golopot:
Fixes #2884.


ljharb (approved):
Thanks!

PR #2906, opened 2021-01-23
[Fix] `jsx-pascal-case`: support regex ignore option

### LINKED ISSUE #2905: Feature request: [jsx-pascal-case] Support regex/glob match
Like the ESLint camelcase rule: https://github.com/eslint/eslint/blob/master/docs/rules/camelcase.md

This is to support suffixes like `_DEPRECATED`, etc.

bcherny:
Fixes #2905.

Note: This will potentially relax existing `ignore`s:

- **Before**, `ignore`only checked exact matches
- **After**, `ignore` checks partial matches as well

I think this is fine, since it mirrors ESLint's behavior for the `camelcase` rule. If we prefer to be extra careful, I can make this a new option (eg. `ignoreRegex`).

### DIFF
--- a/docs/rules/jsx-pascal-case.md
+++ b/docs/rules/jsx-pascal-case.md
@@ -46,7 +46,7 @@ Examples of **correct** code for this rule:
 
 * `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
 * `allowAllCaps`: optional boolean set to `true` to allow components name in all caps (default to `false`).
-* `ignore`: optional string-array of component names to ignore during validation.
+* `ignore`: optional string-array of component names to ignore during validation (supports [minimatch](https://github.com/isaacs/minimatch)-style globs).
 
 ### `allowAllCaps`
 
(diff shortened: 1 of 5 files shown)


ljharb (changes requested):
While It's impressive you found what I believe is the only option in eslint core that accepts regex, it's very dangerous to do `new RegExp` from user-supplied values, and we won't be supporting any regex options here. It would be reasonable to support globs instead, however.

As to this rule specifically, I'd love to understand the use case a bit more - why do you need "deprecated" to be in the name on the consuming side? `import { Foo_DEPRECATED as Foo }`, for example, works fine.

bcherny:
Updated per convo in https://github.com/yannickcr/eslint-plugin-react/issues/2905

docs/rules/jsx-pascal-case.md
```diff
@@ -46,7 +46,7 @@ Examples of **correct** code for this rule:
 
 * `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
 * `allowAllCaps`: optional boolean set to `true` to allow components name in all caps (default to `false`).
-* `ignore`: optional string-array of component names to ignore during validation.
+* `ignore`: optional string-array of component names to ignore during validation (supports [minimatch](https://github.com/isaacs/minimatch)).
```
ljharb, line 49:
```suggestion
* `ignore`: optional string-array of component names to ignore during validation (supports [minimatch](https://github.com/isaacs/minimatch)-style globs).
```

ljharb (approved):
Seems straightforward now, thanks!

PR #2908, opened 2021-01-27
[Docs] Improve legibility of lists of available rules in readme

motato1:
The README contains three different lists of rules (basic rules, JSX-specific ones, recommended ones). I've found that looking through those lists to find out what rule is part of the recommended ruleset is not as simple as it could be.

This PR replaces the lists of available rules with tables containing indicators for whether a rule is enabled in the recommended ruleset and if it is fixable (just like the official [ESLint page](https://eslint.org/docs/rules/) does).

I have removed the list of the recommended rules entirely as the information can now be found in the tables mentioned above, but I can undo that change if you prefer to keep that list in (or transform it into a table as well).

The result looks like this:

✔: Enabled in the [`recommended`](#recommended) configuration.\
🔧: Fixable with [`eslint --fix`](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems).

| ✔ | 🔧 | Rule | Description |
| :---: | :---: | :--- | :--- |
| ✔ |  | [react/display-name](docs/rules/display-name.md) | Prevent missing displayName in a React component definition |
|  | 🔧 | [react/function-component-definition](docs/rules/function-component-definition.md) | Standardize the way function component get defined |
|  |  | [react/forbid-component-props](docs/rules/forbid-component-props.md) | Forbid certain props on components |

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -50,5 +50,11 @@
           "no-template-curly-in-string": 1,
         },
       },
+      {
+        "files": "markdown.config.js",
+        "rules": {
+          "no-console": 0,
+        },
+      },
     ],
 }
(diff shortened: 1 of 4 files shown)


ljharb (approved):
This looks amazing, thanks!

PR #2912, opened 2021-01-29
[Fix] `destructuring-assignment`: fix a false prositive for local prop named `context` in SFC

### LINKED ISSUE #2911: Local prop named `context` erroneously handled as `this.context`
Example of broken code:

```
import React from 'react';

const Op = ({context}) => {
  const d = context.describe();
  return (
    <div className="op">{JSON.stringify(d)}</div>
  );
};

Op.propTypes = {
  context: PropTypes.shape({
    describe: PropTypes.func,
  }).isRequired,
};

export default Op;
```

which results in:

```
error: Must use destructuring context assignment (react/destructuring-assignment):
  2 | 
  3 | const Op = ({context}) => {
> 4 |   const d = context.describe();
    |             ^
  5 |   return (
  6 |     <div className="op">{JSON.stringify(d)}</div>
  7 |   );
```

Note: I edited to fix a typo I made while transcribing the test code, but the problem remains the same.

SyMind:
Fixes #2911.


lib/rules/destructuring-assignment.js
```diff
@@ -10,6 +10,44 @@ const isAssignmentLHS = require('../util/ast').isAssignmentLHS;
...
+    },
+    get contextName() {
+      for (const params of queue) {
+        const context = params[1];
+        if (context && !context.destructuring && context.name) {
+          return context.name;
+        }
+      }
+      return null;
+    }
```
ljharb, line 40: getters are slow and the indirection is less maintainable; let's just make these functions.
```suggestion
    propsName() {
      for (const params of queue) {
        const props = params[0];
        if (props && !props.destructuring && props.name) {
          return props.name;
        }
      }
      return null;
    },
    contextName() {
      for (const params of queue) {
        const context = params[1];
        if (context && !context.destructuring && context.name) {
          return context.name;
        }
      }
      return null;
    }
```

lib/rules/destructuring-assignment.js
```diff
@@ -38,21 +76,22 @@ module.exports = {
   create: Components.detect((context, components, utils) => {
     const configuration = context.options[0] || DEFAULT_OPTION;
     const ignoreClassFields = (context.options[1] && (context.options[1].ignoreClassFields === true)) || false;
+    const sfcParams = createSFCParams();
```
ljharb, line 79: hm, does eslint itself not have utilities for determining the relationship to function params?

lib/rules/destructuring-assignment.js
SyMind: ok, but why getters are slow and the indirection is less maintainable, please tell me more detail. please do tell me!

ljharb: They’re slow because it takes more effort for the engine to use them; they’re less maintainable because it looks like a property at the usage site, but it’s actually invoking a function. Making it a real function means you add `()` and suddenly everything is faster and more obvious. Getters and setters are bad and should be avoided.

cha0s: I haven't contributed code to this issue I originally reported (thanks @SyMind), but I am just reviewing what has been done so far and I feel compelled to say: that's a bad take. Getters and setters are quite a bit more performant than you might think (I use them in real-time applications even with emitters/streams to great effect), but I can respect that since you maintain this project you may have a different opinion.

ljharb:
@SyMind are you no longer interested in completing this PR? it's just awaiting some tweaks.

ljharb:
Although it looks like you deleted your fork, which is very sad, because I think that's an unrecoverable state.

SyMind:
@ljharb Sorry, I did a stupid `git push --force`. I will pull request again.

ljharb:
ok, thanks

PR #2916, opened 2021-02-02
[react/jsx-uses-vars]: False negative on HTML elements used in JSX

karlhorky:
This PR is to add a failing test for the following situation:

```js
// False Negative:
// The `div` parameter below is not recognized as an unused variable
function App(div) {
  return <div />
}
```

This also happens with other variables (eg. variables declared separately):

```js
const div = 123;

export function App() {
  return <div />;
}
```

Incorrectly reported here: https://github.com/typescript-eslint/typescript-eslint/issues/2985

### DIFF
--- a/tests/lib/rules/jsx-uses-vars.js
+++ b/tests/lib/rules/jsx-uses-vars.js
@@ -215,6 +215,19 @@ ruleTester.run('no-unused-vars', ruleNoUnusedVars, {
         React.render(<lowercase />);
       `,
       errors: [{message: '\'lowercase\' is defined but never used.'}]
+    }, {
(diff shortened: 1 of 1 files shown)


ljharb:
Thanks; there's some work that should likely be done here with our `variableUtil`.

ljharb:
This test is now passing; it was fixed in #3070. Will merge to close.

karlhorky:
Great, thanks! Will look for this in the next patch (`7.26.1`) or minor (`7.27.0`)

ljharb:
It’ll be a minor, since there’s already new rules landed.

PR #2562, opened 2020-02-02
[Fix] restore eslint < 4.15 compatibility

toshi-toma:
already seems do not work in ESLint v3. and CI isn't checking ESLint v3.

<details><summary>errors</summary>
4245 passing (5s)
  1 failing

  1) jsx-no-undef
       valid
         /*eslint no-undef:1*/ var React, App; React.render(<App />);:
     TypeError: Cannot read property 'variables' of null
      at checkIdentifierInJSX (lib/rules/jsx-no-undef.js:9:2040)
      at EventEmitter.JSXOpeningElement (lib/rules/jsx-no-undef.js:9:4242)
      at NodeEventGenerator.applySelector (node_modules/eslint/lib/util/node-event-generator.js:265:26)
      at NodeEventGenerator.applySelectors (node_modules/eslint/lib/util/node-event-generator.js:294:22)
      at NodeEventGenerator.enterNode (node_modules/eslint/lib/util/node-event-generator.js:308:14)
      at CodePathAnalyzer.enterNode (node_modules/eslint/lib/code-path-analysis/code-path-analyzer.js:602:23)
      at CommentEventGenerator.enterNode (node_modules/eslint/lib/util/comment-event-generator.js:98:23)
      at Traverser.enter (node_modules/eslint/lib/eslint.js:929:36)
      at Traverser.__execute (node_modules/estraverse/estraverse.js:330:31)
      at Traverser.traverse (node_modules/estraverse/estraverse.js:434:28)
      at Traverser.traverse (node_modules/eslint/lib/util/traverser.js:31:22)
      at EventEmitter.module.exports.api.verify (node_modules/eslint/lib/eslint.js:926:23)
      at runRuleForItem (node_modules/eslint/lib/testers/rule-tester.js:387:38)
      at testValidTemplate (node_modules/eslint/lib/testers/rule-tester.js:420:28)
      at Context.<anonymous> (node_modules/eslint/lib/testers/rule-tester.js:548:25)
      at processImmediate (internal/timers.js:439:21)
</details>

PR #2918, opened 2021-02-06
[Fix] move error messages to meta property

### LINKED ISSUE #2915: Rule messages not defined in meta properties
Is there a reason some rule messages aren't defined through the meta properties, or is this just a small oversight? For example, `jsx-indent` and `jsx-indent-props` have their messages defined in the `create` function.

Is there any interest in converting these to use the meta message property to make writing tests easier in the future?

acdvs:
This moves all error messages for every rule to the built-in `meta.messages` property to support easier rule writing in the future.

Resolves #2915

### DIFF
--- a/lib/rules/boolean-prop-naming.js
+++ b/lib/rules/boolean-prop-naming.js
@@ -14,6 +14,10 @@ const propWrapperUtil = require('../util/propWrapper');
 // Rule Definition
 // ------------------------------------------------------------------------------
 
+// Predefine message for use in context.report conditional.
+// messageId will still be usable in tests.
+const PATTERN_MISMATCH_MSG = 'Prop name ({{propName}}) doesn\'t match rule ({{pattern}})';
+
 module.exports = {
   meta: {
     docs: {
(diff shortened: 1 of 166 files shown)


ljharb:
I’ve fixed the test failures on master and rebase this; the current failures are due to this PR.

acdvs:
@ljharb Took a few extra commits to fix the test cases that weren't caught with local unit tests. Should be good to go after the rebase issue.

acdvs:
@ljharb Could you provide some insight as to why the automatic rebase action is failing here? The logs say there's some conflicts but I'm not actually seeing anything.

ljharb:
I don't see a conflict, and I'm not sure, but don't worry about it - i can merge through that if needed.

ljharb:
Somehow it's finding 487 merge commits, and there's zero, so that's clearly a bug in the action :-)

ljharb:
I've rebased it manually, which should resolve the issue.

ljharb:
Turns out this further broke eslint < v4.15, since v4.15 added `messageId` support, but we weren't testing on eslint 3 so we didn't catch it. #2562 will restore support for it.

PR #2921, opened 2021-02-10
[New] Symmetric useState hook variable names

### LINKED ISSUE #2417: Feature request: enforce naming convention for useState()

duncanbeevers:
Ensure two symmetrically-named variables are destructured from useState hook calls

### Discussion

This is a re-opening of https://github.com/yannickcr/eslint-plugin-react/pull/2873
I've rebased and updated the README using `npm run generate-list-of-rules` to follow the new table-ish layout.

* Supports both `useState` and `React.useState`
* Tested with TypeScript parser and `useState<>` type annotation
* Robust against weird `useState` uses like destructuring too many items, or not destructuring at all
* Fixable when 2nd destructured arg does not match `set<X>` naming convention

Fixes #2417.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ Enable the rules that you would like to use.
 |  |  | [react/forbid-foreign-prop-types](docs/rules/forbid-foreign-prop-types.md) | Forbid using another component's propTypes |
 |  |  | [react/forbid-prop-types](docs/rules/forbid-prop-types.md) | Forbid certain propTypes |
 |  | 🔧 | [react/function-component-definition](docs/rules/function-component-definition.md) | Standardize the way function component get defined |
+|  |  | [react/hook-use-state](docs/rules/hook-use-state.md) | Ensure symmetric naming of useState hook value and setter variables |
 |  |  | [react/no-access-state-in-setstate](docs/rules/no-access-state-in-setstate.md) | Reports when this.state is accessed within setState |
 |  |  | [react/no-adjacent-inline-elements](docs/rules/no-adjacent-inline-elements.md) | Prevent adjacent inline elements not separated by whitespace. |
 |  |  | [react/no-array-index-key](docs/rules/no-array-index-key.md) | Prevent usage of Array index in keys |
--- a/docs/rules/hook-use-state.md
+++ b/docs/rules/hook-use-state.md
@@ -1,0 +1,46 @@
+# Ensure destructuring and symmetric naming of useState hook value and setter variables (react/hook-use-state)
+
+## Rule Details
+
+This rule checks whether the value and setter variables destructured from a `React.useState()` call are named symmetrically.
+
+Examples of **incorrect** code for this rule:
+
+```js
+import React from 'react';
+export default function useColor() {
+  // useState call is not destructured into value + setter pair
+  const useStateResult = React.useState();
+  return useStateResult;
+}
+```
+
+```js
+import React from 'react';
+export default function useColor() {
+  // useState call is destructured into value + setter pair, but identifier
+  // names do not follow the [thing, setThing] naming convention
+  const [color, updateColor] = React.useState();
+  return useStateResult;
+}
+```
+
+Examples of **correct** code for this rule:
+
+```js
+import React from 'react';
+export default function useColor() {
+  // useState call is destructured into value + setter pair whose identifiers
+  // follow the [thing, setThing] naming convention
+  const [color, setColor] = React.useState();
+  return [color, setColor];
+}
+```
+
+```js
+import React from 'react';
+export default function useColor() {
+  // useState result is directly returned
+  return React.useState();
+}
+```
--- a/lib/rules/hook-use-state.js
+++ b/lib/rules/hook-use-state.js
@@ -1,0 +1,152 @@
+/**
+ * @fileoverview Ensure symmetric naming of useState hook value and setter variables
+ * @author Duncan Beevers
+ */
+
+'use strict';
+
+const Components = require('../util/Components');
+const docsUrl = require('../util/docsUrl');
+const report = require('../util/report');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+const messages = {
+  useStateErrorMessage: 'useState call is not destructured into value + setter pair',
+};
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Ensure symmetric naming of useState hook value and setter variables',
+      category: 'Best Practices',
+      recommended: false,
+      url: docsUrl('hook-use-state'),
+    },
+    messages,
+    schema: [],
+    type: 'suggestion',
+    hasSuggestions: true,
+  },
+
+  create: Components.detect((context, components, util) => ({
+    CallExpression(node) {
+      const isImmediateReturn = node.parent
+        && node.parent.type === 'ReturnStatement';
+
+      if (isImmediateReturn || !util.isReactHookCall(node, ['useState'])) {
+        return;
+      }
+
+      const isDestructuringDeclarator = node.parent
+        && node.parent.type === 'VariableDeclarator'
+        && node.parent.id.type === 'ArrayPattern';
+
+      if (!isDestructuringDeclarator) {
+        report(
+          context,
+          messages.useStateErrorMessage,
+          'useStateErrorMessage',
+          { node }
+        );
+        return;
+      }
+
+      const variableNodes = node.parent.id.elements;
+      const valueVariable = variableNodes[0];
+      const setterVariable = variableNodes[1];
+
+      const valueVariableName = valueVariable
+        ? valueVariable.name
+        : undefined;
+
+      const setterVariableName = setterVariable
+        ? setterVariable.name
+        : undefined;
+
+      const expectedSetterVariableName = valueVariableName ? (
+        `set${valueVariableName.charAt(0).toUpperCase()}${valueVariableName.slice(1)}`
+      ) : undefined;
+
+      const isSymmetricGetterSetterPair = valueVariable
+        && setterVariable
+        && setterVariableName === expectedSetterVariableName
+        && variableNodes.length === 2;
+
+      if (!isSymmetricGetterSetterPair) {
+        const suggestions = [
+          {
+            desc: 'Destructure useState call into value + setter pair',
+            fix: (fixer) => {
+              const fix = fixer.replaceTextRange(
+                node.parent.id.range,
+                `[${valueVariableName}, ${expectedSetterVariableName}]`
+              );
+
+              return fix;
+            },
+          },
+        ];
+
+        const defaultReactImports = components.getDefaultReactImports();
+        const defaultReactImportSpecifier = defaultReactImports
+          ? defaultReactImports[0]
+          : undefined;
+
+        const defaultReactImportName = defaultReactImportSpecifier
+          ? defaultReactImportSpecifier.local.name
+          : undefined;
+
+        const namedReactImports = components.getNamedReactImports();
+        const useStateReactImportSpecifier = namedReactImports
+          ? namedReactImports.find((specifier) => specifier.imported.name === 'useState')
+          : undefined;
+
+        const isSingleGetter = valueVariable && variableNodes.length === 1;
+        const isUseStateCalledWithSingleArgument = node.arguments.length === 1;
+        if (isSingleGetter && isUseStateCalledWithSingleArgument) {
+          const useMemoReactImportSpecifier = namedReactImports
+            && namedReactImports.find((specifier) => specifier.imported.name === 'useMemo');
+
+          let useMemoCode;
+          if (useMemoReactImportSpecifier) {
+            useMemoCode = useMemoReactImportSpecifier.local.name;
+          } else if (defaultReactImportName) {
+            useMemoCode = `${defaultReactImportName}.useMemo`;
+          } else {
+            useMemoCode = 'useMemo';
+          }
+
+          suggestions.unshift({
+            desc: 'Replace useState call with useMemo',
+            fix: (fixer) => [
+              // Add useMemo import, if necessary
+              useStateReactImportSpecifier
+                  && (!useMemoReactImportSpecifier || defaultReactImportName)
+                  && fixer.insertTextAfter(useStateReactImportSpecifier, ', useMemo'),
+              // Convert single-value destructure to simple assignment
+              fixer.replaceTextRange(node.parent.id.range, valueVariableName),
+              // Convert useState call to useMemo + arrow function + dependency array
+              fixer.replaceTextRange(
+                node.range,
+                `${useMemoCode}(() => ${context.getSourceCode().getText(node.arguments[0])}, [])`
+              ),
+            ].filter(Boolean),
+          });
+        }
+
+        report(
+          context,
+          messages.useStateErrorMessage,
+          'useStateErrorMessage',
+          {
+            node: node.parent.id,
+            suggest: suggestions,
+          }
+        );
+      }
+    },
+  })),
+};
--- a/tests/lib/rules/hook-use-state.js
+++ b/tests/lib/rules/hook-use-state.js
@@ -1,0 +1,549 @@
+/**
+ * @fileoverview Ensure symmetric naming of setState hook value and setter variables
+ * @author Duncan Beevers
+ */
+
+'use strict';
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+const RuleTester = require('eslint').RuleTester;
+const rule = require('../../../lib/rules/hook-use-state');
+const parsers = require('../../helpers/parsers');
+
+// ------------------------------------------------------------------------------
+// Tests
+// ------------------------------------------------------------------------------
+
+const ruleTester = new RuleTester({
+  parserOptions: {
+    ecmaVersion: 2018,
+    sourceType: 'module',
+  },
+});
+
+const tests = {
+  valid: parsers.all([
+    {
+      code: `
+        import { useState } from 'react'
+        function useColor() {
+          const [color, setColor] = useState()
+          return [color, setColor]
+        }
+      `,
+    },
+    {
+      code: `
+        import { useState } from 'react'
+        function useColor() {
+          const [color, setColor] = useState('#ffffff')
+          return [color, setColor]
+        }
+      `,
+    },
+    {
+      code: `
+        import React from 'react'
+        function useColor() {
+          const [color, setColor] = React.useState()
+          return [color, setColor]
+        }
+      `,
+    },
+    {
+      code: `
+        import { useState } from 'react'
+        function useColor() {
+          const [color1, setColor1] = useState()
+          return [color1, setColor1]
+        }
+      `,
+    },
+    {
+      code: 'useState()',
+    },
+    {
+      code: 'const result = useState()',
+    },
+    {
+      code: 'const [color, setFlavor] = useState()',
+    },
+    {
+      code: `
+        import React from 'react'
+        import useState from 'someOtherUseState'
+        const [color, setFlavor] = useState()
+      `,
+    },
+    {
+      code: `
+        import { useRef } from 'react'
+        const result = useState()
+      `,
+    },
+    {
+      code: `
+        import { useState as useStateAlternativeName } from 'react'
+        function useColor() {
+          const [color, setColor] = useStateAlternativeName()
+          return [color, setColor]
+        }
+      `,
+    },
+    {
+      code: 'const result = React.useState()',
+    },
+    {
+      code: `
+        import { useState } from 'react'
+        function useColor() {
+          return useState()
+        }
+      `,
+    },
+    {
+      code: `
+        import { useState } from 'react'
+        function useColor() {
+          function useState() { // shadows React's useState
+            return null
+          }
+
+          const result = useState()
+        }
+      `,
+    },
+    {
+      code: `
+        import React from 'react'
+        function useColor() {
+          const React = {
+            useState: () => {
+              return null
+            }
+          }
+
+          const result = React.useState()
+        }
+      `,
+    },
+    {
+      code: `
+        import { useState } from 'react';
+        function useColor() {
+          const [color, setColor] = useState<string>()
+          return [color, setColor]
+        }
+      `,
+      features: ['ts', 'no-babel-old'],
+    },
+    {
+      code: `
+        import { useState } from 'react';
+        function useColor() {
+          const [color, setColor] = useState<string>('#ffffff')
+          return [color, setColor]
+        }
+      `,
+      features: ['ts'],
+    },
+  ]),
+  invalid: parsers.all([
+    {
+      code: `
+        import { useState } from 'react';
+        const result = useState()
+      `,
+      errors: [{ message: 'useState call is not destructured into value + setter pair' }],
+    },
+    {
+      code: `
+        import { useState } from 'react';
+        function useColor() {
+          const result = useState()
+          return result
+        }
+      `,
+      errors: [{ message: 'useState call is not destructured into value + setter pair' }],
+    },
+    {
+      code: `
+        import { useState as useStateAlternativeName } from 'react'
+        function useColor() {
+          const result = useStateAlternativeName()
+          return result
+        }
+      `,
+      errors: [{ message: 'useState call is not destructured into value + setter pair' }],
+    },
+    {
+      code: `
+        import React from 'react'
+        function useColor() {
+          const result = React.useState()
+          return result
+        }
+      `,
+      errors: [{ message: 'useState call is not destructured into value + setter pair' }],
+    },
+    {
+      code: `
+        import ReactAlternative from 'react'
+        function useColor() {
+          const result = ReactAlternative.useState()
+          return result
+        }
+      `,
+      errors: [{ message: 'useState call is not destructured into value + setter pair' }],
+    },
+    {
+      code: `
+        import { useState } from 'react'
+        function useColor() {
+          const result = useState()
+          return result
+        }
+      `,
+      errors: [{ message: 'useState call is not destructured into value + setter pair' }],
+    },
+    {
+      code: `
+        import { useState } from 'react'
+        function useColor() {
+          const [, , extra1] = useState()
+        }
+      `,
+      errors: [{ message: 'useState call is not destructured into value + setter pair' }],
+    },
+    {
+      code: `
+        import { useState } from 'react'
+        function useColor() {
+          const [, setColor] = useState()
+        }
+      `,
+      errors: [{ message: 'useState call is not destructured into value + setter pair' }],
+    },
+    {
+      code: `
+        import { useState } from 'react'
+        function useColor() {
+          const { color } = useState()
+        }
+      `,
+      errors: [{ message: 'useState call is not destructured into value + setter pair' }],
+    },
+    {
+      code: `
+        import { useState } from 'react'
+        function useColor() {
+          const [] = useState()
+        }
+      `,
+      errors: [{
+        message: 'useState call is not destructured into value + setter pair',
+      }],
+    },
+    {
+      code: `
+        import { useState } from 'react'
+        function useColor() {
+          const [, , , ,] = useState()
+        }
+      `,
+      errors: [{ message: 'useState call is not destructured into value + setter pair' }],
+    },
+    {
+      code: `
+        import { useState } from 'react'
+        function useColor() {
+          const [color] = useState()
+        }
+      `,
+      errors: [
+        {
+          message: 'useState call is not destructured into value + setter pair',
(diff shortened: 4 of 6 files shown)


ljharb:
i would have strongly preferred reopening #2873 instead of leaving it permanently orphaned on the repo forever :-/ please strive to avoid opening unnecessary PRs.

duncanbeevers:
> i would have strongly preferred reopening #2873

Same; sorry about the premature close+delete. After I restored the branch, I wasn't able to re-open the PR.
I'll tread more lightly in the future.

tests/lib/rules/hook-use-state.js
```diff
@@ -0,0 +1,172 @@
...
+  parserOptions: {
+    ecmaVersion: 2018,
+    sourceType: 'module'
+  }
+});
+
+const tests = {
+  valid: [
+    {
+      code: 'const [color, setColor] = useState()'
```
ljharb, line 30: none of these validate that `useState` is used inside a react component. to avoid false positives, the rule should be limited to our react component detection.

ljharb (changes requested):
What about `React.useState` calls?

tests/lib/rules/hook-use-state.js
duncanbeevers: It's not uncommon to use these hooks outside of React components; for example, when writing custom hooks.

duncanbeevers:
> What about React.useState calls?

Those are covered in the documentation and in the tests;

https://github.com/yannickcr/eslint-plugin-react/pull/2921/files#diff-21ec40762f2f5ec7df5499db37c5971f0dc3f590927c1f49fddd72b0e103e057R35-R37


```ts
    {
      code: 'const [color, setColor] = React.useState()'
    },
```

tests/lib/rules/hook-use-state.js
ljharb: That's a fair point - but in that case, we need to validate that `useState` is *react's* `useState`.

duncanbeevers: In the explicit `React.useState` case, do you feel it's important to verify that `React` is actually React (eg, is preceded by `import React from 'react';` or somesuch?

duncanbeevers:
I've added some primitive support for ensuring that React / useState is imported (though not necessarily in-scope).
`React.useState` calls without an imported React are not flagged.

I tried using the [ScopeManager](https://eslint.org/docs/developer-guide/scope-manager-interface) to determine at the `CallExpression` whether `React` or `useState` were in scope, but didn't succeed in getting any Scope information at all out of it.

I've never seen a `ScopeManager` usage in the wild so I may simply have been holding it wrong.

ljharb:
Could we add some tests for `useState` etc that *aren't* the hook, and also some tests with a pragma that isn't "React"?

duncanbeevers:
Yeah, I'll take another pass at this.

ljharb:
No, but the ts feature runs on both the old and new TS parsers - it’s fine to add a `no-ts-old` feature if the old one doesn’t support what’s valid syntax.

duncanbeevers:
> Could we add some tests for useState etc that aren't the hook, and also some tests with a pragma that isn't "React"?

* Added valid specs where a non-React `useState` function is invoked.
* Added invalid specs where `React` is imported with an alternative identifier
* Added invalid specs where `useState` is imported with an alternative identifier

I've updated the specs to use the new `parsers.all` helper, annotating TS-specific tests with the `ts` feature.
I had to annotate a few cases with `no-babel-old`, all related to Babel(old)'s failure to recognize a typed, 0-argument `useState<T>()` expression

lib/rules/hook-use-state.js
```diff
@@ -0,0 +1,124 @@
...
+      url: docsUrl('hook-use-state'),
+    },
+    fixable: 'code',
+    messages: {
+      [USE_STATE_ERROR_MESSAGE]: 'useState call is not destructured into value + setter pair',
+    },
+    schema: [{
+      type: 'object',
+      additionalProperties: false,
+    }],
```
ljharb, line 31:
```suggestion
    schema: [],
```
otherwise passing a config of `{}` is allowed

lib/rules/hook-use-state.js
```diff
@@ -0,0 +1,124 @@
...
+  },
+
+  create(context) {
+    let reactImportLocalName;
+    let reactUseStateLocalName;
+
+    return {
+      CallExpression(node) {
```
ljharb, line 39: This should all be done inside a Components.detect call, to ensure it only runs on things that are detected as SFCs.

lib/rules/hook-use-state.js
```diff
@@ -0,0 +1,124 @@
...
+          : undefined;
+
+        const setterVariableName = setterVariable
+          ? setterVariable.name
+          : undefined;
+
+        const expectedSetterVariableName = valueVariableName ? (
+          `set${
+            valueVariableName.charAt(0).toUpperCase()
+          }${valueVariableName.slice(1)}`
```
ljharb, line 85:
```suggestion
          `set${valueVariableName.charAt(0).toUpperCase()}${valueVariableName.slice(1)}`
```

tests/lib/rules/hook-use-state.js
```diff
@@ -0,0 +1,233 @@
...
+      const [color, setFlavor, extraneous] = useState()`,
+      errors: [{
+        message: 'useState call is not destructured into value + setter pair',
+      }],
+      output: `import { useState } from 'react';
+      const [color, setColor] = useState()`,
+    },
+    {
+      code: `import { useState } from 'react';
+      const [color, setFlavor] = useState()`,
```
ljharb, line 193: `useState` can only be called inside an SFC that React renders (or inside another hook), so i don't think these tests are valid unless they're inside a component or inside a custom hook

ljharb (changes requested):

tests/lib/rules/hook-use-state.js
duncanbeevers: Do you feel that these cases as-written should be flagged as invalid by this rule?
Or should I just add the dressing to these test cases?

ljharb: As written, the rule should ignore them, because they're not actually a valid hook call - it will throw at runtime already.

Also, eslint-plugin-react-hooks will, i believe, warn on it as-is.

lib/rules/hook-use-state.js
duncanbeevers: Okay, I wrapped the rule definition in `Components.detect`

duncanbeevers:
I wrapped most test-cases in a little more realistic code; a custom hook that acts as pass-through for the useState return value.
It highlights the problem that I'm able to rename one instance of the variable (within the destructuring assignment) but don't fix other instances of that identifier within the scope.

docs/rules/hook-use-state.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Rule Details
+
+This rule checks whether the value and setter variables destructured from a `React.useState()` call are named symmetrically.
+
+Examples of **incorrect** code for this rule:
+
+```js
+import React from 'react';
+const useStateResult = React.useState();
```
ljharb, line 13: we probably should update these examples also

lib/rules/hook-use-state.js
```diff
@@ -0,0 +1,120 @@
...
+      url: docsUrl('hook-use-state'),
+    },
+    fixable: 'code',
+    messages: {
+      [USE_STATE_ERROR_MESSAGE]: 'useState call is not destructured into value + setter pair',
+    },
+    schema: [],
+  },
+
+  create: Components.detect((context) => {
```
ljharb, line 32:
```suggestion
  create: Components.detect((context, components, utils) => {
```

lib/rules/hook-use-state.js
```diff
@@ -0,0 +1,120 @@
...
+    let reactImportLocalName;
+    let reactUseStateLocalName;
+
+    return {
+      CallExpression(node) {
+        const isReactUseStateCall = (
```
ljharb, line 38:
```suggestion
        if (utils.isES5Component(node) || utils.isES6Component(node) || utils.isExplicitComponent(node) || utils.isPureComponent(node)) {
          return;
        }
        const isReactUseStateCall = (
```

this makes me want a function in `utils` that covers all the `if` things, like `utils.areHooksInvalid(node)`.

lib/rules/hook-use-state.js
```diff
@@ -0,0 +1,120 @@
+/**
+ * @fileoverview Ensure symmetric naming of useState hook value and setter variables
+ * @author Duncan Beevers
+ */
+
+'use strict';
+
+const Components = require('../util/Components');
+const docsUrl = require('../util/docsUrl');
```
ljharb, line 9:
```suggestion
const docsUrl = require('../util/docsUrl');
const report = require('../util/report');
```

lib/rules/hook-use-state.js
```diff
@@ -0,0 +1,120 @@
...
+          return;
+        }
+
+        const isDestructuringDeclarator = (
+          node.parent.type === 'VariableDeclarator'
+          && node.parent.id.type === 'ArrayPattern'
+        );
+
+        if (!isDestructuringDeclarator) {
+          context.report({ node, messageId: USE_STATE_ERROR_MESSAGE });
```
ljharb, line 64:
```suggestion
          report(
            context,
            messages.useStateErrorMessage,
            'useStateErrorMessage',
            { node },
          );
```

lib/rules/hook-use-state.js
```diff
@@ -0,0 +1,120 @@
...
+'use strict';
+
+const Components = require('../util/Components');
+const docsUrl = require('../util/docsUrl');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+const USE_STATE_ERROR_MESSAGE = 'useStateErrorMessage';
```
ljharb, line 15:
```suggestion
const messages = {
  useStateErrorMessage: 'useState call is not destructured into value + setter pair',
};
```

lib/rules/hook-use-state.js
```diff
@@ -0,0 +1,120 @@
...
+    docs: {
+      description: 'Ensure symmetric naming of useState hook value and setter variables',
+      category: 'Best Practices',
+      recommended: false,
+      url: docsUrl('hook-use-state'),
+    },
+    fixable: 'code',
+    messages: {
+      [USE_STATE_ERROR_MESSAGE]: 'useState call is not destructured into value + setter pair',
+    },
```
ljharb, line 28:
```suggestion
    messages,
```

lib/rules/hook-use-state.js
```diff
@@ -0,0 +1,120 @@
...
+          || variableNodes.length !== 2
+        ) {
+          context.report({
+            node: node.parent.id,
+            messageId: USE_STATE_ERROR_MESSAGE,
+            fix: valueVariableName ? (fixer) => fixer.replaceTextRange(
+              [node.parent.id.range[0], node.parent.id.range[1]],
+              `[${valueVariableName}, ${expectedSetterVariableName}]`
+            ) : undefined,
+          });
```
ljharb, line 97:
```suggestion
          report(
            context,
            messages.useStateErrorMessage,
            'useStateErrorMessage',
            {
              node: node.parent.id,
              fix: valueVariableName ? (fixer) => fixer.replaceTextRange(
                [node.parent.id.range[0], node.parent.id.range[1]],
                `[${valueVariableName}, ${expectedSetterVariableName}]`
              ) : undefined,
            },
          );
```

lib/rules/hook-use-state.js
```diff
@@ -0,0 +1,120 @@
...
+        }
+      },
+      ImportDeclaration(node) {
+        const isReactImported = node.source.type === 'Literal' && node.source.value === 'react';
+        const reactDefaultSpecifier = node.specifiers.find((specifier) => specifier.type === 'ImportDefaultSpecifier');
+        reactImportLocalName = reactDefaultSpecifier ? reactDefaultSpecifier.local.name : undefined;
```
ljharb, line 103: i think this stuff needs to be stored in eslint state, rather than in local variables, since rules can be run with an eslint server in an IDE, or in watch mode via jest-eslint-runner, etc.

ljharb (changes requested):

lib/rules/hook-use-state.js
duncanbeevers: `react-hooks/rules-of-hooks` enforces that hooks are not used in inappropriate contexts. For performance and simplicities sake, I think it would be better not to add these additional scope checks in this rule.

lib/rules/hook-use-state.js
duncanbeevers: I extended the `Components.detect` work to track the names of React imports the same way it currently tracks the list of possible components.
It tracks the default React import as well as all named React imports; I only need the `useState` import for the feature in this PR, but a general approach seemed more appropriate.
The `hook-set-state-names` rule accesses the named imports and searches for the `useState` import.

I isolated the `Components.detect` changes to their own commit and rebased it to the head of this PR.
Let me know if you think this is the right approach, and if so, whether the `Components.detect` change should be its own PR, or should remain on this one.

lib/rules/hook-use-state.js
ljharb: We definitely can not and should not rely on that rule being enforced; performance is wildly unimportant, and i think simplicity would dictate that the rule doesn't run on places where it makes no sense for it to run.

lib/rules/hook-use-state.js
ljharb: "own PR" was the right way :-)

lib/rules/hook-use-state.js
duncanbeevers: I stabbed around a little bit writing a test case for this, to inform and drive the implementation.
To be clear, Is the expected behavior difference from this recommendation something like this:
 * `useState` hook is used in an invalid context, this rule *does not* flag that usage
 * `useState` hook used in a valid context, this rule *does* flag that usage

```js
import React from 'react'
export default class MyClassComponent extends React.Component {
  render() {
    // CURRENTLY flagged as invalid
    //  useState call is not destructured into value + setter pair
    // EXPECTED not flagged
    //  useState call is invoked in a weird context so this rule ignores the invocation
    const result = React.useState()
  }
}
```

I can implement this kind of fencing, but the suggested utils helpers (eg; `isES5Component`, and friends) aren't quite up to the task. We'd need some other helpers (similar to `getStatelessComponent`) to establish the context.

`eslint-plugin-react-hooks` already has some plumbing for detecting this context; maybe we could leverage that work.

Ultimately though, I think the final behavior suggested here isn't really beneficial.
If we flag variable mis-namings in a bad context, so what? I don't feel that means the tool is leading users towards bad behaviors, nor do I feel that we'd actually be protecting them *better* be remaining silent in known-bad scenarios.
Ultimately, I feel the warning should be raised by `eslint-plugin-react-hooks`, and that this rule shouldn't be over-concerned with the hook's invocation context.

ljharb: It means that code in a non-react part of the codebase will get false positives; it's the reason we have so much code doing component detection.

In other words, it SHOULD be up to react and eslint-plugin-react-hooks to make sure hooks only exist in the proper place; our rule should only be warning on valid hooks.

duncanbeevers: I don't think this will trigger false positives because it's already keyed to the names of the React imports. Unless you consider the scenario above a false positive.

duncanbeevers: I provided a scenario above where I thought these additional checks would cause a change in this rule's behavior; in that scenario, I don't think that flagging is a bad call, even if the context is "bad" or otherwise inhospitable to hooks.

Can you help me understand a scenario where this rule might trigger a false positive, now that the React import guardrails are in place?

ljharb: Hmm, I think you might be right.

I suppose if we're certain that `useState` came from React, then it's not ever going to have a false positive, so maybe this is fine as-is. I was stuck on envisioning things for custom hooks, but that doesn't apply here.

ljharb:
Are there other hooks that might want naming guidelines? `useReducer`, for example? If so, i'd rather a single rule to cover both instead of one for each hook.

lib/rules/hook-use-state.js
duncanbeevers: It is still possible to get a false positive when using `useState`, for example, when the imported `useState` is shadowed.

```ts
import { useState } from 'react'
function useState() {
  return 'whatever'
}
function MyComponent() {
  const result = useState() // flagged by hook-set-state-names
}
```

However, checking the is-a-valid-hook-context wouldn't prevent this kind of error.

ljharb: Very true. To handle that, we'll need to use the available variable scope tools to ensure that the variable isn't shadowing the import/require.

duncanbeevers: I added handling and test cases for shadowed `useState` and `React` references.

tests/lib/rules/hook-use-state.js
ljharb: Yes.

ljharb:
Seems pretty good overall.

duncanbeevers:
I see there are quite a few CI failures around various node versions, all related to the new shadowed-references code examples I introduced.

In my local testing I had to use the `no-default` feature in order to avoid the same error while running node 14.17.0, and I expected that same escape hatch would serve for all node versions.

What should my next steps be, if any, in order to address these CI failures?

For now, I'm going to link this in to my day work and see how the rule holds up in day-to-day use on a large codebase.

ljharb:
so like, in this example:
```js
 import React from 'react'
      const React = {
        useState: () => {
          return null
        }
      }

      export default function useColor() {
        const result = React.useState()
      }
```
that's not valid javascript, because of the repeated identifier in a top-level scope. To fix that, it'd need to be shadowed inside a nested scope - like this:
```js
 import React from 'react'
      const React = {
        useState: () => {
          return null
        }
      }

      export default function useColor() {
        { 
        const React = {
          useState: () => {
            return null
          }
        }
        const result = React.useState()
      }
      }
```
but you basically can't shadow an import at the top level in the first place.

duncanbeevers:
Thanks for the suggestion. I haven't had to deal with shadowing in the wild in a LONG time!
After the implementing the update, I was able to remove the `no-default` features and it looks like the tests are much happier. :smile:

duncanbeevers:
I found some usages in our codebase where we use only the first member of `setState` as an ad-hoc memoization.

```ts
const [id] = useState(getUniqueElementId())
```

As it stands, this rule will flag this usage and recommend the following correction
```ts
// Now `setId` is unused
// Maybe this is good when you're first writing the code
const [id, setId] = useState(getUniqueElementId())
```

However, in this case I think the correct change is to directly memoize the value.

```ts
// Now id hook-use-state is happy, but useMemo may not be imported
const id = useMemo(() => getUniqueElementId(), [])
```

### Discussion

Should this rule:
* do the naive change, and add the `setId` member
* attempt the bigger refactor
* something else (ignore, warn-but-not-fix, etc…)

ljharb:
I think it should offer a suggestion to change it to useMemo, but no autofix at all - if someone writes that, we can’t know their intention, so the safe thing to do is force them to manually intervene.

Since you know which AST node imported useState, can the suggestion change that also so it adds useMemo? (one of the no-unused rules can clean up the useState separately, if it becomes unused)

duncanbeevers:
> Since you know which AST node imported useState, can the suggestion change that also so it adds useMemo? (one of the no-unused rules can clean up the useState separately, if it becomes unused)

Yep; it shouldn't even be super difficult since with the new Components react-imports tracking, I can tell whether or not `useMemo` has already been imported.

duncanbeevers:
> Are there other hooks that might want naming guidelines? useReducer, for example? If so, i'd rather a single rule to cover both instead of one for each hook.

I certainly use some naming conventions, but in my experience there's nothing as universally-used as this `value+setter` idiom.

Here are some conventions I use regularly, just as a concrete example of stuff I *would* enforce with a lint rule. These are simply personal preferences, and I don't feel they represent common practice.
* `const myRef = useRef()` – `-Ref` suffix makes it easy to remember to use `myRef.current` when handling this value
* `const onMyEvent = useCallback(() => …` – `on-` prefix (or `handle-` prefix, in some projects) makes it easy to spot functions, and to sort callback props together in JSX Components.

ljharb:
Fair - I'm more thinking that if there were ever to be anything in the future, it'd be nicer to add it as an option to a single hook naming rule than to have to make a whole new rule.

duncanbeevers:
### Config / Other Hooks

Yes, I tried to consider some real-world cases, and how this rule might accommodate those conventions, but I couldn't come up with anything that was both compelling, and that felt amenable to being expressed here alongside this rule.

One thing I really like about this rule as it stands is zero-configuration.

If naming (or other) conventions emerge around other hooks in the future, adding new rules for those hooks while keeping the config for this one clean seems like a reasonable approach. This code supporting this `hook-use-state` rule is already pretty complicated!

### Use in the Wild

I hooked this rule up to a couple of work repositories; as expected I saw a few usages flagged. Somewhat unexpected was that it didn't crash on anything! (*pats self on back*)

The automatic variable renaming works, but falls short of a complete refactor. Here's a small example.

https://user-images.githubusercontent.com/7367/145723395-b9e3a66e-ea0b-410a-9890-c2620a602928.mov

This behavior makes me feel that all the fixes should be changed to suggestions, since I can't reliably rename later usages of the variable.

Maybe there's an easy way to do this properly, but if so I feel that `rename-this-variable-throughout-the-scope` should live in a helper function for other rules to use, and not be implemented as a one-off inside this rule.

### Next Steps

Let me know which way you'd like to see this go. If I don't hear anything right away, I'll probably convert the fixes to suggestions, and leave the larger variable-rename-refactor work for another PR.

ljharb:
I agree such a thing would need to be a shared helper, and i agree that they shouldn’t be autofixes if they’re going to leave things in a broken state. Your plan seems fine.

duncanbeevers:
Okay, this rule is now suggestions-only, including the `useMemo` suggestion (only applied when `useState` is only destructured into a 1-element array, and is called with 1 argument).

Specs are updated, and the internal implementation is simplified since everything is suggestions now (no more fix/suggest split)

duncanbeevers:
I think the next step for this might be actually do the correct variable rename within its scope. It seems like a a chunk of work, but maybe there's something already out there in the eslint ecosystem to accomplish this.

I saw [this comment](https://github.com/eslint/eslint/issues/13706#issuecomment-696667545) about the general technique of expanding the fix range to cover the entire scope of the changed variable.

README.md
```diff
@@ -132,6 +132,7 @@ Enable the rules that you would like to use.
 |  |  | [react/forbid-foreign-prop-types](docs/rules/forbid-foreign-prop-types.md) | Forbid using another component's propTypes |
 |  |  | [react/forbid-prop-types](docs/rules/forbid-prop-types.md) | Forbid certain propTypes |
 |  | 🔧 | [react/function-component-definition](docs/rules/function-component-definition.md) | Standardize the way function component get defined |
+|  | 🔧 | [react/hook-use-state](docs/rules/hook-use-state.md) | Ensure symmetric naming of useState hook value and setter variables |
```
ljharb, line 135: is this still fixable? i don't think suggestions count here

lib/rules/hook-use-state.js
```diff
@@ -0,0 +1,153 @@
...
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Ensure symmetric naming of useState hook value and setter variables',
+      category: 'Best Practices',
+      recommended: false,
+      url: docsUrl('hook-use-state'),
+    },
+    fixable: 'code',
```
ljharb, line 29: is this still true?

lib/rules/hook-use-state.js
```diff
@@ -0,0 +1,153 @@
...
+          const useMemoReactImportSpecifier = namedReactImports
+            && namedReactImports.find((specifier) => specifier.imported.name === 'useMemo');
+
+          suggestions.unshift({
+            desc: 'Replace useState call with useMemo',
+            fix: (fixer) => {
+              const useMemoCode = (useMemoReactImportSpecifier
+                && useMemoReactImportSpecifier.local.name)
+              || (defaultReactImportName
+                && `${defaultReactImportName}.useMemo`)
```
ljharb, line 121:
```suggestion
              || (defaultReactImportName && `${defaultReactImportName}.useMemo`)
```
not stoked here on using logical operators here to select a value; could we use a ternary instead?

lib/rules/hook-use-state.js
```diff
@@ -0,0 +1,153 @@
...
+const docsUrl = require('../util/docsUrl');
+const report = require('../util/report');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+const messages = {
+  useStateErrorMessage: 'useState call is not destructured into value + setter pair',
+
```
ljharb, line 18:
```suggestion
```

tests/lib/rules/hook-use-state.js
ljharb: i thought we decided this should be ignored? or did we decide instead that its placement should be ignored.

lib/rules/hook-use-state.js
duncanbeevers: It'd be nested. I'll break it up into an `if-elseif-else`. Looks like it's also safe to move it out of the scope of the `fix` function.

lib/rules/hook-use-state.js
duncanbeevers: Nope; I've removed all the *fixable* / *fix* references.

tests/lib/rules/hook-use-state.js
duncanbeevers: `eslint-plugin-react-hooks#rules-of-hooks` is almost entirely dedicated to detecting
* [where a hook is being called](https://github.com/facebook/react/blob/3dc41d8a2590768a6ac906cd1f4c11ca00417eee/packages/eslint-plugin-react-hooks/src/RulesOfHooks.js#L91-L105)
* [whether it's called in an appropriate context](https://github.com/facebook/react/blob/3dc41d8a2590768a6ac906cd1f4c11ca00417eee/packages/eslint-plugin-react-hooks/src/RulesOfHooks.js#L351-L354)

I see a few options here.
* *Do nothing*; flag this case as invalid, and rely on `eslint-plugin-react-hooks` to flag the bad usage
* *Do something simple*, like ignore top-level invocations of the hook
* *Do something ambitious* like port (most of) FBs RulesOfHooks over to this repo
* *Do something ambitious* like integrate FBs RulesOfHooks into `util/Components`
* *Do something ambitious* like extract FBs RulesOfHooks into a new module that can be shared by both repos
* *Do something **even more ambitious*** which I haven't thought of.

I'm in favor of *Do nothing*
The price of a mis-flag here is low; neither this rule nor `eslint-plugin-react-hooks#rules-of-hooks` attempt to **fix** this situation (suggestions-only), so it doesn't add any extra fixer passes.

duncanbeevers: I checked out copying over pieces from rules-of-hooks, but its architected a bit differently.

Whereas this rule relies solely on visiting `CallExpression` nodes, `rules-of-hooks`:
 * [accumulates info about `CallExpressions`](https://github.com/facebook/react/blob/3dc41d8a2590768a6ac906cd1f4c11ca00417eee/packages/eslint-plugin-react-hooks/src/RulesOfHooks.js#L516-L528)
 * and then does its heavy-lifting, including checking the `isInsideComponentOrHook` status [when the rule's `onCodePathEnd` callback fires](https://github.com/facebook/react/blob/3dc41d8a2590768a6ac906cd1f4c11ca00417eee/packages/eslint-plugin-react-hooks/src/RulesOfHooks.js#L127-L142).

ljharb:
I rebased this with some tweaks, and there's a failing test - namely, the one where it should ignore hook calls outside of components (unless we agreed not to treat those differently, in which case, please point me to that thread). Thanks!

duncanbeevers:
> i thought we decided this should be ignored? or did we decide instead that its placement should be ignored.

I don't think there was an explicit decision made. I first switched to using `Components.detect` in order to leverage functionality that might make this easier, but there isn't quite enough implemented in that util to get us there.

I wrote up big-ish reply to why I thought we should leave things as-is, but it's not showing up in this view. (I could only see it in the `Files` view on here, not in the `Conversation` view, but maybe you can see the comments in your email or notifications.)

Basically, `eslint-plugin-react-hooks#rules-of-hooks` has the code to determine whether a hook is used in an appropriate scope; we'd have to port a significant chunk of its innards over here to get the same functionality, for what I see as very marginal benefit.

ljharb (approved):

PR #2925, opened 2021-02-16
[Fix] `jsx-no-target-blank`: Allow rel="noreferrer" when `allowReferrer` is true

### LINKED ISSUE #2924: react/jsx-no-target-blank should allow noreferrer when allowReferrer is true
[The spec](https://html.spec.whatwg.org/multipage/links.html#link-type-noreferrer) says that `noreferrer` implies `noopener`, so it should always suffice to write just `rel="noreferrer"`.

However, if you configure rule `react/jsx-no-target-blank` to `allowReferrer`, then links with `rel="noreferrer"` specified fail, even though it's equivalent to `rel="noreferrer noopener"`.

Confusingly, the error message says `Using target="_blank" without rel="noreferrer" is a security risk` which makes it particularly tricky to figure out what's going on.  (In my case, I forgot that I'd turned on `allowReferrer`.)  I imagine it's not possible to configure the error message depending on the configuration?

Related discussion in #2054.

edemaine:
Fix #2924 by allowing either `rel="noreferrer"` or `rel="noopener"` when `allowReferrer` is true.

Also add a missing negative test with `allowReferrer` set to true.


ljharb (approved):
Thanks!

PR #2927, opened 2021-02-18
feat: `jsx-no-newline`: Enforce the removal of newlines around jsx tags

sladyn98:
Enforces a rules that removes empty lines around jsx elements
```
return (

  <button/>

)
```


ljharb (changes requested):
This seems like it'd be simpler as an option to https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-newline.md, rather than an entirely new rule that would conflict if both of the rules were enabled.

sladyn98:
> This seems like it'd be simpler as an option to https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-newline.md, rather than an entirely new rule that would conflict if both of the rules were enabled.

Yeah if that is the case some of the code could be reused. Let me see if I can give it a shot.

lib/rules/jsx-newline.js
```diff
@@ -35,6 +48,27 @@ module.exports = {
...
+                && /\n\s*\n/.test(firstAdjacentSibling.value)
+                ) {
+                  context.report({
+                    node: secondAdjacentSibling,
+                    messageId: 'noNewLine',
+                    fix(fixer) {
+                      return fixer.replaceText(
+                        firstAdjacentSibling,
+                        sourceCode.getText(firstAdjacentSibling)
+                          .replace(/\n/g, '')
```
sladyn98, line 66: @ljharb This regex looks a bit sketchy. I guess it should just be like one `\n` rather than many

lib/rules/jsx-newline.js
```diff
@@ -22,41 +22,68 @@ module.exports = {
...
         jsxElementParents.forEach((parent) => {
-          parent.children.forEach((element, index, elements) => {
-            if (element.type === 'JSXElement' || element.type === 'JSXExpressionContainer') {
-              const firstAdjacentSibling = elements[index + 1];
-              const secondAdjacentSibling = elements[index + 2];
-              if (
-                firstAdjacentSibling
+          if (configuration.removeNewLines) {
+            parent.children.forEach((element) => {
+              if (element.type === 'JSXText') {
```
sladyn98, line 49: @ljharb I need some help here. My logic for detecting new lines was like iterating through the parents and for each child I am checking if they kind of match a new line. My concern is that the new lines will get counted multiple times as the parser moves through the nodes.
For something like this 
```
<div>

       <button></button>
       
</div>
```
I tried printing like for each parent what the new lines was and I got something like this
```
Parent identified
Newline found
Parent identified
Parent identified
Parent identified
Newline found
```
I am confused like how is the newline found after the last div instead it should be found after the last button right ?

tests/lib/rules/jsx-newline.js
```diff
@@ -166,6 +166,21 @@ const tests = {
...
+       <button></button>
+       
+      </div>
+      `,
+      options: [{
+        removeNewLines: true
+      }],
+      errors: [{
+        messageId: 'noNewLine'
+      }]
```
sladyn98, line 183:
```
  1) jsx-newline
       invalid
         
      <div>

       <button></button>
       
      </div>
      :

      + expected - actual

      -0
      +1
      
      at testInvalidTemplate (node_modules/eslint/lib/rule-tester/rule-tester.js:666:24)
      at Context.<anonymous> (node_modules/eslint/lib/rule-tester/rule-tester.js:893:25)
      at processImmediate (internal/timers.js:439:21)
```

sladyn98: Even when the error is being reported it is saying not reported. I am a bit confused here too :/

lib/rules/jsx-newline.js
ljharb: line endings could also be `\r\n` or whichever, i suppose

lib/rules/jsx-newline.js
ljharb: In your example, there's 5 newlines, but 2 blank lines - i think what you want to find is a *blank* line, ie, `\n\s*\n`, not a *newline*, which is `\n`?

tests/lib/rules/jsx-newline.js
ljharb: it's an "invalid" test but it has 0 errors, but it should have 1.

lib/rules/jsx-newline.js
sladyn98: Yeah

PR #2943, opened 2021-03-09
Detect the React and Flow versions relative to the file being linted.

### LINKED ISSUE #2218: Warning: React version not specified when used with monorepo

jcrosetto:
This fixes #2218.  This changes version.js to detect the React and Flow versions relative to the file being linted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,12 +20,14 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 * [`destructuring-assignment`]: fix a false positive for local prop named `context` in SFC ([#2929][] @SyMind)
 * [`jsx-no-target-blank`]: Allow rel="noreferrer" when `allowReferrer` is true ([#2925][] @edemaine)
 * [`boolean-prop-naming`]: add check for typescript "boolean" type ([#2930][] @vedadeepta)
+* version detection: Add tests that verify versioning works for sibling and child projects ([#2943][] @jcrosetto)
 
 ### Changed
 * [Docs] [`jsx-no-constructed-context-values`][]: fix invalid example syntax ([#2910][] @kud)
 * [readme] Replace lists of rules with tables in readme ([#2908][] @motato1)
 * [Docs] added missing curly braces ([#2923][] @Muditxofficial)
 
+[#2943]: https://github.com/yannickcr/eslint-plugin-react/pull/2943
 [#2930]: https://github.com/yannickcr/eslint-plugin-react/pull/2930
 [#2929]: https://github.com/yannickcr/eslint-plugin-react/pull/2929
 [#2925]: https://github.com/yannickcr/eslint-plugin-react/pull/2925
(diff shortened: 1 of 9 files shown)


ljharb (commented):
Thanks! Can you perhaps add some tests without altering the existing ones?

jcrosetto:
You're welcome! I can't wait to get this in. :)

Basically all I did was swap out using `process.cwd()` with `context.getFilename()` when determining the version number. The existing tests needed to be updated because they relied on `process.cwd()` being used. I changed them to work with `context.getFilename()` instead. 

I didn't add any new tests since the changes I made don't really change the expected output. There wasn't anything new or different to test per say. If you can think of something that needs testing I would be happy to add it.

jcrosetto:
That makes sense. I'll put together a test that replicates it.

jcrosetto:
Added tests that verify versions are correctly detected for both sibling and child projects.

ljharb:
I'm still a bit confused; the existing tests use `process.chdir`, but presumably your fix would make them work even _without_ that - so why can't the chdir be preserved?

ljharb (approved):
I tweaked the test a bit to use sinon instead of reassignment, otherwise LGTM!

PR #2949, opened 2021-03-23
[Fix] version detection: support processor virtual filename

### LINKED ISSUE #2948: `lib/util/version.js:31:31` is not compatible with processors with virtual filename
A lot of rules are incompatible.

```log

ESLint: 7.22.0

TypeError: Error while loading rule 'react/display-name': Provided basedir "/Users/JounQin/Workspaces/GitHub/eslint-mdx/README.md" is not a directory, or a symlink to a directory
Occurred while linting /Users/JounQin/Workspaces/GitHub/eslint-mdx/README.md/3_3.js
    at Function.resolveSync [as sync] (/Users/JounQin/Workspaces/GitHub/eslint-mdx/node_modules/eslint-plugin-react/node_modules/resolve/lib/sync.js:91:24)
    at detectReactVersion (/Users/JounQin/Workspaces/GitHub/eslint-mdx/node_modules/eslint-plugin-react/lib/util/version.js:31:31)
    at getReactVersionFromContext (/Users/JounQin/Workspaces/GitHub/eslint-mdx/node_modules/eslint-plugin-react/lib/util/version.js:55:25)
    at Object.testReactVersion (/Users/JounQin/Workspaces/GitHub/eslint-mdx/node_modules/eslint-plugin-react/lib/util/version.js:124:35)
    at usedPropTypesInstructions (/Users/JounQin/Workspaces/GitHub/eslint-mdx/node_modules/eslint-plugin-react/lib/util/usedPropTypes.js:285:48)
    at Function.componentRule (/Users/JounQin/Workspaces/GitHub/eslint-mdx/node_modules/eslint-plugin-react/lib/util/Components.js:966:37)
    at createRuleListeners (/Users/JounQin/Workspaces/GitHub/eslint-mdx/node_modules/eslint/lib/linter/linter.js:761:21)
    at /Users/JounQin/Workspaces/GitHub/eslint-mdx/node_modules/eslint/lib/linter/linter.js:931:31
    at runRules (/Users/JounQin/Workspaces/GitHub/eslint-mdx/node_modules/eslint/lib/linter/linter.js:876:34)
```

Can we add a check like following?

https://github.com/prettier/eslint-plugin-prettier/pull/401/files#diff-e9a90fe7b2d3420b8e5f64e24d5f56ffe6383e47b07461172cc29ed21b768905R79-R94

JounQin:
close #2948

cc @ljharb

### DIFF
--- a/lib/util/version.js
+++ b/lib/util/version.js
@@ -5,6 +5,7 @@
 
 'use strict';
 
+const fs = require('fs');
 const resolve = require('resolve');
 const path = require('path');
 const error = require('./error');
@@ -21,14 +22,37 @@ function resetDetectedVersion() {
   cachedDetectedReactVersion = undefined;
 }
 
+function resolveBasedir(context) {
+  let basedir = process.cwd();
+  if (context) {
+    const filename = context.getFilename();
+    const dirname = path.dirname(filename);
+    try {
+      if (fs.statSync(filename).isFile()) {
+        // dirname must be dir here
+        basedir = dirname;
+      }
+    } catch (err) {
+      // https://github.com/eslint/eslint/issues/11989
+      if (err.code === 'ENOTDIR') {
+        // the error code alreay indicates that dirname is a file
+        basedir = path.dirname(dirname);
+      }
+    }
+  }
+  return basedir;
+}
+
 // TODO, semver-major: remove context fallback
 function detectReactVersion(context) {
   if (cachedDetectedReactVersion) {
     return cachedDetectedReactVersion;
   }
 
+  const basedir = resolveBasedir(context);
+
   try {
-    const reactPath = resolve.sync('react', {basedir: context ? path.dirname(context.getFilename()) : process.cwd()});
+    const reactPath = resolve.sync('react', {basedir});
     const react = require(reactPath); // eslint-disable-line global-require, import/no-dynamic-require
     cachedDetectedReactVersion = react.version;
     return cachedDetectedReactVersion;
@@ -70,8 +94,10 @@ function getReactVersionFromContext(context) {
 
 // TODO, semver-major: remove context fallback
 function detectFlowVersion(context) {
+  const basedir = resolveBasedir(context);
+
   try {
-    const flowPackageJsonPath = resolve.sync('flow-bin/package.json', {basedir: context ? path.dirname(context.getFilename()) : process.cwd()});
+    const flowPackageJsonPath = resolve.sync('flow-bin/package.json', {basedir});
     const flowPackageJson = require(flowPackageJsonPath); // eslint-disable-line global-require, import/no-dynamic-require
     return flowPackageJson.version;
   } catch (e) {
--- a/tests/util/version.js
+++ b/tests/util/version.js
@@ -79,6 +79,14 @@ describe('Version', () => {
         ['Warning: Flow version was set to "detect" in eslint-plugin-react settings, but the "flow-bin" package is not installed. Assuming latest Flow version for linting.']
       ];
     });
+
+    it('works with virtual filename', () => {
+      sinon.stub(context, 'getFilename').callsFake(() => path.resolve(base, 'detect-version-sibling', 'test.js/0_fake.js'));
+
+      assert.equal(versionUtil.testReactVersion(context, '2.3.4'), true);
+      assert.equal(versionUtil.testReactVersion(context, '2.3.5'), false);
+      assert.equal(versionUtil.testFlowVersion(context, '2.92.0'), true);
+    });
   });
 
   describe('string version', () => {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ## Unreleased
 
+### Fixed
+* version detection: support processor virtual filename ([#2949][] @JounQin)
+
 ## [7.23.0] - 2021.03.22
 
 ### Added
@@ -33,6 +36,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 * [readme] Replace lists of rules with tables in readme ([#2908][] @motato1)
 * [Docs] added missing curly braces ([#2923][] @Muditxofficial)
 
+[#2949]: https://github.com/yannickcr/eslint-plugin-react/pull/2949
 [#2943]: https://github.com/yannickcr/eslint-plugin-react/pull/2943
 [#2935]: https://github.com/yannickcr/eslint-plugin-react/pull/2935
 [#2933]: https://github.com/yannickcr/eslint-plugin-react/pull/2933


lib/util/version.js
```diff
@@ -21,14 +21,33 @@ function resetDetectedVersion() {
...
 }
 
+function resolvePkgPath(pkg, context) {
+  const filepath = context && context.getFilename();
+  const basedir = filepath ? path.dirname(filepath) : process.cwd();
+  try {
+    return resolve.sync(pkg, {basedir});
+  } catch (err) {
+    // https://github.com/eslint/eslint/issues/11989
+    if (err.code !== 'INVALID_BASEDIR' || !/[/\\]\d+_[^/\\]*\.[\da-z]+$/i.test(filepath)) {
```
ljharb, line 31: i'm really not confident about this regex. why do we even need to do the `resolve` call if there's a virtual filename?

tests/util/version.js
```diff
@@ -79,6 +79,22 @@ describe('Version', () => {
...
+    it('works with virtual filename', () => {
+      sinon.stub(context, 'getFilename').callsFake(() => path.resolve(base, 'detect-version-sibling', 'test.js/0_fake.js'));
+
+      assert.equal(versionUtil.testReactVersion(context, '2.3.4'), true);
+      assert.equal(versionUtil.testReactVersion(context, '2.3.5'), false);
+      assert.equal(versionUtil.testFlowVersion(context, '2.92.0'), true);
+    });
+
+    it('throws on invalid virtual filename', () => {
+      sinon.stub(context, 'getFilename').callsFake(() => path.resolve(base, 'detect-version-sibling', 'test.js/fake.js'));
```
ljharb, line 92: i'm confused. what makes this invalid, and the other one valid?

tests/util/version.js
```diff
@@ -79,6 +79,22 @@ describe('Version', () => {
         ['Warning: Flow version was set to "detect" in eslint-plugin-react settings, but the "flow-bin" package is not installed. Assuming latest Flow version for linting.']
       ];
     });
+
+    it('works with virtual filename', () => {
+      sinon.stub(context, 'getFilename').callsFake(() => path.resolve(base, 'detect-version-sibling', 'test.js/0_fake.js'));
```
ljharb, line 84: there's nothing about this filename that actually ensures it's "virtual" - this could absolutely be a real file path.

ljharb (changes requested):

lib/util/version.js
JounQin: `{index}_` prefix is hard coded at https://github.com/eslint/eslint/blob/master/lib/linter/linter.js#L1303

tests/util/version.js
JounQin: As above

tests/util/version.js
JounQin: No, a filename can never be a directory.

`test.js` a true filename, there will never be a `test.js/foo` file.

lib/util/version.js
ljharb: so it kind of seems like, if `path.dirname()` exists and is a file, and if the file does *not* exist, then it's a virtual filename?

ljharb:
What I've been playing with is:
```diff
diff --git a/lib/util/version.js b/lib/util/version.js
index c34b7c55..7f99dd96 100644
--- a/lib/util/version.js
+++ b/lib/util/version.js
@@ -7,6 +7,7 @@
 
 const resolve = require('resolve');
 const path = require('path');
+const fs = require('fs');
 const error = require('./error');
 
 let warnedForMissingVersion = false;
@@ -27,8 +28,20 @@ function detectReactVersion(context) {
     return cachedDetectedReactVersion;
   }
 
+  let basedir = process.cwd();
+  if (context) {
+    const filename = context.getFilename();
+    const fileDir = path.dirname(filename);
+    try {
+      const isFile = fs.statSync(filename).isFile();
+      const fileDirIsDir = fs.statSync(fileDir).isDirectory();
+      if (fileDirIsDir && isFile) {
+        basedir = fileDir;
+      }
+    } catch (e) { /**/ }
+  }
   try {
-    const reactPath = resolve.sync('react', {basedir: context ? path.dirname(context.getFilename()) : process.cwd()});
+    const reactPath = resolve.sync('react', {basedir});
     const react = require(reactPath); // eslint-disable-line global-require, import/no-dynamic-require
     cachedDetectedReactVersion = react.version;
     return cachedDetectedReactVersion;
```

tests/util/version.js
JounQin: Yes, and the test here `test.js` is a true file.

lib/util/version.js
JounQin: @ljharb 

If `path.dirname()` exists and is a file, then the file can never exist, because dirname can not be file and dir at the same time.

I just want to ensure it is a virtual filename generated by ESLint itself specifically instead of passing accidentally by call this package's API.

If you think a loose test is fine, the regexp test can be removed, change to test dirname directly.

ljharb: Given that eslint itself is being very loose by providing this virtual filename, i'd rather rely on fs methods to detect when this is happening. Nobody is likely to be using this API but us, so I'm not worried about constraining it.

JounQin:
This a flow version test too, so an extra util function is preferred.

lib/util/version.js
JounQin: OK, I'll simply the implementation.

ljharb:
Indeed; my implementation isn't ideal, i just meant the overall approach of using `fs` methods.

JounQin:
@ljharb Done!

lib/util/version.js
```diff
@@ -21,14 +22,36 @@ function resetDetectedVersion() {
...
+
 // TODO, semver-major: remove context fallback
 function detectReactVersion(context) {
   if (cachedDetectedReactVersion) {
     return cachedDetectedReactVersion;
   }
 
   try {
-    const reactPath = resolve.sync('react', {basedir: context ? path.dirname(context.getFilename()) : process.cwd()});
+    const reactPath = resolve.sync('react', {basedir: resolveBasedir(context)});
```
ljharb, line 54: can `resolveBasedir` be moved up out of this try block?

lib/util/version.js
```diff
@@ -21,14 +22,36 @@ function resetDetectedVersion() {
...
+    } catch (err) {
+      // https://github.com/eslint/eslint/issues/11989
+      if (err.code === 'ENOTDIR') {
+        // the error code alreay indicates that dirname is a file
+        basedir = path.dirname(dirname);
+      }
+    }
+  }
+  // fallback to cwd only if basedir not resolved
+  return basedir || process.cwd();
```
ljharb, line 44: how is this any different than `let basedir = process.cwd()` at the beginning?

lib/util/version.js
JounQin: I don't see any difference. The caught error will be rethrown.

lib/util/version.js
JounQin: `process.cwd()` will only be called when necessary.

ljharb: It costs nothing to call it, though.

lib/util/version.js
ljharb: Code that never throws shouldn't be inside a `try`, or it will run more slowly for no reason.

JounQin: Is there any reference that `try` is `slow`? I'd prefer cleaner codes.

lib/util/version.js
JounQin: Well, less function calls seems better to me.

lib/util/version.js
ljharb: Either way, it's far cleaner to only put "code that can throw" inside of `try` than to avoid calling `process.cwd()`.

lib/util/version.js
ljharb: then every node program would be prohibitively slow.

JounQin:
@ljharb Although I disagree with your opinion, but these changes are just simple, so Done!

ljharb (approved):
Thanks, this looks great

JounQin:
@ljharb Is there a schedule to release a patch version? I think it is a serious issue because the original version breaks.

ljharb:
I've just published v7.23.1.

JounQin:
Thanks, although the CHANGELOG seems incorrect.

image

ljharb:
That's quite correct. There's nothing unreleased at the moment.

JounQin:
I mean `[7.23.1]`, see #2950

ljharb:
ah yes, thanks, a copypaste error.

PR #2953, opened 2021-03-24
fix handling of warnOnSpreadAttributes being false

### LINKED ISSUE #2952: jsx-no-target-blank fails when spread props is before target even when warnOnSpreadAttributes is false
Starting in 7.23.0, the jsx-no-target-blank rule will fail when `warnOnSpreadAttributes` is false and there's a spread before a non-`_blank` target, which is supposed to pass even if `warnOnSpreadAttributes` is true.

Example:
When `warnOnSpreadAttributes` is false (or default):
```javascript
 <a href='some-link' {...otherProps} target='some-non-blank-target'>
```
Would fail, while
```javascript
 <a href='some-link' target='some-non-blank-target' {...otherProps} >
```
would pass. If `warnOnSpreadAttributes` was true, both would fail.

I believe the issue is the `if/elseif` logic here:
https://github.com/yannickcr/eslint-plugin-react/blob/v7.23.1/lib/rules/jsx-no-target-blank.js#L149

Nokel81:
fixes #2952

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,12 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ### Fixed
 * [`jsx-max-depth`]: Prevent getting stuck in circular references ([#2957][] @AriPerkkio)
+* [`jsx-no-target-blank`]: fix handling of `warnOnSpreadAttributes` being false ([#2953][] @Nokel81)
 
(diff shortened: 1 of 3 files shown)


ljharb (approved):
Looks great, thanks!

Nokel81:
I too was glad that it was such a simple fix.

PR #2955, opened 2021-03-25
Object literal also applies to array literal

andreechristian:
Since array is object?


ljharb:
Arrays don't use curly braces, and this rule isn't about objects, it's about curly braces. Arrays use square brackets.

(also, this doesn't have a test)

PR #2961, opened 2021-04-06
forbid-dom-props: support `JSXNamespacedName`

mrtnzlml:
This rather strange pattern is common in [FBT](https://facebook.github.io/fbt/docs/params) and valid according to [JSX](http://facebook.github.io/jsx/) (see `JSXNamespacedName`). Without this change the rule fails on the following error:

```
TypeError: Cannot read property 'toUpperCase' of undefined
```

See: https://github.com/adeira/universe/issues/2005


AriPerkkio:
Related crashes also spotted here https://github.com/AriPerkkio/eslint-remote-tester/issues/29#issuecomment-811961482.

ljharb (approved):
Thanks!

PR #2963, opened 2021-04-10
[Tests]: add weekly scheduled smoke tests

AriPerkkio:
Context: https://github.com/AriPerkkio/eslint-remote-tester/issues/29

### What

Adds weekly scheduled smoke tests to be run on every Sunday at 00:00. Tests can also be trigger manually (`workflow_dispatch`). These tests will be looking for rule crashes.

Previously these tests have been run in `eslint-remote-tester`'s repository. So far bugs below have been found by automated CI runs:

1. `jsx-no-constructed-context-values: Cannot read property 'set' of undefined` yannickcr/eslint-plugin-react#2894
2. `jsx-no-constructed-context-values: Cannot read property 'type' of null` yannickcr/eslint-plugin-react#2895
3. `no-unknown-property: allowedTags.indexOf is not a function` yannickcr/eslint-plugin-react#2879
4. `jsx-max-depth: Maximum call stack size exceeded` yannickcr/eslint-plugin-react#2880
5. `jsx-wrap-multilines: RangeError: Invalid count value` yannickcr/eslint-plugin-react#2875
6. `jsx-no-script-url: Cannot read property 'type' of null` yannickcr/eslint-plugin-react#2871
7. `no-typos: Cannot read property 'toLowerCase' of undefined` yannickcr/eslint-plugin-react#2870
8. `jsx-max-depth: Cannot read property 'length' of undefined` yannickcr/eslint-plugin-react#2869
9. Unreleased: `Add new rule "react/no-referential-type-default-prop"` yannickcr/eslint-plugin-react#2848
10. `forbid-dom-props` https://github.com/yannickcr/eslint-plugin-react/pull/2961#issuecomment-814917710

### How

[Plugin maintainer making sure all existing rules do not crash](https://github.com/AriPerkkio/eslint-remote-tester#plugin-maintainer-making-sure-all-existing-rules-do-not-crash)

Utilizes https://github.com/marketplace/actions/eslint-remote-tester-runner. Each test run will check as many repositories as it can in **5 hours 30 minutes** (default value of `timeLimit`). 
There are 10K repositories in `test/smoke-test-repositories.json`. The first ~3300 are specifically picked for this plugin as they have `/react/` in the repository name. These were collected from [libraries.io](https://github.com/librariesio/libraries.io).

At first the rules will be `extends: ['plugin:react/all']` but it is absolutely OK to include `rules: {}` with settings for specific rules if wanted. 

Available configuration options:
- `eslint-remote-tester-run-action`: https://github.com/marketplace/actions/eslint-remote-tester-runner#action-parameters
- `eslint-remote-tester`: https://github.com/AriPerkkio/eslint-remote-tester#configuration-options

When test fails the @github-actions bot will open new issue with title `Results of weekly scheduled smoke test`. It will reuse the existing issue if previous one hasn't yet been closed. It's recommended to close the issues once their findings are fixed.

🎉

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -1,0 +1,28 @@
+name: Smoke test
+
+on:
+  schedule:
+    - cron: '0 0 * * SUN'
+  workflow_dispatch:
+
+jobs:
+  lint:
+    if: ${{ github.repository == 'yannickcr/eslint-plugin-react' || github.event_name == 'workflow_dispatch' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 14
+      - uses: ljharb/actions/node/install@main
+        name: 'nvm install lts/* && npm install'
+        with:
+          node-version: 'lts/*'
+          skip-ls-check: true
+      - run: |
+          npm link
+          npm link eslint-plugin-react
+      - uses: AriPerkkio/eslint-remote-tester-run-action@v1
+        with:
+          issue-title: 'Results of weekly scheduled smoke test'
+          eslint-remote-tester-config: test/eslint-remote-tester.config.js
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ node_modules
 !tests/**/node_modules
 npm-debug.log
 sftp-config.json
+eslint-remote-tester-results
 
 # Only apps should have lockfiles
 yarn.lock
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "@types/eslint": "=7.2.10",
     "@types/estree": "^0.0.47",
     "@types/node": "^14.14.37",
-    "@typescript-eslint/parser": "^2.34.0",
+    "@typescript-eslint/parser": "^2.34.0 || ^3.10.1",
     "aud": "^1.1.5",
     "babel-eslint": "^8.2.6",
     "eslint": "^3 || ^4 || ^5 || ^6 || ^7",
--- a/package.json
+++ b/package.json
@@ -53,6 +53,8 @@
     "eslint-config-airbnb-base": "^14.2.1",
     "eslint-plugin-eslint-plugin": "^2.3.0 || ^3.3.1",
     "eslint-plugin-import": "^2.23.4",
+    "eslint-remote-tester": "^1.3.0",
+    "eslint-remote-tester-repositories": "^0.0.2",
     "espree": "^3.5.4",
     "istanbul": "^0.4.5",
     "markdown-magic": "^2.0.0",
--- a/.github/workflows/node-4+.yml
+++ b/.github/workflows/node-4+.yml
@@ -62,7 +62,7 @@ jobs:
         name: 'nvm install ${{ matrix.node-version }} && npm install'
         with:
           node-version: ${{ matrix.node-version }}
-          after_install: npm install --no-save "eslint@${{ matrix.eslint }}"
+          after_install: npm install --no-save "eslint@${{ matrix.eslint }}" "@typescript-eslint/parser@${{ matrix.node-version >= 10 && '3' || '2' }}"
           skip-ls-check: true
         env:
           NPM_CONFIG_LEGACY_PEER_DEPS: true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 * [Refactor] `utils/Components`: correct spelling and delete unused code ([#3026][] @ohhoney1)
 * [Docs] [`jsx-uses-react`], [`react-in-jsx-scope`]: document [`react/jsx-runtime`] config ([#3018][] @pkuczynski @ljharb)
 * [Docs] [`require-default-props`]: fix small typo ([#2994][] @evsasse)
+* [Tests] add weekly scheduled smoke tests ([#2963][] @AriPerkkio)
 
 [#3038]: https://github.com/yannickcr/eslint-plugin-react/pull/3038
 [#3036]: https://github.com/yannickcr/eslint-plugin-react/issues/3036
@@ -35,6 +36,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 [#2998]: https://github.com/yannickcr/eslint-plugin-react/pull/2998
 [#2994]: https://github.com/yannickcr/eslint-plugin-react/pull/2994
 [#2992]: https://github.com/yannickcr/eslint-plugin-react/pull/2992
+[#2963]: https://github.com/yannickcr/eslint-plugin-react/pull/2963
 [#1617]: https://github.com/yannickcr/eslint-plugin-react/pull/1617
 [#1547]: https://github.com/yannickcr/eslint-plugin-react/pull/1547
 
--- a/test/eslint-remote-tester.config.js
+++ b/test/eslint-remote-tester.config.js
@@ -1,0 +1,39 @@
+'use strict';
+
+const eslintRemoteTesterRepositories = require('eslint-remote-tester-repositories');
+
+module.exports = {
+  repositories: eslintRemoteTesterRepositories.getRepositories({randomize: true}),
+
+  pathIgnorePattern: eslintRemoteTesterRepositories.getPathIgnorePattern(),
+
+  extensions: ['js', 'jsx', 'ts', 'tsx'],
+
+  concurrentTasks: 3,
+
+  logLevel: 'info',
+
+  /** Optional boolean flag used to enable caching of cloned repositories. For CIs it's ideal to disable caching. Defauls to true. */
+  cache: false,
+
+  eslintrc: {
+    root: true,
+    env: {
+      es6: true
+    },
+    parser: '@typescript-eslint/parser',
+    parserOptions: {
+      ecmaVersion: 2020,
+      sourceType: 'module',
+      ecmaFeatures: {
+        jsx: true
+      }
+    },
+    settings: {
+      react: {
+        version: '16.13.1'
+      }
+    },
+    extends: ['plugin:react/all']
+  }
+};
--- a/tests/fixtures/version/detect-version-missing/node_modules/react/index.js
+++ b/tests/fixtures/version/detect-version-missing/node_modules/react/index.js
@@ -1,0 +1,5 @@
+'use strict';
+
+const error = new Error();
+error.code = 'MODULE_NOT_FOUND';
+throw error;
--- a/tests/fixtures/version/detect-version-missing/node_modules/react/package.json
+++ b/tests/fixtures/version/detect-version-missing/node_modules/react/package.json
@@ -1,0 +1,4 @@
+{
+    "name": "react",
+    "main": "index.js"
+  }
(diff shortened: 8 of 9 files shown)


AriPerkkio:
CI's pretest steps are failing due to eslint version conflicts between dependencies.

```
npm ERR!   peer eslint@"^5.0.0 || ^6.0.0" from @typescript-eslint/parser@2.34.0

npm ERR! Conflicting peer dependency: eslint@7.24.0
npm ERR!   peer eslint@">=7" from eslint-remote-tester@1.1.0
```

The reason for `eslint-remote-tester`'s `eslint@^7` requirement is the usage of new [`ESLint class` Nodejs API](https://eslint.org/docs/developer-guide/nodejs-api#eslint-class) + https://eslint.org/docs/user-guide/migrating-to-7.0.0#deprecate-cliengine.

I'm not sure how to resolve this conflict here.

AriPerkkio:
Second and last conflict: version util test was checking actual installed dependencies. Now that `react` is installed (required by `ink`) the test fails.
Fixed by `[Tests]: version util test to mock "no react installed" case`.

package.json
```diff
@@ -45,13 +45,14 @@
     "@types/eslint": "^7.2.8",
     "@types/estree": "^0.0.47",
     "@types/node": "^14.14.37",
-    "@typescript-eslint/parser": "^2.34.0",
+    "@typescript-eslint/parser": "^3.10.1",
```
ljharb, line 48: this seems like it would affect our ability to test on v2 of the parser. any thoughts/concerns?

AriPerkkio: I've been looking at this [changelog of `2.34.0` -> `3.0.0`](https://github.com/typescript-eslint/typescript-eslint/commits/v3.0.0/packages/parser) and don't think there's anything special in v2 that is being missing after upgrading to v3. Most important thing is the node v8 support being dropped out but I think that's it. But I must say I'm not expert when it comes to AST parsers. 🙃

ljharb:
hm, my other comment got lost. can the "list of repos" be fetched remotely? i don't really want to have to maintain a manual list of other people's repos if i can avoid it. (it's great to be _able_ to override it locally if needed, ofc)

AriPerkkio:
That's something other developers have been asking too. These repositories have been fetched with [a script for libraries.io](https://github.com/AriPerkkio/eslint-remote-tester/blob/master/configs/repository-query-script/fetch-libraries.js) but their service timeouts often and is extremely unstable, https://github.com/librariesio/libraries.io/issues/2647. Maybe I should finally build some kind of centralized solution for this.

Plugin projects could fetch these repositories from a hosted gist.github but that means they would need `node-fetch` or similar request library as dependency. 
Maybe a better solution would be if I maintained a separate npm package which provided these repositories? Updating them would be easy, and overriding a specific repository in plugin project would be possible, e.g.

```js
// eslint-remote-tester.config.js
const repositories = require('eslint-remote-tester-repositories');

module.exports = {
   repositories: repositories.filter(repository => repository !== 'org/some-repository-which-we-dont-want'),
   ...
}
```

I would like to avoid providing these from `eslint-remote-tester` in order to avoid publishing it often, so a separate npm package would be better.

ljharb:
In the github action, it could be fetched with `curl` and then cached using the actions cache? but yes a separate npm package would also be fine.

AriPerkkio:
Now using repositories from [`eslint-remote-tester-repositories` package](https://www.npmjs.com/package/eslint-remote-tester-repositories).

Also a notable change - removed passing `${{ secrets.GITHUB_TOKEN }}`. It is no longer required by `eslint-remote-tester-run-action`. It's using `github.token` internally which is available by default.

.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,23 @@
+name: Smoke test
+
+on:
+  schedule:
+    - cron: '0 0 * * SUN'
+  workflow_dispatch:
```
ljharb, line 6: can we make sure this only runs on the main repo, and not on everyone's fork?

.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v1
+      with:
+        node-version: 14
+    - run: |
+        npm install
+        npm link
+        npm link eslint-plugin-react
```
ljharb, line 19: i'm a bit confused why this is needed - `npm link` alone ensures that the binary is globally available

.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v1
+      with:
+        node-version: 14
+    - run: |
+        npm install
```
ljharb, line 17: i'd prefer to use my node/install action here (which the other actions in the repo also use)
```suggestion
    - uses: ljharb/actions/node/install@main
        name: 'nvm install lts/* && npm install'
        with:
          node-version: 'lts/*'
          skip-ls-check: true
    - run: |
```

.gitignore
```diff
@@ -19,6 +19,8 @@ node_modules
 !tests/**/node_modules
 npm-debug.log
 sftp-config.json
+.cache-eslint-remote-tester
```
ljharb, line 22: any reason this cache isn't kept inside of `node_modules`? that way it gets blown away (as expected) with `npm ci`, and gitignored/eslintignored automatically

package.json
ljharb: That means we wouldn't be able to use it on node 4 and 6 tests :-/

.github/workflows/smoke-test.yml
AriPerkkio: ESLint is unable load the plugin without this. As far as I understand, the `npm link <package>` creates the actual symlink under projects `node_modules`. Does this mean ESLint is unable to use globally available packages?

Here is example run without this line. https://github.com/AriPerkkio/eslint-plugin-react/runs/2477044926?check_suite_focus=true

```
Error: Configuration validation errors:
- eslintrc: Failed to load plugin 'react' declared in 'CLIOptions': Cannot find module 'eslint-plugin-react'
```

.gitignore
AriPerkkio: I think the cache was initially hidden under `node_modules`. However as the cache can grow into tens of GBs it is often needed to be manually cleared in order to get some disk space for local development machines. Having the 10-20GB cache hidden under `node_modules` didn't provide the best _transparency_. Now that it's clearly visible it's easier to remove - in my opinion.

I'll reconsider moving it under `node_modules` but I don't really like that idea.

Since this repository will only run the tool in CI it is not mandatory to add this line to `.gitignore`. But I would recommend to include it just in case.

.github/workflows/smoke-test.yml
AriPerkkio: The `on.schedule` does not support similar `branches` option as `push|pull_request` but it can be limited by checking the repository from github context. 

```yml
jobs:
  lint:
    if: ${{ github.repository == 'yannickcr/eslint-plugin-react' }}
    runs-on: ubuntu-latest
```

Example run here https://github.com/AriPerkkio/eslint-plugin-react/runs/2477243324?check_suite_focus=true

package.json
AriPerkkio: That's definitely a drawback. Not sure how to resolve this. 😕 

A bit off-topic but just to clarify - how was the `v2.34.0` be able to run on node 4? It [requires node v8 + eslint v5](https://github.com/typescript-eslint/typescript-eslint/blob/v2.34.0/packages/parser/package.json#L12-L14). ESLint v5 [requires node 6](https://github.com/eslint/eslint/blob/5.x/package.json#L125-L127). Am I reading these requirements correctly?

.github/workflows/smoke-test.yml
ljharb: ahhh yes, because eslint `require`s it.

in that case, an alternative to "link to global, then link back down" is if we add to devDeps `"eslint-plugin-react": "link:./"`, perhaps?

.gitignore
ljharb: i agree it should be in gitignore, but also it seems like it should be automatically pruned by the tool without the need for manual removal?

package.json
ljharb: actually you're right; the tests are set up to skip the TS tests on node < 5 (or maybe <= 5).

(this project supports down to eslint 3, which requires node 4)

.github/workflows/smoke-test.yml
AriPerkkio: `"link:./"` seems to fail but `"file:./"` works. No need to run link steps in workflow now. https://docs.npmjs.com/cli/v7/configuring-npm/package-json#local-paths

```
"eslint-plugin-react": "file:./",
```

AriPerkkio: Oh actually I'll need to revert this change. Some older node versions seem to fail. Some fail in unit-test step, some fail in installation step.

https://github.com/yannickcr/eslint-plugin-react/actions/runs/803786800

.gitignore
AriPerkkio: Issue created https://github.com/AriPerkkio/eslint-remote-tester/issues/100. Thanks for the improvement idea. 👍 

The current location of the cache does not block this PR.

package.json
AriPerkkio: Alright so previously with `v2.34.0` the TS tests have been run for `eslint@^5` + `node@^8`. 
Now the `v3` supports `eslint@^5` and `node@^10`. The only thing that is lost is being able to test the typescript parser against node v8. ESLint version is still the same. That doesn't sound too bad, right?

.github/workflows/smoke-test.yml
ljharb: seems it behaves differently in npm 6 vs 7 :-/

.gitignore
AriPerkkio: Cache is now located at `node_modules/eslint-remote-tester/.cache-eslint-remote-tester`. Removed this line from `.gitignore`.

package.json
ljharb: I think we can fix this by depending on `^2.34.0 || ^3.10.1`, and in the CI config, manually installing v2 when it's node < 10.

AriPerkkio: Now using `"@typescript-eslint/parser": "^2.34.0 || ^3.10.1"` and CI seems to be passing. Does it still require changes to CI workflow configurations? 

Should the version be forced here?

https://github.com/yannickcr/eslint-plugin-react/blob/f233eb77c5c55662287991a914f3a7e2c4172db0/.github/workflows/node-4%2B.yml#L65

AriPerkkio:
FYI, while testing this setup with latest `master` it seems that some **unreleased** changes are causing a lot of crashes for `no-typos` rule. Also `no-unstable-nested-components` crashed at least once. Could be related to #3001. 

Test has been running for 35mins and found hundreds of crashes: `[INFO/STATUS] Errors (617)`.

package.json
ljharb: Yes, I think that's precisely the solution that's needed :-)

AriPerkkio: CI will now resolve version based on `matrix.node`: `"@typescript-eslint/parser@${{ matrix.node-version >= 10 && '3' || '2' }}"`.
The "ternary" syntax here is a bit weird but it works.

Logs from jobs:
- `latest majors (10, 7)`: `* AFTER_INSTALL: npm install --no-save "eslint@7" "@typescript-eslint/parser@3"`
- `latest majors (9, 6)`: `* AFTER_INSTALL: npm install --no-save "eslint@6" "@typescript-eslint/parser@2"`

ljharb (approved):
Awesome, thanks!

PR #2965, opened 2021-04-13
[Fix] version detection: support recursive processor virtual filename

JounQin:
continue to work on #2949

---

According to https://github.com/eslint/eslint/pull/14227#discussion_r602802758, virtual filename could be recursive, this PR fix this possible issue.


ljharb (approved):
if we're sure there's no chance of an infinite loop here, lgtm

JounQin:
> if we're sure there's no chance of an infinite loop here, lgtm

Sure, that error indicates the filename is a virtual filename and what means a true filename must exist.

JounQin:
@ljharb Any luck to merge?

JounQin:
@ljharb Any luck to merge?

PR #2967, opened 2021-04-14
[New] `jsx-no-useless-fragment`: add `ignoreUnsafeChildren` flag

sbdchd:
With `ignoreUnsafeChildren` enabled, this rule will only warn about
useless fragments that removing would guarantee no change in behavior.

addresses some of: https://github.com/yannickcr/eslint-plugin-react/issues/2584

### DIFF
--- a/docs/rules/jsx-no-useless-fragment.md
+++ b/docs/rules/jsx-no-useless-fragment.md
@@ -57,14 +57,14 @@ const cat = <>meow</>
 
 ```js
 ...
-"react/jsx-no-useless-fragments": [<enabled>, { "ignoreNeedsMoreChildren": <boolean> }]
+"react/jsx-no-useless-fragments": [<enabled>, { "ignoreUnsafeChildren": <boolean> }]
 ...
 ```
 
-### `ignoreNeedsMoreChildren` (default: `false`)
+### `ignoreUnsafeChildren` (default: `false`)
 
 When `true` the rule will ignore errors related to fragments having enough
-children.
+children when removing the fragment could result in a runtime error.
 
 Examples of **correct** code for this rule:
 
--- a/docs/rules/jsx-no-useless-fragment.md
+++ b/docs/rules/jsx-no-useless-fragment.md
@@ -52,3 +52,28 @@ const cat = <>meow</>
 
 <Fragment key={item.id}>{item.value}</Fragment>
 ```
+
+## Rule Options
+
+```js
+...
+"react/jsx-no-useless-fragments": [<enabled>, { "ignoreNeedsMoreChildren": <boolean> }]
+...
+```
+
+### `ignoreNeedsMoreChildren` (default: `false`)
+
+When `true` the rule will ignore errors related to fragments having enough
+children.
+
+Examples of **correct** code for this rule:
+
+```jsx
+<></>
+
+<>{children}</>
+
+<>{foo && <Foo/>}</>
+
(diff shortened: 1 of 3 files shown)


docs/rules/jsx-no-useless-fragment.md
```diff
@@ -52,3 +52,28 @@ const cat = <>meow</>
 
 <Fragment key={item.id}>{item.value}</Fragment>
 ```
+
+## Rule Options
+
+```js
+...
+"react/jsx-no-useless-fragments": [<enabled>, { "ignoreNeedsMoreChildren": <boolean> }]
```
sbdchd, line 60: I think the `ignoreNeedsMoreChildren` name could use some work, open to ideas

sbdchd: maybe something along the lines of `ignorePossiblyUnsafeExprs`?

ljharb (changes requested):
Per https://github.com/yannickcr/eslint-plugin-react/issues/2584#issuecomment-595978999, the intention was to add an option to only allow a *prop value*, and nothing else, to contain an otherwise useless fragment - because that's the only case where removing the fragment could in fact cause a runtime error.

sbdchd:
@ljharb which fragments should we warning about?

```jsx
// function could be expecting a react node and raise an error at runtime if it got null instead
<></>
// children could be undefined
<>{children}</>
// foo could be undefined
<>{foo && <Foo/>}</>
// foo could be undefined
<>{foo?.map(x => <Bar x={x}/>)}</>
```

ljharb:
In that case, you should be doing:
```jsx
<></> // fair enough
// children could be undefined
{children ?? <></>}
// foo could be undefined
{foo ? <Foo/> : <></>}
// foo could be undefined
{foo?.map(x => <Bar x={x}/>) ?? <></>}
```

In general, if a component is so badly written that it crashes when it receives a nullish value instead of an element, i don't think it makes much sense for the consumer to need to worry about it. I also think that's an appropriate place for an eslint override comment, where you explain that the override is necessary because the other component is incompetently authored.

fregante:
I think this would solve the only issue I have with the rule.

JSX.Element, React.Element and React.ReactNode are not interchangeable types, and however you pull them you end up with type errors.

If you have a `React.FunctionComponent` type for example you just cannot return a raw `children`, but you can return `<>{children}</>`

While `<>` is technically useless, TypeScript disagrees

```
Type '({ children, LoginPage, ignoreApiError, }: PropsWithChildren<RequireAuthProps>) => ReactNode' is not assignable to type 'FC<RequireAuthProps>'.
  Type 'ReactNode' is not assignable to type 'ReactElement<any, any>'.
    Type 'string' is not assignable to type 'ReactElement<any, any>'

ljharb:
@fregante can you elaborate (maybe with a ts playground link) on what you're hoping this PR would fix?

fregante:
I haven't looked at the code, but I assume `ignoreUnsafeChildren` means that `<>{children}<>` or `<>{props.children}</>` would be allowed, without allowing _all_ expressions via `allowExpressions`

The setting might make more sense as `allowExpressions: "children"`

I think the entirety of https://github.com/jsx-eslint/eslint-plugin-react/issues/2584 (except https://github.com/jsx-eslint/eslint-plugin-react/issues/2584#issuecomment-627002431) has this single complaint, so you have a lot more examples there.

PR #2976, opened 2021-04-27
Reduce memory allocation in rules and utils

willheslam:
Hello!
eslint-plugin-react is an amazing plugin - it catches so many problems that I consider it indispensable :) 

Recently on a large-ish project I noticed some of the rules running relatively slowly, so I decided to see if I could speed them up.
In the end I was able to find quite a few opportunities to do less work, mainly by avoiding memory allocations.

I'd really appreciate any feedback on better ways to benchmark the plugin, and how safe these optimisations are.
They pass all the tests, though I appreciate it's not always possible to be 100% confident changes are correct, especially given the large number of changes here - luckily most of them are relatively localised.
I've tried to avoid changing the existing logic as much as possible, though sometimes speed has been picked over readability - I'd appreciate any thoughts on that too.
I've tried to make sure the changes will work on node 4, but that probably needs more thorough testing.

There's a list of the changes summarised at the bottom of this message. Thanks!

---

Benchmark results:

The project I benchmarked this against using `cloc` registers as about 550 files, with about 50k lines of code in .tsx files all containing React components.
The changes knock off about 70 - 100ms off the slowest rules, roughly 30% faster, with a sample result here:
```
Rule                                       | Time (ms) | Relative
:------------------------------------------|----------:|--------:
react/jsx-indent                           |   365.723 |     5.8%
react/no-deprecated                        |   296.861 |     4.7%
react/boolean-prop-naming                  |   286.611 |     4.6%
react/destructuring-assignment             |   283.235 |     4.5%
react/no-unstable-nested-components        |   172.198 |     2.7%
react/no-access-state-in-setstate          |   169.871 |     2.7%
react/jsx-no-bind                          |   161.568 |     2.6%
react/no-direct-mutation-state             |   155.857 |     2.5%
react/prefer-stateless-function            |   150.361 |     2.4%
react/display-name                         |   147.024 |     2.3%
react/no-string-refs                       |   144.832 |     2.3%
react/jsx-sort-props                       |   142.020 |     2.3%
react/no-typos                             |   141.018 |     2.2%
react/default-props-match-prop-types       |   140.545 |     2.2%
react/void-dom-elements-no-children        |   139.404 |     2.2%
```
to
```
Rule                                       | Time (ms) | Relative
:------------------------------------------|----------:|--------:
react/jsx-indent                           |   264.386 |     6.4%
react/boolean-prop-naming                  |   176.753 |     4.3%
react/destructuring-assignment             |   146.563 |     3.6%
react/jsx-sort-props                       |   142.513 |     3.5%
react/no-access-state-in-setstate          |   117.366 |     2.8%
react/no-deprecated                        |   109.879 |     2.7%
react/no-direct-mutation-state             |   107.110 |     2.6%
react/void-dom-elements-no-children        |    98.467 |     2.4%
react/display-name                         |    97.823 |     2.4%
react/no-unstable-nested-components        |    95.287 |     2.3%
react/jsx-no-bind                          |    94.966 |     2.3%
react/default-props-match-prop-types       |    92.893 |     2.3%
react/sort-comp                            |    91.416 |     2.2%
react/prefer-stateless-function            |    86.666 |     2.1%
react/no-typos                             |    84.784 |     2.1%
```

I used the following .eslintrc.js:
```
module.exports = {
  root: true,
  parser: "@typescript-eslint/parser",
  settings: {
    react: {
      pragma: "React",
      fragment: "Fragment",
      version: "detect",
    },
  },
  extends: ["plugin:react/all"],
}
```
with the versions:
```
    "eslint": "^7.25.0",
    "eslint-plugin-react": "^7.23.2",
    "@typescript-eslint/eslint-plugin": "^4.22.0",
    "@typescript-eslint/parser": "^4.22.0",
```
across the .tsx files in the project, 200 times by running:
```
$ script timings
$ for i in {1..200}; do TIMING=100 npx eslint . --ext tsx -o eslintlog; done
````
for both `7.23.2` and then with the `lib` directory swapped in from this branch.

I then averaged all 200 runs in this google sheet:
https://docs.google.com/spreadsheets/d/1Y5QZ_tsAyXrzgw5Rre21Gmzl95jT9irZmxfwSeF1oI0/edit?usp=sharing

---
Optimisations:

- Avoid allocating empty objects or arrays when possible
- Combine filter -> forEach chains
- Replace Object.keys with values/entries where possible
- Extract deeply nested functions to avoid reallocation
- Push expensive operations below guards / early returns
- Promote shared or mostly-constant conditionals
- Swap compound boolean expressions to proriotise cheap short circuiting
- Avoid dynamically generating RegExps when possible
- Lazily initialise and cache complex objects and necessarily dynamic RegExps
- Move expensive and shared calculations to higher scopes
- Replace string splits and trims with RegExps
- Avoid redundant array concatenations
- Split expensive conditional checks across multiple guards
- Move expensive booleans into conditional expressions to exploit short circuiting
- Swap out filters for finds where possible
- Turn arrays into objects/map/dictionary lookups where possible
- Swap array function orders where safe and efficient
- Cache string transformations

### DIFF
--- a/lib/rules/boolean-prop-naming.js
+++ b/lib/rules/boolean-prop-naming.js
@@ -227,7 +227,9 @@ module.exports = {
       if (!node || !Array.isArray(args)) {
         return;
       }
-      args.filter((arg) => arg.type === 'ObjectExpression').forEach((object) => validatePropNaming(node, object.properties));
+      args.forEach((object) => object.type === 'ObjectExpression'
+        && validatePropNaming(node, object.properties)
+      );
     }
 
     // --------------------------------------------------------------------------
--- a/lib/rules/boolean-prop-naming.js
+++ b/lib/rules/boolean-prop-naming.js
@@ -5,6 +5,7 @@
 
 'use strict';
 
+const values = require('object.values');
 const Components = require('../util/Components');
 const propsUtil = require('../util/props');
 const docsUrl = require('../util/docsUrl');
@@ -312,18 +313,18 @@ module.exports = {
         }
 
         const list = components.list();
-        Object.keys(list).forEach((component) => {
+        values(list).forEach((component) => {
           // If this is a functional component that uses a global type, check it
           if (
             (
-              list[component].node.type === 'FunctionDeclaration'
-              || list[component].node.type === 'ArrowFunctionExpression'
+              component.node.type === 'FunctionDeclaration'
+              || component.node.type === 'ArrowFunctionExpression'
             )
-            && list[component].node.params
-            && list[component].node.params.length
-            && list[component].node.params[0].typeAnnotation
+            && component.node.params
+            && component.node.params.length
+            && component.node.params[0].typeAnnotation
           ) {
-            const typeNode = list[component].node.params[0].typeAnnotation;
+            const typeNode = component.node.params[0].typeAnnotation;
             const annotation = typeNode.typeAnnotation;
             let propType;
             if (annotation.type === 'GenericTypeAnnotation') {
@@ -336,14 +337,14 @@ module.exports = {
 
             if (propType) {
               validatePropNaming(
-                list[component].node,
+                component.node,
                 propType.properties || propType.members
               );
             }
           }
 
-          if (list[component].invalidProps && list[component].invalidProps.length > 0) {
-            reportInvalidNaming(list[component]);
+          if (component.invalidProps && component.invalidProps.length > 0) {
+            reportInvalidNaming(component);
           }
         });
 
--- a/lib/rules/boolean-prop-naming.js
+++ b/lib/rules/boolean-prop-naming.js
@@ -174,17 +174,17 @@ module.exports = {
      * @param {Function} addInvalidProp callback to run for each error
      */
     function runCheck(proptypes, addInvalidProp) {
-      proptypes = proptypes || [];
-
-      proptypes.forEach((prop) => {
-        if (config.validateNested && nestedPropTypes(prop)) {
-          runCheck(prop.value.arguments[0].properties, addInvalidProp);
-          return;
-        }
-        if (flowCheck(prop) || regularCheck(prop) || tsCheck(prop)) {
-          addInvalidProp(prop);
-        }
-      });
+      if (proptypes && proptypes.length > 0) {
+        proptypes.forEach((prop) => {
+          if (config.validateNested && nestedPropTypes(prop)) {
+            runCheck(prop.value.arguments[0].properties, addInvalidProp);
+            return;
+          }
+          if (flowCheck(prop) || regularCheck(prop) || tsCheck(prop)) {
+            addInvalidProp(prop);
+          }
+        });
+      }
     }
 
     /**
--- a/lib/rules/default-props-match-prop-types.js
+++ b/lib/rules/default-props-match-prop-types.js
@@ -94,11 +94,13 @@ module.exports = {
         const list = components.list();
 
         // If no defaultProps could be found, we don't report anything.
-        values(list).filter((component) => component.defaultProps).forEach((component) => {
-          reportInvalidDefaultProps(
-            component.declaredPropTypes,
-            component.defaultProps || {}
-          );
+        values(list).forEach((component) => {
+          if (component.defaultProps) {
+            reportInvalidDefaultProps(
+              component.declaredPropTypes,
+              component.defaultProps || {}
+            );
+          }
         });
       }
     };
--- a/lib/rules/default-props-match-prop-types.js
+++ b/lib/rules/default-props-match-prop-types.js
@@ -6,6 +6,7 @@
 
 'use strict';
 
+const values = require('object.values');
 const Components = require('../util/Components');
 const docsUrl = require('../util/docsUrl');
 
@@ -93,10 +94,10 @@ module.exports = {
         const list = components.list();
 
         // If no defaultProps could be found, we don't report anything.
-        Object.keys(list).filter((component) => list[component].defaultProps).forEach((component) => {
+        values(list).filter((component) => component.defaultProps).forEach((component) => {
           reportInvalidDefaultProps(
-            list[component].declaredPropTypes,
-            list[component].defaultProps || {}
+            component.declaredPropTypes,
+            component.defaultProps || {}
           );
         });
       }
--- a/lib/rules/jsx-indent-props.js
+++ b/lib/rules/jsx-indent-props.js
@@ -138,8 +138,8 @@ module.exports = {
      */
     function getNodeIndent(node) {
       let src = context.getSourceCode().getText(node, node.loc.start.column + extraColumnStart);
-      const lines = src.split('\n');
-      src = lines[0];
+      const matches = src.match(/^(.*)\n/);
+      src = matches ? matches[1] : src;
 
       let regExp;
       if (indentType === 'space') {
@@ -150,14 +150,16 @@ module.exports = {
 
       const indent = regExp.exec(src);
       const useOperator = /^([ ]|[\t])*[:]/.test(src) || /^([ ]|[\t])*[?]/.test(src);
-      const useBracket = /^([ ]|[\t])*[<]/.test(src);
 
       line.currentOperator = false;
       if (useOperator) {
         line.isUsingOperator = true;
         line.currentOperator = true;
-      } else if (useBracket) {
-        line.isUsingOperator = false;
+      } else {
+        const useBracket = /^([ ]|[\t])*[<]/.test(src);
+        if (useBracket) {
+          line.isUsingOperator = false;
+        }
       }
 
       return indent ? indent[0].length : 0;
--- a/lib/rules/jsx-indent.js
+++ b/lib/rules/jsx-indent.js
@@ -277,13 +277,11 @@ module.exports = {
      */
     function checkNodesIndent(node, indent, excludeCommas) {
       const nodeIndent = getNodeIndent(node, false, excludeCommas);
-      const isCorrectRightInLogicalExp = isRightInLogicalExp(node) && (nodeIndent - indent) === indentSize;
-      const isCorrectAlternateInCondExp = isAlternateInConditionalExp(node) && (nodeIndent - indent) === 0;
       if (
         nodeIndent !== indent
         && astUtil.isNodeFirstInLine(context, node)
-        && !isCorrectRightInLogicalExp
-        && !isCorrectAlternateInCondExp
+        && !(isRightInLogicalExp(node) && (nodeIndent - indent) === indentSize)
+        && !(isAlternateInConditionalExp(node) && (nodeIndent - indent) === 0)
       ) {
         report(node, indent, nodeIndent);
       }
--- a/lib/rules/jsx-indent.js
+++ b/lib/rules/jsx-indent.js
@@ -149,20 +149,19 @@ module.exports = {
       excludeCommas = excludeCommas || false;
 
       let src = context.getSourceCode().getText(node, node.loc.start.column + extraColumnStart);
-      const lines = src.split('\n');
+      let matches;
       if (byLastLine) {
-        src = lines[lines.length - 1];
+        matches = src.match(/.*\n(.*)$/);
       } else {
-        src = lines[0];
+        matches = src.match(/^(.*)\n/);
       }
-
-      const skip = excludeCommas ? ',' : '';
+      src = matches ? matches[1] : src;
 
       let regExp;
       if (indentType === 'space') {
-        regExp = new RegExp(`^[ ${skip}]+`);
+        regExp = excludeCommas ? /^[ ,]+/ : /^[ ]+/;
       } else {
-        regExp = new RegExp(`^[\t${skip}]+`);
+        regExp = excludeCommas ? /^[\t,]+/ : /^[\t]+/;
       }
 
       const indent = regExp.exec(src);
--- a/lib/rules/jsx-no-useless-fragment.js
+++ b/lib/rules/jsx-no-useless-fragment.js
@@ -19,7 +19,7 @@ function isJSXText(node) {
  * @returns {boolean}
  */
 function isOnlyWhitespace(text) {
-  return text.trim().length === 0;
+  return !/\S/.test(text);
 }
 
 /**
--- a/lib/rules/jsx-no-useless-fragment.js
+++ b/lib/rules/jsx-no-useless-fragment.js
@@ -27,7 +27,7 @@ function isOnlyWhitespace(text) {
  * @returns {boolean}
  */
 function isNonspaceJSXTextOrJSXCurly(node) {
-  return (isJSXText(node) && !isOnlyWhitespace(node.raw)) || node.type === 'JSXExpressionContainer';
+  return node.type === 'JSXExpressionContainer' || (isJSXText(node) && !isOnlyWhitespace(node.raw));
 }
 
 /**
--- a/lib/rules/no-deprecated.js
+++ b/lib/rules/no-deprecated.js
@@ -24,6 +24,16 @@ const MODULES = {
   'react-addons-perf': ['ReactPerf', 'Perf']
 };
 
+const VERSION_MAP = {
+  '0.12.0': [0, 12, 0],
+  '0.13.0': [0, 13, 0],
+  '0.14.0': [0, 14, 0],
+  '15.0.0': [15, 0, 0],
+  '15.5.0': [15, 5, 0],
+  '15.6.0': [15, 6, 0],
+  '16.9.0': [16, 9, 0]
+};
+
 // ------------------------------------------------------------------------------
 // Rule Definition
 // ------------------------------------------------------------------------------
@@ -47,7 +57,13 @@ module.exports = {
   create: Components.detect((context, components, utils) => {
     const pragma = pragmaUtil.getFromContext(context);
 
+    const reactVersion = versionUtil.getReactVersionFromContext(context);
+
+    let deprecatedCache;
     function getDeprecated() {
+      if (deprecatedCache !== undefined) {
+        return deprecatedCache;
+      }
       const deprecated = {};
       // 0.12.0
       deprecated[`${pragma}.renderComponent`] = ['0.12.0', `${pragma}.render`];
@@ -101,6 +117,7 @@ module.exports = {
         'https://reactjs.org/docs/react-component.html#unsafe_componentwillupdate. '
         + 'Use https://github.com/reactjs/react-codemod#rename-unsafe-lifecycles to automatically update your components.'
       ];
+      deprecatedCache = deprecated;
       return deprecated;
     }
 
@@ -108,10 +125,9 @@ module.exports = {
       const deprecated = getDeprecated();
 
       return (
-        deprecated
-        && deprecated[method]
-        && deprecated[method][0]
-        && versionUtil.testReactVersion(context, deprecated[method][0])
+        // eslint-disable-next-line no-prototype-builtins
+        deprecated.hasOwnProperty(method) // necessary to avoid 'constructor'
+        && versionUtil.isLessThan(VERSION_MAP[deprecated[method][0]], reactVersion)
       );
     }
 
@@ -141,8 +157,10 @@ module.exports = {
         return moduleName;
       }
 
+      const matchName = (name) => name === node.init.name;
+
       values(MODULES).some((moduleNames) => {
-        moduleName = moduleNames.find((name) => name === node.init.name);
+        moduleName = moduleNames.find(matchName);
         return moduleName;
       });
 
--- a/lib/rules/no-typos.js
+++ b/lib/rules/no-typos.js
@@ -4,7 +4,8 @@
 
 'use strict';
 
-const PROP_TYPES = Object.keys(require('prop-types'));
+const PROP_TYPES = require('prop-types');
+const fromEntries = require('object.fromentries');
 const Components = require('../util/Components');
 const docsUrl = require('../util/docsUrl');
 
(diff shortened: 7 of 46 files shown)

### CODE REVIEW

File: lib/rules/boolean-prop-naming.js [hunk 1]
```diff
@@ -226,7 +227,9 @@ module.exports = {
       if (!node || !Array.isArray(args)) {
         return;
       }
-      args.filter((arg) => arg.type === 'ObjectExpression').forEach((object) => validatePropNaming(node, object.properties));
+      args.forEach((object) => object.type === 'ObjectExpression'
+        && validatePropNaming(node, object.properties)
+      );
```
ljharb, line 232: how meaningful is this specific change? the engine should be able to optimize this down to the equivalent for loop.

lib/rules/default-props-match-prop-types.js
```diff
@@ -93,11 +94,13 @@ module.exports = {
         const list = components.list();
 
         // If no defaultProps could be found, we don't report anything.
-        Object.keys(list).filter((component) => list[component].defaultProps).forEach((component) => {
```
ljharb, line 96: same with this .filter change to an if check inside forEach?

ljharb (changes requested):
It might be easier to review if this was split up into a separate commit for each conceptual kind of change - since some of them are no-brainers (like the Object.keys -> values) and others are more of a tradeoff between clarity and performance.

lib/rules/boolean-prop-naming.js
willheslam: Good question.
I'm not convinced v8 can optimise a filter and forEach into a for loop - some (admittedly slapdash and micro) benchmarking with node v14.15.3 iterating many times over an array shows a singular forEach with a built-in conditional as reliably faster than a filter, forEach chain:
```
$ ITERATIONS=100 SIZE=10000 node arraytest-filter-for-each.js
filter, forEach: time taken 58.374 ms
$ ITERATIONS=100 SIZE=10000 node arraytest-just-for-each.js
just forEach: time taken 43.283 ms
$ ITERATIONS=100 SIZE=10000 node arraytest-for-of-loop.js
for of loop: time taken 41.101 ms
```
The singular forEach is consistently faster - depending on the arrangement of the code seems to be around 20 -> 40% faster.
Going with more iterations but a smaller array shows similar results too, just in case:
```
$ ITERATIONS=10000 SIZE=100 node arraytest-filter-for-each.js
filter, forEach: time taken 45.790 ms
$ ITERATIONS=10000 SIZE=100 node arraytest-just-for-each.js
just forEach: time taken 40.437 ms
$ ITERATIONS=10000 SIZE=100 node arraytest-for-of-loop.js
for of loop: time taken 36.737 ms
```
Here's the gist:
https://gist.github.com/willheslam/d45a1d183d5d778d82c5113ffb8a4ab8

Whether this specific change is meaningful, I don't honestly have concrete answer - it depends on the frequency the function is called, the average size of `args` (and how many of them fulfil the condition)
and also how much the surrounding environment is generating memory pressure.

Benchmarking each individual change is difficult, given the noise and iteration time and how some effects of memory allocation are only obvious in concert, so as a blanket rule tried to find any quick, localised opportunities to reduce memory allocation, where the change didn't massively impact the flow of the code.

willheslam:
> It might be easier to review if this was split up into a separate commit for each conceptual kind of change - since some of them are no-brainers (like the Object.keys -> values) and others are more of a tradeoff between clarity and performance.

Thanks for suggesting this - I've now split this PR into 12 commits, roughly across the kinds of changes made, or, if hard to disentangle, across the individual modules that have been modified.

lib/util/linkComponents.js
```diff
@@ -12,7 +12,9 @@ const DEFAULT_LINK_ATTRIBUTE = 'href';
 function getLinkComponents(context) {
   const settings = context.settings || {};
   const linkComponents = /** @type {typeof DEFAULT_LINK_COMPONENTS} */ (
-    DEFAULT_LINK_COMPONENTS.concat(settings.linkComponents || [])
+    settings.linkComponents
+      ? DEFAULT_LINK_COMPONENTS.concat(settings.linkComponents)
+      : DEFAULT_LINK_COMPONENTS
```
ljharb, line 17: repeating `DEFAULT_LINK_COMPONENTS` seems unfortunate; is creating an empty array really that costly?

lib/util/propWrapper.js
```diff
@@ -5,21 +5,24 @@
...
-  const propWrapperFunctions = getPropWrapperFunctions(context);
-  const splitName = name.split('.');
-  return Array.from(propWrapperFunctions).some((func) => {
-    if (splitName.length === 2 && func.object === splitName[0] && func.property === splitName[1]) {
-      return true;
-    }
-    return name === func || func.property === name;
-  });
+  const propWrapperFunctions = context.settings.propWrapperFunctions;
+  if (propWrapperFunctions && propWrapperFunctions.length > 0) {
```
ljharb, line 16: is this check needed? calling `[].some()` immediately returns `false`.

lib/rules/jsx-indent-props.js
```diff
@@ -150,14 +150,16 @@ module.exports = {
...
-      const useBracket = /^([ ]|[\t])*[<]/.test(src);
 
       line.currentOperator = false;
       if (useOperator) {
         line.isUsingOperator = true;
         line.currentOperator = true;
-      } else if (useBracket) {
-        line.isUsingOperator = false;
+      } else {
+        const useBracket = /^([ ]|[\t])*[<]/.test(src);
```
ljharb, line 159: can we make this regex at module level, rather than nesting it deeper? (same with the useOperator regexes)

lib/rules/jsx-indent.js
```diff
@@ -149,20 +149,19 @@ module.exports = {
...
       let src = context.getSourceCode().getText(node, node.loc.start.column + extraColumnStart);
-      const lines = src.split('\n');
+      let matches;
       if (byLastLine) {
-        src = lines[lines.length - 1];
+        matches = src.match(/.*\n(.*)$/);
       } else {
-        src = lines[0];
+        matches = src.match(/^(.*)\n/);
       }
```
ljharb, line 157:
```suggestion
      const matches = byLastLine ? src.match(/.*\n(.*)$/) : src.match(/^(.*)\n/);
```

lib/rules/jsx-no-useless-fragment.js
```diff
@@ -19,7 +19,7 @@ function isJSXText(node) {
  * @returns {boolean}
  */
 function isOnlyWhitespace(text) {
-  return text.trim().length === 0;
+  return !/\S/.test(text);
```
ljharb, line 22: these aren't necessarily always the same thing; `trim` can be polyfilled to be more correct, but regexes can't be.

Additionally, is this better than `/^[^\S]*$/.test(text)`, which might be clearer?

lib/rules/no-unstable-nested-components.js
```diff
@@ -35,7 +35,7 @@ function generateErrorMessageWithParentName(parentName) {
  * @returns {Boolean}
  */
 function startsWithRender(text) {
-  return (text || '').startsWith('render');
+  return text !== undefined && text.startsWith('render');
```
ljharb, line 38:
```suggestion
  return text && text.startsWith('render');
```
since `null` and `false` and a few other non-undefined things are falsy.

lib/util/Components.js
```diff
@@ -27,7 +27,12 @@ function usedPropTypesAreEquivalent(propA, propB) {
     if (!propA.allNames && !propB.allNames) {
       return true;
     }
-    if (Array.isArray(propA.allNames) && Array.isArray(propB.allNames) && propA.allNames.join('') === propB.allNames.join('')) {
+    if (Array.isArray(propA.allNames)
+      && Array.isArray(propB.allNames)
+      && (
+        (propA.allNames === propB.allNames)
+        || propA.allNames.join('') === propB.allNames.join('')
+      )) {
```
ljharb, line 35:
```suggestion
    if (
      Array.isArray(propA.allNames)
      && Array.isArray(propB.allNames)
      && (
        propA.allNames === propB.allNames
        || propA.allNames.join('') === propB.allNames.join('')
      )
    ) {
```

lib/util/Components.js
```diff
@@ -27,7 +27,12 @@ function usedPropTypesAreEquivalent(propA, propB) {
     if (!propA.allNames && !propB.allNames) {
       return true;
     }
-    if (Array.isArray(propA.allNames) && Array.isArray(propB.allNames) && propA.allNames.join('') === propB.allNames.join('')) {
+    if (Array.isArray(propA.allNames)
+      && Array.isArray(propB.allNames)
+      && (
+        (propA.allNames === propB.allNames)
```
ljharb, line 33: when will this check ever actually be true?

lib/util/Components.js
```diff
@@ -173,35 +173,41 @@ class Components {
...
-    Object.keys(thisList).filter((i) => thisList[i].confidence < 2).forEach((i) => {
-      let component = null;
-      let node = null;
-      node = thisList[i].node;
-      while (!component && node.parent) {
-        node = node.parent;
-        // Stop moving up if we reach a decorator
-        if (node.type === 'Decorator') {
-          break;
+    for (const val of values(thisList)) {
```
ljharb, line 183: i am very skeptical that `for..of` is ever _faster_ than alternatives, and i have a very strong preference to avoid loops. I'd really prefer this to remain a filter + forEach, unless you can persuade me that the clarity loss is worth the speedup..

lib/util/Components.js
```diff
@@ -173,35 +173,41 @@ class Components {
...
-    });
+    }
 
     // Assign used props in not confident components to the parent component
-    Object.keys(thisList).filter((j) => thisList[j].confidence >= 2).forEach((j) => {
-      const id = getId(thisList[j].node);
-      list[j] = thisList[j];
-      if (usedPropTypes[id]) {
-        list[j].usedPropTypes = mergeUsedPropTypes(list[j].usedPropTypes || [], usedPropTypes[id]);
+    for (const j of Object.keys(thisList)) {
```
ljharb, line 208: same here

lib/util/Components.js
```diff
@@ -41,15 +41,18 @@ function usedPropTypesAreEquivalent(propA, propB) {
...
 
 function mergeUsedPropTypes(propsList, newPropsList) {
-  const propsToAdd = [];
-  newPropsList.forEach((newProp) => {
-    const newPropisAlreadyInTheList = propsList.some((prop) => usedPropTypesAreEquivalent(prop, newProp));
-    if (!newPropisAlreadyInTheList) {
-      propsToAdd.push(newProp);
+  if (propsList === undefined || propsList.length === 0) {
+    return newPropsList;
+  }
```
ljharb, line 46: when would `propsList` ever be undefined? the previous code assumed it was always an array.

lib/util/Components.js
```diff
@@ -403,9 +404,11 @@ function componentRule(rule, context) {
               && requireExpression.callee
               && requireExpression.callee.name === 'require'
               && requireExpression.arguments[0]
-              && requireExpression.arguments[0].value === pragma.toLocaleLowerCase()
             ) {
-              return true;
+              pragmaLowerCase = pragmaLowerCase || pragma.toLocaleLowerCase();
```
ljharb, line 408: if you're looking for perf, why do an unconditional assignment?
```suggestion
              if (!pragmaLowerCase) { pragmaLowerCase = pragma.toLocaleLowerCase(); }
```

note as well that using `if` actually matches the semantics of `||=`.

lib/util/Components.js
```diff
@@ -415,9 +418,11 @@ function componentRule(rule, context) {
           if (
             latestDef.parent
             && latestDef.parent.type === 'ImportDeclaration'
-            && latestDef.parent.source.value === pragma.toLocaleLowerCase()
           ) {
-            return true;
+            pragmaLowerCase = pragmaLowerCase || pragma.toLocaleLowerCase();
```
ljharb, line 422: same here

lib/util/isFirstLetterCapitalized.js
```diff
@@ -9,8 +9,7 @@ function isFirstLetterCapitalized(word) {
   if (!word) {
     return false;
   }
-  const firstLetter = word.charAt(0);
-  return firstLetter.toUpperCase() === firstLetter;
+  return /^[A-Z]/.test(word);
```
ljharb, line 12: a-z aren't the only letters that can be capitalized; i don't think this is a correct change.

tests/util/isFirstLetterCapitalized.js
```diff
@@ -20,4 +20,11 @@ describe('isFirstLetterCapitalized', () => {
     assert.equal(isFirstLetterCapitalized('IsCapitalized'), true);
     assert.equal(isFirstLetterCapitalized('UPPERCASE'), true);
   });
+
+  it('should return false for non-letters', () => {
```
ljharb, line 24: you haven't checked non-english language characters. see https://stackoverflow.com/a/38123020/632724

lib/rules/no-typos.js
```diff
@@ -69,7 +83,7 @@ module.exports = {
     }
 
     function checkValidPropType(node) {
-      if (node.name && !PROP_TYPES.some((propTypeName) => propTypeName === node.name)) {
+      if (node.name && !(node.name in PROP_TYPES)) {
```
ljharb, line 86: why not use Sets here rather than `in`? also, using `in` here ensures we'll get false positives for things like `toString`, or anything else on Object.prototype.

lib/rules/no-unsafe.js
```diff
@@ -73,22 +73,13 @@ module.exports = {
...
-
     /**
      * Checks if a passed method is unsafe
      * @param {string} method Life cycle method
      * @returns {boolean} Returns true for unsafe methods, otherwise returns false
      */
     function isUnsafe(method) {
-      const unsafeMethods = getUnsafeMethods();
-      return unsafeMethods.indexOf(method) !== -1;
+      return (method in unsafe);
```
ljharb, line 82: same here

lib/rules/no-deprecated.js
```diff
@@ -141,8 +157,10 @@ module.exports = {
         return moduleName;
       }
 
+      const matchName = (name) => name === node.init.name;
```
ljharb, line 160: might as well hoist this to module level

lib/util/version.js
```diff
@@ -127,34 +136,45 @@ function getFlowVersionFromContext(context) {
...
+  }
+  const higherMinor = a[0] === b[0] && a[1] < b[1];
+  if (higherMinor) {
+    return true;
+  }
+  const higherOrEqualPatch = a[0] === b[0]
+    && a[1] === b[1]
+    && a[2] <= b[2];
+  return higherOrEqualPatch;
 }
```
ljharb, line 155: this logic doesn't seem like it will work for prerelease versions. i think we need to use the `semver` package for this.

lib/rules/no-unstable-nested-components.js
AriPerkkio, line 38: I'm specifically interested in changes related to this rule. Thanks for testing the performance @willheslam!

While this improvement is definitely welcome I'm having a hard time believing it actually speeds up this rule by ~40%.
Is it really that slow to create new Strings? This method is not even accessed that often. 
Why shouldn't other Strings also be declared once in module scope, e.g. `render`, `CallExpression`?

> Rule                                       | Time (ms) | Relative
> :------------------------------------------|----------:|--------:
> react/no-unstable-nested-components        |   172.198 |     2.7%
> 
> to
> 
> Rule                                       | Time (ms) | Relative
> :------------------------------------------|----------:|--------:
> react/no-unstable-nested-components        |    95.287 |     2.3%

lib/util/Components.js
willheslam: From my testing, most of the time.
This ended up being a strict optimisation above and beyond the (expensive) joins.
Ideally, we'd probably be better off swapping out the join comparison for some kind of fast array of strings equality check as it's still run a good chunk of the time.

Please check my logic below:

With this basic benchmarking on a project with about 550 jsx files:
```
let conditionChecked = 0
let arraysDoHaveIdentityEquality = 0
let arraysDoNotHaveIdentityEquality = 0
function usedPropTypesAreEquivalent(propA, propB) {
  if (propA.name === propB.name) {
    if (!propA.allNames && !propB.allNames) {
      return true;
    }
    if (Array.isArray(propA.allNames) && Array.isArray(propB.allNames)) {
      conditionChecked ++
    }
    if (Array.isArray(propA.allNames) && Array.isArray(propB.allNames) && propA.allNames.join('') === propB.allNames.join('')) {
      if(propA.allNames === propB.allNames) {
        arraysDoHaveIdentityEquality ++
      } else {
        arraysDoNotHaveIdentityEquality ++
      }
      console.log('Arrays have identity equality:', arraysDoHaveIdentityEquality)
      console.log('Arrays do not have identity equality', arraysDoNotHaveIdentityEquality)
      console.log('Ratio', arraysDoHaveIdentityEquality / arraysDoNotHaveIdentityEquality)
      console.log('Total times condition reached', arraysDoHaveIdentityEquality + arraysDoNotHaveIdentityEquality)
      console.log("Condition checked", conditionChecked)
      return true;
    }
    return false;
  }
  return false;
}
```

With an otherwise unaltered lib/util/Components.js,
I sometimes got a ratio of arrays having identity equality of 4 to 1 when the condition had been checked nearly 80,000 times:
```
Arrays have identity equality: 60555
Arrays do not have identity equality 14723
Ratio 4.112952523262922
Total times condition reached 75278
Condition checked 78661
```

but this settled down to about 2 to 1 after the condition had been checked almost a million times.
```
Arrays have identity equality: 572805
Arrays do not have identity equality 291390
Ratio 1.9657675280551838
Total times condition reached 864195
Condition checked 990195
```

lib/util/Components.js
willheslam: On a large project with 500 files, from some very quick checking, the `list` function runs around 23,000 times so my intuition says it's worth looking at, but I'll do some specific benchmarking of this change to check.

lib/util/Components.js
willheslam: All the callsites for `mergedUsedPropTypes` have been reworked to avoid passing in empty arrays, e.g.
```
        usedPropTypes: mergeUsedPropTypes(component.usedPropTypes || [], props.usedPropTypes || [])
```
into
```
        usedPropTypes: mergeUsedPropTypes(component.usedPropTypes, props.usedPropTypes) || []
```

This means less array copies, and we can do the least amount of work - if you're calling `mergeUsedPropTypes(['a', 'b', 'c'], undefined)` it'll directly return the first arg, instead of creating an extra copy.

lib/util/Components.js
willheslam: Good catch, I'll fix this.

lib/util/isFirstLetterCapitalized.js
willheslam: Good catch - I was too dependent on the existing unit tests for this function.
It'd be great if we could avoid doing some string manipulations here, so it could be that we can use the regex as a fast path, but preserve the chatAt & toUpperCase as a slow path.

lib/rules/no-typos.js
willheslam: Good catch on false positives with `in`, I definitely hadn't considered that.

Set looks like a good fit here!

lib/util/version.js
willheslam: Interesting - it's equivalent (or should be - will double check this) to what was already here - note the removed lines from function `test` below.

Happy to consider using `semver`, but I was hoping to keep this PR functionally equivalent and just a pure performance boost - what do you think?

willheslam (commented):
Responded to comments - thanks for these, very good points.

Will fix what has been pointed out and do some more specific benchmarking to double check some changes.

lib/rules/jsx-no-useless-fragment.js
ljharb: I guess I'm not convinced that `.trim()` is meaningfully slower than using a custom regex. Do you have data to support that?

lib/util/Components.js
ljharb: I guess I'm surprised that the AST parser would generate two memoized `allNames` arrays for different nodes.

lib/util/version.js
ljharb: I mean, we could add `semver` prior to this PR to correct the behavior, if you prefer :-p

lib/util/linkComponents.js
ljharb: Creating a copy of an array should be basically free.

lib/rules/jsx-indent-props.js
ljharb: I'm relatively sure that's not unique to ES5, that's eternally been the semantics in JS.

lib/rules/no-unstable-nested-components.js
AriPerkkio: Right, that makes perfect sense. Those utilities are perfect spot for optimization as they are being used widely across the code base. 
That's impressive improvement! 💯

lib/rules/jsx-no-useless-fragment.js
willheslam: Sorry about the late reply!

I was really convinced a regex would be faster because there'd be less work (stop the moment non-whitespace is found, rather than doing the work of constructing a new string, then checking its length), but after constructing some basic micro benchmarks to check it, they're basically equivalent in speed, and in Chrome 91 + latest V8 sometimes trim was faster.

Thanks for pushing back on this, I'm glad it was double checked - will revert this change.

ljharb:
@willheslam are you interested in continuing this effort? It seems like there's a number of comments to address, and a few potential PRs to land separately, plus this needs a rebase.

willheslam:
@ljharb Sorry for disappearing on this, work + life got in the way and I ran out of time.

Yeah - I've got some time this week so I'll take a look at the changes I've made to address the comments, changes pending, and what the rebase will entail.

I appreciate your following up on this!

PR #2979, opened 2021-04-30
Add jsx-sort-props ignoreTags option

fracz:
This PR adds an ability to set ignored tags when checking the `jsx-sort-props` rule.

Generally, we like the idea of sorting props but in some cases it does not make sense. For example, when we use the `Can` component from [React CASL](https://casl.js.org/v4/en/package/casl-react):

```jsx
<Can I="create" a="post">
```

The rule forces us to sort props.

```jsx
<Can a="post" I="create">
```

We'd love to have an option to ignore some tags - which I propose here.

```
ignoreTags: ["Can"]
```

Tests and docs provided. The tag name detection code inside the rule has been "borrowed" from [jsx-props-no-spreading](https://github.com/yannickcr/eslint-plugin-react/blob/f24c353aca1225dbdd0d75a8e1c02d6b711b94c9/lib/rules/jsx-props-no-spreading.js).

### DIFF
--- a/docs/rules/jsx-sort-props.md
+++ b/docs/rules/jsx-sort-props.md
@@ -98,6 +98,16 @@ With `reservedFirst: ["key"]`, the following will **not** warn:
 <Hello key={'uuid'} name="John" ref="ref" />
 ```
 
+### `ignoreTags`
+
+An array of element (tag) names to ignore when checking the properties order.
+
+With `ignoreTags: ["Can"]`, the following will **not** warn:
+
+```jsx
+<Can i="create" a="post" />
+```
+
 ## When not to use
 
 This rule is a formatting preference and not following it won't negatively affect the quality of your code. If alphabetizing props isn't a part of your coding standards, then you can leave this rule off.
--- a/lib/rules/jsx-sort-props.js
+++ b/lib/rules/jsx-sort-props.js
@@ -249,6 +249,9 @@ module.exports = {
         },
         reservedFirst: {
           type: ['array', 'boolean']
+        },
+        ignoreTags: {
+          type: 'array'
         }
       },
       additionalProperties: false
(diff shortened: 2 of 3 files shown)


ljharb:
oof, that sounds like a very strange API. The ordering of differently-named props is definitely not supposed to matter, but I see the whimsical (but ruby-like rather than JS-like) approach they're trying for here.

docs/rules/jsx-sort-props.md
```diff
@@ -98,6 +98,16 @@ With `reservedFirst: ["key"]`, the following will **not** warn:
 <Hello key={'uuid'} name="John" ref="ref" />
 ```
 
+### `ignoreTags`
+
+An array of element (tag) names to ignore when checking the properties order.
```
ljharb, line 103: these aren't tag names, they're custom component names, so the terminology needs tweaking

docs/rules/jsx-sort-props.md
```diff
@@ -98,6 +98,16 @@ With `reservedFirst: ["key"]`, the following will **not** warn:
...
+
+With `ignoreTags: ["Can"]`, the following will **not** warn:
+
+```jsx
+<Can i="create" a="post" />
```
ljharb, line 108: i think rather than ignoring, it would be better to specify a custom ordering that applies to a specific custom component name. What do you think?

lib/rules/jsx-sort-props.js
```diff
@@ -249,6 +249,9 @@ module.exports = {
         },
         reservedFirst: {
           type: ['array', 'boolean']
+        },
+        ignoreTags: {
+          type: 'array'
```
ljharb, line 254: this would need to ensure the items are unique, and are valid custom component names

ljharb (changes requested):

docs/rules/jsx-sort-props.md
fracz: Is it necessary to be a custom component name? The rule is not limited to sort props on the components only. It currently sorts the attributes of "normal" HTML elements, too. If you accept the ignore option, it should be possible to whitelist normal tags, too.

docs/rules/jsx-sort-props.md
fracz: Well, it makes sense in the case of the CASL's Can component. 

I can also think about using this in [FontAwesome](https://fontawesome.com/how-to-use/on-the-web/using-with/react) components (I'd like to have the `fixedWitdth` after `icon` because it is better to read then: `<FontAwesomeIcon icon="coffee" fixedWidth />`.

However, I'm reluctant because it will be less generic (uou don't always know all possible props that can be used with an element). Moreover, this will make the configuration much more complicated.

lib/rules/jsx-sort-props.js
fracz: Agree as for the uniqueness. Not sure about validity of component names - see my comment above.

fracz:
> The ordering of differently-named props is definitely not supposed to matter

It does not matter - component works regardless of the order. But in some cases - it is better to read like "an english sentence".

docs/rules/jsx-sort-props.md
ljharb: This is a very very rare use case; I'm not aware of anything in the entire JS ecosystem besides CASL that has a props API like this. I think it's totally fine to be not generic, since I very much doubt there will ever be a second use case (where the order is sentence-like).

In the case of FontAwesome, just ignoring it wouldn't enforce your preference, it'd just remove enforcement - which is strictly less useful, even if the alternative would require more config.

ljharb (changes requested):
See https://github.com/yannickcr/eslint-plugin-react/pull/2979/files#r624281044; ignoring is not very useful, instead, specifying a custom prop ordering for an arbitrary list of custom component names seems better. Given that this CASL thing seems like the only instance of this pattern in existence, I'd prefer something generic that also allows for enforcement, rather than just ignoring.

PR #2992, opened 2021-05-25
Improve component detection

### LINKED ISSUE #2960: `destructuring-assignment` shouldn't trigger on method calls
The rule `react/destructuring-assignment` is also triggered when using methods of params.

```jsx
      formatter: (val) => {
        return <span>{val.toPrecision(2)}</span>;
      },
```

image

Might also be related to #1950

Wesitos:
+ Added [estraverse](https://github.com/estools/estraverse) as dependency.
+ Implemented function `traverseReturns` using `estraverse` to improve return value analysis.
+ Rewrote `isReturningJSX` function to handle return statements nested inside other structures (`IfConditions`, `SwitchStatements`, etc)
+ This should also fix #2960

### DIFF
--- a/lib/util/ast.js
+++ b/lib/util/ast.js
@@ -4,6 +4,29 @@
 
 'use strict';
 
+const estraverse = require('estraverse');
+
+/**
+ * Wrapper for estraverse.traverse
+ *
+ * @param {ASTNode} ASTnode The AST node being checked
+ * @param {Object} visitor Visitor Object for estraverse
+ */
+function traverse(ASTnode, visitor) {
+  const opts = Object.assign({}, {
+    fallback(node) {
+      return Object.keys(node).filter((key) => key === 'children' || key === 'argument');
+    }
+  }, visitor);
+
+  opts.keys = Object.assign({}, visitor.keys, {
+    JSXElement: ['children'],
+    JSXFragment: ['children']
+  });
+
+  estraverse.traverse(ASTnode, opts);
+}
+
 /**
  * Find a return statment in the current node
  *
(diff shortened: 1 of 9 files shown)


lib/util/ast.js
```diff
@@ -4,6 +4,37 @@
...
+    fallback(node) {
+      return Object.keys(node).filter((key) => {
+        switch (key) {
+          case 'children':
+          case 'argument':
+            return true;
+          default:
+            return false;
+        }
+      });
```
ljharb, line 26: switch statements are awful, let's go with this instead:
```suggestion
      return Object.keys(node).filter((key) => key === 'children' || key === 'argument');
```

lib/util/ast.js
```diff
@@ -37,6 +68,49 @@ function findReturnStatement(node) {
...
+    case 'ArrowFunctionExpression':
+      if (ASTnode.expression) {
+        enterFunc(ASTnode.body);
+        return;
+      }
+      break;
+    default:
+      // We expect only to receive function nodes
+      return;
+  }
```
ljharb, line 96:
```suggestion
  if (ASTNode.type === 'ReturnStatement') {
    enterFunc(ASTnode);
  } else if (ASTNode.type === 'ArrowFunctionExpression') {
    if (ASTNode.expression) {
      enterFunc(ASTnode.body);
    }
  } else if (ASTNode.type === 'ReturnStatement') {
    enterFunc(ASTnode);
  } else if (ASTNode.type !== 'FunctionExpression' && ASTNode.type !== 'FunctionDeclaration' && ASTNode.type !== 'MethodDefinition') {
    throw new TypeError('only function nodes are expected');
  }
```

ljharb (changes requested):
Thanks, hopefully estraverse runs in all the eslint and node versions we support.

lib/util/ast.js
Wesitos: I refactored this with "If-return"s. I think it's a bit cleaner than a chain of 'else if".

`isReturningJSX` (lib/util/Components.js) sometimes is called with `ClassDeclaration` nodes. Now only a few `no-typos` tests were broken by this (I fixed it). IMHO, doing nothing here is safer than throwing an exception.

ljharb (approved):
I really like the code deletion; I'm nervous about the delegation of responsibility; I like farming responsibility out to a dep; that all existing tests pass makes me confident.

Issue #2956, opened 2021-03-29
Optional Chaining report destructuring-assignment


yoyo837:
`"eslint": "7.23.0"`
`"eslint-plugin-react": "7.23.1"`

```jsx
render: val => {
  //
  //
  //
  //
  if (val?.url) {
    return (
      <a href={val.url} rel="noopener noreferrer" target="_blank">
        {val.fileName || 'PDF'}
      </a>
    );
  }
  return null;
},
```
Screenshot:

![image](https://user-images.githubusercontent.com/6134547/112786322-9124eb80-9088-11eb-9c86-9f1854e3dfbe.png)

However, if change it to the following, there is no such problem as above:

```jsx
render: val => {
  //
  //
  //
  //
  if (val?.url) {
    return (
      <a href={val.url} rel="noopener noreferrer" target="_blank">
        {val.fileName || 'PDF'}
      </a>
    );
  }
  // return null; Comment this line
},
```
Screenshot:

![image](https://user-images.githubusercontent.com/6134547/112786511-fe388100-9088-11eb-8fb4-ee59a03d3733.png)

ljharb:
I'm a bit confused. Why do you have a render _arrow function_, ever?

What's the context here? Is this a class component? Class components' render functions never, ever get called with a single argument.

yoyo837:
I’m sorry. The background is to use a react ui library component, which has a render api. https://ant.design/components/table/#components-table-demo-basic

ljharb:
I'm afraid that doesn't help me. A component either is a function, or is a class that has a `render` method. `render` takes no arguments.

Can you please provide the entire code that surrounds this `render` function? Like, the entire contents of the file?

yoyo837:
Does this link help you: https://codesandbox.io/s/ll33o ?

ljharb:
Yes, thank you. In that case, those `render` functions shouldn't be detected as components, despite returning jsx, since they're not defined as part of a createClass or a class component.

randyshoopman:
I'm also seeing this issue.  Wasn't a problem in 7.22.0.  Appeared when upgrading to 7.23.1

PR #3001, opened 2021-06-05
[Fix] destructuring-assignment, no-multi-comp, `no-unstable-nested-components`, component detection: improve component detection

vedadeepta:
Fixes: #2956

Description: 

Example code that shows incorrect error:
```
const columns = [
  {
    render: (val) => {
      if (val.url) {
        return (
          <a href={val.url} rel="noopener noreferrer" target="_blank">
            {val.fileName || 'PDF'}
          </a>
        );
      }
      return null;
    },
  },
];
```

Error shows up because `getParentComponent` calls `getParentStatelessComponent` that calls  [`getStatelessComponent`](https://github.com/yannickcr/eslint-plugin-react/blob/faadccb65d9a4d813fe58efc86293f97447a96da/lib/util/Components.js#L637) on the `ArrowFunctionExpression` (val =>...) in the example code above which incorrectly returns the same `ArrowFunctionExpression` instead of `undefined`.

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -651,6 +651,8 @@ function componentRule(rule, context) {
           return undefined;
         }
         if (utils.isInAllowedPositionForComponent(node) && utils.isReturningJSXOrNull(node)) {
+          if (utils.isParentComponentNotStatelessComponent(node)) return undefined;
+
           const isMethod = node.parent.type === 'Property' && node.parent.method;
 
           if (isMethod && !isFirstLetterCapitalized(node.parent.key.name)) {
@@ -798,6 +800,18 @@ function componentRule(rule, context) {
 
       // Return the component
       return components.add(componentNode, 1);
+    },
+
+    isParentComponentNotStatelessComponent(node) {
+      return (
+        node.parent
+        && node.parent.key
+        && node.parent.key.type === 'Identifier'
+        // custom component functions must start with a capital letter (returns false otherwise)
+        && node.parent.key.name.charAt(0) === node.parent.key.name.charAt(0).toLowerCase()
+        // react render function cannot have params
+        && !!(node.params || []).length
+      );
     }
   };
 
@@ -846,6 +860,7 @@ function componentRule(rule, context) {
         components.add(node, 0);
         return;
       }
+
       const component = utils.getParentComponent();
       if (
         !component
@@ -863,6 +878,7 @@ function componentRule(rule, context) {
         components.add(node, 0);
         return;
       }
+
       node = utils.getParentComponent();
       if (!node) {
         return;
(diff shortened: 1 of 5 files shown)


lib/util/Components.js
```diff
@@ -798,6 +800,17 @@ function componentRule(rule, context) {
...
       return components.add(componentNode, 1);
+    },
+
+    isArrowOrFunctionExpressionRenderWithParams(node) {
+      return (
+        node.parent
+        && node.parent.key
+        && node.parent.key.type === 'Identifier'
+        && node.parent.key.name === 'render'
+        // react render function cannot have params
```
ljharb, line 811:
```suggestion
        && node.parent.key.name.charAt(0) === node.parent.key.name.charAt(0).toLowerCase()
        // custom component functions must start with a capital letter
```
what about something like this? there's nothing inherentky magical about the word "render".

vedadeepta: Yeah, it makes sense, thanks

vedadeepta: Now that I test it - it breaks tests like these

no-unstable-nested-components

```
      function ParentComponent() {
        const rows = [
          {
            name: 'A',
            notPrefixedWithRender: (props) => <Row {...props} />
          },
        ];

        return <Table rows={rows} />;
      }
```
This should return an error since `notPrefixedWithRender:` is an unstable nested component but does not since according to this new logic proposed here  `notPrefixedWithRender:` is not a component. 

This test can be re-written but IMO a lot of external libraries like ant-d, mui have API like above, so it can introduce breaking changes.

```
<some-lowercase-firstletter>Anything: () => jsx
```

ljharb: That’s not a component at all - it’s a render prop. Components must start with a capital letter.

ljharb: Specifically, in react, anything that starts with a lowercase letter is an html element, not a component.

ljharb (approved):

yoyo837:
Thanks for your work.

ThiefMaster:
Not sure if it's caused by the changes in this PR, but after updating fro 7.24 to 7.26, I noticed a bunch of false positives showing up in my code:

```
.../indico/modules/events/editing/client/js/editing/timeline/util.js
  55:14  warning  Must use destructuring revision assignment  react/destructuring-assignment

.../indico/modules/rb/client/js/common/map/reducers.js
  59:9  warning  Must use destructuring action assignment  react/destructuring-assignment
  64:9  warning  Must use destructuring action assignment  react/destructuring-assignment
```

- https://github.com/indico/indico/blob/676dc3d1fede9bd4417982e79310753691be2472/indico/modules/rb/client/js/common/map/reducers.js#L59
- https://github.com/indico/indico/blob/676dc3d1fede9bd4417982e79310753691be2472/indico/modules/rb/client/js/common/map/reducers.js#L64
- https://github.com/indico/indico/blob/676dc3d1fede9bd4417982e79310753691be2472/indico/modules/events/editing/client/js/editing/timeline/util.js#L55

ljharb:
@ThiefMaster i agree those are false positives. Can you file a new issue so we can track fixing it?

PR #3006, opened 2021-06-17
[react/jsx-no-useless-fragments] add option to allow single expressions in fragments

mattdarveniza:
# Problem

When writing JSX in typescript, a common problem is that Typescript expects the return type of a component to be `ReactElement`. This excludes other types that a component could return, for example just a plain string. This is a problem that's been discussed with the TS community here: [21699](https://github.com/microsoft/TypeScript/issues/21699), however my understanding is there are technical barriers to overcome in order to support this.

A common workaround for this is to wrap a variable in a fragment and expression to get the desired result:

```tsx
type Person = { name: string }

// type error
const WhatsMyName: React.FC<{ person: Person }> = ({ person }) => person.name

// workaround
const WhatsMyName: React.FC<{ person: Person }> = ({ person }) => <>{person.name}</>
```

This works well (even if it introduces a small amount of overhead in React), but triggers the `jsx-no-useless-fragments` rule, because as far as eslint is concerned, this _is_ a useless fragment in javascript.

# Solution
I'm proposing that we add an option to allow single expressions for Typescript users that want to employ this workaround. I went with an extra option because the original behaviour still makes sense in plain Javascript, and this is a behaviour only Typescript users would find useful.

(also fixed a typo from `NeedsMoreChidren` -> `NeedsMoreChildren`. Took me a while to figure out why the tests weren't passing thanks to that typo 😅)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ## Unreleased
 
+### Added
+* [`jsx-no-useless-fragments`]: add option to allow single expressions in fragments ([#3006][] @mattdarveniza)
+
 ### Fixed
 * component detection: use `estraverse` to improve component detection ([#2992][] @Wesitos)
 * [`destructuring-assignment`], [`no-multi-comp`], [`no-unstable-nested-components`], component detection: improve component detection ([#3001][] @vedadeepta)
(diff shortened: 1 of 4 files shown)


mattdarveniza:
Thanks for the speedy review and merge @ljharb! Appreciate it.

anton-johansson:
Any plans on releasing a new release so we can work with this? :D

ljharb:
There’s always a plan to do a release. There’s just no schedule.

dzek69:
@ljharb @yannickcr if there is no schedule, can you release it like now? :) it's been 3 months without a release and I'd enjoy this one be released, thanks

PR #3018, opened 2021-07-12
[Docs] `jsx-uses-react`, `react-in-jsx-scope`: document `react/jsx-runtime` config

pkuczynski:


docs/rules/react-in-jsx-scope.md
```diff
@@ -43,4 +43,4 @@ var Hello = <div>Hello {this.props.name}</div>;
 
 ## When Not To Use It
 
-If you are setting `React` as a global variable you can disable this rule.
+If you are setting `React` as a global variable or using [new JSX transform from React 17](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#removing-unused-react-imports), you can disable this rule.
```
ljharb, line 46: rather than explicitly disabling rules manually in this case, one should extend `react/jsx-runtime` which disables those rules for you.

ljharb (changes requested):

ljharb (approved):
went ahead and made the changes, in both rules and in the readme

Issue #3008, opened 2021-06-21
`react/no-unused-prop-types` false positve in render function with other components


redeyes2015:
```tsx
  import React, { FunctionComponent } from "react";
  import "./styles.css";
  import Slide from "./Slide";

  interface SlideProps {
    isActive: boolean; // (*)
  }

  const MyCard: FunctionComponent<{ doo: boolean }> = ({ doo }) => {
    return doo ? <div>1</div> : <div>0</div>;
  }

  export default function App() {
    return (
      <div className="App">
        <Slide>{({ isActive }: SlideProps) => <MyCard doo={isActive} /> }</Slide>
      </div>
    );
  }
```

I found that if the children of `<Slide>` return anything including another component, like `<MyCard>` here,
`react/no-unused-prop-types` would report a failure on `isActive` in `SlideProps` (*), which I think is a false positive.

Changing the render function to `({ isActive }: SlideProps) => isActive ? "a" : "b"` would not trigger the error.

Another work-around I found is change it to: `(props: SlideProps) => <MyCard doo={props.isActive} />` 

It seems eslint-plugin-react v7.22.0 and v7.24.0 both have this issue.

I tried to make a minimal reproducing repo here: https://github.com/redeyes2015/no-unused-prop-types-false-positive

I wonder why this rule is triggered and how to fix it properly.

Thanks for reading. I browsed the issue list, but did not find the exact match one. If I missed it or not asking at the right place, please accept my apology and just close this issue.

ljharb:
I agree, this is a false positive.

doublethefish:
Seen in v7.25.1

I-GiM:
I get this error in my components too. Are there any fixes for this? Is there a way to disable it from the eslint config file?
The only thing that seems to work is not to destructure the map object and disabling the error with a comment.

PR #3023, opened 2021-07-22
Document #3008 adds a test to no-unused-prop-types

thinklikeadesigner:
#3008 request to add a test for maintainers
- figure out why there's a false positive
- adds an easy way to keep  investigating the issue
- No side effects

@DianaSuvorova

### DIFF
--- a/tests/lib/rules/no-unused-prop-types.js
+++ b/tests/lib/rules/no-unused-prop-types.js
@@ -3948,6 +3948,22 @@ ruleTester.run('no-unused-prop-types', rule, {
       `,
       features: ['types'],
     },
+    {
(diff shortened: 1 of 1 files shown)


DianaSuvorova:
Nice, Great job!

Would you please check if this is a case of [SFC](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unused-prop-types.md#false-positives-sfc)

The example in the doc is using flow but the same should apply to TS. Ie if you try to use this prop not within a function would there still be an error?

ljharb:
This is an SFC, but the place being warned on is a render prop - which isn't a component.

PR #2929, opened 2021-02-19
[Fix] `destructuring-assignment`: fix a false prositive for local prop named `context` in SFC

SyMind:
Fixes #2911.

Related pull request: https://github.com/yannickcr/eslint-plugin-react/pull/2912

> hm, does eslint itself not have utilities for determining the relationship to function params?

I adjusted my code. `createSFCParams` collect the props and context info in SFC.

```javascript
function handleStatelessComponent(node) {
      const params = evalParams(node.params);

      const SFCComponent = components.get(context.getScope(node).block);
      if (!SFCComponent) {
        return;
      }
      sfcParams.push(params);

      // ......
}
```

PR #3025, opened 2021-07-27
[Fix] `destructuring-assignment`: get the contextName correctly

ohhoney1:
correct #2929, for issue #2911


ljharb (changes requested):
Thanks! Could we get a regression test for this?

ohhoney1:
> Thanks! Could we get a regression test for this?

@ljharb done

ljharb (approved):

PR #3037, opened 2021-08-10
[new] Add no-update-forcing-props rule

### LINKED ISSUE #1633: New rule to avoid prop value with array or object creation in render

wojtekmaj:
Replaces #2024

Warns you if you have passed a prop with a value that will be [referentially unequal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators#Identity_strict_equality_()) on each render. This is bad for performance, as it will result in the garbage collector being invoked way more than is necessary. It may also cause unnecessary re-renders if a brand new object or function is passed as a prop to a component that uses reference equality check on the prop to determine if it should update.

Fixes #1633. Closes #1888.

### DIFF
--- a/lib/rules/jsx-no-update-forcing-props.js
+++ b/lib/rules/jsx-no-update-forcing-props.js
@@ -1,0 +1,117 @@
+/**
+ * @fileoverview Prevent usage of props that would cause the component to render every time its parent renders
+ * @author Wojciech Maj
+ */
+
+'use strict';
+
+const propName = require('jsx-ast-utils/propName');
+const docsUrl = require('../util/docsUrl');
+const jsxUtil = require('../util/jsx');
+
+const typeToName = {
+  object: 'Objects',
+  array: 'Arrays',
+  func: 'Functions',
+  arrowFunc: 'Functions',
+  new: 'Constructor expressions',
+  call: 'Function calls'
+};
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Prevent usage of props that would cause the component to render every time its parent renders',
+      category: 'Best Practices',
+      recommended: false,
+      url: docsUrl('jsx-no-update-forcing-props')
+    },
+    schema: [{
+      type: 'object',
+      properties: {
+        objectExpressions: {
+          default: false,
+          type: 'boolean'
+        },
+        arrayExpressions: {
+          default: false,
+          type: 'boolean'
+        },
+        functionExpressions: {
+          default: false,
+          type: 'boolean'
+        },
+        arrowFunctionExpressions: {
+          default: false,
+          type: 'boolean'
+        },
+        newExpressions: {
+          default: false,
+          type: 'boolean'
+        },
+        callExpressions: {
+          default: false,
+          type: 'boolean'
+        },
+        ignoreRefs: {
+          default: true,
+          type: 'boolean'
+        }
+      },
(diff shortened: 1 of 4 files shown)


lib/rules/jsx-no-update-forcing-props.js
```diff
@@ -0,0 +1,117 @@
...
+          type: 'boolean'
+        },
+        callExpressions: {
+          default: false,
+          type: 'boolean'
+        },
+        ignoreRefs: {
+          default: true,
+          type: 'boolean'
+        }
```
ljharb, line 59: maybe rather than 7 booleans, 6 of which are named "allow" and default to false, and the 7th named "ignore" and defaults to true - we could use an object like:
```
{
  functions: 'allow',
  arrowFunctions: 'allow',
  objects: 'allow',
  arrays: 'allow',
  newExpressions: 'allow',
  callExpressions: 'allow',
  refs: 'ignore',
}
```
?

ljharb (changes requested):

lib/rules/jsx-no-update-forcing-props.js
wojtekmaj: The problem is that ignoreRefs isn't like other flags. If true (default), it will ignore checking ref prop for all of the above value types.

wojtekmaj:
Self-burn: unless we can verify if these objects, arrays etc. are not created in render function at all, instead of directly in prop values, this rule is only half done.

lib/rules/jsx-no-update-forcing-props.js
ljharb: Ahhh ok, it’s like a parallel setting. Comment still applies to the 6 booleans tho; seems like an object might be easier to manage?

wojtekmaj: I thought it's already an object with a bunch of properties? I'm a noob when it comes to building ESLint rules though.

ljharb: You're right that the config object has a bunch of properties - i'm saying that rather than add them at the top level, it might be better to nest them under a single property, so that *future* options to the rule can go alongside that single property.

ljharb:
@wojtekmaj you deleted your fork, which means this PR is unrecoverable :-/ that's very unfortunate.

wojtekmaj:
The PR stays and so [the diff](https://patch-diff.githubusercontent.com/raw/jsx-eslint/eslint-plugin-react/pull/3037.diff) will remain available. I think since this PR didn't get any interest in 15+ months and I didn't find time to work on this any further it's safe to say this would never have been finalized anyway.

ljharb:
I've landed PRs that others have come by later and finished after 8 years before, so no, it's not safe to say that.

douglasjunior:
> Self-burn: unless we can verify if these objects, arrays etc. are not created in render function at all, instead of only directly in prop values, this rule is only half done.

@wojtekmaj How about the react-perf behavior?

https://github.com/cvazac/eslint-plugin-react-perf

Issue #3024, opened 2021-07-22
false positive in a simple function react/destructuring-assignment


danosaure:
I have the following code:

```javascript
import fn from '../some-file';
export default (fileName) => {
  const match = fileName.match(/some expression/);
  if (match) {
    return fn;
  }
  return null;
};
```

it thinks that `const match = fileName.match()` is causing a destructuring issue.

ljharb:
This certainly shouldn't be detected as a component, since it has no hooks, and no jsx, and the argument isn't named "props".

PR #3038, opened 2021-08-10
[Fix] destructuring-assignment, component detection: handle default export edge case

vedadeepta:
Fixes: #3024 

Added code to deal with `ExportDefaultDeclaration` as the parent node of a `ArrowFunctionExpression`


ljharb (approved):
Thanks, looks great!

PR #3039, opened 2021-08-11
[New] `jsx-pascal-case`: add `allowLeadingUnderscore` option

### LINKED ISSUE #2603: jsx-pascal-case: Allow leading underscore
I often find myself defining a component which wraps a component of the same name. In this case I usually import the wrapped component with an alias, like this:

```jsx
import { Scene as _Scene } from '../scene/scene'

function Scene() {
  return <_Scene {...} />
}
```

This is triggering jsx-pascal-case.

I see there have been a few PRs to address edge cases around jsx-pascal-case to accommodate components called `_` or `SCREAMING_SNAKE`. What do you think about allowing `_PascalCase`?

I'd be happen to make a PR.

mwaddell:
Fixes #2603

* Adds a new *optional* `allowLeadingUnderscore` property to the existing `jsx-pascal-case` rule. 
* Updates documentation to clearly explain when *not* to use it
* Updates unit tests to ensure coverage

### DIFF
--- a/lib/rules/jsx-pascal-case.js
+++ b/lib/rules/jsx-pascal-case.js
@@ -90,6 +90,9 @@ module.exports = {
         allowAllCaps: {
           type: 'boolean'
         },
+        allowLeadingUnderscore: {
+          type: 'boolean'
+        },
         allowNamespace: {
           type: 'boolean'
         },
@@ -111,6 +114,7 @@ module.exports = {
   create(context) {
     const configuration = context.options[0] || {};
     const allowAllCaps = configuration.allowAllCaps || false;
+    const allowLeadingUnderscore = configuration.allowLeadingUnderscore || false;
     const allowNamespace = configuration.allowNamespace || false;
     const ignore = configuration.ignore || [];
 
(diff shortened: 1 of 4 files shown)


lib/rules/jsx-pascal-case.js
```diff
@@ -132,10 +136,12 @@ module.exports = {
         do {
           const splitName = checkNames[index];
           if (splitName.length === 1) return undefined;
-          const isPascalCase = testPascalCase(splitName);
-          const isAllowedAllCaps = allowAllCaps && testAllCaps(splitName);
           const isIgnored = ignoreCheck(ignore, splitName);
 
+          const checkName = allowLeadingUnderscore && splitName.startsWith('_') ? splitName.substring(1) : splitName;
```
ljharb, line 141:
```suggestion
          const checkName = allowLeadingUnderscore && splitName.startsWith('_') ? splitName.slice(1) : splitName;
```

ljharb (approved):
Looks great, thanks

ljharb:
@mwaddell however, because you made this PR from a fork you don't own, i can't rebase and force push to the PR branch (due to a github bug). can you add me to https://github.com/pangaeatech/eslint-plugin-react so i can land this?

mwaddell:
Replaced by #3041 which is from my own repo instead of my company's

lib/rules/jsx-pascal-case.js
mwaddell: Addressed in #3041

Issue #2860, opened 2020-11-18
New rule proposal: react/prefer-functional-component


Bowser65:
With the addition of hooks, it is possible now to write stateful React applications using only functions. However, unless I'm mistaken, there are not ways currently to enforce the use of functional components over class components.

The `prefer-stateless-function` rule is not enough for this case, since it allows class components provided they have a state or additional methods (like handlers defined as a class methods). It is not possible to forbid the use of state altogether (you can disable `setState`, but defining `state` is still allowed, silencing `prefer-stateless-function`), and filtering class methods using eslint rules without interfering with other code is complicated.

The only exception that should be taken into account are components making use of `componentDidCatch`, as there are currently no hook alternative for React (although there is a hook if you're using Preact, so it might be worth making this a configurable field).

Depending on how this rule gets implemented, it could even deprecate the `prefer-stateless-function` rule, in favor of a more generic rule covering more cases. In this case, the configuration for this new rule could follow the following schema:
```json
"react/prefer-functional-component": [ "error", "always | allow-stateful | never" ]
```
Where `always` enforces the use of functional components everywhere, `allow-stateful` behaves as the current `prefer-stateless-function` rule, and `never` enforces the use of class components.

ljharb:
It seems like you could ensure this by configuring existing linter rules to ban an import of `Component` from `'react'`, as well as a property of `React.Component`, although you're right that this wouldn't handle error boundaries.

The reality, though, is that I don't see the benefit of this rule. Just because you *can* implement something in an SFC doesn't mean you *should*, just like just because you *can* implement something in a class component doesn't mean you *should*. 

What is the motivation here to ban class components almost entirely, beyond subjective preference?

Bowser65:
Most of the reasons I have in mind follows React's own motivations for introducing hooks (https://reactjs.org/docs/hooks-intro.html#motivation):

Hooks enforce making reusable code, rather than packing everything in a single class component that's heavier (therefore harder to understand), and harder to reuse. You can quickly end up making components that are either wrappers that'd better fit as a hook (and reduce component tree complexity), or start writing subcomponents as class methods (e.g. `this.renderHeader`, `this.renderBody`, `this.renderFooter`), which isn't super reusable and will become an issue in the event of a refactor.

[...]

PR #3040, opened 2021-08-11
[New Rule] prefer-function-component

tatethurston:
Adds a new rule to enforce the use of function components over class components. Related issue: https://github.com/yannickcr/eslint-plugin-react/issues/2860

**Users wanting to try this lint rule today:**

This functionality exists in [eslint-plugin-react-prefer-function-component](https://www.npmjs.com/package/eslint-plugin-react-prefer-function-component).

`npm install eslint-plugin-react-prefer-function-component`

### DIFF
--- a/docs/rules/prefer-function-component.md
+++ b/docs/rules/prefer-function-component.md
@@ -1,0 +1,89 @@
+# Forbid class Components (react/prefer-function-component)
+
+This rule prevents the use of class components.
+
+Since the addition of hooks, it has been possible to write stateful React components
+using only functions. Mixing both class and function components in a code base adds unnecessary hurdles for sharing reusable logic.
+
+By default, class components that use `componentDidCatch` are enabled because there is currently no hook alternative for React. This option is configurable via `allowComponentDidCatch`.
+
+## Rule Details
+
+This rule will flag any React class components that don't use `componentDidCatch`.
+
+Examples of **incorrect** code for this rule:
+
+```jsx
+import { Component } from 'react';
+
+class Foo extends Component {
+  render() {
+    return <div>{this.props.foo}</div>;
+  }
+}
+```
+
+Examples of **correct** code for this rule:
+
+```jsx
+const Foo = function (props) {
+  return <div>{props.foo}</div>;
+};
+```
+
+```jsx
+const Foo = ({ foo }) => <div>{foo}</div>;
+```
+
+## Rule Options
+
+```js
+...
+"prefer-function-component": [<enabled>, { "allowComponentDidCatch": <allowComponentDidCatch> }]
+...
+```
+
+- `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
+- `allowComponentDidCatch`: optional boolean set to `true` if you would like to ignore class components using `componentDidCatch` (default to `true`).
+
+### `allowComponentDidCatch`
+
+When `true` (the default) the rule will ignore components that use `componentDidCatch`
+
+Examples of **correct** code for this rule:
+
+```jsx
+import { Component } from 'react';
+
+class Foo extends Component {
+  componentDidCatch(error, errorInfo) {
+    logErrorToMyService(error, errorInfo);
+  }
+
+  render() {
+    return <div>{this.props.foo}</div>;
+  }
+}
+```
+
+When `false` the rule will also flag components that use `componentDidCatch`
+
+Examples of **incorrect** code for this rule:
+
+```jsx
+import { Component } from 'react';
+
+class Foo extends Component {
+  componentDidCatch(error, errorInfo) {
+    logErrorToMyService(error, errorInfo);
+  }
+
+  render() {
+    return <div>{this.props.foo}</div>;
+  }
+}
+```
+
+### Related rules
+
+- [prefer-stateless-function](./prefer-stateless-function)
--- a/docs/rules/prefer-function-component.md
+++ b/docs/rules/prefer-function-component.md
@@ -1,4 +1,4 @@
-# Forbid class Components (react/prefer-function-component)
+# Prefer function components over class components (react/prefer-function-component)
 
 This rule prevents the use of class components.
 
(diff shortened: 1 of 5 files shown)


ljharb:
#2860 doesn't have a "help wanted" or "accepted" label on it - I continue to be unconvinced this is actually a good rule to have. Some things must be class components, and other things are more clear/better as class components; I don't think a lint rule is naunced enough to be appropriate.

docs/rules/prefer-function-component.md
```diff
@@ -0,0 +1,89 @@
+# Prefer function components over class components (react/prefer-function-component)
+
+This rule prevents the use of class components.
+
+Since the addition of hooks, it has been possible to write stateful React components
+using only functions. Mixing both class and function components in a code base adds unnecessary hurdles for sharing reusable logic.
+
+By default, class components that use `componentDidCatch` are enabled because there is currently no hook alternative for React. This option is configurable via `allowComponentDidCatch`.
```
ljharb, line 8: i'm not clear on why this is an option. A class component that's an error boundary *can never* be a functional component (pending changes from React itself), so there'd never be any value in this rule warning on it. I think that this option shouldn't be an option, it should just be hardcoded to true.

lib/rules/prefer-function-component.js
```diff
@@ -0,0 +1,86 @@
...
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Enforce components are written as function components',
+      category: 'Stylistic Issues',
+      recommended: false,
+      suggestion: false,
+      url: docsUrl('prefer-function-component')
+    },
+    fixable: false,
```
ljharb, line 25: is there a reason this rule isn't fixable?

Any component with a constructor body or class methods, of course, couldn't be fixable, but the example in the readme could.

ljharb (changes requested):
I wonder if perhaps this could be a more generic "prefer-component-style" rule, that can be used to force class components *or* force function components?

tatethurston:
> I wonder if perhaps this could be a more generic "prefer-component-style" rule, that can be used to force class components _or_ force function components?

Yeah that seems reasonable to me, the motivating intent here is to keep consistency in a code base.

Looking at the recent work by the Relay team (and their close collaboration with the React team) and the React concurrent mode work my impression is that the longer term direction of React is going to be towards all function components, but that is speculative right now.

tatethurston:
> #2860 doesn't have a "help wanted" or "accepted" label on it - I continue to be unconvinced this is actually a good rule to have. Some things must be class components, and other things are more clear/better as class components; I don't think a lint rule is naunced enough to be appropriate.
> 
> That said, I did ask you to open the PR, so I'll give it a good faith review.

I don't have any expectations here, just something to consider. This can always sit or be closed and reopened in the future. I just want things to be code complete (or close) while I have the time.

docs/rules/prefer-function-component.md
tatethurston: Many codebases I've seen don't implement error boundaries from scratch in their code base, instead they import library solutions that use error boundaries internally. I made this configurable (though with the default value of `true` so most clients can ignore it) for clients who want _all_ class component usage to be flagged. This also provides a migration path should React release functional component error boundaries.

lib/rules/prefer-function-component.js
ljharb: Lots of rules are partially fixable; obv it's ideal when things are fully fixable tho.

Actually wait, I'd forgotten about prefer-stateless-function. How is this rule different from that one? https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prefer-stateless-function.md

docs/rules/prefer-function-component.md
tatethurston: Also, Preact does offer functional error boundaries via a hook, so this would be valuable for those users.

lib/rules/prefer-function-component.js
tatethurston: I think https://github.com/yannickcr/eslint-plugin-react/issues/2860 explains the difference well:

> The prefer-stateless-function rule is not enough for this case, since it allows class components provided they have a state or additional methods (like handlers defined as a class methods). It is not possible to forbid the use of state altogether (you can disable setState, but defining state is still allowed, silencing prefer-stateless-function), and filtering class methods using eslint rules without interfering with other code is complicated.

ljharb: i see, so this is a stronger thing than that, where prefer-stateless-function pushes SFCs when possible, this rule would ban class components entirely in favor of hook-using SFCs. Gotcha.

Successful83 (approved):

nebrius:
Just wanted to pop in to say I'd like to have this lint rule too. As pointed out elsewhere, the biggest value here is in enabling consistency across a codebase and the ability to reuse logic between components easily.

In my case, another big use is to use these lint rules to gauge how well we are/are not doing in paying down tech debt over time on an old-ish codebase where we're about 50/50 on function components versus class components. Given the large-ish size of our org and codebase, this has to be automated via tooling so we can create dashboards/reports/etc.

ljharb:
Does the `function-component-definition` rule meet everyone's use cases here?

tatethurston:
> Does the `function-component-definition` rule meet everyone's use cases here?

No that rule enforces the [function definition](https://github.com/jsx-eslint/eslint-plugin-react/blob/4b4bba91ddebe27cf6bdd9f86d2b9a824a07a8fb/docs/rules/function-component-definition.md) and the request here is to enforce function components instead of class components.

nebrius:
> > Does the `function-component-definition` rule meet everyone's use cases here?
> 
> No that rule enforces the [function definition](https://github.com/jsx-eslint/eslint-plugin-react/blob/4b4bba91ddebe27cf6bdd9f86d2b9a824a07a8fb/docs/rules/function-component-definition.md) and the request here is to enforce function components instead of class components. It’s complementary but none overlapping.

Agree with Tate. This rule, while useful, doesn't solve the issue of preventing the introduction of class components into a codebase.

ljharb:
@robwierzbowski nope, nothing's changed there, and ErrorBoundary in particular still is required to be a class component.

robwierzbowski:
That's true. I believe this code allows [error boundaries to remain class components](https://github.com/tatethurston/eslint-plugin-react-prefer-function-component#faq). As you discussed above, whether this is hardcoded true or an option, the issue could be addressed.

PR #3041, opened 2021-08-12
[New] `jsx-pascal-case`: add `allowLeadingUnderscore` option

### LINKED ISSUE #2603: jsx-pascal-case: Allow leading underscore

mwaddell:
Fixes #2603
Replaces #3039

* Adds a new *optional* `allowLeadingUnderscore` property to the existing `jsx-pascal-case` rule. 
* Updates documentation to clearly explain when *not* to use it
* Updates unit tests to ensure coverage


ljharb:
I’ll keep both PRs in sync and land them together :-) thanks!

PR #3048, opened 2021-08-24
jsx-no-bind: Report usage of local functions as violations

### LINKED ISSUE #3047: jsx-no-bind: consider local function declaration usage a violation
Currently the `jsx-no-bind` rule allows the following:

```jsx
function MyComponent() {
  function doStuff() {}
  return <OtherComponent onClick={doStuff} />;
}
```

I think it would be beneficial if it would consider that to be a violation, since I see this done relatively often.

Was it a decision to ignore function declarations (and `let` and `var`) to avoid needing extra analyses to do with reassignment? If not, would a pull request to treat `function a() {}` the same as `const a = function () {}` be accepted?

p7g:
Fixes #3047 

I implemented this in a pretty straightforward way, so there are some trade-offs:

1. Any assignments to a function declaration are not considered. Consequently, a function that is unconditionally re-assigned with something that doesn't bind will still trigger a violation.
2. It uses the same configuration value and error message as the usage of a function expression.

To fix 1 would require some fancy analysis, and is probably not worth the effort, though 2 would be easy to fix and I can do so if needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 * [`no-typos`]: prevent crash on styled components and forwardRefs ([#3036][] @ljharb)
 * [`destructuring-assignment`], component detection: handle default exports edge case ([#3038][] @vedadeepta)
 * [`no-typos`]: fix crash on private methods ([#3043][] @ljharb)
+* [`jsx-no-bind`]: handle local function declarations ([#3048][] @p7g)
 
 ### Changed
 * [Docs] [`jsx-no-bind`]: updates discussion of refs ([#2998][] @dimitropoulos)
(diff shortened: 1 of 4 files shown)


ljharb (approved):
Seems great.

Arguably this is a breaking change since it will cause more warnings to appear, but equally arguably, it's a bug fix that function declarations were ever ignored (it just never came up because the airbnb styleguide forbids function declarations entirely).

p7g:
Cheers!

andy128k:
This incorrectly marks a valid code as incorrect in BDD-style tests.

```js
describe('<Outer/>', () => {
   const Inner = () => <div/>;

   it('wraps', () => {
      render(
        <Outer inner={Inner} />
      );
   });
});
```

JustFly1984:
In your case you can extract Inner outside of describe, cos it has no closure

ljharb:
@andy128k i don't think there's as much value for this rule to apply to tests at all, since it's a rule about runtime performance.

andy128k:
@JustFly1984 `describe` block create a scope. It is not a rare case when a single test file contains multiple `describe`s and sometimes they are nested. They may define components with the same name, so scoping is needed.

```js
describe('<Outer/>', () => {
  describe('<div/>, () => {
     const Inner = () => <div/>;
  
     it('wraps', () => {
        render(
          <Outer inner={Inner} />
        );
     });
  });

  describe('<span/>', () => {
     const Inner = () => <span/>;
  
     it('wraps', () => {
        render(
          <Outer inner={Inner} />
        );
     });
  });
});
```

Extraction just breaks code style.

---

@ljharb This is what I did -- ignored this rule for affected tests. But having a separated configuration of eslint for tests just feels wrong.

ljharb:
It’s totally normal and expected and it’s weird to have the same exact eslint config for tests and prod code - that said, if there’s a way to handle this case that doesn’t special-case BDD globals, I’m down.

JustFly1984:
you can rewrite rules in eslintrc for /*.{test|spec}.{j|t}sx?/ files

Issue #2777, opened 2020-08-29
(react props) is missing in props validation ERROR


S-Amin:
I just tested the version `7.20.6` and I got the error `... is missing in props validation `.
It's a `Nextjs` project and I want to add the linting now but I got the error on my components
this is my package dependency 

```
        "eslint": "^7.7.0",
        "eslint-config-airbnb-typescript": "^9.0.0",
        "eslint-config-prettier": "^6.11.0",
        "eslint-plugin-import": "^2.22.0",
        "eslint-plugin-jsx-a11y": "^6.3.1",
        "eslint-plugin-prettier": "^3.1.4",
        "eslint-plugin-react": "^7.20.6",
        "eslint-plugin-react-hooks": "^4.0.8",

```

this is the `eslint` consfig file 
```

{
    "env": {
        "browser": true,
        "es2020": true
    },
    "extends": [
        "airbnb-typescript",
        // "eslint:recommended",
        // "plugin:react/recommended",
        "plugin:@typescript-eslint/recommended",
        "prettier",
        "prettier/react",
        "prettier/@typescript-eslint",
        "plugin:prettier/recommended"
    ],
    "parser": "@typescript-eslint/parser",
    "parserOptions": {
        "project": "./tsconfig.json",
        "ecmaFeatures": {
            "jsx": true
        },
        "ecmaVersion": 12,
        "sourceType": "module"
    },
    "plugins": ["react", "@typescript-eslint"],
    "rules": {
        "prettier/prettier": ["error"],
        "react/react-in-jsx-scope": "off",
        "import/prefer-default-export": "warn",
        "no-underscore-dangle": "off",
        "no-nested-ternary": "off",
        "@typescript-eslint/no-empty-interface": "warn",
        "@typescript-eslint/naming-convention": [
            "error",
            {
                "selector": "interface",
                "format": ["camelCase", "UPPER_CASE", "PascalCase"]
            }
        ]
    }
}

```

and there are multiple components that `eslint` threw this error `... is missing in props validation `

ljharb:
Please provide exact component code that's being warned on, and the exact text of the warnings.

What's likely happening is that a bunch of missing propTypes are being properly detected now that component detection has been improved to catch a new class of cases.

S-Amin:
One of the components is this: 



```
interface Props {
    posts: string[];
}
const Home: NextPage<Props> = ({ posts }) => {
    function onChange(date, dateString) {
        console.log(date, dateString);
[...]

PR #3049, opened 2021-08-25
[Fix] `prop-types`, `propTypes`: handle React.* TypeScript types

### LINKED ISSUE #3035: [feature request] jsx-no-constructed-context-values but for all props
It will be very helpful to prevent non-stable values for **all** kinds of props to prevent mistakes that can cause needless re-renders. 

For example, to prevent such things:
```
<MySuperHeavyComponent options={{ foo: 'bar'}} />
```

### LINKED ISSUE #3045: [Bug] Enable Plugin to Properly Acknowledge Props Provided Via `React.FC`
Carryover from #2353, intended to remove the noise from the problem.

As stated in the opening comment of the original issue, this plugin does not properly validate props when `React.FC` is used. Information was provided at typescript-eslint/typescript-eslint#3075 on how to address this issue. The suggested approach just needs to be implemented.

cc: @ljharb 

## Original Opening Comment from @dchebakov 

- ESLint Version: v6.0.1
- Node Version: v10.16.0
- npm Version: v6.9.0

**Configuration**

```js
module.exports = {
  parser: '@typescript-eslint/parser',
  parserOptions: {
    project: './tsconfig.json',
  },
  plugins: ['@typescript-eslint'],
  extends: [
    'plugin:react/recommended',
    'plugin:@typescript-eslint/recommended',
  ],
}
```

```tsx
interface PersonProps {
    username: string;
}
const Person: React.FunctionComponent<PersonProps> = (props): React.ReactElement => (
    <div>{props.username}</div>
);
```

```bash
npx eslint src --ext .tsx
```

`eslint` show no errors, because type definition actually present in generic type `React.FunctionComponent<PersonProps>`

```js
error    'username' is missing in props validation       react/prop-types
```

vedadeepta:
Fixes #3045 and related issues.

Implements logic discussed in this comment: https://github.com/yannickcr/eslint-plugin-react/issues/2777#issuecomment-683944481

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 * [`destructuring-assignment`], component detection: handle default exports edge case ([#3038][] @vedadeepta)
 * [`no-typos`]: fix crash on private methods ([#3043][] @ljharb)
 * [`jsx-no-bind`]: handle local function declarations ([#3048][] @p7g)
+* [`prop-types`], `propTypes`: handle React.* TypeScript types ([#3049][] @vedadeepta)
 
 ### Changed
 * [Docs] [`jsx-no-bind`]: updates discussion of refs ([#2998][] @dimitropoulos)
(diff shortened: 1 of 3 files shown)


vedadeepta:
hmm seems like some tests are failing cause `Array.includes` isn't supported in old envs?

ljharb (approved):
Thanks, looks great

vedadeepta:
> Does this solve #2777 as well?

Nope, it will show an error for code like this
```
interface Props {
    posts: string[];
}
const Home: NextPage<Props> = ({ posts }) => {
  return <span>jsx</span>
}
```

because `NextPage` is not in our `allowedGenericTypes` and this [line 945](https://github.com/yannickcr/eslint-plugin-react/blob/d74a7d82d92ba984f7bcbf62d7ebe9272b3d5154/lib/util/propTypes.js#L945) will block the construction of `declaredPropTypes` from TS interface definition.

If we want this case to be handled we need to make `allowedGenericTypes` configurable by the user.

ljharb:
ok - we can look into that in a followup. it'd be great to allow all generic types and actually respect what they declare.

PR #3056, opened 2021-08-29
[Fix] False positive in `no-this-in-sfc`

### LINKED ISSUE #3054: react/no-this-in-sfc: eslint-plugin-react 7.25.0 interprets an old legacy jQuery plugin as a sfc
I get lot's of false positive `react/no-this-in-sfc` warnings in Typescript (*.ts) Files.
It seems as if TypeScript files with ts (not just tsx) extension would be regarded as react files and of type `$.fn.<lowercase>... = function (): <type> | null` would be interpreted as an `sfc`:

```ts
$.fn.getValueAsStringWeak = function (): string | null {
	const val = this.length === 1 ? this.val() : null;

	return typeof val === 'string' ? val : null;
};
```

Wesitos:
+ Modified `getStatelessComponent` method
+ Added check for functions assigned to properties (property name must be PascalCase)

Fixes #3054

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -608,6 +608,7 @@ function componentRule(rule, context) {
      * @returns {ASTNode | undefined}
      */
     getStatelessComponent(node) {
+      const parent = node.parent;
       if (
         node.type === 'FunctionDeclaration'
         && (!node.id || isFirstLetterCapitalized(node.id.name))
@@ -617,6 +618,13 @@ function componentRule(rule, context) {
       }
 
       if (node.type === 'FunctionExpression' || node.type === 'ArrowFunctionExpression') {
+        const isMethod = parent.type === 'Property' && parent.method;
+        const isPropertyAssignment = parent.type === 'AssignmentExpression'
+          && parent.left.type === 'MemberExpression';
+        const isModuleExportsAssignment = isPropertyAssignment
+          && parent.left.object.name === 'module'
+          && parent.left.property.name === 'exports';
+
         if (node.parent.type === 'ExportDefaultDeclaration') {
           if (utils.isReturningJSX(node)) {
             return node;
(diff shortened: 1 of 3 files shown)


lib/util/Components.js
```diff
@@ -617,6 +618,13 @@ function componentRule(rule, context) {
       }
 
       if (node.type === 'FunctionExpression' || node.type === 'ArrowFunctionExpression') {
+        const isMethod = parent.type === 'Property' && parent.method;
+        const isPropertyAssignment = parent.type === 'AssignmentExpression'
+              && parent.left.type === 'MemberExpression';
+        const isModuleExportsAssignment = isPropertyAssignment
+              && parent.left.object.name === 'module'
+              && parent.left.property.name === 'exports';
```
ljharb, line 626:
```suggestion
          && parent.left.type === 'MemberExpression';
        const isModuleExportsAssignment = isPropertyAssignment
          && parent.left.object.name === 'module'
          && parent.left.property.name === 'exports';
```

lib/util/Components.js
```diff
@@ -630,31 +638,36 @@ function componentRule(rule, context) {
...
         }
 
-        // Case like `React.memo(() => <></>)` or `React.forwardRef(...)`
-        const pragmaComponentWrapper = utils.getPragmaComponentWrapper(node);
-        if (pragmaComponentWrapper) {
-          return pragmaComponentWrapper;
+        if (isPropertyAssignment
+            && !isModuleExportsAssignment
+            && !isFirstLetterCapitalized(parent.left.property.name)) {
+          return undefined;
```
ljharb, line 667:
```suggestion
        if (
          isPropertyAssignment
          && !isModuleExportsAssignment
          && !isFirstLetterCapitalized(parent.left.property.name)
        ) {
          return undefined;
```

ljharb (approved):

PR #3061, opened 2021-08-30
[Fix] `jsx-no-useless-fragments`: Allow whitespace when `allowExpressions` is `true`

benj-dobs:
Bug fix.

Problem: When `react/jsx-no-useless-fragment` has `allowExpression` set to `true`, the following code produces an error:

```jsx
<>
  {foo ? (
    <Bar />
  ) : (
    <Baz />
  )}
</>
```

This PR amends the `allowExpressions` rule added in #3006 to allow insignificant whitespace around the expression.


ljharb (approved):
awesome, thanks!

Issue #3062, opened 2021-08-30
'children' is missing in props validation(react/prop-types) when using `React.PropsWithChildren` as of 7.25.0


FritsvanCampen:
When you have a typescript component like:

```
const MyComponent = (props: React.PropsWithChildren<{}>): React.ReactElement => {
	return <>{props.children}</>;
};
```

As of 7.25 you get the warning:

> 'children' is missing in props validation(react/prop-types)

`React.PropsWithChildren` is defined as: 
```
type PropsWithChildren<P> = P & { children?: ReactNode | undefined };
```

If you change the Component props type to:

```
const MyComponent = (props: React.PropsWithChildren<{ children?: ReactNode | undefined }>): React.ReactElement => {
	return <>{props.children}</>;
};
```

The warning disappears.

I believe this is a mistake. The warning should not trigger in both cases.
It did not trigger in 7.24. It does as of 7.25.

ljharb:
cc @Wesitos

vedadeepta:
Not sure if this is a bug, `children` is indeed not defined right? @ljharb @FritsvanCampen

vedadeepta:
Will look into it

johnhunter:
I'm seeing the same with `React.FC` when upgrading from 7.24.0 to 7.25.0

vedadeepta:
@johnhunter the PR fixes `React.FC` also
this should work
```
          import React from 'react';
          const Person: React.FC<{}> = (props): React.ReactElement => (
              <>{props.children}</>
          );

```

ljharb:
Fixed in #3064, which will be released soon.

FritsvanCampen:
Thanks for the quick fix!

johnhunter:
I still see `'children' is missing in props validation  react/prop-types` using `React.FC` with v7.25.1

ljharb:
@johnhunter you’ll see that for eternity, since that version doesn’t contain the fix.

PR #3064, opened 2021-08-30
[Fix] `prop-types`, `propTypes`: handle implicit `children` prop in react's generic types

vedadeepta:
#3062 

All react generic type extends from `type PropsWithChildren<P> = P & { children?: ReactNode | undefined };`. So whenever a generic type from React is used I've added an optional children prop to `this.declaredPropTypes`.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L824


ljharb (approved):
Thanks, seems good!

Yohandah:
this PR introduced a regression : https://github.com/jsx-eslint/eslint-plugin-react/issues/3524

PR #3065, opened 2021-09-01
[Fix] `display-name`: fix misinterpreted function components

danielfinke:
- when determining if an `ArrowFunctionExpression` returns JSX, do not check arguments to a returned `CallExpression`

Arrow function expressions that return the result of a function call taking JSX arguments are incorrectly deemed to be returning JSX. This leads to some misinterpretations of arrow functions as function components, which are then flagged as "missing display name".

### `ArrowFunctionExpression` as `Property` value
```
const obj = {
  prop: () => console.log(<a>something</a>)
};
const obj2 = {
  prop: () => { return console.log(<a>something</a>); }
};
```

### `ArrowFunctionExpression` as RHS of `AssignmentExpression`
```
let obj3;
obj3 = () => console.log(<a>something</a>);
```

I think any other examples of `ArrowFunctionExpressions` that pass `utils.isInAllowedPositionForComponent` at https://github.com/yannickcr/eslint-plugin-react/blob/810806e58d5a611e6743227c91d859c2c02563bb/lib/util/Components.js#L648-L650 will hit this too (excl. a `VariableDeclarator` as it's handled on [L635](https://github.com/yannickcr/eslint-plugin-react/blob/810806e58d5a611e6743227c91d859c2c02563bb/lib/util/Components.js#L635) ahead of time).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ### Fixed
 * [`jsx-no-useless-fragments`]: Handle insignificant whitespace correctly when `allowExpressions` is `true` ([#3061][] @benj-dobs)
 * [`prop-types`], `propTypes`: handle implicit `children` prop in react's generic types ([#3064][] @vedadeepta)
+* [`display-name`]: fix arrow function returning result of function call with JSX arguments being interpreted as component ([#3065][], @danielfinke)
 
+[#3065]: https://github.com/yannickcr/eslint-plugin-react/pull/3065
 [#3064]: https://github.com/yannickcr/eslint-plugin-react/pull/3064
 [#3061]: https://github.com/yannickcr/eslint-plugin-react/pull/3061
 
--- a/tests/util/jsx.js
+++ b/tests/util/jsx.js
@@ -27,9 +27,7 @@ describe('jsxUtil', () => {
     const assertValid = (codeStr) => assert(
       isReturningJSX(() => false, parseCode(codeStr), mockContext)
     );
-    const assertInValid = (codeStr) => assert(
-      !!isReturningJSX(() => false, parseCode(codeStr), mockContext)
-    );
+
     it('Works when returning JSX', () => {
       assertValid(`
         function Test() {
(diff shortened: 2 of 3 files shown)


tests/util/jsx.js
```diff
@@ -71,11 +69,27 @@ describe('jsxUtil', () => {
     });
 
     it('Can ignore null', () => {
-      assertInValid(`
+      assertValid(`
```
ljharb, line 72: Does this mean that `Test` is changing from "not a component" to "a component", or the reverse?

tests/util/jsx.js
```diff
@@ -71,11 +69,27 @@ describe('jsxUtil', () => {
...
       `);
     });
+
+    it('Ignores JSX arguments to function calls used as return value of arrow functions', () => {
+      let astNode = parseCode(`const obj = {
+        prop: () => test(<a>something</a>)
+      }`);
+      let arrowFunctionExpression = astNode.declarations[0].init.properties[0].value;
+
+      assert(!isReturningJSX(() => false, arrowFunctionExpression, mockContext));
```
ljharb, line 85: maybe these could use the `assertValid`/`assertInvalid` helpers?

tests/util/jsx.js
danielfinke: `assertInValid` wasn't doing anything - `isReturningJSX` was being double-negated on L31 - it was identical in function to `assertValid`

ljharb: gotcha

tests/util/jsx.js
danielfinke: I don't think I can. The node passed to `isReturningJSX` needs to be the `ArrowFunctionExpression` to illustrate the issue. But in the examples I provided, it ends up being the `AssignmentExpression` if I try to use `assertValid` as-is IIRC. And just putting in the arrow function by itself passes a `FunctionExpressionStatement` or something like that. It doesn't match what's being passed from `Components.js`, in short

ljharb: You can't pass just `() => test(<a>something</a>)`? or are you saying it needs the context provided by being inside the object literal

danielfinke: Yeah that context is relevant. In the case of the objects it's the being in the `Property` context and in the second example it's the being in the `AssignmentExpression` context

ljharb (approved):

ljharb:
@danielfinke unfortunately since you made this PR from a fork you don't own, i can't force push to it due to a github bug.

Please do *not* close this PR and open a new one; what would be ideal is if you can add me to https://github.com/Unity-Technologies/eslint-plugin-react for long enough that I can get this PR landed.

danielfinke:
@ljharb I'll see what I can do - not sure what hoops I have to jump thru for that.

If there's anything simple you need changed to the commit message, etc., I can do that on your behalf (as that may be quicker). Let me know.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ## Unreleased
 
 ### Fixed
+* [`display-name`]: fix arrow function returning result of function call with JSX arguments being interpreted as component ([#3065][], @danielfinke)
```
ljharb, line 9: this should be added at the bottom of the "fixed" list, and you'll need to add a PR link for `[#3065]` (see `[#3061]` as an example)

ljharb:
You may also be able to check the "allow edits" box in the RHS of the PR, but github's bug means that probably won't work (but is worth a shot)

CHANGELOG.md
danielfinke: @ljharb done!

> You may also be able to check the "allow edits" box in the RHS of the PR
Nah, I tried looking for this when I first submitted the PR and saw the failure. Doesn't seem to be an option anywhere for me

ljharb:
Thanks! It'd be great if in the future you could make PRs from your own personal fork :-)

PR #3070, opened 2021-09-08
[fix] `jsx-uses-vars` ignores lowercase tag names

alanorozco:
Context: eslint/eslint#15040

### DIFF
--- a/lib/rules/jsx-uses-vars.js
+++ b/lib/rules/jsx-uses-vars.js
@@ -11,6 +11,9 @@ const docsUrl = require('../util/docsUrl');
 // Rule Definition
 // ------------------------------------------------------------------------------
 
+const isTagNameRe = /^[a-z]/;
+const isTagName = (name) => isTagNameRe.test(name);
+
 module.exports = {
   meta: {
     docs: {
@@ -33,6 +36,10 @@ module.exports = {
         if (node.name.name) {
           // <Foo>
           name = node.name.name;
+          // Exclude lowercase tag names like <div>
+          if (isTagName(name)) {
+            return;
+          }
         } else if (node.name.object) {
           // <Foo...Bar>
           let parent = node.name.object;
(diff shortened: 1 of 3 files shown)


lib/rules/jsx-uses-vars.js
```diff
@@ -33,6 +36,10 @@ module.exports = {
         if (node.name.name) {
           // <Foo>
           name = node.name.name;
+          // Exclude lowercase tag names like <div>
+          if (isTagName(name)) {
```
ljharb, line 40: what does this do with namespaced tags? Is `<a.X />` not allowed?

alanorozco: `a.X` is not handled on this codepath, I believe that's under the condition below for `node.name.object`. [They're interpreted as references](https://babeljs.io/repl#?browsers=ie%2011&build=&builtIns=false&corejs=3.6&spec=false&loose=false&code_lz=DwPgUABBwGYPZwgenFYAxBzXXnAdAEICGATtpLgvgEZkXApA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=react&prettier=true&targets=&version=7.15.5&externalPlugins=%40babel%2Fplugin-syntax-object-rest-spread%407.8.3&assumptions=%7B%7D) so they should work fine.

If you would like, I could add the following valid test cases.

```js
    {
      code: `
        /* eslint jsx-uses-vars: 1 */
        var object;
        React.render(<object.Tag />);
      `
    }, {
      code: `
        /* eslint jsx-uses-vars: 1 */
        var object;
        React.render(<object.tag />);
      `
    }
```

ljharb: Thanks, adding those would be great :-)

ljharb (approved):

yoyo837:
Nice fix.

mrtnzlml:
Hola @alanorozco! 👋 It seems like this PR caused a breaking change in a patch release, see: https://github.com/yannickcr/eslint-plugin-react/issues/3080

Could you please have a look? I am not sure what would be the right thing to do in this case. 🤔

PR #3071, opened 2021-09-10
[readme] Update broken link for configuration files

prateek3255:
Fix broken link to ESLint docs in README


ljharb (approved):
unfortunate it broke in the first place - cool URLs don't change - but thanks for the update

PR #3073, opened 2021-09-11
fix jsx-no-bind.md example

thisconnect:
useCallback should take an array with the value of useState and not the setter of useState

### DIFF
--- a/docs/rules/jsx-no-bind.md
+++ b/docs/rules/jsx-no-bind.md
@@ -193,7 +193,7 @@ const Button = () => {
   const [text, setText] = useState("Before click");
   const onClick = useCallback(() => {
     setText("After click");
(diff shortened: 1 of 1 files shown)


ljharb:
Since it uses the setter, that's what it should take.

I think React lets state setters be optional in dependency arrays, but it's still better to provide it.

Let's open an issue in the future first to make sure a change is valid :-) thanks!

thisconnect:
Thanks I just realized that it was intentional after I opened the PR, still find it a bit strange to use it with the setter, and not too much documentation on it https://reactjs.org/docs/hooks-reference.html#usecallback i.e. should useState setters always be "cached" with useCallback.

ljharb:
React allows you to omit them, but logic and explicitness dictates they should be included.

PR #3076, opened 2021-09-18
[Fix] `prop-types` `propTypes`: bail out unknown generic types inside func params

### LINKED ISSUE #3074: Version 7.25.2 missing prop-types with typescript parametrised props
### The following works alright in 7.25.1:

```jsx
type Props<ValueType> = {
  value: ValueType;
  onClick: (value: ValueType) => void;
};

// prettier-ignore
const Button = <T,>({ onClick, value }: Props<T>) => {
  return <button onClick={() => onClick(value)}>BUTTON</button>;
};

const valueObj = {
  aprop: 'avalue'
};

  return (
    <div className="App">
      <h1>a button</h1>
      <Button onClick={(value) => console.log(value)} value={valueObj} />
    </div>
  );
}
```

### With version 7.25.2 same code results in react/prop-types errors
```
7:23  error  'onClick' is missing in props validation  react/prop-types
7:32  error  'value' is missing in props validation    react/prop-types
```

vedadeepta:
Fixes #3074 

when `const Comp = (props: NonReactGeneric<T>) => {...}` proceed as usual instead of early return.

### DIFF
--- a/lib/util/propTypes.js
+++ b/lib/util/propTypes.js
@@ -970,16 +970,6 @@ module.exports = function propTypesInstructions(context, components, utils) {
           }
         });
       } else {
-        // check if its a valid generic type when `X<{...}>`
-        if (
-          param.typeAnnotation
-          && param.typeAnnotation.typeAnnotation
-          && param.typeAnnotation.typeAnnotation.type === 'TSTypeReference'
-          && param.typeAnnotation.typeAnnotation.typeParameters != null
-          && !isValidReactGenericTypeAnnotation(param.typeAnnotation.typeAnnotation)
-        ) {
-          return;
-        }
         markPropTypesAsDeclared(node, resolveTypeAnnotation(param));
       }
     } else {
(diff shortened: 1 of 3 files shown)


lib/util/propTypes.js
```diff
@@ -970,16 +970,6 @@ module.exports = function propTypesInstructions(context, components, utils) {
...
-        // check if its a valid generic type when `X<{...}>`
-        if (
-          param.typeAnnotation
-          && param.typeAnnotation.typeAnnotation
-          && param.typeAnnotation.typeAnnotation.type === 'TSTypeReference'
-          && param.typeAnnotation.typeAnnotation.typeParameters != null
-          && !isValidReactGenericTypeAnnotation(param.typeAnnotation.typeAnnotation)
-        ) {
-          return;
-        }
```
ljharb, line 982: i'm confused why this was here at all if removing it didn't cause a test to fail :-/

ljharb (approved):
Thanks!

lib/util/propTypes.js
vedadeepta: Yes, I was wondering the same. I put it there because I thought we were only handling generic types that came from react for components declared as `const Component = (props: IProps) => {...}`. This broke the earlier behavior, my fault - will be careful from next time.

PR #3078, opened 2021-09-19
[New] `jsx-max-props-per-line`: allow defining both `multi` and `single` line prop count

SIL0RAK:
# Problem 
You can only define single or multiple line prop count. 

# Solution 
Allow to config rule to check both multiple and single line tags while maintaining backward compatibility.

### DIFF
--- a/lib/rules/jsx-max-props-per-line.js
+++ b/lib/rules/jsx-max-props-per-line.js
@@ -8,6 +8,13 @@
 const docsUrl = require('../util/docsUrl');
 const report = require('../util/report');
 
+function getPropName(context, propNode) {
+  if (propNode.type === 'JSXSpreadAttribute') {
+    return context.getSourceCode().getText(propNode.argument);
+  }
+  return propNode.name.name;
+}
+
 // ------------------------------------------------------------------------------
 // Rule Definition
 // ------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


lib/rules/jsx-max-props-per-line.js
```diff
@@ -29,8 +29,22 @@ module.exports = {
...
+            type: 'object',
+            properties: {
+              single: {
+                type: 'integer',
+                minimum: 1
+              },
+              multi: {
+                type: 'integer',
+                minimum: 1
+              }
```
ljharb, line 45: are single/multi both required? what happens if i only pass one of them?

what about a value of zero? (as in, what if i want to force all props to be single-line, or all multi-line)

lib/rules/jsx-max-props-per-line.js
```diff
@@ -43,7 +57,19 @@ module.exports = {
...
+    const isExtendedConfig = typeof maximum !== 'number';
+
+    if (isExtendedConfig) {
+      maximumSingle = maximum.single || 1;
+      maximumMulti = maximum.multi || 1;
+    }
+
+    const when = isExtendedConfig
+      ? 'always'
+      : configuration.when || 'always';
```
ljharb, line 72: nothing in the schema seems to prevent passing a single/multi object and also setting `when`. if it's going to be ignored then we should make that schema disallowed.

ljharb (changes requested):

lib/rules/jsx-max-props-per-line.js
SIL0RAK: No, other will default to Infinity.
So `maximum: {multi: 2}}` will apply maximum length to multiple line tags and ignore single line tags.

Zero value is a nice idea for another PR.  This rule originally didn't support such a case.

ljharb (approved):
tweaked the implementation a bit. thanks!

PR #3083, opened 2021-09-24
[Docs] `jsx-max-props-per-line`: fix options example

MrRaiter:
Small issues with `jsx-max-props-per-line`  docs


ljharb (approved):
Thanks!

Issue #3044, opened 2021-08-17
noopener alone not allowed


mauro1855:
Hi,

I've been reading the discussion surrounding noopener / noreferrer, including what was done in #2043.
However, I still find the behaviour strange.

Here is my link:
``` <a href="https://some-external-link.com" target="_blank" rel="noopener">Link</a> ```

I'm getting a warning:
``` Using target="_blank" without rel="noreferrer" is a security risk: see https://html.spec.whatwg.org/multipage/links.html#link-type-noopener  react/jsx-no-target-blank ```

The html spec link redirects me to ```noopener``` documentation, and yet it is still requiring me to add ```noreferrer```. As far as I understand, the use or ```noreferrer``` or ```noopener``` removes the security risk, so the rule should not be triggered if one of them is present (no matter which). In this case, since I'm using ```noopener``` (thus removing the security risk), the warning should not be shown, imho.

Why don't I use ```noreferrer```? I think it's important for websites to know where their traffic is coming from, so I don't intend on hiding the referrer, and I don't think you should either. The referrer information helps bloggers and independent content creators properly monitor the traffic arriving to their websites.

ljharb:
See https://github.com/yannickcr/eslint-plugin-react/issues/2022#issuecomment-432032005:
> Due to the fact that noreferrer predates noopener, there are cases where noopener is ignored (because the behavior is not implemented).

That information also violates the privacy of the user, and I don't think your opinion is the common one here about what people should do - referrers aren't reliable and anyone relying on that info is getting a very incomplete picture, no matter what you choose to do. You are welcome to disable the rule, of course.

mauro1855:
Hi @ljharb,

I've read a lot of comments related to this here in the project, and it's not always obvious if the comments are referring to the use of ```noopener``` isolated, or in conjunction with ```noreferrer```, like ```noreferrer noopener```. That issue was initially opened because the linter was requiring both ```noreferrer noopener```, which is not really needed since ```noreferrer``` implies ```noopener```.

At the time, you even had an opinion which is compatible with mine: https://github.com/yannickcr/eslint-plugin-react/issues/2022#issuecomment-624339760:
> Given that, it seems we should modify the rule - by default, even - to no longer require both (but, to continue to allow both).

[...]

PR #3088, opened 2021-09-29
[Fix] `jsx-no-target-blank`: improve error messages

cutiful:
The current error message is confusing because it complains about links missing `rel="noreferrer"` even if `allowReferrer` is set to `true` and so `rel="noopener"` is enough to satisfy it. I made it report different messages depending on `allowReferrer` value and state that even if `rel="noopener"` is enough, `rel="noreferrer"` is preferred as per https://github.com/yannickcr/eslint-plugin-react/issues/3044#issuecomment-916801701. Also clarified the docs for `allowReferrer` to prevent misunderstandings like #3044.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,12 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ### Fixed
 * [`no-namespace`]: fix crash on non-string React.createElement name ([#3082] @ljharb)
 * [`no-namespace`]: avoid crash on non-string createElement values ([#3085] @ljharb)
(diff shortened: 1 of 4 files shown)


cutiful:
Not fully sure about the phrasing of `noTargetBlankWithoutNoreferrer`, it may sound like the rule can be safely disabled if you don't support older browsers. On the other hand, it is soon going to be true, as in newer browsers, `rel="noopener"` is implied with `target="_blank"`; and either way the [link](https://mathiasbynens.github.io/rel-noopener/#recommendations) (taken from the MDN [article](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noopener)) explains the details.

ljharb (approved):
Thanks, seems like an improvement

PR #3092, opened 2021-10-01
[Fix] `propTypes`: add `VoidFunctionComponent` to react generic list

### LINKED ISSUE #3090: bug is missing in props validation
hi why this is not valid ? it should ok no! 
image
image

here the demo from your example
```tsx
interface Props {
	age: number
 }
 const Hello: VFC<Props> = function Hello(props) {
	const { age } = props;

	return <div>Hello {name}</div>;
	// 'name' type is missing in props validation
 }

```

vedadeepta:
Fixes #3090


ljharb (approved):
looks great, thank you

PR #3120, opened 2021-11-03
[Dev Deps] update `eslint-remote-tester-run-action`

AriPerkkio:
Adds support for ESLint v8. Works with ESLint v7 too. 

Example run: https://github.com/AriPerkkio/eslint-plugin-react/actions/runs/1415490892. The `no-invalid-html-attribute` rule crashed couple of times there.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -19,7 +19,7 @@ jobs:
       - run: |
           npm link
           npm link eslint-plugin-react
-      - uses: AriPerkkio/eslint-remote-tester-run-action@v1
+      - uses: AriPerkkio/eslint-remote-tester-run-action@v2
         with:
           issue-title: 'Results of weekly scheduled smoke test'
           eslint-remote-tester-config: test/eslint-remote-tester.config.js


AriPerkkio:
Also one thing worth to mention:

The `eslint-remote-test-run-acton@v2` is using Github Docker Action with Node 14. For some reason `npm link` doesn't work if workflow is using node 14, but node 16 works. 

Basically this kind of change is required in order to have `npm link` working between the workflow and Github Docker Action. 
```diff
      - uses: actions/setup-node@v1
        with:
-         node-version: 14
+         node-version: 16
```

@ljharb any ideas why `node@14` package links do not pass to Docker containers, even if both host and container are using the same version? Node version of Github Docker action doesn't matter as long as the workflow is running Node 16.

The current workflow here is working OK since `ljharb/actions/node/install@main` ends up with Node 16:

https://github.com/yannickcr/eslint-plugin-react/blob/d62ab20baa6251cf29933f5633453984bb749c13/.github/workflows/smoke-test.yml#L14-L18

ljharb:
My guess is that npm 6 has a bug that npm 7 fixes. Are you installing the latest npm version always? If not, try that :-)

AriPerkkio:
Ah yes it seems there is a bug in npm 6. Updating npm in the workflow/host works. 
After debugging this for too long it's great to find a root cause. Thanks! 👍 🎉 

Container / action `npm@6` and `node14`:
- Host / workflow `npm@6` and `node14`. npm link fails with `Cannot find module 'eslint-plugin-react'`
- Host / workflow `npm@7` and `node14`. npm link works.
- Host / workflow `npm@8` and `node14`. npm link works.

ljharb (approved):

PR #3129, opened 2021-11-11
[New] `function-component-definition`: support `namedComponents` option being an array

petersendidit:
This adds support to the `function-component-definition`  rule to have the
`namedComponents` and `unnamedComponents` options be an array.

### DIFF
--- a/docs/rules/function-component-definition.md
+++ b/docs/rules/function-component-definition.md
@@ -31,13 +31,15 @@ function getComponent() {
 
 ## Rule Options
 
-This rule takes an options object as a second parameter where the preferred function type for components can be specified. The first property of the options object is `"namedComponents"` which can be `"function-declaration"`, `"function-expression"`, or `"arrow-function"` and has `'function-declaration'` as its default. The second property is `"unnamedComponents"` that can be either `"function-expression"` or `"arrow-function"`, and has `'function-expression'` as its default.
+This rule takes an options object as a second parameter where the preferred function type for components can be specified.
+The first property of the options object is `"namedComponents"` which can be `"function-declaration"`, `"function-expression"`, `"arrow-function"`, or an array containing any of those, and has `'function-declaration'` as its default.
+The second property is `"unnamedComponents"` that can be either `"function-expression"`, `"arrow-function"`, or an array containing any of those, and has `'function-expression'` as its default.
 
 ```js
 ...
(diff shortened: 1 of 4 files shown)


lib/rules/function-component-definition.js
```diff
@@ -109,23 +109,44 @@ module.exports = {
...
       },
-    }],
+    ],
   },
 
   create: Components.detect((context, components) => {
     const configuration = context.options[0] || {};
 
-    const namedConfig = configuration.namedComponents || 'function-declaration';
+    const namedConfig = Array.isArray(configuration.namedComponents) ? configuration.namedComponents : [configuration.namedComponents] || ['function-declaration'];
```
ljharb, line 149:
```suggestion
    const namedConfig = [].concat(configuration.namedComponents || 'function-declaration');
```

docs/rules/function-component-definition.md
```diff
@@ -31,12 +31,13 @@ function getComponent() {
...
+This rule takes an options object as a second parameter where the preferred function type for components can be specified. The first property of the options object is `"namedComponents"` which can be `"function-declaration"`, `"function-expression"`, `"arrow-function"`, or an array
+containing any of those and has `'function-declaration'` as its default. The second property is `"unnamedComponents"` that can be either `"function-expression"` or `"arrow-function"`, and has `'function-expression'` as its default.
 
 ```js
 ...
 "react/function-component-definition": [<enabled>, {
-  "namedComponents": "function-declaration" | "function-expression" | "arrow-function",
+  "namedComponents": "function-declaration" | "function-expression" | "arrow-function" | Array<string>,
```
ljharb, line 40: the type of the array isn’t string, it’s the same enum

ljharb (changes requested):
Thanks, this is great. Let’s make the same change on unnamed components too.

docs/rules/function-component-definition.md
petersendidit: Wasn't sure the correct syntax that should be used to define that here.

ljharb: I think we might need:
```suggestion
  "namedComponents": "function-declaration" | "function-expression" | "arrow-function" | Array<"function-declaration" | "function-expression" | "arrow-function">,
```

ljharb (approved):

PR #3131, opened 2021-11-11
[Fix] `jsx-no-literals`: properly handle HTML entities when allowed

### LINKED ISSUE #3130: jsx-no-literals: HTML entities in allowedValues are ignored
Hello! I'm using the `jsx-no-literals` rule with some exceptions (in the `allowedStrings` option). In particular I want to allow the `&mdash;` literal. However, the rule is still flagging the usage as an error even with this set.

Consider this simple file:

```jsx
<div>&mdash;</div>
```

And the following configuration:

```javascript
{
  noStrings: true,
  allowedStrings: [
    '&mdash;',
   ],
}
```

This will result in an error:

```javascript
  {
    ruleId: 'jsx-no-literals',
    severity: 1,
    message: 'Strings not allowed in JSX files: "&mdash;"',
    line: 2,
    column: 14,
    nodeType: 'JSXText',
    messageId: 'noStringsInJSX',
    endLine: 2,
    endColumn: 21
  }
```

I took a quick look at the code for this rule.

- When the `JSXText` check is done, `node.value` no longer is `&mdash;` but rather the actual symbol it represents (`—`). This causes the error
- Later when the `Literal` check is done, `node.value` is the original `&mdash;`. This passes the check.

So the `JSXText` check causes the rule to flag the error.

If I change the check, in the `getValidation` function, in `jsx-no-literals.js` to this:

```javascript
      if (config.allowedStrings.has(trimIfString(node.raw)))
```

then `node.raw` is still `&mdash;` even for the `JSXText` check. This small change seems to fix the issue, but I was hesitant to submit a PR because I'm not sure what other implications using the `raw` vs. the `value` has.

For what it's worth, after this change, all the test cases for `jsx-no-literals` still pass.

joeattardi:
When an HTML entity such as &mdash; is checked with ESLint, the `value`
property becomes the actual unicode character instead of the raw
literal.

This means if `&mdash;` is in the allowed strings array, and `&mdash;`
is in the code being checked, the actual value used in the JSXText check
will be `—`, which will not match against `&mdash;`, so it causes a
false positive error.

When checking against the set of allowed strings, this will now check
both the `value` and the `raw` properties of the node. For `&mdash;`, the
`raw` property will still be `&mdash;`.

If _either_ of these match the list of allowed values, then no error
will be returned.

I am very new to the project so if this is not the right fix, I will not be offended if you close the PR. 😆 

Fixes #3130

### DIFF
--- a/tests/lib/rules/jsx-no-literals.js
+++ b/tests/lib/rules/jsx-no-literals.js
@@ -284,6 +284,18 @@ ruleTester.run('jsx-no-literals', rule, {
         <img alt='blank image'></img>
       `,
     },
+    {
+      code: `
+        <div>&mdash;</div>
+      `,
+      options: [{ noStrings: true, allowedStrings: ['&mdash;', '—'] }],
+    },
+    {
+      code: `
+        <div>—</div>
+      `,
+      options: [{ noStrings: true, allowedStrings: ['&mdash;', '—'] }],
+    },
   ]),
 
   invalid: parsers.all([
--- a/lib/rules/jsx-no-literals.js
+++ b/lib/rules/jsx-no-literals.js
@@ -89,9 +89,11 @@ module.exports = {
     }
 
     function getValidation(node) {
-      if (config.allowedStrings.has(trimIfString(node.value))) {
+      const values = [trimIfString(node.raw), trimIfString(node.value)];
+      if (values.some((value) => config.allowedStrings.has(value))) {
         return false;
       }
+
       const parent = getParentIgnoringBinaryExpressions(node);
 
       function isParentNodeStandard() {


tests/lib/rules/jsx-no-literals.js
```diff
@@ -284,6 +284,10 @@ ruleTester.run('jsx-no-literals', rule, {
         <img alt='blank image'></img>
       `,
     },
+    {
+      code: '<div>&mdash;</div>',
+      options: [{ noStrings: true, allowedStrings: ['&mdash;'] }],
+    },
```
ljharb, line 290: let's add a few more test cases covering the behavior of `—` in the jsx and `'&mdash;'` in the config, as well as the reverse.

ljharb (changes requested):
Thanks!

tests/lib/rules/jsx-no-literals.js
joeattardi: Will do! Good idea. I'll update the PR in the next day or so.

ljharb: @joeattardi any update?

joeattardi: Sorry for the lack of responsiveness on this.

I was working on adding some more tests and I'm not sure what's going on.

I added the following test to the `valid` section:

```
    {
      code: '<div>&mdash;</div>',
      only: true,
      options: [{ noStrings: true, allowedStrings: ['—'] }],
    },
```

When I run the tests, this fails:

```
 AssertionError [ERR_ASSERTION]: Should have no errors but had 1: [
  {
    ruleId: 'jsx-no-literals',
    severity: 1,
    message: 'Strings not allowed in JSX files: "&mdash;"',
    line: 1,
    column: 6,
    nodeType: 'Literal',
    messageId: 'noStringsInJSX',
    endLine: 1,
    endColumn: 13
  }
]
```

However if I add the exact same test to the `invalid` section, with the expected error:

```
{
      code: '<div>&mdash;</div>',
      only: true,
      options: [{ noStrings: true, allowedStrings: ['—'] }],
      errors: [
        {
          messageId: 'noStringsInJSX',
          data: { text: '&mdash;' },
        },
      ],
    },
```

This test fails, too:

```
```

Maybe I'm missing something about how the tests work?

ljharb: On which parser? It’s possible different parsers produce different results.

joeattardi:
Hey just wanted to check in, I haven't forgotten about finishing this PR I've just had a crazy week. Hoping to have an update soon, thanks for your patience!

ljharb:
Thanks, appreciate the followup. This last week's a busy week for many, no worries - just mark it as ready for review when it is.

joeattardi:
Hey, I really apologize for the lack of updates on this. I have a lot going on right now and I don't think I will have the time to finish this PR currently. Feel free to close it or if you want to add the missing tests that's great as well.

Sorry again for the lack of communication on this.

ljharb:
@joeattardi no worries, thanks. I'd prefer to keep the PR open for as long as it takes for you, me, or someone else to finish it (anyone can post a link to a branch, and I'll pull it in)

ROSSROSALES:
Hey @ljharb, I'm going to try and tackle this in code spaces. I'll add more test cases and try to help close this PR

ljharb:
@ROSSROSALES sounds great! please post a link here to your branch (instead of opening a new PR) and i'll pull in the changes.

ROSSROSALES:
Hi @ljharb I made changes in a new branch on codespaces, how do I share the link to this branch?

ljharb:
@ROSSROSALES navigate to the branch on your fork, and share that link - eg, you have two codespace branches, and one of them is https://github.com/ROSSROSALES/eslint-plugin-react/tree/codespace-c172

ROSSROSALES:
> @ROSSROSALES navigate to the branch on your fork, and share that link - eg, you have two codespace branches, and one of them is https://github.com/ROSSROSALES/eslint-plugin-react/tree/codespace-c172

Ah I see thank you for teaching me! Are you able to see the changes from this link now?
https://github.com/ROSSROSALES/eslint-plugin-react/tree/codespace-d03a
I made the mistake and worked on a branch off of @joeattardi's fork so the changes could not be pushed.

ljharb:
I've pulled in the changes, and rebased.

ljharb (approved):
Looks great!

PR #3133, opened 2021-11-12
[Docs] HTTP => HTTPS

Schweinepriester:
Checked the links, corrected them and skipping the redirect HTTP => HTTPS this way 0:-)

For convenience:
- https://eslint.org/docs/user-guide/configuring/language-options#specifying-parser-options
- https://opensource.org/licenses/mit-license.php

… want me to check and change [the rest](https://github.com/yannickcr/eslint-plugin-react/search?l=Markdown&q=http%3A) as well? :)


ljharb:
Yes, we might as well change them all at once.

Schweinepriester:
There we go, should have got them all.

Not sure about the failing pipeline…? Is there anything I can do?

I'm not totally sure about http://facebook.github.io/react/docs/more-about-refs.html#the-ref-callback-attribute => https://reactjs.org/docs/refs-and-the-dom.html#callback-refs, might want to have a look, but I think it's the correct one :)

I assume you were going to do that anyway, but given my commits and messages, I'd recommend squashing 0:-)

ljharb (approved):
Rebased, and converted the remaining http links to https in the codebase.

PR #3142, opened 2021-11-17
[Fix] `prop-types`: fix false positives on renames in object destructuring

### LINKED ISSUE #2944: [react/prop-types] false positive when renaming in destructuring context
When following `react/destructuring-assignment`, if my prop shares a name with another constant I rename it during the destructure (`prop.styles` to `propStyles` below) to avoid the collision, I then start getting warned that `styles.posterImage' is missing in props validation` etc. for all keys of the `styles` object.

```javascript
import styles from '../styles/Video'

// ...

  render() {
    const { altText, styles: propStyles, posterUrl, videoUrl, subtitleUrl } = this.props

    const poster = (
      <>
        <img
          className={css(propStyles, styles.posterImage, this.state.isHovering && styles.zoomed)}
          src={posterUrl}
          alt={altText}
        />
```

Expected behavior: `styles` is not recognized as a prop after destructuring but `propStyles` is. No false positives on constants that share name with renamed prop's original name on the `props` object.

golopot:
Fixes #2944.

```js
const { styles: x } = props;
```
Our code incorrectly picks up `styles` as a prop, while `x` is the correct one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,12 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ### Fixed
 * [`no-invalid-html-attribute`]: allow `link` `rel` to have `apple-touch-icon`, `mask-icon` ([#3132][] @ljharb)
 * [`no-unused-class-component-methods`]: add `getChildContext` lifecycle method ([#3136][] @yoyo837)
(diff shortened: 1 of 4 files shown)


ljharb (approved):
Thanks, seems great!

PR #3149, opened 2021-11-26
[New] component detection: track React imports

duncanbeevers:
This PR adds a couple of basic tests to the existing `Components.detect` functionality.
Next it adds a facility for detecting the imported React identifiers during a parse, and adds a simple test for that detection.

This is an extraction of work initiated in https://github.com/yannickcr/eslint-plugin-react/pull/2921 and supports that PR

I think having a test runner for the `Components.detect` utility was worthwhile in itself, and separated that test-runner functionality into its own commit.

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -47,6 +47,7 @@ function mergeUsedPropTypes(propsList, newPropsList) {
 }
 
 const Lists = new WeakMap();
+const ReactImports = new WeakMap();
 
 /**
  * Components
@@ -54,6 +55,7 @@ const Lists = new WeakMap();
 class Components {
   constructor() {
     Lists.set(this, {});
+    ReactImports.set(this, {});
   }
 
   /**
@@ -178,6 +180,52 @@ class Components {
   length() {
     const list = Lists.get(this);
     return Object.keys(list).filter((i) => list[i].confidence >= 2).length;
+  }
+
+  /**
+   * Return the node naming the default React import
+   * It can be used to determine the local name of import, even if it's imported
+   * with an unusual name.
+   *
+   * @returns {ASTNode} React default import node
+   */
+  getDefaultReactImports() {
+    return ReactImports.get(this).defaultReactImports;
+  }
+
+  /**
+   * Return the nodes of all React named imports
+   *
+   * @returns {Object} The list of React named imports
+   */
+  getNamedReactImports() {
+    return ReactImports.get(this).namedReactImports;
+  }
+
+  /**
+   * Add the default React import specifier to the scope
+   *
+   * @param {ASTNode} specifier The AST Node of the default React import
+   * @returns {void}
+   */
+  addDefaultReactImport(specifier) {
+    const info = ReactImports.get(this);
+    ReactImports.set(this, Object.assign({}, info, {
+      defaultReactImports: (info.defaultReactImports || []).concat(specifier),
+    }));
+  }
+
+  /**
+   * Add a named React import specifier to the scope
+   *
+   * @param {ASTNode} specifier The AST Node of a named React import
+   * @returns {void}
+   */
+  addNamedReactImport(specifier) {
+    const info = ReactImports.get(this);
+    ReactImports.set(this, Object.assign({}, info, {
+      namedReactImports: (info.namedReactImports || []).concat(specifier),
+    }));
   }
 }
 
@@ -857,6 +905,25 @@ function componentRule(rule, context) {
     },
   };
 
+  // Detect React import specifiers
+  const reactImportInstructions = {
+    ImportDeclaration(node) {
+      const isReactImported = node.source.type === 'Literal' && node.source.value === 'react';
+      if (!isReactImported) {
+        return;
+      }
+
+      node.specifiers.forEach((specifier) => {
+        if (specifier.type === 'ImportDefaultSpecifier') {
+          components.addDefaultReactImport(specifier);
+        }
+        if (specifier.type === 'ImportSpecifier') {
+          components.addNamedReactImport(specifier);
+        }
+      });
+    },
+  };
+
   // Update the provided rule instructions to add the component detection
   const ruleInstructions = rule(context, components, utils);
   const updatedRuleInstructions = Object.assign({}, ruleInstructions);
@@ -866,7 +933,8 @@ function componentRule(rule, context) {
   const allKeys = new Set(Object.keys(detectionInstructions).concat(
     Object.keys(propTypesInstructions),
     Object.keys(usedPropTypesInstructions),
-    Object.keys(defaultPropsInstructions)
+    Object.keys(defaultPropsInstructions),
+    Object.keys(reactImportInstructions)
   ));
 
   allKeys.forEach((instruction) => {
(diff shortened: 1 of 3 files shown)


lib/util/Components.js
```diff
@@ -179,6 +181,53 @@ class Components {
...
+  /**
+   * Return the node naming the default React import
+   * It can be used to determine the local name of import, even if it's imported
+   * with an unusual name.
+   *
+   * @returns {ASTNode} React default import node
+   */
+  getDefaultReactImport() {
+    const info = ReactImports.get(this);
+    return info.defaultReactImport;
```
ljharb, line 194:
```suggestion
    return ReactImports.get(this).defaultReactImport;
```

lib/util/Components.js
```diff
@@ -179,6 +181,53 @@ class Components {
...
+  }
+
+  /**
+   * Return the nodes of all React named imports
+   *
+   * @returns {Object} The list of React named imports
+   */
+  getNamedReactImports() {
+    const info = ReactImports.get(this);
+    return info.namedReactImports;
```
ljharb, line 204:
```suggestion
    return ReactImports.get(this).namedReactImports;
```

lib/util/Components.js
```diff
@@ -179,6 +181,53 @@ class Components {
...
+
+  /**
+   * Add the default React import specifier to the scope
+   *
+   * @param {ASTNode} specifier The AST Node of the default React import
+   * @returns {void}
+   */
+  addDefaultReactImport(specifier) {
+    const info = ReactImports.get(this);
+    info.defaultReactImport = specifier;
```
ljharb, line 215:
```suggestion
    ReactImports.get(this).defaultReactImport = specifier;
```
altho i'm not a fan of the mutation here. why not `ReactImports.set(this, { ...info, defaultReactImport: specifier })`?

also tho, when would this function be called when there's already a defaultReactImport? Should it throw if it's called when `defaultReactImport` is already set?

tests/util/Component.js
```diff
@@ -0,0 +1,85 @@
...
+    ecmaFeatures: {
+      jsx: true,
+    },
+  },
+});
+
+describe('Components', () => {
+  describe('static detect', () => {
+    function makeTest(name, test, done) {
+      return it(name, () => {
```
ljharb, line 20: i'm not much of a fan of dynamically created tests like this; could we make the `it`s all hardcoded, and do the rest dynamically?

even better might be to avoid using `it`s at all, and just use ruletester with a dynamically built-up array of test cases?

tests/util/Component.js
```diff
@@ -0,0 +1,85 @@
...
+        const rule = Components.detect((context, components, util) => ({
+          'Program:exit'(node) {
+            done(name, context, components, util, node);
+          },
+        }));
+
+        const tests = {
+          valid: [Object.assign(test, {
```
ljharb, line 28:
```suggestion
          valid: [Object.assign({}, test, {
```
to avoid mutation

tests/util/Component.js
```diff
@@ -0,0 +1,85 @@
...
+      Object.values(components.list()).forEach((component) => {
+        assert(
+          component.node.id.name === 'MyStatelessComponent',
+          `${name} should be named MyStatelessComponent`
+        );
+      });
+    });
+
+    makeTest('Class Components', {
+      code: `import React from 'react'
```
ljharb, line 57: all these tests should use `parsers.all` with `features` so we get maximum test coverage.

tests/util/Component.js
```diff
@@ -0,0 +1,85 @@
...
+    makeTest('React imports', {
+      code: 'import React, { useCallback, useState } from \'react\'',
+    }, (_name, _context, components) => {
+      const defaultReactImport = components.getDefaultReactImport();
+      assert(defaultReactImport.local.name === 'React', 'default React import identifier should be "React"');
+
+      const namedReactImports = components.getNamedReactImports();
+      assert(namedReactImports.length === 2, 'named React import identifiers should be "useCallback" and "useState"');
+      assert(namedReactImports[0].local.name === 'useCallback', 'named React import identifiers should be "useCallback" and "useState"');
+      assert(namedReactImports[1].local.name === 'useState', 'named React import identifiers should be "useCallback" and "useState"');
```
ljharb, line 82:
```suggestion
      assert.deepEqual(
        namedReactImports.length,
        [
          { local: { name: 'useCallback' } },
          { local: { name: 'useState' } },
        ], 
        'named React import identifiers should be "useCallback" and "useState"'
      );
```

ljharb (changes requested):

lib/util/Components.js
duncanbeevers: It could throw, but it seems like it should be another rule's responsibility to simply flag the double-import, not this rules responsibility to throw a fit.

In normal cases the first / only definition is detected.
In abnormal cases the final detected instance wins; a little weird but at least predictable.

tests/util/Component.js
duncanbeevers: I feel the same way, but this was the most expedient.
I'll see about teasing this apart a little bit.

tests/util/Component.js
duncanbeevers: This doesn't really work in this instance because the returned identifier nodes have so much other stuff attached to them, that we end up with a failing `deepEqual` and an unreadable error message.

I could assert over all that stuff, but it's not really the point of the test.

```
      named React import identifiers should be "useCallback" and "useState"
Occurred while linting <input>:1
Rule: "import React, { useCallback, useState } from 'react'"
      + expected - actual

       [
         {
      -    "imported": {
      -      "loc": {
      -        "end": {
      -          "column": 27
      -          "line": 1
      -        }
      -        "start": {
      -          "column": 16
      -          "line": 1
      -        }
      -      }
      -      "name": "useCallback"
      -      "parent": [Circular]
```

lib/util/Components.js
ljharb: A rule couldn’t handle that unless this utility tracked every occurrence.

It seems like it’d be better then to keep an array, and rules can assume there’s one, or check, as they like.

tests/util/Component.js
ljharb: Maybe deepEqual to `namedReactImports.map(x => x.local.name)`?

lib/util/Components.js
duncanbeevers: I went ahead and made `defaultReactImports` an accumulating array the same as `namedReactImports`, and assert the same way using `deepEqual` over the mapped local identifiers of each of the returned arrays.

tests/util/Component.js
```diff
@@ -0,0 +1,94 @@
...
+  describe('static detect', () => {
+    function testComponentsDetect(test, done) {
+      const rule = Components.detect((context, components, util) => ({
+        'Program:exit'() {
+          done(context, components, util);
+        },
+      }));
+
+      const tests = {
+        valid: parsers.all([Object.assign({}, test, {
```
duncanbeevers, line 28: Added the `parsers.all` call here; no special treatment of the test cases was necessary (whew!)

ljharb (approved):

ljharb:
Note: I'm not treating this as semver-minor, since it's not really part of the public API.

PR #3155, opened 2021-12-11
[New Rule] no-undefined-initial-state

hjoelh:
Hey 👋

Thought this rule would be handy for some folks,  would appreciate any feedback 🙏

### DIFF
--- a/docs/rules/no-undefined-initial-state.md
+++ b/docs/rules/no-undefined-initial-state.md
@@ -1,0 +1,20 @@
+# Prevents initialising useState with undefined (react/no-undefined-initial-state)
+
+Some developers consider explicitly setting undefined to be a a bad idea and prefer to only allow JavaScript itself to set undefined values. This rule is aimed at reinforcing this idea via not allowing useState to be initialised with undefined.
+
+## Rule Details
+
+Examples of **incorrect** code for this rule:
+
+```jsx
+const [state, setState] = useState();
+const [state, setState] = useState(undefined);
+```
+
+Examples of **correct** code for this rule:
+
+```jsx
+const [state, setState] = useState(null);
+const [state, setState] = useState(true);
+const [state, setState] = useState({});
+```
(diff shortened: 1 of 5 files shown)


docs/rules/no-undefined-initial-state.md
```diff
@@ -0,0 +1,20 @@
+# Prevents initialising useState with undefined (react/no-undefined-initial-state)
+
+Some developers consider explicitly setting undefined to be a a bad idea and prefer to only allow JavaScript itself to set undefined values. This rule is aimed at reinforcing this idea via not allowing useState to be initialised with undefined.
```
ljharb, line 3: can you elaborate why some think of this as a bad idea? i certainly don't.

If you don't want explicit undefined references, there's an eslint rule for this that covers every context. What's the value of banning it just in `useState` but not everywhere else?

ljharb (changes requested):
Please file an issue discussing things like this before opening a PR in the future.

docs/rules/no-undefined-initial-state.md
hjoelh: hey, makes sense to open a discussion first 👍

My understanding is that some prefer to enforce using null as a explicit 'nothing/empty' value and leave undefined to mean not declared or not given a value

no-undefined rule does not appear to catch useState() -but does catch useState(undefined)

ljharb: How is that preference specific to useState? It seems like that preference would apply to everywhere in the language, and since https://eslint.org/docs/rules/no-undefined exists, why would this plugin need to address it?

hjoelh: agreed but no-undefined rule does not catch useState()

ljharb: Then that’s a bug with the rule, because it should be forcing use of void there.

You can also use no-restricted-syntax to forbid any mention of the keyword undefined *and* void.

PR #3156, opened 2021-12-13
[New] Component detection: add `util.isReactHookCall`

duncanbeevers:
* Rename Components test suite filename to match sibling lib/util/Components filename.
* Extend Components `testComponentsDetect` function to accept custom instructions, and to accumulate the results of processing those instructions.
* Add utility to check whether a `CallExpression` is a React hook call.

### 🗣 Discussion

This work extracts the somewhat complicated guard clause from #2921
It also makes hooks-usage detection generic, so the util can be used to check for:
* call to any React hook
* call to one or more specific React hooks; this will be used to check for `useState` calls in #2921

I converted this PR to a Draft while I work to extract this in a way that leaves the `hook-use-state` rule clean and easy to process.

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -7,6 +7,7 @@
 
 const doctrine = require('doctrine');
 const arrayIncludes = require('array-includes');
+const fromEntries = require('object.fromentries');
 const values = require('object.values');
 
 const variableUtil = require('./variable');
@@ -45,6 +46,8 @@ function mergeUsedPropTypes(propsList, newPropsList) {
 
   return propsList.concat(propsToAdd);
 }
+
+const USE_HOOK_PREFIX_REGEX = /^use[A-Z]/;
 
 const Lists = new WeakMap();
 const ReactImports = new WeakMap();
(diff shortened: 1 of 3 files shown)


ljharb:
I renamed the test file in the default branch so when you rebase it'll show the diff properly :-)

lib/util/Components.js
```diff
@@ -787,6 +789,93 @@ function componentRule(rule, context) {
...
+      const defaultReactImportSpecifier = defaultReactImports
+        ? defaultReactImports[0]
+        : undefined;
+
+      const defaultReactImportName = defaultReactImportSpecifier
+        ? defaultReactImportSpecifier.local.name
+        : undefined;
+
+      const reactHookImportSpecifiers = namedReactImports
+        ? namedReactImports.filter((specifier) => specifier.imported.name.match(USE_HOOK_PREFIX_REGEX))
```
ljharb, line 817:
```suggestion
        ? namedReactImports.filter((specifier) => USE_HOOK_PREFIX_REGEX.test(specifier.imported.name))
```

lib/util/Components.js
```diff
@@ -787,6 +789,93 @@ function componentRule(rule, context) {
...
+        : undefined;
+      const reactHookImportNames = reactHookImportSpecifiers
+        ? reactHookImportSpecifiers.reduce(
+          (acc, specifier) => {
+            acc[specifier.local.name] = specifier.imported.name;
+            return acc;
+          },
+          {}
+        )
+        : undefined;
```
ljharb, line 827:
```suggestion
      const reactHookImportNames = reactHookImportSpecifiers
        && fromEntries(reactHookImportSpecifiers.map(({ local, imported }) => [local.name, imported.name]);
```
and pull in `object.fromentries` for fromEntries

lib/util/Components.js
```diff
@@ -787,6 +789,93 @@ function componentRule(rule, context) {
...
+            && node.callee.object.name === defaultReactImportName
+            && node.callee.property.type === 'Identifier'
+            && node.callee.property.name.match(USE_HOOK_PREFIX_REGEX)
+      );
+
+      const isPotentialHookCall = !!(
+        reactHookImportNames
+          && node.callee.type === 'Identifier'
+          && node.callee.name.match(USE_HOOK_PREFIX_REGEX)
+      );
```
ljharb, line 842:
```suggestion
      const isPotentialHookCall = reactHookImportNames
          && node.callee.type === 'Identifier'
          && node.callee.name.match(USE_HOOK_PREFIX_REGEX);
```

lib/util/Components.js
```diff
@@ -787,6 +789,93 @@ function componentRule(rule, context) {
         && !!(node.params || []).length
       );
     },
+
+    /**
+     * Identify whether a node (CallExpression) is a call to a React hook
+     *
+     * @param {ASTNode} node The AST node being searched. (expects CallExpression)
+     * @param {('useCallback'|'useContext'|'useDebugValue'|'useEffect'|'useImperativeHandle'|'useLayoutEffect'|'useMemo'|'useReducer'|'useRef'|'useState')[]} [expectedHookNames] React hook names to which search is limited.
```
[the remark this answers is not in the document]
duncanbeevers: I'm not sure; this is the first time I've messed about with jsdoc.
I don't know of any union type exported from React that captures the union of all the exported hook names.

The code is reasonably robust and will just as happily accept an array of arbitrary strings; I just wanted to guide plugin writers towards the happy path.

Sorry I don't have a better answer for this.

tests/util/Components.js
```diff
@@ -19,12 +21,32 @@ const ruleTester = new eslint.RuleTester({
...
+
+      const rule = Components.detect((_context, components, util) => {
+        const instructionResults = [];
+
+        const augmentedInstructions = fromEntries(
+          entries(instructions || {}).map((nodeTypeAndHandler) => {
+            const nodeType = nodeTypeAndHandler[0];
+            const handler = nodeTypeAndHandler[1];
+            return [nodeType, (node) => {
+              instructionResults.push({ type: nodeType, result: handler(node, context, components, util) });
```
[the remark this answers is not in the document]
duncanbeevers: The purpose of the `Components` test scaffolding is to allow exfiltration of the `Components.detect` locals to a scope in the test where `assert` was available, but without having to write a complete rule with a `Program:exit` clause.

The original implementation exfiltrated the `Components.detect` locals, but didn't make it easy to actually test some of those locals; in particular I couldn't test `utils` because the utility functions expect `ASTNodes`, and by the time `Program:exit` fires, I've lost the `CallExpression` nodes I need.

These changes are intended to aid in writing new test cases for `Components` functionality, and in particular, for testing `utils` functionality, and those utils may expect many different ASTNode types, not just `CallExpression`.

Custom instructions passed into the `testComponentsDetect` get augmented and then incorporated into a custom rule, and when that rule runs over code the augmented instructions do two things:
1. Provide the `Components.detect` locals to the instruction callback. Ordinarily the instruction callback only receives a `node` as an argument and relies on having been defined in the `Components.detect` closure to access these local.
2. Accumulate a "log" as the visitor traverses the parsed AST; the log tracks which node type was traversed, and the result of that visit.

Taken together, these allow creation test cases intended to read something like:

> The visitor visited a `CallExpression` node
> The visitor had access to the `util` local
> The `util` function `isReactHookCall` returned the following result for the visited `CallExpression` node

lib/util/Components.js
```diff
@@ -46,6 +46,8 @@ function mergeUsedPropTypes(propsList, newPropsList) {
   return propsList.concat(propsToAdd);
 }
 
+const USE_HOOK_PREFIX_REGEX = /^use/i;
```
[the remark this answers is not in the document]
duncanbeevers: Oh yep, and I should lose the `'i'` flag on there too.

lib/util/Components.js
ljharb: whoops, oh well, we can just name the argument then

tests/util/Components.js
```diff
@@ -19,12 +21,32 @@ const ruleTester = new eslint.RuleTester({
...
+            }];
+          })
+        );
+
+        return Object.assign({}, augmentedInstructions, {
+          'Program:exit'(node) {
+            if (augmentedInstructions['Program:exit']) {
```
duncanbeevers, line 43: This isn't yet exercised, but I figured if someone needed to test `util` functionality when handling the `Program:exit` instruction it should behave the same way as handling any other instruction type.

duncanbeevers: I added a self-test section to the `test/util/Components` file in order to exercise the logging behavior while visiting `Program:exit` instructions.

ljharb (approved):

PR #3157, opened 2021-12-14
[Fix] `jsx-indent-props`: Reset `line.isUsingOperator` correctly after ternary (#3146)

### LINKED ISSUE #3146: jsx-indent-props: Enforcing incorrect indents when using a component with props in ternary before
I experience some odd behavior of `jsx-indent-props` depending on a ternary before the actual component.
When having a component with props within the ternary and not having the whole ternary in one line, the rule enforces more indents for the subsequent component than it should.

I would expect the ternary not having any impact on the returning component.

Tested versions: 7.27.0 and 7.27.1

```
import React from 'react';

// works as expected - no props for the component in ternary
export const LintExample = () => {
  const foo = true
    ? <div>test</div>
    : false;

  return <div
    id="id"
  >
    test
  </div>;
};

// works as expected - props for the component in ternary but ternary in one line
export const LintExample1 = () => {
  const foo = true ? <div id="id">test</div> /* added prop "id" here*/ : false;

  return <div
    id="id"
  >
    test
  </div>;
};

// does not work as expected - props for the component in ternary and ternary not in one line
export const LintExample2 = () => {
  const foo = true
    ? <div id="id">test</div> // added prop "id" here
    : false;

  return <div
    id="id" // linter complains: ESLint: Expected indentation of 6 space characters but found 4.(react/jsx-indent-props)
  >
    test
  </div>;
};
```

tobiaswaltl:
Closes #3146 

When having a JSX element after a ternary and there are not only spaces before the opening bracket (e.g. due to a `return` keyword) `line.isUsingOperator` is not reset correctly and the rule enforces a higher indent.

Example:
```
export const LintExample2 = () => {
  const foo = true
    ? <div id="id">test</div>
    : false;

  return <div
    id="id" // linter complains: ESLint: Expected indentation of 6 space characters but found 4.(react/jsx-indent-props)
  >
    test
  </div>;
};
```


tests/lib/rules/jsx-indent-props.js
```diff
@@ -195,6 +195,48 @@ ruleTester.run('jsx-indent-props', rule, {
...
+       const F = () => {
+         const foo = true
+          ? <div id="id">test</div>
+          : false;
+
+          return <div
+            id="id"
+          >
+            test
+          </div>
```
ljharb, line 209:
```suggestion
        return <div
          id="id"
        >
          test
        </div>
```
the return statement should always be at the same indentation level as the `const`

tests/lib/rules/jsx-indent-props.js
```diff
@@ -195,6 +195,48 @@ ruleTester.run('jsx-indent-props', rule, {
...
```
ljharb, line 230:
```suggestion
        return <div
          id="id"
        >
          test
        </div>
```

ljharb (changes requested):
Just looking at the test cases, this doesn't seem right.

tests/lib/rules/jsx-indent-props.js
tobiaswaltl: You are right - somehow I've screwed up the indents for the whole code within the template string. I've amended the fix.

lib/rules/jsx-indent-props.js
```diff
@@ -153,7 +153,7 @@ module.exports = {
 
       const indent = regExp.exec(src);
       const useOperator = /^([ ]|[\t])*[:]/.test(src) || /^([ ]|[\t])*[?]/.test(src);
-      const useBracket = /^([ ]|[\t])*[<]/.test(src);
```
ljharb, line 156: Can we add some test cases that use tabs, since that's what's being removed here?

tobiaswaltl: Yes, I've added them.

ljharb (approved):

PR #3163, opened 2021-12-31
[Fix] `prop-types`, `propTypes`: add support for exported type inference

### LINKED ISSUE #3015: forwardRef example has error in TypeScript
```
    "@types/react": "^17.0.11",
    "@types/react-dom": "^17.0.8",
    "@typescript-eslint/eslint-plugin": "^4.28.0",
    "@typescript-eslint/parser": "^4.28.0",
```

The code from https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/forward_and_create_ref/ on forwardRef has the error `'children' is missing in props validation` based on the `react/propTypes` ESLint rule https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md

This rule says it validates on prop-types typing, as well as TS types. I am not sure if this is an issue in ESLint or TypeScript

vedadeepta:
Fixes #3015

### DIFF
--- a/lib/util/propTypes.js
+++ b/lib/util/propTypes.js
@@ -547,6 +547,30 @@ module.exports = function propTypesInstructions(context, components, utils) {
     if (node.right) return getRightMostTypeName(node.right);
   }
 
+  /**
+   * Returns true if the node is either a interface or type alias declaration
+   * @param {ASTNode} node
+   * @return {boolean}
+   */
+  function filterInterfaceOrTypeAlias(node) {
+    return (
+      astUtil.isTSInterfaceDeclaration(node) || astUtil.isTSTypeAliasDeclaration(node)
+    );
+  }
+
+  /**
+   * Returns true if the interface or type alias declaration node name matches the type-name str
+   * @param {ASTNode} node
+   * @param {string} typeName
+   * @return {boolean}
+   */
+  function filterInterfaceOrAliasByName(node, typeName) {
+    return (
+      (node.id && node.id.name === typeName)
+      || (node.declaration && node.declaration.id && node.declaration.id.name === typeName)
+    );
+  }
+
   class DeclarePropTypesForTSTypeAnnotation {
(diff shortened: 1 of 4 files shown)


lib/util/propTypes.js
```diff
@@ -547,6 +547,35 @@ module.exports = function propTypesInstructions(context, components, utils) {
...
+   * @param {ASTNode} node
+   * @param {string} typeName
+   * @return {boolean}
+   */
+  function filterInterfaceOrAliasByName(node, typeName) {
+    if (node.id && node.id.name) {
+      return node.id.name === typeName;
+    }
+
+    if (node && node.declaration && node.declaration.id && node.declaration.id.name) {
```
ljharb, line 572: if `node` is falsy then this function would have crashed on line 568

lib/util/propTypes.js
```diff
@@ -547,6 +547,35 @@ module.exports = function propTypesInstructions(context, components, utils) {
...
+      return node.declaration.id.name === typeName;
+    }
+
+    return false;
```
ljharb, line 576:
```suggestion
    return node && (
      (node.id && node.id.name === typeName)
      || (node.declaration && node.declaration.id && node.declaration.id.name === typeName)
    );
```
?

ljharb (approved):
LGTM!

lib/util/propTypes.js
vedadeepta: yes, my bad - the check should not contain `node && ` part. `node` would never be falsy.

lib/util/propTypes.js
vedadeepta: sure, but i feel sometimes one liners are less readable.

ljharb: yeah true, in this case it'd be much more readable with optional chaining, but we can't use that yet :-/

PR #3165, opened 2022-01-04
Add --fix-type support

### LINKED ISSUE #2039: [ESLint v5.9.0] Fix types
A new option was added to ESLint. Reference: https://eslint.org/blog/2018/11/eslint-v5.9.0-released#the-fix-type-option

Should we start supporting this? If so, looks like "layout" is pretty obvious, but "problem" vs "suggestion" could be harder to tell. Or should we just set all recommended non-layout rules to "problem"?

### LINKED ISSUE #2618: Autofix --fix-type layout

sryze:
Add `type` property to all rules to support usage of eslint's `--fix-type` option

Fixes #2618. Fixes #2039.

Successor of this PR: https://github.com/yannickcr/eslint-plugin-react/pull/2619

### DIFF
--- a/lib/rules/jsx-closing-bracket-location.js
+++ b/lib/rules/jsx-closing-bracket-location.js
@@ -19,13 +19,14 @@ const messages = {
 
 module.exports = {
   meta: {
+    type: 'layout',
     docs: {
       description: 'Validate closing bracket location in JSX',
       category: 'Stylistic Issues',
       recommended: false,
       url: docsUrl('jsx-closing-bracket-location'),
     },
-    fixable: 'code',
+    fixable: 'whitespace',
 
     messages,
 
--- a/lib/rules/jsx-closing-bracket-location.js
+++ b/lib/rules/jsx-closing-bracket-location.js
@@ -26,7 +26,7 @@ module.exports = {
       recommended: false,
       url: docsUrl('jsx-closing-bracket-location'),
     },
-    fixable: 'whitespace',
+    fixable: 'code',
 
     messages,
 
--- a/lib/rules/jsx-curly-spacing.js
+++ b/lib/rules/jsx-curly-spacing.js
@@ -43,7 +43,7 @@ module.exports = {
       recommended: false,
       url: docsUrl('jsx-curly-spacing'),
     },
-    fixable: 'whitespace',
+    fixable: 'code',
 
     messages,
 
--- a/lib/rules/jsx-curly-spacing.js
+++ b/lib/rules/jsx-curly-spacing.js
@@ -36,13 +36,14 @@ const messages = {
 
 module.exports = {
   meta: {
+    type: 'layout',
     docs: {
       description: 'Enforce or disallow spaces inside of curly braces in JSX attributes',
       category: 'Stylistic Issues',
       recommended: false,
       url: docsUrl('jsx-curly-spacing'),
     },
-    fixable: 'code',
+    fixable: 'whitespace',
 
     messages,
 
--- a/lib/rules/jsx-equals-spacing.js
+++ b/lib/rules/jsx-equals-spacing.js
@@ -28,7 +28,7 @@ module.exports = {
       recommended: false,
       url: docsUrl('jsx-equals-spacing'),
     },
-    fixable: 'whitespace',
+    fixable: 'code',
 
     messages,
 
--- a/lib/rules/jsx-equals-spacing.js
+++ b/lib/rules/jsx-equals-spacing.js
@@ -21,13 +21,14 @@ const messages = {
 
 module.exports = {
   meta: {
+    type: 'layout',
     docs: {
       description: 'Disallow or enforce spaces around equal signs in JSX attributes',
       category: 'Stylistic Issues',
       recommended: false,
       url: docsUrl('jsx-equals-spacing'),
     },
-    fixable: 'code',
+    fixable: 'whitespace',
 
     messages,
 
--- a/lib/rules/jsx-first-prop-new-line.js
+++ b/lib/rules/jsx-first-prop-new-line.js
@@ -19,13 +19,14 @@ const messages = {
 
 module.exports = {
   meta: {
+    type: 'layout',
     docs: {
       description: 'Ensure proper position of the first property in JSX',
       category: 'Stylistic Issues',
       recommended: false,
       url: docsUrl('jsx-first-prop-new-line'),
     },
-    fixable: 'code',
+    fixable: 'whitespace',
 
     messages,
 
--- a/lib/rules/jsx-first-prop-new-line.js
+++ b/lib/rules/jsx-first-prop-new-line.js
@@ -26,7 +26,7 @@ module.exports = {
       recommended: false,
       url: docsUrl('jsx-first-prop-new-line'),
     },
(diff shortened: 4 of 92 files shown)


sryze:
Rebased

lib/rules/jsx-closing-bracket-location.js
```diff
@@ -19,13 +19,14 @@ const messages = {
...
   meta: {
+    type: 'layout',
     docs: {
       description: 'Validate closing bracket location in JSX',
       category: 'Stylistic Issues',
       recommended: false,
       url: docsUrl('jsx-closing-bracket-location'),
     },
-    fixable: 'code',
+    fixable: 'whitespace',
```
ljharb, line 29: why is this change correct?

(if it is correct, i'd like to land it separately, before this PR)

lib/rules/jsx-curly-spacing.js
```diff
@@ -36,13 +36,14 @@ const messages = {
...
   meta: {
+    type: 'layout',
     docs: {
       description: 'Enforce or disallow spaces inside of curly braces in JSX attributes',
       category: 'Stylistic Issues',
       recommended: false,
       url: docsUrl('jsx-curly-spacing'),
     },
-    fixable: 'code',
+    fixable: 'whitespace',
```
ljharb, line 46: same here?

lib/rules/jsx-equals-spacing.js
```diff
@@ -21,13 +21,14 @@ const messages = {
...
   meta: {
+    type: 'layout',
     docs: {
       description: 'Disallow or enforce spaces around equal signs in JSX attributes',
       category: 'Stylistic Issues',
       recommended: false,
       url: docsUrl('jsx-equals-spacing'),
     },
-    fixable: 'code',
+    fixable: 'whitespace',
```
ljharb, line 31: and here?

lib/rules/jsx-first-prop-new-line.js
```diff
@@ -19,13 +19,14 @@ const messages = {
...
   meta: {
+    type: 'layout',
     docs: {
       description: 'Ensure proper position of the first property in JSX',
       category: 'Stylistic Issues',
       recommended: false,
       url: docsUrl('jsx-first-prop-new-line'),
     },
-    fixable: 'code',
+    fixable: 'whitespace',
```
ljharb, line 29: and here

lib/rules/jsx-indent-props.js
```diff
@@ -44,13 +44,14 @@ const messages = {
...
   meta: {
+    type: 'layout',
     docs: {
       description: 'Validate props indentation in JSX',
       category: 'Stylistic Issues',
       recommended: false,
       url: docsUrl('jsx-indent-props'),
     },
-    fixable: 'code',
+    fixable: 'whitespace',
```
ljharb, line 54: and here

lib/rules/jsx-max-props-per-line.js
```diff
@@ -25,13 +25,14 @@ const messages = {
...
   meta: {
+    type: 'layout',
     docs: {
       description: 'Limit maximum of props on a single line in JSX',
       category: 'Stylistic Issues',
       recommended: false,
       url: docsUrl('jsx-max-props-per-line'),
     },
-    fixable: 'code',
+    fixable: 'whitespace',
```
ljharb, line 35: and here

lib/rules/jsx-props-no-multi-spaces.js
```diff
@@ -19,13 +19,14 @@ const messages = {
...
   meta: {
+    type: 'layout',
     docs: {
       description: 'Disallow multiple spaces between inline JSX props',
       category: 'Stylistic Issues',
       recommended: false,
       url: docsUrl('jsx-props-no-multi-spaces'),
     },
-    fixable: 'code',
+    fixable: 'whitespace',
```
ljharb, line 29: and here

lib/rules/jsx-space-before-closing.js
```diff
@@ -24,14 +24,15 @@ const messages = {
...
+    type: 'layout',
     deprecated: true,
     docs: {
       description: 'Validate spacing before closing bracket in JSX',
       category: 'Stylistic Issues',
       recommended: false,
       url: docsUrl('jsx-space-before-closing'),
     },
-    fixable: 'code',
+    fixable: 'whitespace',
```
ljharb, line 35: and here

lib/rules/no-array-index-key.js
```diff
@@ -21,9 +21,10 @@ const messages = {
 
 module.exports = {
   meta: {
+    type: 'problem',
     docs: {
       description: 'Prevent usage of Array index in keys',
-      category: 'Best Practices',
+      category: 'Possible Errors',
```
ljharb, line 27: changing the category is a breaking change; also, this is indeed a best practice, and does not necessarily indicate a problem.

lib/rules/no-typos.js
```diff
@@ -29,9 +29,10 @@ const messages = {
 
 module.exports = {
   meta: {
+    type: 'problem',
     docs: {
       description: 'Prevent common typos',
-      category: 'Stylistic Issues',
+      category: 'Possible Errors',
```
ljharb, line 35: this is a breaking change, and also should be done in a separate PR.

lib/rules/void-dom-elements-no-children.js
```diff
@@ -49,9 +49,10 @@ const noChildrenInVoidEl = 'Void DOM element <{{element}} /> cannot receive chil
 
 module.exports = {
   meta: {
+    type: 'problem',
     docs: {
       description: 'Prevent passing of children to void DOM elements (e.g. `<br />`).',
-      category: 'Best Practices',
+      category: 'Possible Errors',
```
ljharb, line 55: also here

ljharb (changes requested):

lib/rules/jsx-closing-bracket-location.js
sryze: it was part of the original PR

I'm 100% not sure what this property means because the documentation doesn't mention what is the difference between `code` and `whitespace`; possibly `whitepsace` means that a fix will only change whitespace characters

sryze: @ljharb I've reverted the changes to `fixable` and `category` in made by the previous PR author as they are not related to `--fix-type` (purpose of this PR); please have a look, thanks!

sryze: looks like this change was correct, because the fix changes only spacing before the bracket (indents + newline), see https://github.com/eslint/eslint/discussions/15489#discussioncomment-1915826

ljharb: @sryze can we make a separate PR that changes all the "code"s to "whitespace"s?

PR #3167, opened 2022-01-05
[New] `jsx-no-target-blank`: Improve fixer with option `allowReferrer`

apepper:
> If you do not support Internet Explorer (any version), Chrome < 49, Opera < 36, Firefox < 52, desktop Safari < 10.1 or iOS Safari < 10.3, you may set `allowReferrer` to `true`, keep the HTTP Referer header and only add `rel="noopener"` to your links.

Source: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md#when-to-override-it

This PR improves the fixer to use `rel="noopener"` if `allowReferrer` is set.


ljharb (approved):
Thanks, looks great!

PR #3169, opened 2022-01-06
[Fix] `jsx-no-target-blank`: Improve readme

apepper:
See https://web.dev/external-anchors-use-rel-noopener/, https://chromestatus.com/feature/6140064063029248 and https://caniuse.com/mdn-html_elements_a_implicit_noopener for some background information on this change.

### DIFF
--- a/docs/rules/jsx-no-target-blank.md
+++ b/docs/rules/jsx-no-target-blank.md
@@ -123,6 +123,8 @@ This rule supports the ability to use custom components for forms. To enable thi
 
 ## When To Override It
 
+Modern browsers (Chrome ≥ 88, Edge ≥ 88, Firefox ≥ 79 and Safari ≥ 12.2) automatically imply `rel="noopener"`. Therefore this rule is no longer needed, if legacy browsers are not supported. See https://web.dev/external-anchors-use-rel-noopener/ and https://caniuse.com/mdn-html_elements_a_implicit_noopener for more details.
+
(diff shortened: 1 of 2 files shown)


docs/rules/jsx-no-target-blank.md
```diff
@@ -123,6 +123,8 @@ This rule supports the ability to use custom components for forms. To enable thi
 
 ## When To Override It
 
+Modern browsers (Chrome ≥ 88, Edge ≥ 88, Firefox ≥ 79 and Safari ≥ 12.2) automatically imply `rel="noopener"`. Therefore this rule is no longer needed, if legacy browsers are not supported. See https://web.dev/external-anchors-use-rel-noopener/ and https://caniuse.com/mdn-html_elements_a_implicit_noopener for more details.
```
ljharb, line 126: How does this interact with the browser support paragraph on line 130?

apepper: At least to my knowledge there are three stages regarding empty `ref`:

* legacy: `noreferrer` is needed (e.g. because of IE11). `noopener` is not good enough.
* in-between: In non IE11 browsers `noopener` is good enough (e.g. with Chrome 49 and newer)
* modern: In modern browser there is no need to specify `noopener` because it is automatically implied (e.g. with Chrome 88).

So basically: If you still support IE11 use the default of this eslint-rule. If you're not supporting IE11 but old versions of chrome or safari use the `allowReferrer` rule. If you're using modern browsers you don't need this rule at all.

ljharb: Ah, i see, thanks.

ljharb (approved):

PR #3173, opened 2022-01-11
Feat/index

HomyeeKing:
allow using `index` with other unique indentifier, like 

```js
 {
      code: 'foo.map((baz, i) => <Foo key={i + baz.id} />)',
    },
    {
      code: 'foo.map((baz, i) => <Foo key={`${baz.id}${i}`} />)',
    },
```

### DIFF
--- a/docs/rules/no-array-index-key.md
+++ b/docs/rules/no-array-index-key.md
@@ -67,6 +67,14 @@ things.map((thing) => (
   <Hello key={thing.id} />
 ));
 
+things.map((thing, i) => (
+  <Hello key={thing.id + i} />
+));
+
+things.map((thing, i) => (
+  <Hello key={`${thing.id}${i}`} />
+));
+
 things.map((thing) => (
   React.cloneElement(thing, { key: thing.id })
 ));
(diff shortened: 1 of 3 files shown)


HomyeeKing:
also, I don't find the relevant code about `MemberExpression ` branch, is that a missing or intend to be
image
https://github.com/HomyeeKing/eslint-plugin-react/blob/151bb2b13892969bea17b334e882eb422152c30a/lib/rules/no-array-index-key.js#L133

docs/rules/no-array-index-key.md
```diff
@@ -67,6 +67,14 @@ things.map((thing) => (
   <Hello key={thing.id} />
 ));
 
+things.map((thing, i) => (
+  <Hello key={thing.id + i} />
+));
+
+things.map((thing, i) => (
+  <Hello key={`${thing.id}${i}`} />
+));
```
ljharb, line 76: this is an antipattern, though - it shouldn't be allowed.

If the index is present in any way in the key, then all the stuff that's not the index is useless.

ljharb (changes requested):

PR #3174, opened 2022-01-11
[Fix] `no-invalid-html-attribute`: allow 'shortcut icon' on `link`

### LINKED ISSUE #3172: `react/no-invalid-html-attribute` seems to report false positives
Hey there, I've got an automatic lint bump in my repo that introduces `react/no-invalid-html-attribute` as a new rule and it fails with two/three errors:

https://github.com/Primajin/eyesbound/runs/4760311505?check_suite_focus=true

> Error - “home” is never a valid “rel” attribute value.

By adding rel="home" to a hyperlink, a page indicates that the destination of that hyperlink is the homepage of the site [...]
https://microformats.org/wiki/rel-home
This is a draft from 2005... Not sure if it was ever official?

> Error - “shortcut” is never a valid “rel” attribute value.

If I read https://mathiasbynens.be/notes/rel-shortcut-icon correctly this is not according to specification but older IE need it when the file is not exactly called `favicon.ico`, is that correct?

So are these arguably false positives, or instead everyone is just using it wrong and it was never a specification - thus I still wonder if everyone uses it kind of makes it a default 🤔 😅 
Thanks for looking into it folks.

Primajin:
Closes #3172 

## Description

For historical reasons, the icon keyword may be preceded by the keyword "shortcut".

[https://html.spec.whatwg.org/multipage/links.html#rel-icon](https://html.spec.whatwg.org/multipage/links.html#rel-icon) (last sentence before the next paragraph about `license`)

## Intentions

1. First of all I'm checking if there are pairs registered for the given attribute name after all.
2. Now I'm splitting the attribute value(s) into *overlapping* pairs. If you want to check on the funky regex: https://regex101.com/r/aJXAEz/1
3. now I'm iterating pair by pair
4. on each pair I check what are the possible siblings for my current item - currently there is only 1:1 but potentially there could be a pair where `foo` can go with `bar` OR `baz` but NEVER `qux`
5. now I check if my pair is separated by a single space or multiple in between
6. Check if the first value has a corresponding pairing
7. Get the last value as this is the "real" second one
8. Check if this is not one of the anticipated siblings
9. set message and message id based on, if the second value is truthy
    1. if it is the attribute was not paired with the right one
    2. if its not it stands alone which makes it invalid.
10. report the issue
11. check if there is only one option what the sibling could be, if there is only one anyways let's autofix it for the user! 🎉 

## Demo

This introduces two new messages:

```html
<link rel="shortcut"></link>
```
Reports: `“shortcut” must be directly followed by “icon”.`

<br><br>

```html
<link rel="shortcut manifest"></link>
```
Reports: `“shortcut” can not be directly followed by “manifest” without “icon”.`<br>

### DIFF
--- a/lib/rules/no-invalid-html-attribute.js
+++ b/lib/rules/no-invalid-html-attribute.js
@@ -39,8 +39,14 @@ const rel = new Map([
   ['prerender', new Set(['link'])],
   ['prev', new Set(['link', 'area', 'a', 'form'])],
   ['search', new Set(['link', 'area', 'a', 'form'])],
+  ['shortcut', new Set(['link'])], // generally allowed but needs pair with "icon"
+  ['shortcut\u0020icon', new Set(['link'])],
   ['stylesheet', new Set(['link'])],
   ['tag', new Set(['area', 'a'])],
+]);
+
+const pairs = new Map([
+  ['shortcut', new Set(['icon'])],
 ]);
 
 /**
@@ -49,6 +55,14 @@ const rel = new Map([
  */
 const VALID_VALUES = new Map([
   ['rel', rel],
+]);
+
+/**
+ * Map between attributes and a mapping between pair-values and a set of values they are valid with
+ * @type {Map<string, Map<string, Set<string>>>}
+ */
+const VALID_PAIR_VALUES = new Map([
+  ['rel', pairs],
 ]);
 
 /**
@@ -216,6 +230,8 @@ const messages = {
   noMethod: 'The ”{{attributeName}}“ attribute cannot be a method.',
   onlyMeaningfulFor: 'The ”{{attributeName}}“ attribute only has meaning on the tags: {{tagNames}}',
   emptyIsMeaningless: 'An empty “{{attributeName}}” attribute is meaningless.',
+  notAlone: '“{{reportingValue}}” must be directly followed by “{{missingValue}}”.',
+  notPaired: '“{{reportingValue}}” can not be directly followed by “{{secondValue}}” without “{{missingValue}}”.',
 };
 
 function splitIntoRangedParts(node, regex) {
@@ -256,10 +272,10 @@ function checkLiteralValueNode(context, attributeName, node, parentNode, parentN
     return;
   }
 
-  const parts = splitIntoRangedParts(node, /([^\s]+)/g);
-  for (const part of parts) {
-    const allowedTags = VALID_VALUES.get(attributeName).get(part.value);
-    const reportingValue = part.reportingValue;
+  const singleAttributeParts = splitIntoRangedParts(node, /(\S+)/g);
+  for (const singlePart of singleAttributeParts) {
+    const allowedTags = VALID_VALUES.get(attributeName).get(singlePart.value);
+    const reportingValue = singlePart.reportingValue;
     if (!allowedTags) {
       report(context, messages.neverValid, 'neverValid', {
         node,
@@ -268,7 +284,7 @@ function checkLiteralValueNode(context, attributeName, node, parentNode, parentN
           reportingValue,
         },
         fix(fixer) {
-          return fixer.removeRange(part.range);
+          return fixer.removeRange(singlePart.range);
         },
       });
     } else if (!allowedTags.has(parentNodeName)) {
@@ -280,20 +296,54 @@ function checkLiteralValueNode(context, attributeName, node, parentNode, parentN
           elementName: parentNodeName,
         },
         fix(fixer) {
-          return fixer.removeRange(part.range);
+          return fixer.removeRange(singlePart.range);
         },
       });
     }
   }
 
+  const allowedPairsForAttribute = VALID_PAIR_VALUES.get(attributeName);
+  if (allowedPairsForAttribute) {
+    const pairAttributeParts = splitIntoRangedParts(node, /(?=(\b\S+\s*\S+))/g);
+    for (const pairPart of pairAttributeParts) {
+      for (const [pairing, siblings] of allowedPairsForAttribute) {
+        const attributes = pairPart.reportingValue.split('\u0020');
+        const [firstValue, secondValue] = attributes;
+        if (firstValue === pairing) {
+          const lastValue = attributes[attributes.length - 1]; // in case of multiple white spaces
+          if (!siblings.has(lastValue)) {
+            const message = secondValue ? messages.notPaired : messages.notAlone;
+            const messageId = secondValue ? 'notPaired' : 'notAlone';
+            report(context, message, messageId, {
+              node,
+              data: {
+                reportingValue: firstValue,
+                secondValue,
+                missingValue: [...siblings].join(', '),
+              },
+            });
+          }
+        }
+      }
+    }
+  }
+
   const whitespaceParts = splitIntoRangedParts(node, /(\s+)/g);
   for (const whitespacePart of whitespaceParts) {
-    if (whitespacePart.value !== ' ' || whitespacePart.range[0] === (node.range[0] + 1) || whitespacePart.range[1] === (node.range[1] - 1)) {
+    if (whitespacePart.range[0] === (node.range[0] + 1) || whitespacePart.range[1] === (node.range[1] - 1)) {
       report(context, messages.spaceDelimited, 'spaceDelimited', {
         node,
         data: { attributeName },
         fix(fixer) {
           return fixer.removeRange(whitespacePart.range);
+        },
+      });
+    } else if (whitespacePart.value !== '\u0020') {
+      report(context, messages.spaceDelimited, 'spaceDelimited', {
+        node,
+        data: { attributeName },
+        fix(fixer) {
+          return fixer.replaceTextRange(whitespacePart.range, '\u0020');
         },
       });
     }
--- a/tests/lib/rules/no-invalid-html-attribute.js
+++ b/tests/lib/rules/no-invalid-html-attribute.js
@@ -130,6 +130,9 @@ ruleTester.run('no-invalid-html-attribute', rule, {
     { code: '<link rel="icon"></link>' },
     { code: 'React.createElement("link", { rel: "icon" })' },
     { code: 'React.createElement("link", { rel: ["icon"] })' },
+    { code: '<link rel="shortcut icon"></link>' },
+    { code: 'React.createElement("link", { rel: "shortcut icon" })' },
+    { code: 'React.createElement("link", { rel: ["shortcut icon"] })' },
     { code: '<link rel="license"></link>' },
     { code: 'React.createElement("link", { rel: "license" })' },
     { code: 'React.createElement("link", { rel: ["license"] })' },
@@ -451,6 +454,26 @@ ruleTester.run('no-invalid-html-attribute', rule, {
       ],
     },
     {
+      code: '<a rel="noreferrer        noopener"></a>',
+      output: '<a rel="noreferrer noopener"></a>',
+      errors: [
+        {
+          messageId: 'spaceDelimited',
+          data: { attributeName: 'rel' },
+        },
+      ],
+    },
+    {
+      code: '<a rel="noreferrer\xa0\xa0noopener"></a>',
+      output: '<a rel="noreferrer noopener"></a>',
+      errors: [
+        {
+          messageId: 'spaceDelimited',
+          data: { attributeName: 'rel' },
+        },
+      ],
+    },
+    {
       code: '<a rel={"noreferrer noopener foobar"}></a>',
       output: '<a rel={"noreferrer noopener "}></a>',
       errors: [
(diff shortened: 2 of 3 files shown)


tests/lib/rules/no-invalid-html-attribute.js
```diff
@@ -130,6 +130,9 @@ ruleTester.run('no-invalid-html-attribute', rule, {
     { code: '<link rel="icon"></link>' },
     { code: 'React.createElement("link", { rel: "icon" })' },
     { code: 'React.createElement("link", { rel: ["icon"] })' },
+    { code: '<link rel="shortcut icon"></link>' },
+    { code: 'React.createElement("link", { rel: "shortcut icon" })' },
+    { code: 'React.createElement("link", { rel: ["shortcut icon"] })' },
```
ljharb, line 135: these should be valid, but shouldn't `shortcut` and `shortcut foo` and `foo shortcut` all be invalid?

Primajin: Yep but I am having a hard time wrapping my head around the code and the error messages.

So normally `neverValid` would trigger - which is kind of not true - because paired with `icon` it is valid on link.
I changed the Map item to `['shortcut\u0020icon', new Set(['link'])],` which also doesn't help because then it still reports `shortcut` to be never valid and I don't want to allow `shortcut` to be valid alone 🤔

ljharb: since this is a special case of two words, we might need to alter the rule where it splits on a space.

Primajin: Not feeling super comfortable doing that on my first contribution to this project though 😅  Wouldn't really know where to start and keep it clean.

ljharb: That’s what code review is for, if you’re up for it :-)

Primajin: Alright I'll take a stab at it 🔪

Primajin: OK did some progress, need to take another look regarding multiple spaces, signing off for today 😄

lib/rules/no-invalid-html-attribute.js
```diff
@@ -51,152 +57,160 @@ const VALID_VALUES = new Map([
...
-  'body',
+  'a',
+  'abbr',
+  'acronym',
   'address',
+  'applet',
+  'area',
   'article',
   'aside',
-  'footer',
```
ljharb, line 70: let's revert sorting this list so the diff is cleaner

lib/rules/no-invalid-html-attribute.js
```diff
@@ -208,14 +222,16 @@ const COMPONENT_ATTRIBUTE_MAP = new Map();
...
-  noEmpty: 'An empty “{{attributeName}}” attribute is meaningless.',
+  emptyIsMeaningless: 'An empty “{{attributeName}}” attribute is meaningless.',
   neverValid: '“{{reportingValue}}” is never a valid “{{attributeName}}” attribute value.',
-  notValidFor: '“{{reportingValue}}” is not a valid “{{attributeName}}” attribute value for <{{elementName}}>.',
-  spaceDelimited: '”{{attributeName}}“ attribute values should be space delimited.',
+  noEmpty: 'An empty “{{attributeName}}” attribute is meaningless.',
   noMethod: 'The ”{{attributeName}}“ attribute cannot be a method.',
+  notAlone: '“{{reportingValue}}” must be directly followed by “{{missingValue}}”.',
+  notPaired: '“{{reportingValue}}” can not be directly followed by “{{secondValue}}” without “{{missingValue}}”.',
+  notValidFor: '“{{reportingValue}}” is not a valid “{{attributeName}}” attribute value for <{{elementName}}>.',
```
ljharb, line 231: also this - let's revert all unrelated formatting/sorting

ljharb (changes requested):
Let's add some tests that show that `shortcut` by itself is invalid.

lib/rules/no-invalid-html-attribute.js
Primajin: sure, this was just for me to get acquainted with the code. Sorting things help me anticipate what the next item in the list will be and allow me to spot missing values, e.g. when there was no `strong` after `strike` I'd directly see that it would be missing.
Reverted in 5d92d00

lib/rules/no-invalid-html-attribute.js
Primajin: Sure 67b05f7

Primajin:
> Let's add some tests that show that `shortcut` by itself is invalid.

Yes that should be already in:
```js
    {
      code: '<link rel="shortcut"></link>',
      output: '<link rel="shortcut icon"></link>',
      errors: [
        {
          messageId: 'notAlone',
          data: {
            reportingValue: 'shortcut',
            missingValue: 'icon',
          },
        },
      ],
    },
```

Just `shortcut` can not be alone and will be auto-fixed to `shortcut icon`.


Now next I need to look into some spacing that is still added:
image

Hope to be able to solve this soon 🤞🏻 
When everything is ready I will mark the PR as ... "not draft" 😅 and will write up some description what I did and what my intentions were.

Primajin:
OK I see that I can't actually do two things in one go, so even

```js
      code: '<a rel={"batgo noopener"}></a>',
      output: '<a rel={" noopener"}></a>',
```

doesn't turn directly into only `noopener` but it is expected that there is an (invalid) leading space, so then I changed the expectations on my newly added code to also expect a trailing space that I just couldn't manage to get rid of in one go (will be picked up by `spaceDelimited` rule in the next cycle).

I added my intentions kind of line by line so that you can understand the code better for making suggestions.
Happy / Ready for round two 😄 👍🏻

Primajin:
Ohh and there was an issue on the white-space part - before we always removed everything that is not exactly one space character - but that means that two spaces between two words will just be removed, instead we should replace all white-space that is not in the beginning or the end to exactly one space character. For security this should also include non-breaking-spaces (e.g. when copied from somewhere else) - so I added another test case that covers that.

lib/rules/no-invalid-html-attribute.js
ljharb: Totally fine to do it in a separate PR, just makes this one harder to review :-)

ljharb:
We should not autofix `shortcut` to `shortcut icon`; that's not a safe autofix because it will change runtime behavior. It should be a warning that requires the user to manually fix it.

tests/lib/rules/no-invalid-html-attribute.js
```diff
@@ -373,6 +376,26 @@ ruleTester.run('no-invalid-html-attribute', rule, {
...
+      output: '<a rel="noreferrer noopener"></a>',
+      errors: [
+        {
+          messageId: 'spaceDelimited',
+          data: { attributeName: 'rel' },
+        },
+      ],
+    },
+    {
+      code: '<a rel="noreferrer  noopener"></a>', // this string contains nbsp
```
ljharb, line 390: could we do this using a template literal and `'\xa0'`?

tests/lib/rules/no-invalid-html-attribute.js
```diff
@@ -537,6 +560,74 @@ ruleTester.run('no-invalid-html-attribute', rule, {
         },
       ],
     },
+    {
+      code: '<link rel="shortcut"></link>',
+      output: '<link rel="shortcut icon"></link>',
```
ljharb, line 565: as commented; this isn't a safe autofix.
```suggestion
      code: '<link rel="shortcut"></link>',
```

tests/lib/rules/no-invalid-html-attribute.js
```diff
@@ -537,6 +560,74 @@ ruleTester.run('no-invalid-html-attribute', rule, {
...
+          data: {
+            reportingValue: 'shortcut',
+            missingValue: 'icon',
+          },
+        },
+      ],
+    },
+    {
+      code: '<link rel="shortcut foo"></link>',
+      output: '<link rel="shortcut icon "></link>',
```
ljharb, line 578:
```suggestion
      code: '<link rel="shortcut foo"></link>',
```

tests/lib/rules/no-invalid-html-attribute.js
```diff
@@ -537,6 +560,74 @@ ruleTester.run('no-invalid-html-attribute', rule, {
...
+      errors: [
+        {
+          messageId: 'spaceDelimited',
+          data: { attributeName: 'rel' },
+        },
+      ],
+    },
+    {
+      code: '<link rel="shortcut  foo"></link>',
+      output: '<link rel="shortcut icon  "></link>',
```
ljharb, line 609:
```suggestion
      code: '<link rel="shortcut  foo"></link>',
      output: '<link rel="shortcut foo"></link>',
```
this is a fine autofix, but the test case should still be invalid after it's done.

Primajin:
Thanks for the feedback! Will work on it! 👍🏻

tests/lib/rules/no-invalid-html-attribute.js
Primajin: It doesn't let me use template literal but it seems to work just fine with single quotes.
eslint-error
single-quotes

tests/lib/rules/no-invalid-html-attribute.js
Primajin: 4de38ef

tests/lib/rules/no-invalid-html-attribute.js
Primajin: 4de38ef

tests/lib/rules/no-invalid-html-attribute.js
Primajin: 4de38ef

tests/lib/rules/no-invalid-html-attribute.js
Primajin: Changed in 4de38ef - should work with single quotes as well (test passing locally)

ljharb: oh sure, that works, i forgot this project doesn’t use babel

tests/lib/rules/no-invalid-html-attribute.js
```diff
@@ -537,6 +560,73 @@ ruleTester.run('no-invalid-html-attribute', rule, {
...
+          data: {
+            reportingValue: 'shortcut',
+            missingValue: 'icon',
+          },
+        },
+      ],
+    },
+    {
+      code: '<link rel="shortcut foo"></link>',
+      output: '<link rel="shortcut "></link>',
```
ljharb, line 577:
```suggestion
      output: '<link rel="shortcut"></link>',
```

ljharb (approved):
Looks great! Just one nit.

tests/lib/rules/no-invalid-html-attribute.js
Primajin: Hmm but it actually does that - it's the opposite of

```js
      code: '<a rel={"batgo noopener"}></a>',
      output: '<a rel={" noopener"}></a>',
```

Primajin:
```
 1) no-invalid-html-attribute
       invalid
         <link rel="shortcut foo"></link>:

      AssertionError [ERR_ASSERTION]: Output is incorrect.
      + expected - actual

      -<link rel="shortcut "></link>
      +<link rel="shortcut"></link>
```

The test fails when I remove it

Primajin: Might already be covered by

```js
      code: '<a rel="noreferrer noopener foobar"></a>',
      output: '<a rel="noreferrer noopener "></a>',
```
Same behaviour - I could just remove it maybe?

Primajin: Ahh no, we shouldn't remove it - it is the testcase for the `notPaired` branch (in contrast to `notAlone`)

ljharb: I'm basically saying that the autofix output should never have leading or trailing spaces.

Primajin: Yeah that also confused me, but apparently that's "how things used to be" in `master` as well:

https://github.com/yannickcr/eslint-plugin-react/blob/0d8946212a1ba560b12e2e01807033e19dc774f7/tests/lib/rules/no-invalid-html-attribute.js#L355

https://github.com/yannickcr/eslint-plugin-react/blob/0d8946212a1ba560b12e2e01807033e19dc774f7/tests/lib/rules/no-invalid-html-attribute.js#L378

https://github.com/yannickcr/eslint-plugin-react/blob/0d8946212a1ba560b12e2e01807033e19dc774f7/tests/lib/rules/no-invalid-html-attribute.js#L403

https://github.com/yannickcr/eslint-plugin-react/blob/0d8946212a1ba560b12e2e01807033e19dc774f7/tests/lib/rules/no-invalid-html-attribute.js#L416

https://github.com/yannickcr/eslint-plugin-react/blob/0d8946212a1ba560b12e2e01807033e19dc774f7/tests/lib/rules/no-invalid-html-attribute.js#L477

Apparently it only runs one cycle when fixing and we would need to find a way to kind of `while` loop so long until it can't fix any more issues - but that sounds like a major refactor to me 🤔

ljharb: fair point. altho it seems like adding `.trim()` to the attribute contents wouldn't be that difficult?

Primajin: Good idea! But where would I run this, I think I haven't fully grasped when/how fix runs. Would this be part of fix? Or would we just always in the end after all checks run another trim as it's never good/valid to keep it?

ljharb: Adding a single fixer on program exit that trimmed any attribute strings that had been previously autofixed (we’d have to track that manually) would be good. I think that should wait for a followup PR, tho, because it was a pre-existing problem. 

I’ll land this tomorrow, as-is.

Primajin:
Yay 🎉 Thanks for your support - in the end it was a lot of fun, even when it was a bit scary first 🚀

ljharb:
Heads up that I did need to land https://github.com/yannickcr/eslint-plugin-react/commit/5f49f51dd237dd4c6758bc0c3e5b300a28770966, due to array destructuring and spread syntax not working on node 4 and 5.

ljharb:
I also filed #3180 to track updating the implementation - I think now that we have "pairs", and good test cases, it'd be worth rewriting it from scratch.

lib/rules/no-invalid-html-attribute.js
Primajin: https://github.com/yannickcr/eslint-plugin-react/pull/3182 😉

PR #3182, opened 2022-01-19
[Refactor] `no-invalid-html-attribute`: sort HTML_ELEMENTS and messages

Primajin:
Sorting things help me anticipate what the next item in the list will be and allow me to spot missing values, e.g. when there was no strong after strike I'd directly see that it would be missing.

> Totally fine to do it in a separate PR, just makes this one harder to review :-)
_Originally posted by @ljharb in https://github.com/yannickcr/eslint-plugin-react/pull/3174#discussion_r785464082_

This is a pure `chore` commit - it only re-arranges/sorts the code - no additions or removals.


ljharb (approved):

ljharb:
(I renamed it because the word "chore" is wildly inappropriate; chores are miserable things that you're forced to do by an authority figure :-) maintenance tasks and refactors are good things, and shouldn't invite that kind of connotation)

Primajin:
Hahaha 😅 OK! 👍🏻
I wasn't sure how much folks are into Conventional Commit style here. Refactor sounds much much better

ljharb:
That tooling is great, but unfortunately the actual style is terrible, and offers no way to configure it.

PR #3195, opened 2022-02-01
Handle TSTypeReference in no-unused-prop-types

niik:
:wave: from GitHub Desktop land!

We're currently upgrading to TypeScript 4.5 (https://github.com/desktop/desktop/pull/13768) and while doing that I wanted to start using the `react/no-unused-props-type` rule instead of the `react-unused-props-and-state` from the obsolete `tslint-microsoft-contrib` package.

Unfortunately I couldn't get the rule to produce any errors. I put up a minimal reproduction repository at https://github.com/niik/no-unused-prop-types-repro which illustrates the problem. To test it just clone and run `yarn lint`

Doing so very haphazardly troubleshooting I noticed that when encountering our prop types in `propTypes.js` they were of type `TSTypeReference` which was not being handled in the switch case.  Spotting that `DeclarePropTypesForTSTypeAnnotation` seemed to be able to handle both `TSTypeAnnotation` and `TSTypeReference` internally I simply tried adding it to the switch and lo-and-behold the rule just started working.

Now, obviously I have no idea what I'm doing here so I'm happy to adjust my approach given some guidance from someone who understands eslint/ts-parser internals but running my minimal reproduction repository with my branch (https://github.com/niik/no-unused-prop-types-repro/tree/with-ts-type-reference-fix) reports the error correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 * [`prop-types`], `propTypes`: add support for exported type inference ([#3163][] @vedadeepta)
 * [`no-invalid-html-attribute`]: allow 'shortcut icon' on `link` ([#3174][] @Primajin)
 * [`prefer-exact-props`] improve performance for `Identifier` visitor ([#3190][] @meowtec)
+* `propTypes`: Handle TSTypeReference in no-unused-prop-type ([#3195][] @niik)
 
 ### Changed
 * [readme] change [`jsx-runtime`] link from branch to sha ([#3160][] @tatsushitoji)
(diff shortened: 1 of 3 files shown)


ljharb:
The fix looks great, but if we don't have a regression test, it will guaranteed be broken again in the future :-)

Your repro looks pretty small, so it should be straightforward to adapt it into a test case. Want to give it a shot?

ljharb:
Looks like v5.10, and we're already testing on `^5`. I'd hope they didn't change the name of the parse nodes in a minor version, but if they had I'd have expected to see failures sooner.

ljharb:
I reran the passing tests on master from 3 days ago, and they continued to pass: https://github.com/yannickcr/eslint-plugin-react/actions/runs/1767687068 - so we do need some kind of new coverage, or possibly, to upgrade to TS 4.5.

That would imply it's a breaking change in TS tho.

ljharb:
So, I couldn't make this test fail locally. Then, I installed different versions of TS and the TS eslint parser, and was able to make it fail. Then, I couldn't get it to NOT fail (absent the fix).

So, I guess we'll land it, but something weird is going on.

ljharb (approved):

Issue #3170, opened 2022-01-06
Support for `multilineLast` in  `jsx-sort-props`


duhamelgm:
Hello! 

We have an use case that its not covered in `jsx-sort-props`. 
For readability purposes we prefer to have multiline props at the bottom. We found that long components become hard to read when following only alphabetical order, and it's prefered to split them by the complexity of the prop instead.

```javascript
// Current jsx-sort-props 
<Field
  float
  className={classNames(classes.inputWidth, {
    [classes.noBorder]: isActive === "values",
  })}
  disabled={isDisabled}
  initialValue={computePercentage(number, count)}
  InputProps={{
    ...customInputProps,
  }}
  inputProps={{
    className: inputClassName,
  }}
  min="0"
  name="fieldName"
  placeholder={getTranslation("field")}
  validate={validate}
/>
```
```javascript
// Using a possible multilineLast rule option
<Field
  float
  disabled={isDisabled}
  initialValue={computePercentage(number, count)}
  min="0"
  name="fieldName"
  placeholder={getTranslation("field")}
  validate={validate}
  className={classNames(classes.inputWidth, {
    [classes.noBorder]: isActive === "values",
  })}
  InputProps={{
    ...customInputProps
  }}
  inputProps={{
    className: inputClassName,
  }}
/>
```

Should this option be implemented in `jsx-sort-props`?

ljharb:
It's an interesting idea, and certainly achievable.

I'm not sure it's common enough to warrant the complexity tho, and it kind of feels like it's papering over the real problem: that a component takes so many props in the first place that reading it alphabetically is difficult.

duhamelgm:
I think there's no easy way to validate if it's common enough or not, but we can see a use case in some components of commonly used libraries like `react-final-form` or `material-ui` which receive multiple props of multilayer complexity.

And I agree that having multiple multiline props in a component might feel like an antipattern for some people, but I think that if you want to enforce that behavior in your codebase you should use a different rule set that checks that for you. IMO `jsx-sort-props` should only care about props readability and not prop complexity conventions.

ljharb:
I suppose an option that treats multiline props as a separate group, allows it to be configured "first" or "last", and sorts each group of (multiline, single-line) props according to all the other rules, would be reasonable.

The PR adding it should please provide extensive test cases, potentially duplicating every other existing tested scenario.

PR #3198, opened 2022-02-06
[New] `jsx-sort-props`: support multiline prop groups

duhamelgm:
Hello :wave: 

This PR Adds a new multiline group for the `jsx-sort-props` rule. Related to the issue: #3170

The shorthand and callback rules take precedence over the multiline rule set

### DIFF
--- a/docs/rules/jsx-sort-props.md
+++ b/docs/rules/jsx-sort-props.md
@@ -29,6 +29,7 @@ Examples of **correct** code for this rule:
   "callbacksLast": <boolean>,
   "shorthandFirst": <boolean>,
   "shorthandLast": <boolean>,
+  "multiline": "ignore" | "first" | "last",
   "ignoreCase": <boolean>,
   "noSortAlphabetically": <boolean>,
   "reservedFirst": <boolean>|<array<string>>,
@@ -68,6 +69,42 @@ When `true`, short hand props must be listed after all other props (unless `call
 
 ```jsx
 <Hello name="John" tel={5555555} active validate />
+```
+
+### `multiline`
+
+Enforced sorting for multiline props
+
+* `ignore`: Multiline props will not be taken in consideration for sorting.
+
+* `first`: Multiline props must be listed before all other props (unless `shorthandFirst` is set), but still respecting the alphabetical order.
+
+* `last`: Multiline props must be listed after all other props (unless either `callbacksLast` or `shorthandLast` are set), but still respecting the alphabetical order.
+
+Defaults to `ignore`.
+
+```jsx
+// 'jsx-sort-props': [1, { multiline: 'first' }]
+<Hello
+  classes={{
+    greetings: classes.greetings,
+  }}
+  active
+  validate
+  name="John"
+  tel={5555555}
+/>
+
+// 'jsx-sort-props': [1, { multiline: 'last' }]
+<Hello
+  active
+  validate
+  name="John"
+  tel={5555555}
+  classes={{
+    greetings: classes.greetings,
+  }}
+/>
 ```
 
 ### `noSortAlphabetically`
--- a/lib/rules/jsx-sort-props.js
+++ b/lib/rules/jsx-sort-props.js
@@ -6,6 +6,7 @@
 'use strict';
 
 const propName = require('jsx-ast-utils/propName');
+const includes = require('array-includes');
 const docsUrl = require('../util/docsUrl');
 const jsxUtil = require('../util/jsx');
 const report = require('../util/report');
@@ -18,6 +19,10 @@ function isCallbackPropName(name) {
   return /^on[A-Z]/.test(name);
 }
 
+function isMultilineProp(node) {
+  return node.loc.start.line !== node.loc.end.line;
+}
+
 const messages = {
   noUnreservedProps: 'A customized reserved first list must only contain a subset of React reserved props. Remove: {{unreservedWords}}',
   listIsEmpty: 'A customized reserved first list must not be empty',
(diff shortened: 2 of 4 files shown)


docs/rules/jsx-sort-props.md
```diff
@@ -29,6 +29,8 @@ Examples of **correct** code for this rule:
   "callbacksLast": <boolean>,
   "shorthandFirst": <boolean>,
   "shorthandLast": <boolean>,
+  "multilineFirst": <boolean>,
+  "multilineFirst": <boolean>,
```
ljharb, line 33:
```suggestion
  "multilineFirst": <boolean>,
  "multilineLast": <boolean>,
```

tests/lib/rules/jsx-sort-props.js
```diff
@@ -494,6 +608,170 @@ ruleTester.run('jsx-sort-props', rule, {
...
+          name="fieldName"
+          placeholder={getTranslation("field")}
+          inputProps={{
+            className: inputClassName,
+          }}
+          onValidate={validate}
+          {...rest}
+        />
+      `,
+      errors: 6,
```
ljharb, line 774: let's use an array of objects with messageId and line number, so we can make sure the right lines are being warned.

ljharb (changes requested):
Rather than two mutually exclusive boolean props, this would be better as an enum value for "multiline" - ignore (default), first, or last, perhaps?

lib/rules/jsx-sort-props.js
```diff
@@ -333,7 +360,7 @@ module.exports = {
             }
             if (!currentValue && previousValue) {
               report(context, messages.listShorthandFirst, 'listShorthandFirst', {
-                node: memo.name,
+                node: decl.name,
```
duhamelgm, line 363: I think we had an issue here with the `shorthandFirst` node returned in the report. I think that we're expecting the report to be assigned to the shorthand prop and not to the prop that's being compared to. Similar to the report that we do for `reservedFirst`

tests/lib/rules/jsx-sort-props.js
```diff
@@ -494,6 +608,214 @@ ruleTester.run('jsx-sort-props', rule, {
...
+      `,
+    },
+    {
+      code: `
+        <App
+          a={{
+            aA: 1,
+          }}
+          b
+          inline={1}
```
duhamelgm, line 683: We have a limitation with the current implementation here, whenever conflicting groups are in order between themselves, but in disorder with other props, only the group with higher hierarchy will be reported.

For example here we enabled `{ multiline: 'last', shorthandLast: true }`, but only the shorthand prop is being reported as an error, since multiline thinks that it's in order.

I feel like there's no easy way to fix this unless we change the way in which `jsx-sort-props` checks and order the props, so that it check props by groups instead of the direct adjacent neighbors of each prop. I wouldn't mind working on that, but I feel like that's out of the scope of this PR...

duhamelgm:
@ljharb I was able to work on the given feedback, now the property is an enum rather than two mutually exclusive props, and the tests are now expecting line numbers for each specific error. There was a little issue regarding how the `shorthandFirst` property gave its report so I changed that as well. Let me know what you think, and thanks for taking the time to look into this!

lib/rules/jsx-sort-props.js
```diff
@@ -75,6 +81,18 @@ function contextCompare(a, b, options) {
     }
   }
 
+  if (options.multiline === 'first' || options.multiline === 'last') {
```
ljharb, line 84:
```suggestion
  if (options.multiline !== 'ignore') {
```

lib/rules/jsx-sort-props.js
```diff
@@ -241,6 +261,10 @@ module.exports = {
         shorthandLast: {
           type: 'boolean',
         },
+        // Whether multiline properties should be listed first or last
+        multiline: {
+          enum: ['ignore', 'first', 'last'],
```
ljharb, line 266:
```suggestion
          enum: ['ignore', 'first', 'last'],
          default: 'ignore',
```
we still need the explicit default, but this is nice documentation

tests/lib/rules/jsx-sort-props.js
ljharb: I agree; that sounds like a great improvement, but belongs in another PR.

tests/lib/rules/jsx-sort-props.js
```diff
@@ -494,6 +608,214 @@ ruleTester.run('jsx-sort-props', rule, {
...
+          line: 13,
+        },
+        {
+          messageId: 'listCallbacksLast',
+          line: 19,
+        },
+        {
+          messageId: 'listCallbacksLast',
+          line: 19,
+        },
```
ljharb, line 817: same here
```suggestion
        {
          messageId: 'listMultilineLast',
          line: 4,
        }
        {
          messageId: 'listReservedPropsFirst',
          line: 12,
        },
        {
          messageId: 'listShorthandFirst',
          line: 13,
        },
        {
          messageId: 'listCallbacksLast',
          line: 19,
        }
```

tests/lib/rules/jsx-sort-props.js
```diff
@@ -494,6 +608,214 @@ ruleTester.run('jsx-sort-props', rule, {
...
+          line: 8,
+        },
+        {
+          messageId: 'listCallbacksLast',
+          line: 8,
+        },
+        {
+          messageId: 'listCallbacksLast',
+          line: 8,
+        },
```
ljharb, line 711: three reports on the same line seems like an issue; we may need to do some memoizing (via a WeakMap) to prevent reporting on a node we've already reported on.
```suggestion
        {
          messageId: 'listShorthandLast',
          line: 6,
        },
        {
          messageId: 'listCallbacksLast',
          line: 8,
        },
```

duhamelgm: Makes sense, I just added a method to memoize the reported node attributes

ljharb (approved):
Nicely done, and very thorough.

ljharb:
One more change, because node < 6 doesn't have array `.includes` :-)

PR #3202, opened 2022-02-11
[Docs] `jsx-key`: improve example 

chnakamura:
Using an index for a key is a violation of the [no-array-index-key](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md) lint rule, so it feel weird to me that we're showing it as an example for "correct" code, even though it isn't strictly a violation of the `js-key` rule.

### DIFF
--- a/docs/rules/jsx-key.md
+++ b/docs/rules/jsx-key.md
@@ -22,7 +22,7 @@ Examples of **correct** code for this rule:
 ```jsx
 [<Hello key="first" />, <Hello key="second" />, <Hello key="third" />];
 
-data.map((x, i) => <Hello key={i}>{x}</Hello>);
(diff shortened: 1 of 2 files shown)


docs/rules/jsx-key.md
```diff
@@ -22,7 +22,7 @@ Examples of **correct** code for this rule:
 ```jsx
 [<Hello key="first" />, <Hello key="second" />, <Hello key="third" />];
 
-data.map((x, i) => <Hello key={i}>{x}</Hello>);
+data.map(x => <Hello key={x.id}>{x}</Hello>);
```
ljharb, line 25:
```suggestion
data.map((x) => <Hello key={x.id}>{x}</Hello>);
```

ljharb (approved):
thanks!

PR #3203, opened 2022-02-12
[New] add `jsx-no-leaked-render`

Belco90:
New rule `jsx-no-leaked-zero` to prevent potential 0 leaked to the DOM from numeric inline conditions. It includes 2 autofix strategies.

### DIFF
--- a/docs/rules/jsx-no-leaked-render.md
+++ b/docs/rules/jsx-no-leaked-render.md
@@ -1,0 +1,208 @@
+# Prevent problematic leaked values from being rendered (react/jsx-no-leaked-render)
+
+Using the `&&` operator to render some element conditionally in JSX can cause unexpected values being rendered, or even crashing the rendering.
+
+
+## Rule Details
+
+This rule aims to prevent dangerous leaked values from being rendered since they can cause unexpected values reaching the final DOM or even crashing your render method.
+
+In React, you might end up rendering unexpected values like `0` or `NaN`. In React Native, your render method will crash if you render `0`, `''`, or `NaN`:
+
+```jsx
+const Example = () => {
+  return (
+    <>
+      {0 && <Something/>}
+      {/* React: renders undesired 0 */}
+      {/* React Native: crashes 💥 */}
+      
+      {'' && <Something/>}
+      {/* React: renders nothing */}
+      {/* React Native: crashes 💥 */}
+      
+      {NaN && <Something/>}
+      {/* React: renders undesired NaN */}
+      {/* React Native: crashes 💥 */}
+    </>
+  )
+}
+```
+
+This can be avoided by:
+- coercing the conditional to a boolean: `{!!someValue && <Something />}`
+- transforming the binary expression into a ternary expression which returns `null` for falsy values: `{someValue ? <Something /> : null}`
+
+This rule is autofixable; check the Options section to read more about the different strategies available.
+
+Examples of **incorrect** code for this rule:
+
+```jsx
+const Component = ({ count, title }) => {
+  return <div>{count && title}</div>
+}
+```
+
+```jsx
+const Component = ({ count }) => {
+  return <div>{count && <span>There are {count} results</span>}</div>
+}
+```
+
+```jsx
+const Component = ({ elements }) => {
+  return <div>{elements.length && <List elements={elements}/>}</div>
+}
+```
+
+```jsx
+const Component = ({ nestedCollection }) => {
+  return (
+    <div>
+      {nestedCollection.elements.length && <List elements={nestedCollection.elements} />}
+    </div>
+  )
+}
+```
+
+```jsx
+const Component = ({ elements }) => {
+  return <div>{elements[0] && <List elements={elements}/>}</div>
+}
+```
+
+```jsx
+const Component = ({ numberA, numberB }) => {
+  return <div>{(numberA || numberB) && <Results>{numberA+numberB}</Results>}</div>
+}
+```
+
+```jsx
+// If the condition is a boolean value, this rule will report the logical expression
+// since it can't infer the type of the condition.
+const Component = ({ someBool }) => {
+  return <div>{someBool && <Results>{numberA+numberB}</Results>}</div>
+}
+```
+
+Examples of **correct** code for this rule:
+
+```jsx
+const Component = ({ elements }) => {
+  return <div>{elements}</div>
+}
+```
+
+```jsx
+// An OR condition it's considered valid since it's assumed as a way
+// to render some fallback if the first value is falsy, not to render something conditionally.
+const Component = ({ customTitle }) => {
+  return <div>{customTitle || defaultTitle}</div>
+}
+```
+
+```jsx
+const Component = ({ elements }) => {
+  return <div>There are {elements.length} elements</div>
+}
+```
+
+```jsx
+const Component = ({ elements, count }) => {
+  return <div>{!count && 'No results found'}</div>
+}
+```
+
+```jsx
+const Component = ({ elements }) => {
+  return <div>{!!elements.length && <List elements={elements}/>}</div>
+}
+```
+
+```jsx
+const Component = ({ elements }) => {
+  return <div>{Boolean(elements.length) && <List elements={elements}/>}</div>
+}
+```
+
+```jsx
+const Component = ({ elements }) => {
+  return <div>{elements.length > 0 && <List elements={elements}/>}</div>
+}
+```
+
+```jsx
+const Component = ({ elements }) => {
+  return <div>{elements.length ? <List elements={elements}/> : null}</div>
+}
+```
+
+### Options
+
+The supported options are:
+
+### `validStrategies`
+An array containing `"coerce"`, `"ternary"`, or both (default: `["ternary", "coerce"]`) - Decide which strategies are considered valid to prevent leaked renders (at least 1 is required). The "coerce" option will transform the conditional of the JSX expression to a boolean. The "ternary" option transforms the binary expression into a ternary expression returning `null` for falsy values. The first option from the array will be the strategy used when autofixing, so the order of the values matters.
+
+It can be set like:
+```json5
+{
+  // ...
+  "react/jsx-no-leaked-render": [<enabled>, { "validStrategies": ["ternary", "coerce"] }]
+  // ...
+}
+```
+
+Assuming the following options: `{ "validStrategies": ["ternary"] }`
+
+Examples of **incorrect** code for this rule, with the above configuration:
+```jsx
+const Component = ({ count, title }) => {
+  return <div>{count && title}</div>
+}
+```
+
+```jsx
+const Component = ({ count, title }) => {
+  return <div>{!!count && title}</div>
+}
+```
+
+Examples of **correct** code for this rule, with the above configuration:
+```jsx
+const Component = ({ count, title }) => {
+  return <div>{count ? title : null}</div>
+}
+```
+
+Assuming the following options: `{ "validStrategies": ["coerce"] }`
+
+Examples of **incorrect** code for this rule, with the above configuration:
+```jsx
+const Component = ({ count, title }) => {
+  return <div>{count && title}</div>
+}
+```
+
+```jsx
+const Component = ({ count, title }) => {
+  return <div>{count ? title : null}</div>
+}
+```
+
+Examples of **correct** code for this rule, with the above configuration:
+```jsx
+const Component = ({ count, title }) => {
+  return <div>{!!count && title}</div>
+}
+```
+
+## When Not To Use It
+
+If you are working in a typed-codebase which encourages you to always use boolean conditions, this rule can be disabled.
+
+## Further Reading
+
+- [React docs: Inline If with Logical && Operator](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator)
+- [Good advice on JSX conditionals - Beware of zero](https://thoughtspile.github.io/2022/01/17/jsx-conditionals/)
+- [Twitter: rendering falsy values in React and React Native](https://twitter.com/kadikraman/status/1507654900376875011?s=21&t=elEXXbHhzWthrgKaPRMjNg)
--- a/lib/rules/jsx-no-leaked-render.js
+++ b/lib/rules/jsx-no-leaked-render.js
@@ -1,0 +1,134 @@
+/**
+ * @fileoverview Prevent problematic leaked values from being rendered
+ * @author Mario Beltrán
+ */
+
+'use strict';
+
+const docsUrl = require('../util/docsUrl');
+const report = require('../util/report');
+const isParenthesized = require('../util/ast').isParenthesized;
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+const messages = {
+  noPotentialLeakedRender: 'Potential leaked value that might cause unintentionally rendered values or rendering crashes',
+};
+
+const COERCE_STRATEGY = 'coerce';
+const TERNARY_STRATEGY = 'ternary';
+const DEFAULT_VALID_STRATEGIES = [TERNARY_STRATEGY, COERCE_STRATEGY];
+const COERCE_VALID_LEFT_SIDE_EXPRESSIONS = ['UnaryExpression', 'BinaryExpression', 'CallExpression'];
+
+function trimLeftNode(node) {
+  // Remove double unary expression (boolean coercion), so we avoid trimming valid negations
+  if (node.type === 'UnaryExpression' && node.argument.type === 'UnaryExpression') {
+    return trimLeftNode(node.argument.argument);
+  }
+
+  return node;
+}
+
+function ruleFixer(context, fixStrategy, fixer, reportedNode, leftNode, rightNode) {
+  const sourceCode = context.getSourceCode();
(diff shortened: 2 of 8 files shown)


Belco90:
I'm not sure how to fix the 2 remaining checks failing.

Belco90:
@yannickcr pinging you for getting some review.

docs/rules/jsx-no-leaked-zero.md
```diff
@@ -0,0 +1,159 @@
...
+  return <div>{elements.length ? <List elements={elements}/> : null}</div>
+}
+```
+
+### Options
+
+The supported options are:
+
+### `fixStrategy`
+`"cast"` or `"ternary"` (default: `"ternary"`) - Decide the strategy to autofix the rule. The "cast" option will cast to boolean the condition of the JSX expression. The "ternary" option transforms the binary expression into a ternary expression returning `null` for falsy values.
```
ljharb, line 124: this does answer the question, but typically rules don't have autofix-specific options - in other words, i'd expect someone who sets this rule to "cast" would see warnings with a ternary, and vice versa.

lib/rules/jsx-no-leaked-zero.js
```diff
@@ -0,0 +1,103 @@
...
+        return fixer.replaceText(logicalExpressionNode, `${leftSideText} ? ${rightSideText} : null`);
+      }
+      throw new Error('Invalid value for "fixStrategy" option');
+    }
+
+    return {
+      'JSXExpressionContainer > LogicalExpression'(node) {
+        const leftSide = node.left;
+        if (node.operator !== '&&') {
+          return;
```
ljharb, line 87: i think the operator can be included in the AST selector

Belco90:
@ljharb my bad, sorry for any inconvenience caused.

lib/rules/jsx-no-leaked-zero.js
```diff
@@ -0,0 +1,103 @@
...
+const docsUrl = require('../util/docsUrl');
+const report = require('../util/report');
+const isParenthesised = require('../util/ast').isParenthesised;
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+const messages = {
+  noPotentialNumericEvaluation: 'Potential numeric evaluation that ends up in leaked zero to the DOM',
```
ljharb, line 17: react isn't just about the DOM; this also applies to server rendering.
```suggestion
  noPotentialNumericEvaluation: 'Potential numeric evaluation resulting in an unintentionally rendered `0`',
```

lib/util/ast.js
```diff
@@ -264,6 +264,23 @@ function getKeyValue(context, node) {
...
 }
 
+/**
+ * Checks if a node is surrounded by parenthesis.
+ *
+ * @param {object} context - Context from the rule
+ * @param {ASTNode} node - Node to be checked
+ * @returns {boolean}
+ */
+function isParenthesised(context, node) {
```
ljharb, line 274:
```suggestion
function isParenthesized(context, node) {
```

tests/lib/rules/jsx-no-leaked-zero.js
```diff
@@ -0,0 +1,313 @@
...
+//------------------------------------------------------------------------------
+
+const ruleTester = new RuleTester({ parserOptions });
+ruleTester.run('jsx-no-leaked-zero', rule, {
+  valid: parsers.all([
+    {
+      code: `
+      const Component = () => {
+        return <div>{customTitle || defaultTitle}</div>
+      }
```
ljharb, line 36:
```suggestion
        const Component = () => {
          return <div>{customTitle || defaultTitle}</div>
        }
```
(same for all the test cases)

ljharb (changes requested):

docs/rules/jsx-no-leaked-zero.md
Belco90: Not sure if I understood this correctly. Do you mean that code not leaking zeros but using the opposite strategy from the one set in the rule option should be reported as incorrect too?

ljharb: No, what i mean is, a rule should behave identically with or without autofix (except for whether the violations are fixed or not). So, I would expect the rule itself to be configurable to allow a cast, or a ternary, or both. An unleaked zero that's using a disallowed mechanism would warn.

Belco90: > What i mean is, a rule should behave identically with or without autofix (except for whether the violations are fixed or not)

Isn't that the opposite of what you are suggesting then? The current implementation will report the same errors no matter which `fixStrategy` the user picks. The option is just to decide which strategy the user prefers for autofixing the errors.

If the user sets `fixStrategy: "cast"` and there is a ternary inside a JSX expression, it won't be reported, since the possibility of getting a leaked zero is prevented.

Changing the behavior to report unleaked zero using the other option not set in `fixStrategy` would make the rule not to behave identically in every case.

ljharb: If they set their strategy to "cast", then I would expect the rule to warn on a ternary (that, if done as a &&, would leak a zero).

The rule shouldn't behave identically with different options - if the options are different, and the rule behaves differently, then why are there options at all

Belco90: Additionally, what would be the autofix strategy if a third "both" value is added for the `fixStrategy` option?

IMO I think we are overcomplicating this. The goal of this option was to offer 2 different ways of autofixing the same error depending on the user's taste. If that's an issue, I'd just stick to one way of always autofixing the error, and remove the rule option and the other autofix implementation.

Belco90: > The rule shouldn't behave identically with different options - if the options are different, and the rule behaves differently, then why are there options at all

The autofix strategy is behaving differently, that was the purpose of the option. So I'm thinking of just removing it and stick to a ternary autofix.

ljharb: eslint rules indeed don't offer fix-specific options.

I also don't think there's enough value for the rule at all if it's going to only force one of the fix strategies.

Belco90: I really don't know what you are suggesting then... If we don't want to offer an autofix option (I don't know why this is a problem even if other rules don't offer it, it shouldn't prevent this one from offering such an option), but enforcing only one fix strategy makes the rule not have enough value, how am I supposed to solve this?

ljharb: My hope is that you'd be willing to modify the rule so that it can be configured to require one strategy or the later, warns on an incorrect strategy, and autofixes to the correct one.

It would be fine, however, to default to "both strategies are OK" via an array of both options, in either order, and to just pick the first one for autofix, and then the user could swap the order if they wanted the other autofix but both to be acceptable.

Belco90: Sounds like a good approach. I'll try to address the proposed changes during the week.

Belco90: This should be addressed now!

Belco90:
Everything finally sorted out!

docs/rules/jsx-no-leaked-zero.md
```diff
@@ -0,0 +1,184 @@
...
+const Component = ({ elements }) => {
+  return <div>{elements.length ? <List elements={elements}/> : null}</div>
+}
+```
+
+### Options
+
+The supported options are:
+
+### `validFixStrategies`
```
ljharb, line 123: I think this option name needs to change - since it's not actually about autofix strategies. It's about which form(s) satisfy the rule. It's just that the autofixer *happens* to pick the first one in this list.

ljharb (changes requested):

docs/rules/jsx-no-leaked-zero.md
Belco90: My intention was to refer to the "strategies considered valid to fix the issue reported", but fair enough: I wasn't really happy with the name. Any suggestion?

Belco90: Renamed to just `validStrategies`

Belco90:
I've renamed the rule to `jsx-no-leaked-render` since:

1. `NaN` is also leaked in React
2. Not casting to bool causes crashes in React Native

docs/rules/jsx-no-leaked-render.md
```diff
@@ -0,0 +1,208 @@
...
+  return <div>{elements.length ? <List elements={elements}/> : null}</div>
+}
+```
+
+### Options
+
+The supported options are:
+
+### `validStrategies`
+An array containing `"cast"`, `"ternary"` or both (default: `["ternary", "cast"]`) - Decide which strategies are considered valid to prevent leaked renders (at least 1 is required). The "cast" option will cast to boolean the condition of the JSX expression. The "ternary" option transforms the binary expression into a ternary expression returning `null` for falsy values. The first option from the array will be used as autofix, so the order of the values matter.
```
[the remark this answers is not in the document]
Belco90: Good idea! Done.

ljharb:
@Belco90 re https://github.com/jsx-eslint/eslint-plugin-react/pull/3203/commits/866f5328f9c7e401718214c6ffe2fff1347fc054, we *should* use fragments - you just need to add `features: ['fragments']` to the test object.

Belco90:
Not sure why the "Automatic Rebase" check keeps failing

ljharb:
because instead of deleting your local branch and checking out my rebase, you merged your original local branch into my rebased one, and that means there's merge conflicts when trying to rebase it. Don't worry about that tho, I can handle it when it's time to land this.

Belco90:
> because instead of deleting your local branch and checking out my rebase, you merged your original local branch into my rebased one, and that means there's merge conflicts when trying to rebase it. Don't worry about that tho, I can handle it when it's time to land this.

Ah, damn! Sorry about that 🤦‍♂️

docs/rules/jsx-no-leaked-render.md
```diff
@@ -0,0 +1,208 @@
...
+      {NaN && <Something/>}
+      {/* React: renders undesired NaN */}
+      {/* React Native: crashes 💥 */}
+    </>
+  )
+}
+```
+
+This can be avoided by:
+- casting the condition to bool: `{!!someValue && <Something />}`
```
ljharb, line 33:
```suggestion
- coercing the conditional to a boolean: `{!!someValue && <Something />}`
```

docs/rules/jsx-no-leaked-render.md
```diff
@@ -0,0 +1,208 @@
...
+  return <div>{elements.length ? <List elements={elements}/> : null}</div>
+}
+```
+
+### Options
+
+The supported options are:
+
+### `validStrategies`
+An array containing `"coerce"`, `"ternary"`, or both (default: `["ternary", "coerce"]`) - Decide which strategies are considered valid to prevent leaked renders (at least 1 is required). The "coerce" option will cast to boolean the condition of the JSX expression. The "ternary" option transforms the binary expression into a ternary expression returning `null` for falsy values. The first option from the array will be used as autofix, so the order of the values matter.
```
ljharb, line 145:
```suggestion
An array containing `"coerce"`, `"ternary"`, or both (default: `["ternary", "coerce"]`) - Decide which strategies are considered valid to prevent leaked renders (at least 1 is required). The "coerce" option will transform the conditional of the JSX expression to a boolean. The "ternary" option transforms the binary expression into a ternary expression returning `null` for falsy values. The first option from the array will be used as autofix, so the order of the values matter.
```

lib/rules/jsx-no-leaked-render.js
```diff
@@ -0,0 +1,137 @@
...
+  },
+
+  create(context) {
+    const config = context.options[0] || {};
+    const validStrategies = config.validStrategies || DEFAULT_VALID_STRATEGIES;
+    const fixStrategy = validStrategies[0];
+    const areBothStrategiesValid = validStrategies.length === 2;
+
+    function trimLeftNode(node) {
+      // Remove double unary expression (boolean cast), so we avoid trimming valid negations
```
ljharb, line 67:
```suggestion
      // Remove double unary expression (boolean coercion), so we avoid trimming valid negations
```

lib/rules/jsx-no-leaked-render.js
```diff
@@ -0,0 +1,137 @@
...
+
+      throw new Error('Invalid value for "fixStrategy" option');
+    }
+
+    return {
+      'JSXExpressionContainer > LogicalExpression[operator="&&"]'(node) {
+        const leftSide = node.left;
+        const CAST_VALID_LEFT_SIDE_EXPRESSIONS = ['UnaryExpression', 'BinaryExpression', 'CallExpression'];
+        const isCastStrategyValid = areBothStrategiesValid || fixStrategy === COERCE_STRATEGY;
+        const isCastValidLeftExpression = CAST_VALID_LEFT_SIDE_EXPRESSIONS.some(
```
ljharb, line 106:
```suggestion
        const isCoerceValidLeftExpression = COERCE_VALID_LEFT_SIDE_EXPRESSIONS.some(
```

lib/rules/jsx-no-leaked-render.js
```diff
@@ -0,0 +1,137 @@
...
+      }
+
+      throw new Error('Invalid value for "fixStrategy" option');
+    }
+
+    return {
+      'JSXExpressionContainer > LogicalExpression[operator="&&"]'(node) {
+        const leftSide = node.left;
+        const CAST_VALID_LEFT_SIDE_EXPRESSIONS = ['UnaryExpression', 'BinaryExpression', 'CallExpression'];
+        const isCastStrategyValid = areBothStrategiesValid || fixStrategy === COERCE_STRATEGY;
```
ljharb, line 105:
```suggestion
        const isCoerceStrategyValid = areBothStrategiesValid || fixStrategy === COERCE_STRATEGY;
```

lib/rules/jsx-no-leaked-render.js
```diff
@@ -0,0 +1,137 @@
...
+        return fixer.replaceText(reportedNode, `${leftSideText} ? ${rightSideText} : null`);
+      }
+
+      throw new Error('Invalid value for "fixStrategy" option');
+    }
+
+    return {
+      'JSXExpressionContainer > LogicalExpression[operator="&&"]'(node) {
+        const leftSide = node.left;
+        const CAST_VALID_LEFT_SIDE_EXPRESSIONS = ['UnaryExpression', 'BinaryExpression', 'CallExpression'];
```
ljharb, line 104:
```suggestion
        const COERCE_VALID_LEFT_SIDE_EXPRESSIONS = ['UnaryExpression', 'BinaryExpression', 'CallExpression'];
```
seems like this list can be hoisted up to the module level?

ljharb (approved):

Belco90:
> @Belco90 and @ljharb – thanks for making this happen ❤️ 

Looking forward to getting it released!

karlhorky:
@Belco90 @ljharb is this rule description still valid for React Native?

I can't get an empty string to crash React Native without a wrapping `<Text>` component 😮 https://twitter.com/karlhorky/status/1576958348288987137

I would do the PR to update the docs if this is no longer a thing...

ljharb:
That would be a major DX change for React Native, so I'd want to see some kind of announcement that it was intentional by the RN team.

karlhorky:
I also was surprised - but couldn't find anything official after a bit of digging. Asking some maintainers here: https://twitter.com/karlhorky/status/1576987525914464256

karlhorky:
Ok, so I didn't really get a clear answer from any maintainers there, but I just started a new React Native app just now (`npx react-native init` with `react-native@0.70.1`) and confirmed empty strings outside of `<Text>` components do not cause this error:

Screen Shot 2022-10-04 at 13 10 49

Whereas a string with content in it does:

Screen Shot 2022-10-04 at 13 14 34

Repo: https://github.com/karlhorky/react-native-empty-string-outside-text-component

@ljharb enough proof?

karlhorky:
Asked some more maintainers, just in case: https://twitter.com/karlhorky/status/1577257935796740096

karlhorky:
Good point about it being intentional, @rickhanlonii is going to look into it: https://twitter.com/rickhanlonii/status/1577291418485362688

karlhorky:
Since I haven't received any feedback yet, I opened an issue on the React Native GitHub repo: https://github.com/facebook/react-native/issues/35002

karlhorky:
Looks like indeed an intentional change - actually in React itself (in version 18): https://github.com/facebook/react-native/issues/35002#issuecomment-1282513818

I would update the docs if you agree @ljharb @Belco90

ljharb:
@karlhorky sounds like we need a PR to update the docs, as well as the rule when react is >= 18

karlhorky:
Docs PR: https://github.com/jsx-eslint/eslint-plugin-react/pull/3468

@Belco90 would you maybe want to do the 2nd PR that @ljharb mentioned above? I think the idea is to modify the rule to not report errors on empty strings if React is detected to be above v18.

Belco90:
> @Belco90 would you maybe want to do the 2nd PR that @ljharb mentioned above? I think the idea is to modify the rule to not report errors on empty strings if React is detected to be above or equal to v18.

I'm afraid I don't have enough time right now to take care of it 😢

karlhorky:
Ok, opened #3479 for the rule change

PR #3205, opened 2022-02-13
chore: Add the .vscode setting

islandryu:
I wanted to configure vscode, so I added .vscode to gitignore.
If, for some reason, you have not added .vscode to gitignore, please close this PR.

In addition, I've added a launch.json template.
Please rename launch.template.json to launch.json.
With the test file open in vscode, press f5 to start the debugger.

### DIFF
--- a/.vscode/launch.template.json
+++ b/.vscode/launch.template.json
@@ -1,0 +1,21 @@
+{
+	"version": "0.2.0",
+	"configurations": [
+    {
+      "type": "node",
+      "protocol": "inspector",
+      "request": "launch",
+      "name": "Mocha Tests (currently opened test)",
+      "program": "${workspaceRoot}/node_modules/mocha/bin/_mocha",
+      "args": [
+        "${command:extension.commandvariable.file.filePosix}"
+      ],
+      "env": {
+        "NODE_ENV": "testing"
+      },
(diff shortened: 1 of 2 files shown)


.vscode/launch.template.json
```diff
@@ -0,0 +1,21 @@
+{
+	"version": "0.2.0",
+	"configurations": [
+    {
+      "type": "node",
+      "protocol": "inspector",
+      "request": "launch",
+      "name": "Mocha Tests (currently opened test)",
+      "program": "${workspaceRoot}/node_modules/mocha/bin/_mocha",
```
ljharb, line 9: I’m not clear on why this config is helpful.

.vscode/launch.template.json
```diff
@@ -0,0 +1,21 @@
...
+      "args": [
+        "${command:extension.commandvariable.file.filePosix}"
+      ],
+      "env": {
+        "NODE_ENV": "testing"
```
ljharb, line 14: NODE_ENV is never “testing” ever anywhere; perhaps you’re thinking of “test”?

.vscode/launch.template.json
islandryu: This is the setting to run the debugger with vscode.

Issue #3221, opened 2022-02-25
Automating github releases and release notes


xt0rted:
In https://github.com/yannickcr/eslint-plugin-react/issues/3215#issuecomment-1051077785 it was mentioned that release notes were missing because it's done manually. I automated this with GitHub Actions and can PR a workflow for it if you'd like.

This one is triggered on release and updates it with the changelog contents so it doesn't have to be done manually. https://github.com/xt0rted/stylelint-actions-formatters/blob/main/.github/workflows/release.yml

If you're not creating the release and instead just pushing a tag then the workflow can be triggered by that and create the release for you. Either way if you'd like to automate this and want help just let me know.

ljharb:
That's an interesting offer. Unfortunately Github's own release note automation ignores PR-less commits, so it's not useful.

The only concern I have is that sometimes I'll push the tag, but then discover something that needs fixing first, so I'll force push the tag later - it's only on `npm publish` that the tag becomes conceptually immutable. Do you have an idea of how to work around that?

xt0rted:
If you trigger off the tag then the workflow will run as soon as it's pushed. In this case it'd be best to manually create a release (maybe let that handle creating that tag too?) and have the workflow run from there.

This could still remove some manual steps though. The changelog is the right format to automatically be copied over, automatic release notes could be added, and publishing to npm could be done as well.

I force pushed a lot to my repo getting some of the recent versions published which is why I'm using the GitHub release as the trigger.

xt0rted:
Looks like there's an endpoint to [delete releases](https://docs.github.com/en/rest/reference/releases#delete-a-release). It might be possible to check if a release already exists, delete it, and then create a new one for the tag. Then the rest of your workflow to publish it wouldn't need to change.

ljharb:
The problem is that I don't remember to create the release; pushing the tag is never something I forget.

I suppose as long as the following hold true:
 - a new tag creates a release
 - an overwritten tag deletes, and recreates, the release
 - a deleted tag deletes the release

that it should be fine. I'd love to review a PR :-)

ljharb:
This was done in #3223, thank you!

ljharb:
@xt0rted it doesn't seem to be working quite right: https://github.com/jsx-eslint/eslint-plugin-react/releases/tag/v7.30.1

xt0rted:
[...]

PR #3223, opened 2022-02-26
[actions] add release workflow

xt0rted:
Based on the discussion in #3221 this adds a workflow that runs when tags are pushed and creates/updates a release of the same name.

After some testing it turns out GitHub detects force pushes of tags and automatically points an existing release to the new commit. This means the extra steps to check for an existing release, delete it if found, etc. aren't needed.

With this in place you should be able to:

- push a `v2.0.0` tag
    - a release will be created automatically
    - the changelog contents will be copied to it
- make some changes to `main`
- move the tag up a few commits
- force push the `v2.0.0` tag
    - GitHub will auto-update the release's tag/sha
    - this workflow will update the release notes with anything new

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,51 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - "v*.*.*"
+
+jobs:
+  release:
+    if: github.event.repository.fork == false
+
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+
+    steps:
+      - uses: step-security/harden-runner@v1
+        with:
+          allowed-endpoints:
+            api.github.com:443
+            github.com:443
+
+      - name: Get version from tag
+        id: tag_name
+        run: echo ::set-output name=current_version::${GITHUB_REF#refs/tags/v}
+        shell: bash
+
+      - uses: actions/checkout@v2
+
+      - uses: mindsers/changelog-reader-action@v2
+        id: changelog_reader
+        with:
+          version: ${{ steps.tag_name.outputs.current_version }}
+
+      - name: Get common links from changelog
+        id: changelog
+        run: |
+          # Parse the changelog for common links
+          _links="$(grep -P '^\[.*]:.+' ${GITHUB_WORKSPACE}/CHANGELOG.md | sort -u)"
+          _links="${_links//'%'/'%25'}"
+          _links="${_links//$'\n'/'%0A'}"
+          _links="${_links//$'\r'/'%0D'}"
+          # Set output 'links' to $_links
+          echo "::set-output name=links::${_links}"
+
+      - uses: softprops/action-gh-release@v1
+        with:
+          body: |
+            ${{ steps.changelog_reader.outputs.changes }}
+            ${{ steps.changelog.outputs.links }}


.github/workflows/release.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+  
+    steps:
+      - name: Get version from tag
+        id: tag_name
+        run: echo ::set-output name=current_version::${GITHUB_REF#refs/tags/v}
+        shell: bash
```
xt0rted, line 19: This is redundant since the ubuntu vm is being used, can be removed if you want

.github/workflows/release.yml
```diff
@@ -0,0 +1,32 @@
...
+
+      - name: Get changelog entry
+        uses: mindsers/changelog-reader-action@v2
+        id: changelog_reader
+        with:
+          version: ${{ steps.tag_name.outputs.current_version }}
+
+      - uses: softprops/action-gh-release@v1
+        with:
+          body: ${{ steps.changelog_reader.outputs.changes }}
```
xt0rted, line 32: This action will create the release if it doesn't exist, or update it if it does. So all that's needed is to pass the release body to it.

.github/workflows/release.yml
```diff
@@ -0,0 +1,32 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - "v*.*.*"
```
ljharb, line 6: this should ideally also make sure it doesn't run on forks

.github/workflows/release.yml
ljharb: fine to keep, explicit > implicit

.github/workflows/release.yml
```diff
@@ -0,0 +1,32 @@
...
+      contents: write
+  
+    steps:
+      - name: Get version from tag
+        id: tag_name
+        run: echo ::set-output name=current_version::${GITHUB_REF#refs/tags/v}
+        shell: bash
+
+      - name: Check out repo
+        uses: actions/checkout@v2
```
ljharb, line 22:
```suggestion
      - uses: actions/checkout@v2
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,32 @@
...
+
+      - name: Get changelog entry
+        uses: mindsers/changelog-reader-action@v2
+        id: changelog_reader
+        with:
+          version: ${{ steps.tag_name.outputs.current_version }}
```
ljharb, line 28: to confirm, this will read CHANGELOG.md?

ljharb (commented):
Could we also incorporate the harden-runner? https://github.com/step-security/harden-runner

Here's an example: https://github.com/nvm-sh/nvm/blob/master/.github/workflows/lint.yml#L11-L16

.github/workflows/release.yml
xt0rted: This won't run on forks unless they push a tag to it with this format. To prevent it from ever running we could do:

```diff
 jobs:
   release:
+    if: github.repository == 'yannickcr/eslint-plugin-react'
     runs-on: ubuntu-latest
```

.github/workflows/release.yml
xt0rted: Correct, by default it pulls from `./CHANGELOG.md`

.github/workflows/release.yml
ljharb: Forks have all tags by default, and I often push tags to my fork first to ensure tests pass before i push it to the main repo. That addition sounds great :-) (unless there's something that can say "if this is not a fork" without hardcoding the repo name)

xt0rted:
Just pushed a change that will grab the common links at the end of the file and append them to the release notes. It required added an html comment to the changelog so they can be easily found. Hopefully that's not an issue. The results of a test run can be seen here https://github.com/xt0rted/eslint-plugin-react/releases/tag/v7.30.0

I'm not sure how best to handle links inside of the version section. The action is [filtering them out](https://github.com/mindsers/changelog-reader-action/blob/66d9e432ff515d3cd6a90e0d8579b454ad45720e/src/parse-entry.js#L17) for some reason. Maybe we could get a setting added that would skip the filter? Another option is to change how PRs are referenced. Instead of doing `[#123][]` you could do `#123` and let github auto link them.

xt0rted:
Issue about the link filter https://github.com/mindsers/changelog-reader-action/issues/52

ljharb:
That wouldn't cause them to be autolinked when viewing the markdown directly, which i presume is why that choice was originally made.

CHANGELOG.md
```diff
@@ -3580,6 +3580,8 @@ If you're still not using React 15 you can keep the old behavior by setting the
 ### Added
  * First revision
 
+<!-- common links -->
```
ljharb, line 3583: this does _work_, but it'd be nice to not rely on it, and instead find a way to locate all referenced links

.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Get common links from changelog
+        id: changelog
+        run: |
+          # Parse the changelog for common links
+          _links="$(grep -Pz -o '(?s)<!-- common links -->.*' ${GITHUB_WORKSPACE}/CHANGELOG.md)"
+          _links="${_links//'%'/'%25'}"
+          _links="${_links//$'\n'/'%0A'}"
+          _links="${_links//$'\r'/'%0D'}"
+          # Set output 'links' to $_links
```
ljharb, line 44: i wonder if it might be simpler to just grab *every* link from the file, sort | uniq, and dump them at the bottom of the release body?

xt0rted: Adjusted to pull all links. Updated test release at https://github.com/xt0rted/eslint-plugin-react/releases/tag/v7.30.0. Looks good to me so far.

ljharb (approved):
k, let's merge this and give it a shot with v7.29.3 :-)

ljharb:
Looks like it worked! https://github.com/yannickcr/eslint-plugin-react/releases/tag/v7.29.3 thank you!

xt0rted:
Got a couple Dependabot PRs for the new version, everything looks good except for the `release notes` section. For some reason they're including every link at the end of the release notes even if they aren't referenced in the text of it. https://github.com/OctoLinker/OctoLinker/pull/1510

ljharb:
That sounds like a dependabot bug; I'd report it to github. (i don't personally use dependabot; i find renovate to be a much better tool)

PR #3226, opened 2022-02-28
[Docs] Recategorize rules in readme

### LINKED ISSUE #3216: Categorize rules in docs as one of Possible Problems, Suggestions, and Formatting
These categories help readers find rules more quickly. A user interested in Possible Problems rules can look at Possible Problems rules. A user interested in formatting rules can look at formatting rules.

golopot:
Closes #3216 .

~~The categories are taken from https://eslint.org/docs/rules/ .~~

Rendered Readme: https://github.com/yannickcr/eslint-plugin-react/blob/793d5374e09beb34dfcbc4ac3e8494f6ecaf48f1/README.md

### DIFF
--- a/lib/rules/no-typos.js
+++ b/lib/rules/no-typos.js
@@ -32,7 +32,7 @@ module.exports = {
   meta: {
     docs: {
       description: 'Disallow common typos',
-      category: 'Stylistic Issues',
+      category: 'Possible Errors',
       recommended: false,
       url: docsUrl('no-typos'),
     },
--- a/lib/rules/boolean-prop-naming.js
+++ b/lib/rules/boolean-prop-naming.js
@@ -24,7 +24,7 @@ const messages = {
 module.exports = {
   meta: {
     docs: {
-      category: 'Stylistic Issues',
+      category: 'Best Practices',
       description: 'Enforces consistent naming for boolean props',
       recommended: false,
       url: docsUrl('boolean-prop-naming'),
--- a/lib/rules/destructuring-assignment.js
+++ b/lib/rules/destructuring-assignment.js
@@ -57,7 +57,7 @@ module.exports = {
   meta: {
     docs: {
       description: 'Enforce consistent usage of destructuring assignment of props, state, and context',
-      category: 'Stylistic Issues',
+      category: 'Best Practices',
       recommended: false,
       url: docsUrl('destructuring-assignment'),
     },
--- a/lib/rules/function-component-definition.js
+++ b/lib/rules/function-component-definition.js
@@ -119,7 +119,7 @@ module.exports = {
   meta: {
     docs: {
       description: 'Enforce a specific function type for function components',
-      category: 'Stylistic Issues',
+      category: 'Best Practices',
       recommended: false,
       url: docsUrl('function-component-definition'),
     },
--- a/lib/rules/jsx-boolean-value.js
+++ b/lib/rules/jsx-boolean-value.js
@@ -72,7 +72,7 @@ module.exports = {
   meta: {
     docs: {
       description: 'Enforce boolean attributes notation in JSX',
-      category: 'Stylistic Issues',
+      category: 'Best Practices',
       recommended: false,
       url: docsUrl('jsx-boolean-value'),
     },
--- a/lib/rules/jsx-filename-extension.js
+++ b/lib/rules/jsx-filename-extension.js
@@ -31,7 +31,7 @@ module.exports = {
   meta: {
     docs: {
       description: 'Disallow file extensions that may contain JSX',
-      category: 'Stylistic Issues',
+      category: 'Best Practices',
       recommended: false,
       url: docsUrl('jsx-filename-extension'),
     },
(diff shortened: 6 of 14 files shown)


lib/rules/forbid-prop-types.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
   meta: {
     docs: {
       description: 'Forbid certain propTypes',
-      category: 'Best Practices',
+      category: 'suggestion',
```
ljharb, line 32: i prefer eslint's previous mechanism for categorizing rules, to be honest. is there any benefit from using this style?

golopot: Do you mean categorizing rules as problems, best practices, and stylistic issues? In that case I agree with you.

ljharb: Yes. I'm saying that i think eslint's old categories were superior, and it's better to stick with those :-)

golopot: Updated.

lib/rules/button-has-type.js
```diff
@@ -32,7 +32,7 @@ module.exports = {
   meta: {
     docs: {
       description: 'Forbid "button" element without an explicit "type" attribute',
-      category: 'Possible Errors',
+      category: 'Best Practices',
```
ljharb, line 35: this is definitely a possible error; button elements must never omit a "type" attribute

lib/rules/jsx-handler-names.js
```diff
@@ -21,7 +21,7 @@ module.exports = {
   meta: {
     docs: {
       description: 'Enforce event handler naming conventions in JSX',
-      category: 'Stylistic Issues',
+      category: 'Best Practices',
```
ljharb, line 24: i'm not sure this really qualifies as a best practice; stylistic feels appropriate here

lib/rules/jsx-max-depth.js
```diff
@@ -24,7 +24,7 @@ module.exports = {
   meta: {
     docs: {
       description: 'Validate JSX maximum depth',
-      category: 'Stylistic Issues',
+      category: 'Best Practices',
```
ljharb, line 27: this is absolutely *not* a best practice :-) it should stay stylistic

lib/rules/jsx-no-useless-fragment.js
```diff
@@ -89,7 +89,7 @@ module.exports = {
     fixable: 'code',
     docs: {
       description: 'Disallow unnecessary fragments',
-      category: 'Possible Errors',
+      category: 'Best Practices',
```
ljharb, line 92: unnecessary fragments _are_ an error

lib/rules/no-invalid-html-attribute.js
```diff
@@ -533,7 +533,7 @@ module.exports = {
   meta: {
     fixable: 'code',
     docs: {
-      description: 'Forbid attribute with an invalid values`',
+      description: 'Forbid attribute with an invalid values',
```
ljharb, line 536:
```suggestion
      description: 'Forbid attribute with an invalid value',
```

lib/rules/no-typos.js
```diff
@@ -31,7 +31,7 @@ module.exports = {
   meta: {
     docs: {
       description: 'Prevent common typos',
-      category: 'Stylistic Issues',
+      category: 'Best Practices',
```
ljharb, line 34: this seems more like a possible error

lib/rules/no-unstable-nested-components.js
```diff
@@ -270,7 +270,7 @@ module.exports = {
   meta: {
     docs: {
       description: 'Prevent creating unstable components inside components',
-      category: 'Possible Errors',
+      category: 'Best Practices',
```
ljharb, line 273: causing unnecessary renders is almost certainly an error

lib/rules/prefer-exact-props.js
```diff
@@ -24,7 +24,7 @@ module.exports = {
   meta: {
     docs: {
       description: 'Prefer exact proptype definitions',
-      category: 'Possible Errors',
+      category: 'Best Practices',
```
ljharb, line 27: this is also a possible error; inexact propTypes allow for bugs.

ljharb (commented):
It still feels useful to distinguish jsx-specific rules; not everyone uses jsx with react.

golopot (commented):
I am thinking that best practice is a subjective word. A rule might be a best practice for me but not for other people. If a rule a considered a good practice for 50% of developers, it might considered a best practice _among_ these developers, and considered a stylistic issue among other developers.

> It still feels useful to distinguish jsx-specific rules; not everyone uses jsx with react.

I think the prefix in rule names should suffice this need.

ljharb:
I'm not sure that's true, given that some of the jsx rules cover non-jsx as well.

As for best practice, of course it's subjective, but its perfectly fine to be opinionated. All of the things I've indicated are prevailing community-wide best practices; if someone disagrees, they are free to disable the rule.

PR #3233, opened 2022-03-05
feat: added "ignore" option to react version check

oscartbeaumont:
I am using this eslint rule in a project which is based on [Preact](https://preactjs.com) and this PR allows the React version not specified error to be disabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ You should also specify settings that will be shared across all the plugin rules
       "version": "detect", // React version. "detect" automatically picks the version you have installed.
                            // You can also use `16.0`, `16.3`, etc, if you want to override the detected value.
                            // It will default to "latest" and warn if missing, and to "detect" in the future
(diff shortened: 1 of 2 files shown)


ljharb:
This is eslint plugin _react_, and the react version is pretty important for a bunch of rules. I'm not sure this really makes sense.

oscartbeaumont:
I can see your point but specifically, I am using `preact/compat` which uses my bundler to replace `react` imports with `preact/compat` imports. So from that perspective, I expect all of the rules which apply to `react` to work on my codebase.

Also, this PR isn't a specific change to this plugin for Preact. It just allows the user to hide a warning about not having set the React version if they want to.

oscartbeaumont:
If the rules rely on the version heavily then you are right it wouldn't make sense to merge this. Would it be possible to specify "latest" as a version number and it default to the newest version of React?

ljharb:
I would suggest figuring out which react version corresponds most closely to the preact version you're using, and explicitly configure that in your config.

PR #3241, opened 2022-03-11
[Fix] `boolean-prop-naming`: handle React.FC, intersection, union types

### LINKED ISSUE #3197: react/boolean-prop-naming interface/type on React.FC validation of boolean with typescript does not work
So this works:
```
type Props = {
  enabled: boolean
}
const Hello = (props: Props) => <div />;
```

But If you want to validate a boolaan prop on an interface or type in a React.FC
```

type Props = {
  enabled: boolean;
}

const Hello: React.FC<Props> = ({ enabled }) => {
```

Then it doesnt work and thinks enabled is allowed.
Is this intended?

mobily:
**Types of changes**

This PR fixes #3197, therefore, the `boolean-prop-naming` rule validates correctly the following code: 

```tsx
import type { FC } from 'react' // or import { FC } from 'react'

type Props = {
  enabled: boolean // ❌ error
} 

export const Component: FC<Props> = props => { 
  return <div /> 
}
```

```tsx
import * as React from 'react' 

type Props = {
  enabled: boolean // ❌ error
} 

export const Component: React.FC<Props> = props => { 
  return <div /> 
}
```

```tsx
import type { FC } from 'react' // or import { FC } from 'react'

type Props = {
  isEnabled: boolean // ✅
} 

export const Component: FC<Props> = props => { 
  return <div /> 
}
```

```tsx
import * as React from 'react' 

type Props = {
  isEnabled: boolean // ✅
} 

export const Component: React.FC<Props> = props => { 
  return <div /> 
}
```

Moreover, previously, the plugin could not handle a case like this:

```tsx
import * as React from 'react'
import { BoxProps } from '@mobily/stacks'

type Props = {
  isEnabled: boolean, // ⬅️ this was completely ignored
} & BoxProps

export const Component = (props: Props) => { 
  return <div /> 
}
```

or

```tsx
import * as React from 'react'

type Props = {
  isEnabled: boolean, 
} & {
  lol: boolean, // ⬅️ same here, both property types were ignored
}

export const Component = (props: Props) => { 
  return <div /> 
}
```

Now, the plugin handles the intersection and union types as well, so the following will be validated correctly:

```tsx
import * as React from 'react'

type Props = {
  isEnabled: boolean, // ✅
} & {
  lol: boolean, // ❌ error
}

export const Component = (props: Props) => { 
  return <div /> 
}
```

```tsx
import * as React from 'react'

type Props = {
  isEnabled: boolean, // ✅
} | {
  lol: boolean, // ❌ error
}

export const Component = (props: Props) => { 
  return <div /> 
}
```

```tsx
import * as React from 'react'

type Props = {
  isEnabled: boolean, // ✅
} & ({
  lol: boolean, // ❌ error
} | {
  hasLOL: boolean, // ✅
})

export const Component = (props: Props) => { 
  return <div /> 
}
```

### DIFF
--- a/lib/rules/boolean-prop-naming.js
+++ b/lib/rules/boolean-prop-naming.js
@@ -228,6 +228,65 @@ module.exports = {
       args.filter((arg) => arg.type === 'ObjectExpression').forEach((object) => validatePropNaming(node, object.properties));
     }
 
+    function getComponentTypeAnnotation(component) {
+      // If this is a functional component that uses a global type, check it
+      if (
+        (component.node.type === 'FunctionDeclaration' || component.node.type === 'ArrowFunctionExpression')
+        && component.node.params
+        && component.node.params.length > 0
+        && component.node.params[0].typeAnnotation
+      ) {
+        return component.node.params[0].typeAnnotation.typeAnnotation;
+      }
+
+      if (
+        component.node.parent
+        && component.node.parent.type === 'VariableDeclarator'
+        && component.node.parent.id
+        && component.node.parent.id.type === 'Identifier'
+        && component.node.parent.id.typeAnnotation
+        && component.node.parent.id.typeAnnotation.typeAnnotation
+        && component.node.parent.id.typeAnnotation.typeAnnotation.typeParameters
+        && (
+          component.node.parent.id.typeAnnotation.typeAnnotation.typeParameters.type === 'TSTypeParameterInstantiation'
+          || component.node.parent.id.typeAnnotation.typeAnnotation.typeParameters.type === 'TypeParameterInstantiation'
+        )
+      ) {
+        return component.node.parent.id.typeAnnotation.typeAnnotation.typeParameters.params.find(
+          (param) => param.type === 'TSTypeReference' || param.type === 'GenericTypeAnnotation'
+        );
+      }
+    }
+
+    function findAllTypeAnnotations(identifier, node) {
+      if (node.type === 'TSTypeLiteral' || node.type === 'ObjectTypeAnnotation') {
+        const currentNode = [].concat(
+          objectTypeAnnotations.get(identifier.name) || [],
+          node
+        );
+        objectTypeAnnotations.set(identifier.name, currentNode);
+      } else if (
+        node.type === 'TSParenthesizedType'
+        && (
+          node.typeAnnotation.type === 'TSIntersectionType'
+          || node.typeAnnotation.type === 'TSUnionType'
+        )
+      ) {
+        node.typeAnnotation.types.forEach((type) => {
+          findAllTypeAnnotations(identifier, type);
+        });
+      } else if (
+        node.type === 'TSIntersectionType'
+        || node.type === 'TSUnionType'
+        || node.type === 'IntersectionTypeAnnotation'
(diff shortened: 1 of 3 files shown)


lib/rules/boolean-prop-naming.js
```diff
@@ -333,7 +361,14 @@ module.exports = {
...
+              propType.forEach((prop) => {
+                validatePropNaming(
+                  list[component].node,
+                  prop.properties || prop.members
+                );
+              });
+            } else if (propType) {
               validatePropNaming(
                 list[component].node,
                 propType.properties || propType.members
```
ljharb, line 374:
```suggestion
            if (propType) {
              [].concat(propType).forEach((prop) => {
                validatePropNaming(
                  list[component].node,
                  prop.properties || prop.members
                );
              });
```

tests/lib/rules/boolean-prop-naming.js
```diff
@@ -417,6 +417,56 @@ ruleTester.run('boolean-prop-naming', rule, {
...
     },
+    {
+      code: `
+        type Props = {
+          isEnabled: boolean
+        } & OtherProps
+        const HelloNew = (props: Props) => { return <div /> };
+      `,
+      options: [{ rule: '^is[A-Z]([A-Za-z0-9]?)+' }],
+      features: ['ts'],
```
ljharb, line 428: does this pass with flow as well?
```suggestion
      features: ['types'],
```

same question throughout

ljharb (commented):
Thanks, this is great!

lib/rules/boolean-prop-naming.js
mobily: ✅

tests/lib/rules/boolean-prop-naming.js
mobily: @ljharb yes, seems it does! 🎉

mobily:
@ljharb in the meantime, I have added handling the union type, looking forward to your review!

lib/rules/boolean-prop-naming.js
```diff
@@ -228,6 +228,46 @@ module.exports = {
...
+        return component.node.parent.id.typeAnnotation.typeAnnotation.typeParameters.params.find(
+          (param) => param.type === 'TSTypeReference'
+        );
+      }
+    }
+
+    function findTypeAnnotations(identifier, node) {
+      if (node.type === 'TSTypeLiteral') {
+        const currentNode = objectTypeAnnotations.get(identifier.name) || [];
+        currentNode.push(node);
```
ljharb, line 262:
```suggestion
        const currentNode = [].concat(objectTypeAnnotations.get(identifier.name) || [], node);
```
this way we avoid mutation

ljharb (approved):
Seems great!

ljharb:
hmm, looks like the latest changes you pushed up (or, possibly, my tiny tweaks, but i doubt it) are failing some tests. can you take a look?

mobily:
✅ and done, thanks, @ljharb for tweaking my code a bit!

mobily:
@ljharb ok, I think the most important cases have been handled, could you please run all test suites once again?

mobily:
thank you @ljharb ❤️

PR #3244, opened 2022-03-14
[Fix] `hook-use-state`: Allow UPPERCASE setState setter prefixes

duncanbeevers:
Closes https://github.com/yannickcr/eslint-plugin-react/issues/3243

This PR allows `useState` destructured variable names to use ALL-UPPER-CASE for the first segment of the setter variable name.

### Example

```js
// Previously invalid, now valid
const [rgbColor, setRGBColor] = useState();
```

### Discussion

The existing suggested variable name `setRgbColor` is still the sole suggestion; `setRGBColor` is considered valid, but is never suggested.

This should be a non-breaking change as all previously-valid code is still considered valid.

### DIFF
--- a/lib/rules/hook-use-state.js
+++ b/lib/rules/hook-use-state.js
@@ -66,13 +66,17 @@ module.exports = {
         ? setterVariable.name
         : undefined;
 
-      const expectedSetterVariableName = valueVariableName ? (
-        `set${valueVariableName.charAt(0).toUpperCase()}${valueVariableName.slice(1)}`
-      ) : undefined;
+      const caseCandidateMatch = valueVariableName ? valueVariableName.match(/(^[a-z]+)(.*)/) : undefined;
+      const upperCaseCandidatePrefix = caseCandidateMatch ? caseCandidateMatch[1] : undefined;
+      const caseCandidateSuffix = caseCandidateMatch ? caseCandidateMatch[2] : undefined;
+      const expectedSetterVariableNames = upperCaseCandidatePrefix ? [
+        `set${upperCaseCandidatePrefix.charAt(0).toUpperCase()}${upperCaseCandidatePrefix.slice(1)}${caseCandidateSuffix}`,
+        `set${upperCaseCandidatePrefix.toUpperCase()}${caseCandidateSuffix}`,
+      ] : [];
 
       const isSymmetricGetterSetterPair = valueVariable
         && setterVariable
-        && setterVariableName === expectedSetterVariableName
+        && expectedSetterVariableNames.indexOf(setterVariableName) !== -1
         && variableNodes.length === 2;
 
       if (!isSymmetricGetterSetterPair) {
@@ -80,9 +84,13 @@ module.exports = {
           {
             desc: 'Destructure useState call into value + setter pair',
             fix: (fixer) => {
+              if (expectedSetterVariableNames.length === 0) {
+                return;
+              }
+
               const fix = fixer.replaceTextRange(
                 node.parent.id.range,
-                `[${valueVariableName}, ${expectedSetterVariableName}]`
+                `[${valueVariableName}, ${expectedSetterVariableNames[0]}]`
               );
 
               return fix;
--- a/tests/lib/rules/hook-use-state.js
+++ b/tests/lib/rules/hook-use-state.js
@@ -38,6 +38,33 @@ const tests = {
     {
       code: `
         import { useState } from 'react'
+        function useRGB() {
+          const [rgb, setRGB] = useState()
+          return [rgb, setRGB]
+        }
+      `,
+    },
+    {
+      code: `
+        import { useState } from 'react'
+        function useRGBValue() {
+          const [rgbValue, setRGBValue] = useState()
+          return [rgbValue, setRGBValue]
+        }
+      `,
+    },
+    {
+      code: `
+        import { useState } from 'react'
+        function useCustomColorValue() {
+          const [customColorValue, setCustomColorValue] = useState()
+          return [customColorValue, setCustomColorValue]
+        }
+      `,
+    },
+    {
+      code: `
+        import { useState } from 'react'
         function useColor() {
           const [color, setColor] = useState('#ffffff')
           return [color, setColor]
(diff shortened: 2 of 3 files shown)


lib/rules/hook-use-state.js
```diff
@@ -66,23 +66,32 @@ module.exports = {
...
+      const expectedSetterVariableNames = upperCaseCandidatePrefix ? [
+        `${upperCaseCandidatePrefix.charAt(0).toUpperCase()}${upperCaseCandidatePrefix.slice(1)}${caseCandidateSuffix}`,
+        `${upperCaseCandidatePrefix.toUpperCase()}${caseCandidateSuffix}`,
+      ].filter(Boolean).map((candidate) => `set${candidate}`) : undefined;
 
       const isSymmetricGetterSetterPair = valueVariable
         && setterVariable
-        && setterVariableName === expectedSetterVariableName
+        && expectedSetterVariableNames
+        && expectedSetterVariableNames.includes(setterVariableName)
```
duncanbeevers, line 80: Oops, I think `Array.protype.includes` is too fresh; reworking to use the Old Ways.

tests/lib/rules/hook-use-state.js
```diff
@@ -35,6 +35,24 @@ const tests = {
...
+        function useRGBValue() {
+          const [rgbValue, setRGBValue] = useState()
+          return [rgbValue, setRGBValue]
+        }
+      `,
+    },
+    {
+      code: `
+        import { useState } from 'react'
+        function useCustomColorValue() {
```
duncanbeevers, line 50: I added this case to ensure the camel-case splitter case captures the correct suffix.

tests/lib/rules/hook-use-state.js
```diff
@@ -35,6 +35,33 @@ const tests = {
         }
       `,
     },
+    {
+      code: `
+        import { useState } from 'react'
+        function useRGB() {
+          const [rgb, setRGB] = useState()
+          return [rgb, setRGB]
```
duncanbeevers, line 43: Testing both:
* bare ALL-UPPERCASE value
* compound UPPERCASE-prefix + arbitrary suffix

lib/rules/hook-use-state.js
```diff
@@ -66,23 +66,32 @@ module.exports = {
...
-      const expectedSetterVariableName = valueVariableName ? (
-        `set${valueVariableName.charAt(0).toUpperCase()}${valueVariableName.slice(1)}`
-      ) : undefined;
+      const caseCandidateMatch = valueVariableName ? valueVariableName.match(/(^[a-z]+)(.*)/) : undefined;
+      const upperCaseCandidatePrefix = caseCandidateMatch ? caseCandidateMatch[1] : undefined;
+      const caseCandidateSuffix = caseCandidateMatch ? caseCandidateMatch[2] : undefined;
+      const expectedSetterVariableNames = upperCaseCandidatePrefix ? [
+        `set${upperCaseCandidatePrefix.charAt(0).toUpperCase()}${upperCaseCandidatePrefix.slice(1)}${caseCandidateSuffix}`,
+        `set${upperCaseCandidatePrefix.toUpperCase()}${caseCandidateSuffix}`,
+      ] : undefined;
```
ljharb, line 75:
```suggestion
      const expectedSetterVariableNames = upperCaseCandidatePrefix ? [
        `set${upperCaseCandidatePrefix.charAt(0).toUpperCase()}${upperCaseCandidatePrefix.slice(1)}${caseCandidateSuffix}`,
        `set${upperCaseCandidatePrefix.toUpperCase()}${caseCandidateSuffix}`,
      ] : [];
```

lib/rules/hook-use-state.js
```diff
@@ -66,23 +66,32 @@ module.exports = {
...
 
       const isSymmetricGetterSetterPair = valueVariable
         && setterVariable
-        && setterVariableName === expectedSetterVariableName
+        && expectedSetterVariableNames
+        && expectedSetterVariableNames.indexOf(setterVariableName) !== -1
```
ljharb, line 80:
```suggestion
        && expectedSetterVariableNames.indexOf(setterVariableName) !== -1
```

ljharb (approved):
LGTM, thanks

lib/rules/hook-use-state.js
```diff
@@ -66,23 +66,31 @@ module.exports = {
...
-        && setterVariableName === expectedSetterVariableName
+        && expectedSetterVariableNames.indexOf(setterVariableName) !== -1
         && variableNodes.length === 2;
 
       if (!isSymmetricGetterSetterPair) {
         const suggestions = [
           {
             desc: 'Destructure useState call into value + setter pair',
             fix: (fixer) => {
+              if (!expectedSetterVariableNames) {
```
ljharb, line 87:
```suggestion
              if (expectedSetterVariableNames.length === 0) {
```

ljharb (approved):

lib/rules/hook-use-state.js
```diff
@@ -72,7 +72,7 @@ module.exports = {
       const expectedSetterVariableNames = upperCaseCandidatePrefix ? [
         `set${upperCaseCandidatePrefix.charAt(0).toUpperCase()}${upperCaseCandidatePrefix.slice(1)}${caseCandidateSuffix}`,
         `set${upperCaseCandidatePrefix.toUpperCase()}${caseCandidateSuffix}`,
-      ] : undefined;
```
duncanbeevers: I usually try to avoid this empty-object fallback pattern as it allocates unnecessary empty objects, and because it aligns really naturally with the conditional-access-operator pattern in more-modern js versions.

Happy to follow with the suggestions, but just thought I'd share my thoughts, particularly because they've been shaped somewhat by this repository's tool-for-all-ages mentality.

ljharb: yeah, that's true, but without optional chaining, it's a lot grosser. i prefer having a constant type for things; it's simpler to reason about, and "object allocation" is a non-problem.

duncanbeevers:
> I think this is a bugfix, not an addition :-)

As you wish to call it… I'll trust your judgement and re-message the commit before squashing.

However, imo…

* The previous behavior was well-specified.
* The new behavior introduces no new regressions from the old behavior.

To me this seems like a refinement / enhancement / new feature, rather than addressing a bug that emerged from using  the old behavior.

ljharb:
It certainly adds a regression if someone was depending on the previous behavior; we just forgot about initialisms in the initial creation of the rule.

PR #3247, opened 2022-03-19
Improve no-unstable-nested-components error message and catch React.memo()

zacharyliu:
Fixes https://github.com/yannickcr/eslint-plugin-react/issues/3231.

The bugfix itself is just one line since `React.memo` wrappers are already being detected as components; this plugin just wasn't searching for `CallExpression`s.

I've also updated the error message to be more specific about the recommended fix. Technically this can also be fixed by calling the function directly (`renderComponent(props)` vs `<Component {...props} />`) but I'm not sure if we want to squeeze that into the error message. 

Memoizing with `useCallback()` is not generally safe since the returned function isn't guaranteed to be stable, even if the deps are empty. I have a branch that can detect this, but that probably has other consequences so I'll leave it to a separate PR.

### DIFF
--- a/lib/rules/no-unstable-nested-components.js
+++ b/lib/rules/no-unstable-nested-components.js
@@ -14,7 +14,6 @@ const report = require('../util/report');
 // Constants
 // ------------------------------------------------------------------------------
 
-const ERROR_MESSAGE_WITHOUT_NAME = 'Declare this component outside parent component or memoize it.';
 const COMPONENT_AS_PROPS_INFO = ' If you want to allow component creation in props, set allowAsProps option to true.';
 const HOOK_REGEXP = /^use[A-Z0-9].*$/;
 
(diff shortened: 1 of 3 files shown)


lib/rules/no-unstable-nested-components.js
```diff
@@ -24,11 +23,11 @@ const HOOK_REGEXP = /^use[A-Z0-9].*$/;
 
 /**
  * Generate error message with given parent component name
- * @param {String} parentName Name of the parent component
+ * @param {String} parentName Name of the parent component, if known
  * @returns {String} Error message with parent component name
  */
 function generateErrorMessageWithParentName(parentName) {
-  return `Declare this component outside parent component "${parentName}" or memoize it.`;
+  return `Do not define components during render. React will see a new component type on every render and destroy the entire subtree’s DOM nodes and state (https://reactjs.org/docs/reconciliation.html#elements-of-different-types). Instead, move this component definition out of the parent component${parentName ? ` “${parentName}” ` : ' '}and pass data as props.`;
```
ljharb, line 30: can this be passed in with `data` instead of generating a different template literal every time?

ljharb (commented):
Looks good overall

lib/rules/no-unstable-nested-components.js
ljharb: on further thought, it could, but it'd make the conditional space trickier, so we won't

ljharb (approved):

PR #3249, opened 2022-03-20
[New] `require-default-props`: add option `functions`

### LINKED ISSUE #2396: `defaultProps` rule to be deprecated on function components
`defaultProps` on function components is getting deprecated by React (see: https://github.com/facebook/react/pull/16210)

For this rule to follow suit, it should only be triggered when missing from class components.

The official recommendation in future versions of React is to use ES6 default values for function components.

nix6839:
Closes #2396

## Todo
- [x] Update function comments.
- [x] Add more tests.
- [x] Update docs.
- [x] Update changelog.

### DIFF
--- a/docs/rules/require-default-props.md
+++ b/docs/rules/require-default-props.md
@@ -36,7 +36,7 @@ HelloWorld.defaultProps = {
 
 // Logs:
 // Invalid prop `name` of type `string` supplied to `HelloWorld`, expected `object`.
-ReactDOM.render(<HelloWorld />,  document.getElementById('app'));
+ReactDOM.render(<HelloWorld />, document.getElementById('app'));
 ```
 
 Without `defaultProps`:
@@ -55,7 +55,7 @@ HelloWorld.propTypes = {
 
 // Nothing is logged, renders:
 // "Hello,!"
-ReactDOM.render(<HelloWorld />,  document.getElementById('app'));
+ReactDOM.render(<HelloWorld />, document.getElementById('app'));
 ```
 
 ## Rule Details
@@ -197,13 +197,21 @@ NotAComponent.propTypes = {
 
 ```js
 ...
-"react/require-default-props": [<enabled>, { forbidDefaultForRequired: <boolean>, ignoreFunctionalComponents: <boolean> }]
+"react/require-default-props": [<enabled>, {
+  "forbidDefaultForRequired": <boolean>,
+  "classes": "defaultProps | "ignore",
+  "functions": "defaultProps" | "defaultArguments" | "ignore"
+  // @deprecated Use `functions` option instead.
+  "ignoreFunctionalComponents": <boolean>,
+}]
 ...
 ```
 
-* `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
-* `forbidDefaultForRequired`: optional boolean to forbid prop default for a required prop. Defaults to false.
-* `ignoreFunctionalComponents`: optional boolean to ignore this rule for functional components. Defaults to false.
+- `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
+- `forbidDefaultForRequired`: optional boolean to forbid prop default for a required prop. Defaults to false.
+- `classes`: optional string to determine which strategy a class component uses for defaulting props. Defaults to "defaultProps".
+- `functions`: optional string to determine which strategy a functional component uses for defaulting props. Defaults to "defaultProps".
+- `ignoreFunctionalComponents`: optional boolean to ignore this rule for functional components. Defaults to false. Deprecated, use `functions` instead.
 
 ### `forbidDefaultForRequired`
 
@@ -279,9 +287,129 @@ MyStatelessComponent.propTypes = {
 };
 ```
 
+### `classes`
+
+- "defaultProps": Use `.defaultProps`. It's default.
+- "ignore": Ignore this rule for class components.
+
+Examples of **incorrect** code for this rule, when set to `defaultProps`:
+
+```jsx
+class Greeting extends React.Component {
+  render() {
+    return (
+      <h1>Hello, {this.props.foo} {this.props.bar}</h1>
+    );
+  }
+
+  static propTypes = {
+    foo: PropTypes.string,
+    bar: PropTypes.string.isRequired
+  };
+}
+```
+
+Examples of **correct** code for this rule, when set to `defaultProps`:
+
+```jsx
+class Greeting extends React.Component {
+  render() {
+    return (
+      <h1>Hello, {this.props.foo} {this.props.bar}</h1>
+    );
+  }
+
+  static propTypes = {
+    foo: PropTypes.string,
+    bar: PropTypes.string.isRequired
+  };
+
+  static defaultProps = {
+    foo: "foo"
+  };
+}
+```
+
+### `functions`
+
+- "defaultProps": Use `.defaultProps`. It's default.
+- "defaultArguments": Use default arguments in the function signature.
+- "ignore": Ignore this rule for functional components.
+
+Examples of **incorrect** code for this rule, when set to `defaultArguments`:
+
+```jsx
+function Hello({ foo }) {
+  return <div>{foo}</div>;
+}
+
+Hello.propTypes = {
+  foo: PropTypes.string
+};
+Hello.defaultProps = {
+  foo: 'foo'
+}
+```
+
+```jsx
+function Hello({ foo = 'foo' }) {
+  return <div>{foo}</div>;
+}
+
+Hello.propTypes = {
+  foo: PropTypes.string
+};
+Hello.defaultProps = {
+  foo: 'foo'
+}
+```
+
+```jsx
+function Hello(props) {
+  return <div>{props.foo}</div>;
+}
+
+Hello.propTypes = {
+  foo: PropTypes.string
+};
+```
+
+Examples of **correct** code for this rule, when set to `defaultArguments`:
+
+```jsx
+function Hello({ foo = 'foo' }) {
+  return <div>{foo}</div>;
+}
+
+Hello.propTypes = {
+  foo: PropTypes.string
+};
+```
+
+```jsx
+function Hello({ foo }) {
+  return <div>{foo}</div>;
+}
+
+Hello.propTypes = {
+  foo: PropTypes.string.isRequired
+};
+```
+
+```jsx
+function Hello(props) {
+  return <div>{props.foo}</div>;
+}
+
+Hello.propTypes = {
+  foo: PropTypes.string.isRequired
+};
+```
+
 ### `ignoreFunctionalComponents`
 
 When set to `true`, ignores this rule for all functional components.
+**Deprecated**, use `functions` instead.
 
 Examples of **incorrect** code for this rule:
 
@@ -345,6 +473,7 @@ MyStatelessComponent.propTypes = {
 If you don't care about using `defaultProps` for your component's props that are not required, you can disable this rule.
 
 # Resources
+
 - [Official React documentation on defaultProps](https://facebook.github.io/react/docs/typechecking-with-proptypes.html#default-prop-values)
 
 [PropTypes]: https://reactjs.org/docs/typechecking-with-proptypes.html
--- a/lib/rules/require-default-props.js
+++ b/lib/rules/require-default-props.js
@@ -5,6 +5,8 @@
 
 'use strict';
 
+const entries = require('object.entries');
+const values = require('object.values');
 const Components = require('../util/Components');
 const docsUrl = require('../util/docsUrl');
 const astUtil = require('../util/ast');
@@ -17,6 +19,9 @@ const report = require('../util/report');
 const messages = {
   noDefaultWithRequired: 'propType "{{name}}" is required and should not have a defaultProps declaration.',
   shouldHaveDefault: 'propType "{{name}}" is not required, but has no corresponding defaultProps declaration.',
+  noDefaultPropsWithFunction: 'Don’t use defaultProps with function components.',
+  shouldAssignObjectDefault: 'propType "{{name}}" is not required, but has no corresponding default argument value.',
+  destructureInSignature: 'Must destructure props in the function signature to initialize an optional prop.',
 };
 
 module.exports = {
@@ -35,6 +40,19 @@ module.exports = {
         forbidDefaultForRequired: {
           type: 'boolean',
         },
+        classes: {
+          allow: {
+            enum: ['defaultProps', 'ignore'],
+          },
+        },
+        functions: {
+          allow: {
+            enum: ['defaultArguments', 'defaultProps', 'ignore'],
+          },
+        },
+        /**
+         * @deprecated
+         */
         ignoreFunctionalComponents: {
           type: 'boolean',
         },
@@ -46,7 +64,15 @@ module.exports = {
   create: Components.detect((context, components) => {
     const configuration = context.options[0] || {};
     const forbidDefaultForRequired = configuration.forbidDefaultForRequired || false;
-    const ignoreFunctionalComponents = configuration.ignoreFunctionalComponents || false;
+    const classes = configuration.classes || 'defaultProps';
+    /**
+     * @todo
+     * - Remove ignoreFunctionalComponents
+     * - Change default to 'defaultArguments'
+     */
+    const functions = configuration.ignoreFunctionalComponents
+      ? 'ignore'
+      : configuration.functions || 'defaultProps';
 
     /**
      * Reports all propTypes passed in that don't have a defaultProps counterpart.
@@ -55,14 +81,10 @@ module.exports = {
      * @return {void}
      */
     function reportPropTypesWithoutDefault(propTypes, defaultProps) {
-      // If this defaultProps is "unresolved", then we should ignore this component and not report
-      // any errors for it, to avoid false-positives with e.g. external defaultProps declarations or spread operators.
-      if (defaultProps === 'unresolved') {
-        return;
-      }
+      entries(propTypes).forEach((propType) => {
+        const propName = propType[0];
+        const prop = propType[1];
 
-      Object.keys(propTypes).forEach((propName) => {
-        const prop = propTypes[propName];
         if (!prop.node) {
           return;
         }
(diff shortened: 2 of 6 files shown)


lib/rules/require-default-props.js
```diff
@@ -17,6 +17,9 @@ const report = require('../util/report');
 const messages = {
   noDefaultWithRequired: 'propType "{{name}}" is required and should not have a defaultProps declaration.',
   shouldHaveDefault: 'propType "{{name}}" is not required, but has no corresponding defaultProps declaration.',
+  noDefaultPropsWithFunction: 'Don\'t use defaultProps with function component.',
```
ljharb, line 20:
```suggestion
  noDefaultPropsWithFunction: 'Don’t use defaultProps with function components.',
```
straight quotes are never correct in prose :-)

lib/rules/require-default-props.js
```diff
@@ -17,6 +17,9 @@ const report = require('../util/report');
...
+  shouldInitializeOptional: '`propType {{name}} is not required, but doesn\'t initialize.`',
```
ljharb, line 21:
```suggestion
  shouldInitializeOptional: '`propType {{name}} is not required, but doesn’t initialize.`',
```
also not confident about the wording here

lib/rules/require-default-props.js
```diff
@@ -17,6 +17,9 @@ const report = require('../util/report');
...
+  destructureInSignature: 'Must destructure props in the function signature to initialize optional property.',
```
ljharb, line 22:
```suggestion
  destructureInSignature: 'Must destructure props in the function signature to initialize an optional prop.',
```

lib/rules/require-default-props.js
```diff
@@ -55,14 +62,7 @@ module.exports = {
...
     function reportPropTypesWithoutDefault(propTypes, defaultProps) {
-      // If this defaultProps is "unresolved", then we should ignore this component and not report
-      // any errors for it, to avoid false-positives with e.g. external defaultProps declarations or spread operators.
-      if (defaultProps === 'unresolved') {
-        return;
-      }
-
-      Object.keys(propTypes).forEach((propName) => {
-        const prop = propTypes[propName];
+      Object.entries(propTypes).forEach(([propName, prop]) => {
```
ljharb, line 65: Object.entries isn’t available in every node we support; please use the `object.entries` package for that.

lib/rules/require-default-props.js
```diff
@@ -95,17 +142,31 @@ module.exports = {
       'Program:exit'() {
         const list = components.list();
 
-        Object.keys(list).filter((component) => {
+        Object.values(list).filter((component) => {
```
ljharb, line 145: Same here; please use the object.values package.

lib/util/ast.js
```diff
@@ -219,6 +219,19 @@ function isFunction(node) {
...
 
+/**
+ * Checks if node is a function declaration or expression or arrow function.
+ * @param {ASTNode} node The node to check
+ * @return {Boolean} true if it's a function-like
+ */
+function isFunctionLike(node) {
+  return (
+    node.type === 'FunctionDeclaration'
+    || node.type === 'FunctionExpression'
```
ljharb, line 230:
```suggestion
    || isFunctionLikeExpression(node)
```

lib/util/ast.js
```diff
@@ -219,6 +219,19 @@ function isFunction(node) {
...
+    || node.type === 'ArrowFunctionExpression'
```
ljharb, line 231:
```suggestion
```

tests/lib/rules/require-default-props.js
```diff
@@ -211,6 +211,32 @@ ruleTester.run('require-default-props', rule, {
...
+          return <div>Hello {props.foo}</div>;
+        }
+        Hello.propTypes = {
+          foo: PropTypes.string
+        };
+        Hello.defaultProps = {
+          foo: 'bar;'
+        }
+      `,
+      options: [{ forbidDefaultForRequired: true }],
```
ljharb, line 226: Shouldn’t this be invalid, since foo is required but has a defaultProp?

ljharb (changes requested):

lib/rules/require-default-props.js
nix6839: What's mean that not confident about the wording?

tests/lib/rules/require-default-props.js
nix6839: > ```ts
> Hello.propTypes = {
>   foo: PropTypes.string
> };
> ```

There is no `.isRequired`. Am I missing something?

lib/rules/require-default-props.js
ljharb: Like I’m not sure if this is the best way to word the message, but i don’t have a concrete suggestion yet.

tests/lib/rules/require-default-props.js
ljharb: oh, no, you’re right :-)

lib/rules/require-default-props.js
nix6839: I refactor that and option name. How about this?

nix6839:
Should we also handle `forbidDefaultForRequired` && `assignDefaultToObjectForFc`?

docs/rules/require-default-props.md
```diff
@@ -197,13 +197,21 @@ NotAComponent.propTypes = {
...
-* `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
-* `forbidDefaultForRequired`: optional boolean to forbid prop default for a required prop. Defaults to false.
-* `ignoreFunctionalComponents`: optional boolean to ignore this rule for functional components. Defaults to false.
+- `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
+- `forbidDefaultForRequired`: optional boolean to forbid prop default for
+  a required prop. Defaults to false.
+- `ignoreFunctionalComponents`: optional boolean to ignore this rule for
+  functional components. Defaults to false.
+- `assignDefaultToObjectForFc`: optional booolean to assign default to object
+  instead of `defaultProps` for functional components. Defaults to false.
```
ljharb, line 214:
```suggestion
- `forbidDefaultForRequired`: optional boolean to forbid prop default for a required prop. Defaults to `false`.
- `ignoreFunctionalComponents`: optional boolean to ignore this rule for functional components. Defaults to `false`.
- `assignDefaultToObjectForFc`: optional boolean to assign default to object instead of `defaultProps` for functional components. Defaults to `false`.
```

ljharb (commented):
The point of this option is to *allow* default arguments on props to "count" alongside `defaultProps`, right? It wouldn't force using default arguments, it would just let them satisfy the requirement.

What about `allowDefaultPropArguments`? It can't apply except in an SFC, so i don't think that needs calling out, and the `=` isn't assignment, and it's not "to an object" either.

nix6839:
> The point of this option is to _allow_ default arguments on props to "count" alongside `defaultProps`, right? It wouldn't force using default arguments, it would just let them satisfy the requirement.
> 
> What about `allowDefaultPropArguments`? It can't apply except in an SFC, so i don't think that needs calling out, and the `=` isn't assignment, and it's not "to an object" either.

Here is my thoughts:
**No alongside, only use default arguments** (this means that forbid to `defaultProps`). Other than that, I have the same thoughts as you.

Therefore, what about `forceDefaultPropArguments`, which means to force more `allowDefaultPropArguments?

ljharb:
If that’s what it does, then that’s a better name for the option.

nix6839:
Is there anything I can do?

nix6839:
I don't know what they are doing, but there is an error when `ignorePropsValidation` and `ignoreUnusedPropTypesValidation` are `false`.

docs/rules/require-default-props.md
```diff
@@ -197,13 +197,18 @@ NotAComponent.propTypes = {
...
 ...
 ```
 
-* `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
-* `forbidDefaultForRequired`: optional boolean to forbid prop default for a required prop. Defaults to false.
-* `ignoreFunctionalComponents`: optional boolean to ignore this rule for functional components. Defaults to false.
+- `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
+- `forbidDefaultForRequired`: optional boolean to forbid prop default for a required prop. Defaults to false.
+- `ignoreFunctionalComponents`: optional boolean to ignore this rule for functional components. Defaults to false.
+- `forceDefaultPropArguments`: optional booolean to assign default to object instead of `defaultProps` for functional components. Defaults to false.
```
ljharb, line 211: i think we should make sure that `forceDefaultPropArguments` can not be set to `true` when `ignoreFunctionalComponents` is `true`?

or maybe, we should deprecate `ignoreFunctionalComponents`, and come up with an alternate schema:
```
{
  "forbidDefaultForRequired": <boolean>,

  // for now, this wouldn't support any other choices, and would be optional, but later could be set to "ignore" as long as `functions` wasn't set to "ignore"
  "classes": "defaultProps",

  "functions": "ignore" | "defaultProps" | "default arguments"
}
```

Thoughts?

lib/rules/require-default-props.js
```diff
@@ -95,17 +154,31 @@ module.exports = {
       'Program:exit'() {
         const list = components.list();
 
-        Object.keys(list).filter((component) => {
+        values(list).filter((component) => {
           if (ignoreFunctionalComponents
-            && (astUtil.isFunction(list[component].node) || astUtil.isFunctionLikeExpression(list[component].node))) {
+              && astUtil.isFunctionLike(component.node)) {
```
ljharb, line 159:
```suggestion
              && astUtil.isFunctionLike(component.node)
            ) {
```

lib/rules/require-default-props.js
```diff
@@ -95,17 +154,31 @@ module.exports = {
...
+            return false;
+          }
+          return component.declaredPropTypes !== undefined;
         }).forEach((component) => {
-          reportPropTypesWithoutDefault(
-            list[component].declaredPropTypes,
-            list[component].defaultProps || {}
-          );
+          if (forceDefaultPropArguments
+              && astUtil.isFunctionLike(component.node)) {
```
ljharb, line 170:
```suggestion
          if (forceDefaultPropArguments && astUtil.isFunctionLike(component.node)) {
```

lib/util/ast.js
```diff
@@ -219,6 +219,15 @@ function isFunction(node) {
...
 }
 
+/**
+ * Checks if node is a function declaration or expression or arrow function.
+ * @param {ASTNode} node The node to check
+ * @return {Boolean} true if it's a function-like
+ */
+function isFunctionLike(node) {
+  return node.type === 'FunctionDeclaration' || isFunctionLikeExpression(node);
+}
```
ljharb, line 229: let's add a unit test for this AST helper

tests/lib/rules/require-default-props.js
```diff
@@ -952,6 +1105,32 @@ ruleTester.run('require-default-props', rule, {
...
+      code: `
+        type Props = {
+          foo?: string,
+          bar?: string
+        }
+        function Hello({ foo = 'asdf', bar = 'qwer' }: Props) {
+          return <div>Hello {foo} and {bar}</div>;
+        }
+      `,
+      features: ['flow'],
```
ljharb, line 1118:
```suggestion
      features: ['types'],
```
i think this syntax works with TS too?

tests/lib/rules/require-default-props.js
```diff
@@ -952,6 +1105,32 @@ ruleTester.run('require-default-props', rule, {
...
+      code: `
+        type Props = {
+          foo?: string,
+          bar: string
+        }
+        function Hello({ foo = 'asdf', bar }: Props) {
+          return <div>Hello {foo} and {bar}</div>;
+        }
+      `,
+      features: ['flow'],
```
ljharb, line 1131:
```suggestion
      features: ['types'],
```
also this one?

tests/lib/rules/require-default-props.js
```diff
@@ -982,6 +1161,19 @@ ruleTester.run('require-default-props', rule, {
       `,
       features: ['ts'],
     },
+    {
+      code: `
+        interface Props {
+          foo?: string,
+          bar?: string
+        }
+        const Hello: React.Fc<Props> = ({ foo = 'asdf', bar = 'qwer' }) => {
```
ljharb, line 1170:
```suggestion
        const Hello: React.FC<Props> = ({ foo = 'asdf', bar = 'qwer' }) => {
```

docs/rules/require-default-props.md
nix6839: > we should deprecate ignoreFunctionalComponents, and come up with an alternate schema

I like this, then should the schema be like following lines? 

```
{
  "forbidDefaultForRequired": <boolean>,
  "classes": "defaultProps",
  "functions": "ignore" | "defaultProps" | "defaultArguments"
}
```

Or should I keep the `forbidDefaultForRequired`?

ljharb: Yep, exactly. And the new stuff would only be permitted if ignoreFunctionalComponents is omitted.

nix6839:
@ljharb [Failed test](https://github.com/jsx-eslint/eslint-plugin-react/runs/6137767129?check_suite_focus=true) is not code error. Can you retrigger ci?

nix6839:
Two failed coverages appear to need no testing. Should I guard it?

docs/rules/require-default-props.md
```diff
@@ -197,13 +197,23 @@ NotAComponent.propTypes = {
...
+}]
 ...
 ```
 
-* `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
-* `forbidDefaultForRequired`: optional boolean to forbid prop default for a required prop. Defaults to false.
-* `ignoreFunctionalComponents`: optional boolean to ignore this rule for functional components. Defaults to false.
+- `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
+- `forbidDefaultForRequired`: optional boolean to forbid prop default for a required prop. Defaults to false.
+- `classes`: For now, only works with "defaultProps".
```
ljharb, line 214: not just for now, really? classes can only ever have `defaultProps`.

docs/rules/require-default-props.md
```diff
@@ -197,13 +197,23 @@ NotAComponent.propTypes = {
...
+- `functions`: optional string that this rule determines strategy for the function. "defaultArguments" is recommended. We will change it to deafult later. Defaults to "defaultProps".
```
ljharb, line 215: defaultArguments is very much NOT recommended; defaultProps is a far superior mechanism.
```suggestion
- `functions`: optional string to determine which strategy a functional component uses for defaulting props. Defaults to "defaultProps".
```

docs/rules/require-default-props.md
```diff
@@ -279,9 +289,88 @@ MyStatelessComponent.propTypes = {
 };
 ```
 
+### `functions`
+
+- "defaultProps": Use `.defaultProps`.
+- "defaultArguments": Use object default arguments
+- "ignore": Ignore this rule for functional components.
+
+"defaultArguments" is recommended. We will change it to deafult later.
```
ljharb, line 298: see above: it's not, and i don't plan to ever change the default. `.defaultProps` will remain the superior approach for the foreseeable future.

docs/rules/require-default-props.md
```diff
@@ -279,9 +289,88 @@ MyStatelessComponent.propTypes = {
...
+
+Hello.propTypes = {
+  foo: PropTypes.string.isRequired
+};
+```
+
 ### `ignoreFunctionalComponents`
 
 When set to `true`, ignores this rule for all functional components.
+**Depreacted**, use `functions` instead.
```
ljharb, line 373:
```suggestion
**Deprecated**, use `functions` instead.
```

lib/rules/require-default-props.js
```diff
@@ -35,6 +40,23 @@ module.exports = {
         forbidDefaultForRequired: {
           type: 'boolean',
         },
+        /**
+         * For now, this wouldn't support any other choices, and would be optional,
+         * but later could be set to "ignore" as long as `functions` wasn't set to "ignore"
+         */
+        classes: {
```
ljharb, line 47: might as well add "ignore" support now?

lib/rules/require-default-props.js
```diff
@@ -87,6 +112,65 @@ module.exports = {
...
+     * @param {Object} defaultProps Object of defaultProps to check used.
+     */
+    function reportFunctionComponent(componentNode, declaredPropTypes, defaultProps) {
+      if (defaultProps) {
+        report(context, messages.noDefaultPropsWithFunction, 'noDefaultPropsWithFunction', {
+          node: componentNode,
+        });
+      }
+      if (componentNode.params.length === 0) {
+        return;
```
ljharb, line 128: this can be covered with a functional component that takes no props, no?

lib/rules/require-default-props.js
```diff
@@ -87,6 +112,65 @@ module.exports = {
...
+              return prop.value.type !== 'AssignmentPattern';
+            }).forEach((prop) => {
+              report(context, messages.shouldAssignObjectDefault, 'shouldAssignObjectDefault', {
+                node: prop,
+                data: { name: prop.key.name },
+              });
+            });
+            break;
+          }
+          default:
```
ljharb, line 168: this should be an if/else instead of a switch.

ljharb (changes requested):

docs/rules/require-default-props.md
nix6839: According to the original issue:
> defaultProps on function components is getting deprecated by React (see: facebook/react#16210)

It has been three-ish years since the original issue was reported. Is there any reason why the `defaultProps` is better now?

ljharb: Three-ish years and no actual deprecation, yes.

The primary reason is the same: `.defaultProps` are introspectable from outside the component, and default arguments are not.

nix6839: Oh, I didn't think that far. Thank you.

lib/rules/require-default-props.js
nix6839: [`f38ffa8` (#3249)](https://github.com/jsx-eslint/eslint-plugin-react/pull/3249/commits/f38ffa80909830fb5945039f7d73d7b1fbf14c2f)
I think I cover it above. But I keep getting warnings. What should I do?

nix6839: Oh sorry. Now I see it has been fixed.

ljharb: it doesn't seem to be fixed?

ljharb (approved):

lib/rules/require-default-props.js
nix6839: https://github.com/jsx-eslint/eslint-plugin-react/pull/3249/commits/f38ffa80909830fb5945039f7d73d7b1fbf14c2f
Oh... but I coverd it in the commit above. What's wrong?

ljharb: Looks like you already filtered out everything that has no declared propTypes in the Program:exit listener, so it never gets here with those test cases.

Removing that filtering, and adding a couple additional guards, covers the line - but i don't think that's necessary, so I'll just remove it.

sheiman-pavlo:
When we can expect this to be tagged and released?

ljharb:
v7.30.0 is released.

PR #3251, opened 2022-03-21
[New] `no-unknown-property`: Allow crossOrigin on image tag (SVG)

### LINKED ISSUE #3250: no-unknown-property does not allow `crossOrigin` on svg elements
It looks like `crossorigin` is valid on all SVG elements (https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/crossorigin), however this plugin reports it as invalid. It's probably not useful on all of them (I'm hitting it on `<image>`) but should probably try to follow spec if possible.

zpao:
Fixes #3250


ljharb (approved):
Thanks!

PR #3254, opened 2022-03-23
[readme] remove global usage and eslint version from readme

aladdin-add:
I've also updated eslint@7 => eslint@8(latest).


ljharb (approved):
Thanks! I just removed the version; the default is "latest" and this way we won't have to keep updating it.

PR #3258, opened 2022-03-28
Fix issue #3240

### LINKED ISSUE #3240: no-unused-state with GDSFP and airbnb plugin causes "Cannot read properties of undefined (reading 'variables')"
I have come across this error after a recent upgrade, and it looks like it may be related to https://github.com/yannickcr/eslint-plugin-react/commit/9b227aa8c8d2ffacd508ff2b281f4d7852409464. This only happens with the `airbnb` plugin enabled, which my company is using. If you comment that line out from the `.eslintrc.js` file, it works fine.

Apologies if this is not the appropriate place for this, I can take it to Airbnb if it's a better fit.

Minimal `.eslintrc.js`:

```
module.exports = {
  parser: '@babel/eslint-parser',
  extends: 'airbnb',
  parserOptions: {
    requireConfigFile: false,
  },
};
```

Minimal `test.js`:
```
class Thing extends Component {
  static getDerivedStateFromProps = (nextProps, prevState) => {
  };
};
```

```
# ./node_modules/eslint/bin/eslint.js test.js

ESLint: 7.32.0

TypeError: Cannot read properties of undefined (reading 'variables')
Occurred while linting /root/test.js:2
    at PropertyDefinition, ClassProperty (/root/node_modules/eslint-plugin-react/lib/rules/no-unused-state.js:373:20)
    at /app/frontend/node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Object.emit (/app/frontend/node_modules/eslint/lib/linter/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (/app/frontend/node_modules/eslint/lib/linter/node-event-generator.js:293:26)
    at NodeEventGenerator.applySelectors (/app/frontend/node_modules/eslint/lib/linter/node-event-generator.js:322:22)
    at NodeEventGenerator.enterNode (/app/frontend/node_modules/eslint/lib/linter/node-event-generator.js:336:14)
    at CodePathAnalyzer.enterNode (/app/frontend/node_modules/eslint/lib/linter/code-path-analysis/code-path-analyzer.js:711:23)
    at /app/frontend/node_modules/eslint/lib/linter/linter.js:960:32
```

WillyLiaoWH:
Try to fix #3240 with the smallest change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 ### Fixed
 * [`hook-use-state`]: Allow UPPERCASE setState setter prefixes ([#3244][] @duncanbeevers)
 * `propTypes`: add `VFC` to react generic type param map ([#3230][] @dlech)
+* [`no-unused-state`]: avoid a crash ([#3258][] @WillyLiaoWH @ljharb)
 
 ### Changed
 * [readme] remove global usage and eslint version from readme ([#3254][] @aladdin-add)
@@ -22,6 +23,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 [#3261]: https://github.com/yannickcr/eslint-plugin-react/pull/3261
 [#3260]: https://github.com/yannickcr/eslint-plugin-react/pull/3260
 [#3259]: https://github.com/yannickcr/eslint-plugin-react/pull/3259
+[#3258]: https://github.com/yannickcr/eslint-plugin-react/pull/3258
 [#3254]: https://github.com/yannickcr/eslint-plugin-react/pull/3254
 [#3251]: https://github.com/yannickcr/eslint-plugin-react/pull/3251
 [#3244]: https://github.com/yannickcr/eslint-plugin-react/pull/3244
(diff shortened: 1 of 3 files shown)


ljharb (changes requested):
Thanks; but first we'd need a test case that fails without the fix.

WillyLiaoWH:
Unfortunately, I cannot reproduce it in the test environments. But the error still occurs in my working environment. Here is the sample code that failed in my working environment. The key point is that if I don't declare it in the arrow function form, it would be fine. It would be great if you could spend a few minutes checking the problem. You can just close the PR if you cannot find the root cause either. Thanks for your work again.
```
import React, { PureComponent } from 'react';

class TestNoUnusedState {
  constructor(props) {
    super(props);
    this.state = {
      id: null,
    };
  }

  static getDerivedStateFromProps = (props, state) => {
    if (state.id !== props.id) {
      return {
        id: props.id,
      };
    }

    return null;
  };

  render() {
    return <h1>{this.state.id}</h1>;
  }
}

export default TestNoUnusedState;
```

ljharb:
Your test case reveals another problem, but it passes without your change.

ljharb:
oops, it does fail in older node versions :-) i'll update this PR.

ljharb (approved):

WillyLiaoWH:
Thanks for the fix 🙏

PR #3262, opened 2022-04-02
Remove `david-dm` badge from `README`

ddzz:
The `david-dm` service is no longer available, so this badge can be removed from the `README`.


ljharb:
It’s not working right now, but that’s not the same thing as no longer available. Can you elaborate?

ddzz:
> It’s not working right now, but that’s not the same thing as no longer available. Can you elaborate?

It's been down for months, and there's no indication that the service is coming back. See https://github.com/alanshaw/david/issues/182

ljharb:
It'd still be nice to keep this open, once we hear back from Alan.

ddzz:
> It'd still be nice to keep this open, once we hear back from Alan.

👍 I just reopened it. I'll try to fix the merge conflict this evening.

ljharb (approved):
I suppose we can always re-add if the service comes back.

PR #3263, opened 2022-04-03
Add caching to linting script

ddzz:
Quick PR to add `--cache` to the `lint` script, which should speed up linting by a few seconds.


ljharb:
Caching also can cause issues with correctness, which is why eslint doesn’t enable it by default; I’m not sure it’s worth it for such a small improvement.

PR #3260, opened 2022-04-01
[New] `jsx-tag-spacing`: Add `multiline-always` option

Nokel81:
- This option enforces that the closingTag or the selfClosingTag is on a
  new line if the entire tag is a multiline tag

Fixes #3210.

PR #3264, opened 2022-04-05
Fix: change multiline-always to proportional-always

Nokel81:
- If the node is multiline then require newline
- If the node is singleline then require space

Signed-off-by: Sebastian Malton <sebastian@malton.name>

### DIFF
--- a/lib/rules/jsx-tag-spacing.js
+++ b/lib/rules/jsx-tag-spacing.js
@@ -101,7 +101,7 @@ function validateBeforeSelfClosing(context, node, option) {
   const leftToken = getTokenBeforeClosingBracket(node);
   const closingSlash = sourceCode.getTokenAfter(leftToken);
 
-  if (node.loc.start.line !== node.loc.end.line && option === 'multiline-always') {
+  if (node.loc.start.line !== node.loc.end.line && option === 'proportional-always') {
     if (leftToken.loc.end.line === closingSlash.loc.start.line) {
       report(context, messages.beforeSelfCloseNeedNewline, 'beforeSelfCloseNeedNewline', {
         node,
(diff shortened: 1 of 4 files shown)


ljharb:
Can you elaborate?

This is a breaking change with respect to #3260, but since it's not yet released, we could safely make it.

lib/rules/jsx-tag-spacing.js
```diff
@@ -178,54 +180,68 @@ function validateAfterOpening(context, node, option) {
...
-            return fixer.insertTextBefore(closingToken, '\n');
-          },
-        });
-      }
-    }
-
-    if (leftToken.loc.start.line !== closingToken.loc.start.line) {
-      return;
-    }
+  if (node.selfClosing) {
```
ljharb, line 183: let's revert this change to keep the diff minimal

lib/rules/jsx-tag-spacing.js
```diff
@@ -178,54 +180,68 @@ function validateAfterOpening(context, node, option) {
...
+    return;
+  }
 
-    const adjacent = !sourceCode.isSpaceBetweenTokens(leftToken, closingToken);
+  const sourceCode = context.getSourceCode();
+  const leftToken = option === 'proportional-always'
+    ? getTokenBeforeClosingBracket(node)
+    : sourceCode.getLastTokens(node, 2)[0];
```
ljharb, line 190: does `getLastTokens` always return an array?

Nokel81: Yes it does, at least that is what the eslint docs imply.

ljharb (approved):

burtek:
Docs not changed? https://github.com/jsx-eslint/eslint-plugin-react/issues/3294

ljharb:
yes, they were overlooked.

PR #3265, opened 2022-04-06
[Refactor]: improve performance for detecting function components

golopot:
Benchmark on a private repo with `time npx eslint .` (real time):

|  | before | after |
| --- | -- | -- |
| run 1 | 6.6s | 4.8s |
| run 2 | 6.5s | 4.7s |
| run 3 | 6.7s | 4.8s |

Benchmark with `TIMING=10 npx eslint .`:

Before:
```
Rule                                | Time (ms) | Relative
:-----------------------------------|----------:|--------:
react/destructuring-assignment      |   345.923 |     7.0%
react/no-unstable-nested-components |   208.838 |     4.2%
react/no-did-mount-set-state        |   189.097 |     3.8%
react/boolean-prop-naming           |   187.663 |     3.8%
react/no-direct-mutation-state      |   186.664 |     3.8%
react/no-deprecated                 |   184.020 |     3.7%
react/hook-use-state                |   146.379 |     3.0%
react/no-access-state-in-setstate   |   145.210 |     3.0%
react/prefer-stateless-function     |   133.564 |     2.7%
react/require-optimization          |   128.301 |     2.6%
```

After:
```
Rule                                | Time (ms) | Relative
:-----------------------------------|----------:|--------:
react/no-did-mount-set-state        |   181.362 |     6.2%
react/destructuring-assignment      |   162.586 |     5.5%
react/no-deprecated                 |   134.652 |     4.6%
react/no-direct-mutation-state      |   124.110 |     4.2%
react/boolean-prop-naming           |   116.978 |     4.0%
react/no-did-update-set-state       |   113.393 |     3.9%
react/no-unknown-property           |    92.802 |     3.2%
react/hook-use-state                |    91.712 |     3.1%
react/no-unstable-nested-components |    90.477 |     3.1%
react/no-access-state-in-setstate   |    87.266 |     3.0%
```

### DIFF
--- a/lib/util/ast.js
+++ b/lib/util/ast.js
@@ -62,24 +62,28 @@ function findReturnStatement(node) {
   return loopNodes(bodyNodes);
 }
 
+// eslint-disable-next-line valid-jsdoc -- valid-jsdoc cannot parse function types.
 /**
  * Helper function for traversing "returns" (return statements or the
  * returned expression in the case of an arrow function) of a function
  *
  * @param {ASTNode} ASTNode The AST node being checked
  * @param {Context} context The context of `ASTNode`.
- * @param {function} enterFunc Function to execute for each returnStatement found
+ * @param {(returnValue: ASTNode, breakTraverse: () => void) => void} onReturn
+ *   Function to execute for each returnStatement found
  * @returns {undefined}
  */
-function traverseReturns(ASTNode, context, enterFunc) {
+function traverseReturns(ASTNode, context, onReturn) {
   const nodeType = ASTNode.type;
 
   if (nodeType === 'ReturnStatement') {
-    return enterFunc(ASTNode);
+    onReturn(ASTNode.argument, () => {});
+    return;
   }
 
   if (nodeType === 'ArrowFunctionExpression' && ASTNode.expression) {
-    return enterFunc(ASTNode.body);
+    onReturn(ASTNode.body, () => {});
+    return;
   }
 
   /* TODO: properly warn on React.forwardRefs having typo properties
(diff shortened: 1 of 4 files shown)


lib/util/ast.js
```diff
@@ -111,15 +115,23 @@ function traverseReturns(ASTNode, context, enterFunc) {
...
+        case 'IfStatement':
+        case 'ForStatement':
+        case 'WhileStatement':
+        case 'SwitchStatement':
+        case 'SwitchCase':
+          return;
         default:
+          this.skip();
       }
     },
```
golopot, line 136: Avoids unnecessary traversals by skipping all node types but the handled cases.

lib/util/jsx.js
```diff
@@ -95,63 +94,48 @@ function isWhiteSpaces(value) {
...
+  };
 
-    return found && estraverse.VisitorOption.Break;
+  let found = false;
+  astUtil.traverseReturns(ASTnode, context, (node, breakTraverse) => {
+    if (isJSXValue(node)) {
+      found = true;
+      breakTraverse();
+    }
   });
```
golopot, line 139: Same as above and simplify by not using `astUtil.traverse`.

tests/util/ast.js
```diff
@@ -96,8 +96,8 @@ describe('ast', () => {
 
       assert.strictEqual(enterCalls.length, 6);
 
-      enterCalls.forEach((node, idx) => {
-        assert.strictEqual(node.lastArg.argument.value, idx);
+      enterCalls.forEach((call, idx) => {
+        assert.strictEqual(call.args[0].value, idx);
```
ljharb, line 100: why the change from last arg to first arg?

ljharb (commented):
I can't pretend I understand this change, but if all the tests pass, then i guess yay?

tests/util/ast.js
golopot: Previously the called function has only one parameter: `(node) => void`, I refactored that function to `(node, fn) => void`. The last arg changed to a different thing.

golopot:
The gist is that `isReturningJSX` tries to find return statements that contains JSX. For that a tree traversing is performed. Previously the tree traversing is inefficient because it make unnecessary traversals to nodes that cannot contain a return statement. This PR makes the tree traversal more efficient by only traverse what is needed.

ljharb (approved):

PR #3266, opened 2022-04-07
[Fix] Use curly single quote in error message for `jsx-no-useless-fragment`

develohpanda:
As per title :) 

Currently live:
image

### DIFF
--- a/lib/rules/jsx-no-useless-fragment.js
+++ b/lib/rules/jsx-no-useless-fragment.js
@@ -79,7 +79,7 @@ function containsCallExpression(node) {
 }
 
 const messages = {
-  NeedsMoreChildren: 'Fragments should contain more than one child - otherwise, there‘s no need for a Fragment at all.',
+  NeedsMoreChildren: 'Fragments should contain more than one child - otherwise, there’s no need for a Fragment at all.',
   ChildOfHtmlElement: 'Passing a fragment to an HTML element is useless.',
 };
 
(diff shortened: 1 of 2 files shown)


lib/rules/jsx-no-useless-fragment.js
```diff
@@ -79,7 +79,7 @@ function containsCallExpression(node) {
 }
 
 const messages = {
-  NeedsMoreChildren: 'Fragments should contain more than one child - otherwise, there‘s no need for a Fragment at all.',
+  NeedsMoreChildren: 'Fragments should contain more than one child - otherwise, there\'s no need for a Fragment at all.',
```
ljharb, line 82: straight quotes are always incorrect in prose; in fact, the curly single quote is the *only* correct character to use.

I do see it's pointing the wrong way tho, which means:
```suggestion
  NeedsMoreChildren: 'Fragments should contain more than one child - otherwise, there’s no need for a Fragment at all.',
```
is the proper fix.

ljharb (changes requested):

lib/rules/jsx-no-useless-fragment.js
develohpanda: > straight quotes are always incorrect in prose; in fact, the curly single quote is the only correct character to use.

Today I learned! Thank you for the review 🙌

ljharb (approved):

PR #3273, opened 2022-04-18
[Refactor] improve performance by avoiding unnecessary component detect

golopot:
Rules using `Component.detect()` are generally slow. Some rules uses `Component.detect(context, components, utils)` solely for the `utils` it provided. These `Component.detect`s can be avoided by extracting some `utils` functions into a module.

Benchmark with config `plugin:react/all`:
```diff
$ time npx eslint .
- 16.429
+ 13.692
```

### DIFF
--- a/lib/rules/jsx-no-bind.js
+++ b/lib/rules/jsx-no-bind.js
@@ -8,7 +8,6 @@
 'use strict';
 
 const propName = require('jsx-ast-utils/propName');
-const Components = require('../util/Components');
 const docsUrl = require('../util/docsUrl');
 const jsxUtil = require('../util/jsx');
 const report = require('../util/report');
@@ -63,7 +62,7 @@ module.exports = {
     }],
   },
 
-  create: Components.detect((context) => {
+  create(context) {
     const configuration = context.options[0] || {};
 
     // Keep track of all the variable names pointing to a bind call,
@@ -198,5 +197,5 @@ module.exports = {
         }
       },
     };
-  }),
+  },
 };
--- a/lib/types.d.ts
+++ b/lib/types.d.ts
@@ -13,9 +13,7 @@ declare global {
   type JSXFragment = ASTNode;
   type JSXSpreadAttribute = ASTNode;
 
-  interface Context extends eslint.SourceCode {
-    getFirstTokens(node: estree.Node | ASTNode, options?: eslint.SourceCode.CursorWithCountOptions): eslint.AST.Token[];
-  }
+  type Context = eslint.Rule.RuleContext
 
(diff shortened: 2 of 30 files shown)


lib/rules/jsx-no-bind.js
```diff
@@ -63,7 +62,7 @@ module.exports = {
     }],
   },
 
-  create: Components.detect((context) => {
+  create(context) {
```
ljharb, line 65: this doesn't seem right - we don't want this rule running on jsx that's outside of a detected component.

ljharb (changes requested):
Before I review the whole thing - this seems like a very scary change to me. I'd much rather make component detection faster, and have *every* rule unconditionally be wrapped with component detection logic, than risk false positive warnings just to eke out a bit more performance from an eslint run (something that even without jest-eslint-runner is pretty fast already)

golopot:
This change does not cause more reports or less reports. This change is a safe refactoring that does not change ouput. The `Component.detect` thing has no effect if you are not using `components` or `utils` it provides. Using `Component.detect` without using `components` or `utils` is analogous to spending lots of time computing a value and not using it.

ljharb:
My expectation would be that using Components.detect would restrict the AST node visitors to only applying within components. If that's not how it works, then I clearly misunderstand the abstraction.

golopot:
> My expectation would be that using Components.detect would restrict the AST node visitors to only applying within components. If that's not how it works, then I clearly misunderstand the abstraction.

Yes some rules uses `Component.detect((context) => { return { ImportDeclaration(node) {}}) }`, event though `ImportDeclaration` cannot possibly be in a component.

ljharb:
Fair enough; I'll rereview.

lib/types.d.ts
```diff
@@ -13,9 +13,7 @@ declare global {
   type JSXFragment = ASTNode;
   type JSXSpreadAttribute = ASTNode;
 
-  interface Context extends eslint.SourceCode {
-    getFirstTokens(node: estree.Node | ASTNode, options?: eslint.SourceCode.CursorWithCountOptions): eslint.AST.Token[];
-  }
+  type Context = eslint.Rule.RuleContext
```
ljharb, line 16: why this change?

lib/util/Components.js
```diff
@@ -375,30 +276,6 @@ function componentRule(rule, context) {
       return isCreateElement(node, context);
```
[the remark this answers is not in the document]
golopot: Yes

lib/types.d.ts
golopot: That Context type definition is a duct tape definition. It lacks many methods.

ljharb: ah, so this one includes getFirstTokens, awesome

ljharb (approved):

PR #3276, opened 2022-04-23
[Refactor] improve performance of component detection

golopot:
Some node visitors for components detection are redundant.

Benchmark with `time npx eslint .` (mean of 5 runs):

```diff
- 5.06 s
+ 5.01 s
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,10 +32,12 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 * [Refactor] add `isParenthesized` AST util ([#3203][] @Belco90)
 * [Docs] `default-props-match-prop-types`, `require-default-props`, `sort-prop-types`: fix typos ([#3279][] @nix6839)
 * [Refactor] improve performance of rule merging ([#3281][] @golopot)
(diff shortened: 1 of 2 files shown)


ljharb:
Can you elaborate on how they're redundant?

golopot:
Our heuristics for detecting function components is checking if the function returns jsx or null (`isReturningJSX`), this function is expensive to execute. In a simple test case `isReturningJSX` is found to run 3 times for a simple function.

This part is redundant
https://github.com/jsx-eslint/eslint-plugin-react/blob/24621e4454109a285519e987d9a425b88339f734/lib/util/Components.js#L829-L840

, because checking `isReturningJSX` is included in `getParentComponent` -> `getParentStatelessComponent` -> `getStatelessComponent` -> `isReturningJSX`.

https://github.com/jsx-eslint/eslint-plugin-react/blob/24621e4454109a285519e987d9a425b88339f734/lib/util/Components.js#L785-L796

ljharb:
Would it be simpler to memoize isReturningJSX?

golopot:
> Would it be simpler to memoize isReturningJSX?

No, it wouldn't. Deleting redundant code is better than adding memoizing.

ljharb (approved):

PR #3279, opened 2022-04-29
[Docs] `default-props-match-prop-types`, `require-default-props`, `sort-prop-types`: fix typos

nix6839:


docs/rules/require-default-props.md
```diff
@@ -165,7 +165,7 @@ MyStatelessComponent.propTypes = {
 };
 
 MyStatelessComponent.defaultProps = {
-    bar: 'some default'
```
ljharb, line 168: let's also fix the indentation in the previous example

ljharb (changes requested):
thanks!

docs/rules/require-default-props.md
nix6839: Where is typo in the previouse example? I don't find that. It's means another document?

ljharb: the previous example seems to have one-space indentation

nix6839: Sorry, I don't understand. Are you talking about revert?

ljharb: I'm talking about line 165 - it looks like the `};` is an extra space indented. is that just github's UI?

nix6839: There is no extra space indented. Just github's UI :)

ljharb (approved):

Issue #3140, opened 2021-11-16
False positives of react/prop-types in forwardRefs with 7.27.0


gunters63:
prop type checking seems to broken inside forwardRef with version 7.27.0, 7.26.1 still worked fine.

The error only happens for type intersection hierarchies (I didn't test interface extensions).

When you have types like this:

````
type ControlProps = {
  className?: string | undefined;
};

type NamedProps = {
  name: string;
};

type ControlPropsWithChildren = ControlProps & {
  children?: ReactNode;
};

type BaseButtonProps = ControlPropsWithChildren &
  NamedProps & {
    onClick?: (() => void) | undefined;
    onMouseDown?: (() => void) | undefined;
    onMouseUp?: (() => void) | undefined;
    disabled?: boolean | undefined;
    width?: number;
    type?: 'submit' | 'reset' | 'button' | undefined;
  };

````

The plugin will now falsely flag all properties in nested type intersections as missing in props validation, in this case name, className and children (see screenshot wiggly lines).

The problem does not happen when the component is not wrapped in forwardRef.
It also didn't happen with version 7.26.1

````
D:\...\BaseButton.tsx
  23:7  error  'name' is missing in props validation       react/prop-types
  24:7  error  'className' is missing in props validation  react/prop-types
  28:7  error  'children' is missing in props validation   react/prop-types
````

![image](https://user-images.githubusercontent.com/90346/141880939-bdd699a5-aae4-4194-b5b3-80729b0fed27.png)

Its likely that this commit: https://github.com/yannickcr/eslint-plugin-react/commit/4bc34995dbab597492c5081493d77755fd3501e8 triggered the problem

**Comment by ljharb** [comment]
cc @vedadeepta

**Comment by vaniyokk** [comment]
Same problem

**Comment by mataspetrikas** [comment]
I have noticed a potential cause/fix. 

The validation fails in the following case:

```
type InputTextProps = InputHTMLAttributes<HTMLInputElement> & {
  label: string;
  helperText?: string;
  error?: boolean;
};

export default forwardRef<HTMLInputElement, InputTextProps>(function InputText(
  props: InputTextProps,
  forwardedRef: Ref<HTMLInputElement>,
) {
  const { name, value, label, error = false, helperText, ...rest } = props;
  const [isFocused, setIsFocused] = useState<boolean>(false);
  const isActive = (typeof value === 'string' && value.length > 0) || isFocused;

[...]

PR #3280, opened 2022-04-30
[Fix] propType: ignore imported types/interface in forwardRef generic

vedadeepta:
Fixes: #3140 

Fix is to ignore when type declarations are not present in the same file.

Turns out imported external interfaces & types were already handled by the searchDeclaration function, the error was happening because i mistakenly hardcoded `ignorePropsValidation: false` when handling the `forwardRef<element, props>` case 💀


ljharb (approved):
Thanks!

PR #3281, opened 2022-05-02
[Refactor] improve performance of rule merging

golopot:
The performance for merged rule in `Component.detect` is improved. In an end-to-end linting benchmark this change is 4% faster.

Benchmark (mean of 5 runs):
```diff
- 5.164 s
+ 4.966 s
```

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -247,6 +247,39 @@ function getWrapperFunctions(context, pragma) {
   ]);
 }
 
+// eslint-disable-next-line valid-jsdoc
+/**
+ * Merge many eslint rules into one
+ * @param {{[_: string]: Function}[]} rules the returned values for eslint rule.create(context)
+ * @returns {{[_: string]: Function}} merged rule
+ */
+function mergeRules(rules) {
+  /** @type {Map<string, Function[]>} */
+  const handlersByKey = new Map();
+  rules.forEach((rule) => {
+    Object.keys(rule).forEach((key) => {
+      const fns = handlersByKey.get(key);
+      if (!fns) {
+        handlersByKey.set(key, [rule[key]]);
+      } else {
+        fns.push(rule[key]);
+      }
+    });
+  });
+
+  /** @type {{[key: string]: Function}} */
+  const rule = {};
+  handlersByKey.forEach((fns, key) => {
+    rule[key] = function mergedHandler(node) {
+      fns.forEach((fn) => {
+        fn(node);
+      });
+    };
+  });
+
+  return rule;
+}
+
 function componentRule(rule, context) {
   const pragma = pragmaUtil.getFromContext(context);
   const sourceCode = context.getSourceCode();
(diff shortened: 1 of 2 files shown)


lib/util/Components.js
```diff
@@ -859,44 +893,21 @@ function componentRule(rule, context) {
...
-    Object.keys(propTypesInstructions),
-    Object.keys(usedPropTypesInstructions),
-    Object.keys(defaultPropsInstructions),
-    Object.keys(reactImportInstructions)
-  ));
 
-  allKeys.forEach((instruction) => {
-    updatedRuleInstructions[instruction] = (node) => {
-      if (instruction in detectionInstructions) {
-        detectionInstructions[instruction](node);
```
golopot, line 878: This property access performed for each AST node turned out to be performance bottleneck.

lib/util/Components.js
```diff
@@ -247,6 +247,40 @@ function getWrapperFunctions(context, pragma) {
...
+      }
+    }
+  }
+
+  /** @type {{[key: string]: Function}} */
+  const rule = {};
+  for (const key of handlersByKey.keys()) {
+    const fns = handlersByKey.get(key);
+    rule[key] = function mergedHandler(node) {
+      for (const fn of fns) {
```
ljharb, line 275: for-of is slow; `fns.forEach((node) => fn(node))` is almost certainly faster (and cleaner; loops are bad)

lib/util/Components.js
```diff
@@ -247,6 +247,40 @@ function getWrapperFunctions(context, pragma) {
...
+        handlersByKey.set(key, [rule[key]]);
+      } else {
+        fns.push(rule[key]);
+      }
+    }
+  }
+
+  /** @type {{[key: string]: Function}} */
+  const rule = {};
+  for (const key of handlersByKey.keys()) {
```
ljharb, line 272: same here; let's use `handlersByKey.forEach` here

lib/util/Components.js
```diff
@@ -247,6 +247,40 @@ function getWrapperFunctions(context, pragma) {
...
+  for (const rule of rules) {
+    for (const key of Object.keys(rule)) {
+      const fns = handlersByKey.get(key);
+      if (!fns) {
+        handlersByKey.set(key, [rule[key]]);
+      } else {
+        fns.push(rule[key]);
+      }
+    }
+  }
```
ljharb, line 268:
```suggestion
  rules.forEach((rule) => {
    Object.keys(rule).forEach((key) => {
      const fns = handlersByKey.get(key);
      if (!fns) {
        handlersByKey.set(key, [rule[key]]);
      } else {
        fns.push(rule[key]);
      }
    });
  });
```

nattapong0643510634 (approved):

nattapong0643510634 (approved):

lib/util/Components.js
golopot: for-of is not slow in node, v8 have done optimizations to make it performs comparable to for loop.

```js
const N = 100000;
const length = 100;

const array = new Array(length).fill().map((_, i) => i);

function benchmarkForLoop() {
  let start = performance.now();
  let x = 0;
  for (let i = 0; i < N; i++) {
    for (let j = 0; j < array.length; j++) {
      x = (x + array[j]) % 7;
    }
  }
  let end = performance.now();
  console.log(end - start);
}

function benchmarkForOfLoop() {
  let start = performance.now();
  let x = 0;
  for (let i = 0; i < N; i++) {
    for (const y of array) {
      x = (x + y) % 7;
    }
  }
  let end = performance.now();
  console.log(end - start);
}

function benchmarkForEach() {
  let start = performance.now();
  let x = 0;
  for (let i = 0; i < N; i++) {
    array.forEach((y) => {
      x = (x + y) % 7;
    });
  }
  let end = performance.now();
  console.log(end - start);
}

benchmarkForLoop();
benchmarkForOfLoop();
benchmarkForEach();

// Output:
// 42.265081003308296
// 47.21948900818825
// 57.372923001646996
```

ljharb: Try those benchmarks without using Array(), which creates a sparse array that .fill may not be able to un-taint.

lib/util/Components.js
```diff
@@ -247,6 +247,39 @@ function getWrapperFunctions(context, pragma) {
   ]);
 }
 
+// eslint-disable-next-line valid-jsdoc
```
ljharb, line 250: is there no way to have this rule pass with this function's signature?

ljharb (approved):

lib/util/Components.js
golopot: That rule has been deprecated for a long time and it lacks support for lots of typescript syntaxes.

ljharb: gotcha; so we'd have to use a similar rule from the typescript plugin?

golopot: You are looking for eslint-plugin-jsdoc .

golopot: typescript plugin does not handle jsdoc as far as I know

PR #3283, opened 2022-05-04
[no-constructed-context-values] Update isJsxContext check to allow for JSXIdentifier ending with Provider

liuyenwei:
`jsx/no-constructed-context-values` currently only works for context providers that are used as `<Context.Provider value={..}>`

We have a lot of use cases where, instead of exporting the context directly, we explicitly export the provider as a named export.  The naming convention for this is typically `${NameOfContext}Provider`

The proposed change here is to update the `isJsxContext` check to take the above scenario into account and run the lint rule for these elements as well.

### DIFF
--- a/lib/rules/jsx-no-constructed-context-values.js
+++ b/lib/rules/jsx-no-constructed-context-values.js
@@ -143,12 +143,15 @@ module.exports = {
     return {
       JSXOpeningElement(node) {
         const openingElementName = node.name;
-        if (openingElementName.type !== 'JSXMemberExpression') {
-          // Has no member
-          return;
-        }
 
-        const isJsxContext = openingElementName.property.name === 'Provider';
+        // Consider an element a context provider if the name is either:
+        // - identifier and ends with Provider (i.e. SomeContextProvider)
+        // - member expression and has Provider as a property (i.e. SomeContext.Provider)
+        const isJsxContext = (openingElementName.type === 'JSXMemberExpression'
+            && openingElementName.property.name === 'Provider')
+          || (openingElementName.type === 'JSXIdentifier'
+            && openingElementName.name.endsWith('Provider'));
+
         if (!isJsxContext) {
           // Member is not Provider
           return;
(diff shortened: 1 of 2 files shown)


lib/rules/jsx-no-constructed-context-values.js
```diff
@@ -143,12 +143,15 @@ module.exports = {
...
-        }
 
-        const isJsxContext = openingElementName.property.name === 'Provider';
+        // Consider an element a context provider if the name is either:
+        // - identifier and ends with Provider (i.e. SomeContextProvider)
+        // - member expression and has Provider as a property (i.e. SomeContext.Provider)
+        const isJsxContext = (openingElementName.type === 'JSXMemberExpression'
+            && openingElementName.property.name === 'Provider')
+          || (openingElementName.type === 'JSXIdentifier'
+            && openingElementName.name.endsWith('Provider'));
```
ljharb, line 153: This isn't a convention we can reasonably hardcode into this plugin, I'm afraid.

ljharb (changes requested):

lib/rules/jsx-no-constructed-context-values.js
liuyenwei: another option here could be to add a config option for specifying a pattern to test an identifier against.  
maybe something like:
```
jsxIdentifierPattern: RegExp
```

what do you think about something like ^?  happy to update the PR if you're okay with this approach

ljharb: Allowing regexes in eslint config is a very bad idea, and is a magnet for ReDOS CVEs.

I don't think it's possible to have your convention work with static analysis, I'm afraid. You could perhaps use a pragma comment like we have to identify React components that extend "not React.Component", but you'd still have to put it in every consuming file.

liuyenwei: thanks for the reply.  

re: the regexp, my thought here was that it would be isolated to just the jsx identifier so the risk would be low.  also, afaik there are other rules (in this project) that do support regex configs so i thought it would be an acceptable approach (ex: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/lib/rules/boolean-prop-naming.js#L45) 

it seems that this _should_ be something that we can detect statically since we know the identifier, it's just a matter of validating that the identifier is a valid context provider.  another idea here would be use a function as a config option (i.e. `isJsxIdentifierContextProvider`) but that feels pretty overkill imo

not trying to be pushy here - just trying to brainstorm ideas to support our use case.  would love to avoid forking this rule since 99% of the logic would be the same.

PR #3293, opened 2022-05-20
Clarify the number of examples.

ioggstream:
## This PR

- clarifies these were three different examples.
- I didn't understand these were three different examples :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,10 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ### Changed
 * [Docs] [`jsx-tag-spacing`]: rename option from [#3264][] ([#3294[] @ljharb)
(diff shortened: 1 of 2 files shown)


ljharb:
Typically it's better not to count the number, because then we have to update it every time we add or remove an example.

Perhaps instead, splitting it into separate code blocks?

ioggstream:
> splitting it into separate code blocks?

That's ok! Can do it in this PR?

> Typically it's better not to count the number, 
> because then we have to update it every time we add or remove an example.

Yep. In some cases though, clarifying that is useful. E.g. see [this excerpt from the latest HTTP spec](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.10.2.3.p.6).

ljharb:
Yes, let's split it in this PR instead of hardcoding a count.

ljharb (approved):
Thanks!

PR #3304, opened 2022-06-01
[Fix] `display-name`: fix false positive when using memo

### LINKED ISSUE #3303: False positive for react/display-name when using memo
After upgrading to 7.30 we have found out that some function component are highlighted by this rule even if they are declared with normal functions rather than arrow functions.

Looking at the code base we found out that these happens only for those components which has an arrow comparison function:

```jsx
function MyComponent(props) {
   return <b>{props.name}</b>;
}

const MemoizedMyComponent = React.memo(
    MyComponent,
    (prevProps, nextProps) => prevProps.name === nextProps.name
)
```

It looks like the linter considers the comparison function as an other component.

golopot:
Fix #3303.

PR #3276 caused the following to be considered a component, while in v7.29.0 it is not considered a component:

```js
const MemoizedMyComponent = React.memo(
  MyComponent,
  () => 1
)
```

This PR adds a check that the function being wrapped must return JSX to be considered a component.

The bug is so subtle and I am not sure this is the right way to fix it.


ljharb (approved):
Thanks!

PR #3305, opened 2022-06-01
[Fix] `display-name`: fix false positive for HOF returning only nulls and literals

### LINKED ISSUE #3300: False-positive for react/display-name after 7.29.4 → 7.30.0 upgrade
After upgrading eslint-plugin-react from 7.29.4 to 7.30.0 a new linting error appeared in my project that doesn't seem right. My project is a big full-stack monolith web app and it contains both Vue.js and React.js code. Now I am getting `react/display-name` for some Vue/Vuex code that has nothing to do with React. I stripped down the file, the original was much longer, but this still triggers the error:

```js
import Vuex from 'vuex'
import Vue from 'vue'

// code omitted

const getters = {
  getChat: state => (scopePath, partnerId) => {
    const selectedChatScope = state.chats && state.chats.find(chat => chat.path === scopePath)

    if (!selectedChatScope) {
      return null
    }

    return 'something'
  }
}

export {
  getters
}
```

```
  7:21  error  Component definition is missing display name  react/display-name
```

golopot:
Fixes #3300.

This bug is caused by #3276.

Stop considering the following a component:
```js
const f = (a) => () => {
  if (a) {
    return null;
  }
  return 1;
};
```

The bug is triggered when a literal like `1` or `"a"` is returned somewhere.

### DIFF
--- a/tests/lib/rules/no-this-in-sfc.js
+++ b/tests/lib/rules/no-this-in-sfc.js
@@ -275,58 +275,5 @@ ruleTester.run('no-this-in-sfc', rule, {
         { messageId: 'noThisInSFC' },
       ],
     },
-    {
-      code: `
-        class Foo {
-          bar() {
-            return () => {
-              this.something();
-              return null;
-            }
-          }
-        }
-      `,
-      errors: [{ messageId: 'noThisInSFC' }],
-    },
-    {
-      code: `
-        class Foo {
-          bar = () => () => {
-            this.something();
-            return null;
-          };
-        }
-      `,
-      features: ['class fields', 'no-ts-old'], // TODO: FIXME: remove `no-ts-old` and fix
-      errors: [{ messageId: 'noThisInSFC' }],
-    },
-    {
-      code: `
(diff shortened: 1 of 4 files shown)


tests/lib/rules/no-this-in-sfc.js
```diff
@@ -275,58 +275,5 @@ ruleTester.run('no-this-in-sfc', rule, {
...
-          bar() {
-            () => () => {
-              this.something();
-              return null;
-            };
-          }
-        }
-      `,
-      errors: [{ messageId: 'noThisInSFC' }],
-    },
```
golopot, line 330: These tests fails after this change.

ljharb (approved):

ljharb:
Looks like after a rebase, there's some tests failing.

golopot:
I have rebased the PR and removed some tests from the incoming commits.

ljharb (approved):

SiljeAlice (approved):
Good

PR #3307, opened 2022-06-08
add allowEmpty option to jsx-no-useless-fragment rule

FrankNorton32:
Optionally allow `<></>`

While attempting to add jsx-no-useless-fragments to my company's repo, I ran into the issue that while we wanted to remove instances where the fragment was only wrapping a single element, this rule also removed fragments with no children. While this can be replaced with `null`  in most cases, since we were working with Typescript, I ran into type issues in many places. For some of these, the typing was not reachable (npm packages) which lead to many exceptions to the rule. Passing empty fragments where a `JSX.Element` is expected but there is not content is common practice and so I created an exception to the rule that allows for this usage to exist.


ljharb:
Generally it's best to discuss this sort of thing in an issue before filing a PR.

Can you elaborate on when you need to pass an element, an empty fragment suffices, and `null` isn't accepted?

ljharb:
@FrankNorton32 id really appreciate more information before just closing this.

FrankNorton32:
> @FrankNorton32 id really appreciate more information before just closing this.

My team ran into an issue where a component who's type was out of reach (NPM package) required a JSX element but we needed it to possibly be empty. One of the possible inputs on this component, used in multiple places, was `<></>`. This linting rule broke this but due to the type requirement, we could not pass `null` as would be the normal solution. Ultimately, I believe we solved this by using `"allowExpressions": true` and passing `<>{null}</>` and in our case, the allowExpressions option is preferred for more than just this issue so it was the right choice for us and it likely makes this PR a fix for a something that no one else will have a problem with. Though, it might just be good practice to allow a more targeted approach for addressing <></> without opening the doors to <>{...}</>.

ljharb:
Gotcha, thanks for the context. It seem like an even better fix is for that npm package to accept `null` instead of forcing you to pass an empty fragment :-)

I'll close.

PR #3311, opened 2022-06-21
Allow multiline JSX expressions when prevent option is true: `jsx-newline`

### LINKED ISSUE #3033: [feature request] react/jsx-newline
Hi,

I would find it interesting to provide a new option to the [react/jsx-newline](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-newline.md) rule, so that there is a newline between JSX expressions ONLY if they are multiline.

I find that single line JSX expressions are easier to read and don't require a space between them, but multiline expressions can quickly become hard to read.

Would anyone be eager to help me implement this feature, as this would be the first rule I would contributing to the project?

Thanks in advance!

TildaDares:
Fixes #3033

This is an initial PR to confirm that this is the change required to allow multi-lines in the `jsx-newlines` rule.

### DIFF
--- a/lib/rules/jsx-newline.js
+++ b/lib/rules/jsx-newline.js
@@ -16,7 +16,12 @@ const report = require('../util/report');
 const messages = {
   require: 'JSX element should start in a new line',
   prevent: 'JSX element should not start in a new line',
+  allowMultilines: 'Multiline JSX elements should start in a new line',
 };
+
+function isMultilined(node) {
+  return node.loc.start.line !== node.loc.end.line;
+}
 
 module.exports = {
   meta: {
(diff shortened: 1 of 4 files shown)


lib/rules/jsx-newline.js
```diff
@@ -16,6 +16,7 @@ const report = require('../util/report');
 const messages = {
   require: 'JSX element should start in a new line',
   prevent: 'JSX element should not start in a new line',
+  allowMultilines: 'Multilined JSX element should start in a new line',
```
ljharb, line 19:
```suggestion
  allowMultilines: 'Multiline JSX elements should start in a new line',
```

lib/rules/jsx-newline.js
```diff
@@ -45,11 +50,20 @@ module.exports = {
   create(context) {
     const jsxElementParents = new Set();
     const sourceCode = context.getSourceCode();
+
+    function isMultilined(node) {
+      return (node.loc.start.line !== node.loc.end.line);
```
ljharb, line 55:
```suggestion
      return node.loc.start.line !== node.loc.end.line;
```

lib/rules/jsx-newline.js
```diff
@@ -45,11 +50,20 @@ module.exports = {
   create(context) {
     const jsxElementParents = new Set();
     const sourceCode = context.getSourceCode();
+
+    function isMultilined(node) {
```
ljharb, line 54: i think this can be hoisted up to module level?

lib/rules/jsx-newline.js
```diff
@@ -45,11 +50,20 @@ module.exports = {
...
+    }
+
     return {
       'Program:exit'() {
         jsxElementParents.forEach((parent) => {
           parent.children.forEach((element, index, elements) => {
             if (element.type === 'JSXElement' || element.type === 'JSXExpressionContainer') {
+              const configuration = context.options[0] || {};
+              const prevent = configuration.prevent || false;
+              const allowMultilines = (configuration.allowMultilines || false) && prevent;
```
ljharb, line 65: does it make sense to have prevent false, and allowMultilines true? if not, can the schema forbid that?

ljharb (commented):
Looks great overall!

ljharb (approved):
Looks great!

PR #3315, opened 2022-06-29
Improved readability of the checkNodesIndent function within the jsx-…

caroline223:
…indent-props.js file

Within the checkNodesIndent function, I attempted to improve readability by indenting the boolean conditions.

### DIFF
--- a/lib/rules/jsx-indent-props.js
+++ b/lib/rules/jsx-indent-props.js
@@ -172,17 +172,23 @@ module.exports = {
      * @param {Number} indent needed indent
      */
     function checkNodesIndent(nodes, indent) {
+      let nestedIndent = indent;
       nodes.forEach((node) => {
         const nodeIndent = getNodeIndent(node);
-        if (line.isUsingOperator && !line.currentOperator && indentSize !== 'first' && !ignoreTernaryOperator) {
-          indent += indentSize;
+        if (
+          line.isUsingOperator
+          && !line.currentOperator
+          && indentSize !== 'first'
+          && !ignoreTernaryOperator
+        ) {
+          nestedIndent += indentSize;
           line.isUsingOperator = false;
         }
         if (
           node.type !== 'ArrayExpression' && node.type !== 'ObjectExpression'
-          && nodeIndent !== indent && astUtil.isNodeFirstInLine(context, node)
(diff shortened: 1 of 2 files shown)


lib/rules/jsx-indent-props.js
```diff
@@ -174,7 +174,8 @@ module.exports = {
     function checkNodesIndent(nodes, indent) {
       nodes.forEach((node) => {
         const nodeIndent = getNodeIndent(node);
-        if (line.isUsingOperator && !line.currentOperator && indentSize !== 'first' && !ignoreTernaryOperator) {
+        if (
+          line.isUsingOperator && !line.currentOperator && indentSize !== 'first' && !ignoreTernaryOperator) {
```
ljharb, line 178: i don't think this is more readable as-is. 

Generally, if the entire if condition can't fit on one line, then each condition needs to be on its own line:
```suggestion
        if (
          line.isUsingOperator
          && !line.currentOperator
          && indentSize !== 'first'
          && !ignoreTernaryOperator
        ) {
```

lib/rules/jsx-indent-props.js
```diff
@@ -173,18 +173,21 @@ module.exports = {
...
+        let nestedIndent = indent;
         const nodeIndent = getNodeIndent(node);
         if (
-          line.isUsingOperator && !line.currentOperator && indentSize !== 'first' && !ignoreTernaryOperator) {
-          // eslint-disable-next-line no-param-reassign
-          indent += indentSize;
+          line.isUsingOperator
+          && !line.currentOperator
+          && indentSize !== 'first'
+          && !ignoreTernaryOperator) {
```
ljharb, line 182:
```suggestion
          && !ignoreTernaryOperator
        ) {
```

lib/rules/jsx-indent-props.js
```diff
@@ -173,16 +173,22 @@ module.exports = {
      */
     function checkNodesIndent(nodes, indent) {
       nodes.forEach((node) => {
+        let nestedIndent = indent;
```
ljharb, line 176:
```suggestion
      let nestedIndent = indent;
      nodes.forEach((node) => {
```
otherwise it isn't persisted across the forEach

PR #3316, opened 2022-06-29
[Refactor] `boolean-prop-naming`, `jsx-indent`: avoid assigning to arguments

caroline223:


lib/rules/jsx-indent-props.js
```diff
@@ -172,8 +172,8 @@ module.exports = {
      * @param {Number} indent needed indent
      */
     function checkNodesIndent(nodes, indent) {
-      let nestedIndent = indent;
       nodes.forEach((node) => {
+        let nestedIndent = indent;
```
ljharb, line 176: this should be reverted; the variable needs to persist across loop iterations

lib/rules/boolean-prop-naming.js
```diff
@@ -172,9 +172,10 @@ module.exports = {
      * @param {Function} addInvalidProp callback to run for each error
      */
     function runCheck(proptypes, addInvalidProp) {
-      proptypes = proptypes || [];
+      let newProptypes = proptypes;
+      newProptypes = newProptypes || [];
 
-      proptypes.forEach((prop) => {
+      newProptypes.forEach((prop) => {
```
ljharb, line 178:
```suggestion
      (proptypes || []).forEach((prop) => {
```

lib/rules/boolean-prop-naming.js
```diff
@@ -172,9 +172,10 @@ module.exports = {
      * @param {Function} addInvalidProp callback to run for each error
      */
     function runCheck(proptypes, addInvalidProp) {
-      proptypes = proptypes || [];
+      let newProptypes = proptypes;
+      newProptypes = newProptypes || [];
```
ljharb, line 176:
```suggestion
```

lib/rules/jsx-indent.js
```diff
@@ -168,18 +168,20 @@ module.exports = {
      * @return {Number} Indent
      */
     function getNodeIndent(node, byLastLine, excludeCommas) {
-      byLastLine = byLastLine || false;
-      excludeCommas = excludeCommas || false;
+      let finalLine = byLastLine;
+      let noCommas = excludeCommas;
+      finalLine = finalLine || false;
+      noCommas = noCommas || false;
```
ljharb, line 174:
```suggestion
      const finalLine = !!byLastLine;
      const noCommas = !!excludeCommas;
```
altho i'm not sure either of these is needed at all, since both variables are used in a truthiness check.

ljharb (changes requested):

lib/rules/boolean-prop-naming.js
caroline223: When I used this solution, an error arose saying that newProptypes is assigned a value but never used. Instead, I wrote it as (newProptypes || []).forEach((prop) => {

lib/rules/jsx-indent-props.js
caroline223: Done.

lib/rules/jsx-indent.js
caroline223: Therefore, should I make the changes that you suggested or leave as is?

lib/rules/boolean-prop-naming.js
ljharb: Right - i don’t think we need a variable at all.

lib/rules/jsx-indent.js
ljharb: I think maybe just remove the variables entirely

lib/rules/boolean-prop-naming.js
caroline223: I'm a bit unclear as to why I should remove this?

ljharb: The argument (`proptypes`) is only actually needed once - in the forEach - and the `|| []` logic is basically the same conceptually as "only try to iterate the array if the argument is truthy".

So, one alternative way to write it is:
```js
(argument || []).forEach…
```
and another would be:
```js
if (argument) {
  argument.forEach…
}
```

and in neither case do we need to a) reassign to an argument, or b) create a variable at all (since it's only used once). Either case reads simpler and cleaner to me than what's in master.

caroline223: Done.

lib/rules/boolean-prop-naming.js
caroline223: @ljharb I have made the suggested changes if you wish to take a look at it.

lib/rules/jsx-indent.js
```diff
@@ -168,18 +168,18 @@ module.exports = {
...
     function getNodeIndent(node, byLastLine, excludeCommas) {
-      byLastLine = byLastLine || false;
-      excludeCommas = excludeCommas || false;
+      const finalLine = !!byLastLine;
+      const noCommas = !!excludeCommas;
 
       let src = context.getSourceCode().getText(node, node.loc.start.column + extraColumnStart);
       const lines = src.split('\n');
-      if (byLastLine) {
+      if (finalLine) {
```
ljharb, line 176:
```suggestion
      if (byLastLine) {
```

lib/rules/jsx-indent.js
```diff
@@ -168,18 +168,18 @@ module.exports = {
...
         src = lines[lines.length - 1];
       } else {
         src = lines[0];
       }
 
-      const skip = excludeCommas ? ',' : '';
+      const skip = noCommas ? ',' : '';
```
ljharb, line 182:
```suggestion
      const skip = excludeCommas ? ',' : '';
```

lib/rules/jsx-indent.js
```diff
@@ -168,18 +168,18 @@ module.exports = {
      * @return {Number} Indent
      */
     function getNodeIndent(node, byLastLine, excludeCommas) {
-      byLastLine = byLastLine || false;
-      excludeCommas = excludeCommas || false;
+      const finalLine = !!byLastLine;
+      const noCommas = !!excludeCommas;
```
ljharb, line 172:
```suggestion
```

caroline223: @ljharb Done

lib/rules/jsx-indent.js
caroline223: @ljharb Done

lib/rules/jsx-indent.js
caroline223: @ljharb Done

ljharb (approved):

PR #3317, opened 2022-07-01
[Fix] `jsx-no-literals`: properly error on children with noAttributeStrings: true

### LINKED ISSUE #3171: [`jsx-no-literals`]: no errors for children when `noAttributeStrings: true`
```js
'react/jsx-no-literals': [2, { noAttributeStrings: true }],
```

```js
import React from 'react';

export const WithAttributes = ({}) => <div title="foo bar" />;
export const WithChildren = ({}) => <div>baz bob</div>;
export const WithAttributesAndChildren = ({}) => (
  <div title="foo bar">baz bob</div>
);

```

Expected: the JSX inside of `WithChildren` produces an error (like it does when `noAttributeStrings: false`)

Actual: the JSX inside of `WithChildren` does not produce an error

TildaDares:
Fixes #3171

### DIFF
--- a/lib/rules/jsx-no-literals.js
+++ b/lib/rules/jsx-no-literals.js
@@ -70,24 +70,14 @@ module.exports = {
     config.allowedStrings = new Set(config.allowedStrings.map(trimIfString));
 
     function defaultMessageId() {
-      if (config.noAttributeStrings) {
+      const ancestorIsJSXElement = arguments.length >= 1 && arguments[0];
+      if (config.noAttributeStrings && !ancestorIsJSXElement) {
         return 'noStringsInAttributes';
       }
       if (config.noStrings) {
         return 'noStringsInJSX';
       }
       return 'literalNotInJSXExpression';
-    }
-
-    function reportLiteralNode(node, messageId) {
-      messageId = messageId || defaultMessageId();
-
-      report(context, messages[messageId], messageId, {
-        node,
-        data: {
-          text: context.getSourceCode().getText(node).trim(),
-        },
-      });
     }
 
     function getParentIgnoringBinaryExpressions(node) {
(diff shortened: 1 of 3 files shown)


lib/rules/jsx-no-literals.js
```diff
@@ -69,8 +69,8 @@ module.exports = {
     const config = Object.assign({}, defaults, context.options[0] || {});
     config.allowedStrings = new Set(config.allowedStrings.map(trimIfString));
 
-    function defaultMessageId() {
-      if (config.noAttributeStrings) {
+    function defaultMessageId(ancestorIsJSXElement = false) {
```
ljharb, line 72: default arguments don't work in node 4
```suggestion
    function defaultMessageId() {
      const ancestorIsJSXElement = arguments.length > 1 && arguments[0];
```

tests/lib/rules/jsx-no-literals.js
```diff
@@ -613,5 +613,41 @@ ruleTester.run('jsx-no-literals', rule, {
...
+        {
+          messageId: 'noStringsInAttributes',
+          data: { text: '"foo bar"' },
+        },
+      ],
+    },
+    {
+      code: `export const WithAttributesAndChildren = ({}) => (
+              <div title="foo bar">baz bob</div>
+            );`,
```
ljharb, line 639:
```suggestion
      code: `
        export const WithAttributesAndChildren = ({}) => (
          <div title="foo bar">baz bob</div>
        );
      `,
```

ljharb (changes requested):
Looks good!

ljharb:
Looks like some of the new test cases are expecting the wrong error message?

ljharb (approved):

PR #3318, opened 2022-07-02
Enforce JSX inline conditional as a ternary

### LINKED ISSUE #1979: prohibit `&&` conditional rendering in the return statement of a component
```jsx
// avoid
const WrapTingy = ({ cond }) => (
  cond && <Tingy/>
);
 
// that would error if cond is `undefined`, use this version instead:
 
const WrapTingy = ({ cond }) => (
  <>{cond && <Tingy/>}</>
);
```

holic:
This fixes a fairly common bug where React will render the falsy left side of a conditional with `&&` operator, resulting in things like `0` showing up in your UI. See https://github.com/jsx-eslint/eslint-plugin-react/pull/2888#issuecomment-751413618.

Fixes #1979
Closes #2888 

cc @ljharb

### DIFF
--- a/index.js
+++ b/index.js
@@ -30,6 +30,7 @@ const allRules = {
   'jsx-handler-names': require('./lib/rules/jsx-handler-names'),
   'jsx-indent': require('./lib/rules/jsx-indent'),
   'jsx-indent-props': require('./lib/rules/jsx-indent-props'),
+  'jsx-inline-conditional': require('./lib/rules/jsx-inline-conditional'),
   'jsx-key': require('./lib/rules/jsx-key'),
   'jsx-max-depth': require('./lib/rules/jsx-max-depth'),
   'jsx-max-props-per-line': require('./lib/rules/jsx-max-props-per-line'),
@@ -124,9 +125,7 @@ module.exports = {
   rules: allRules,
   configs: {
     recommended: {
-      plugins: [
-        'react',
-      ],
+      plugins: ['react'],
       parserOptions: {
         ecmaFeatures: {
           jsx: true,
@@ -134,6 +133,7 @@ module.exports = {
       },
       rules: {
         'react/display-name': 2,
+        'react/jsx-inline-conditional': 2,
         'react/jsx-key': 2,
         'react/jsx-no-comment-textnodes': 2,
         'react/jsx-no-duplicate-props': 2,
@@ -158,9 +158,7 @@ module.exports = {
       },
     },
     all: {
-      plugins: [
-        'react',
-      ],
+      plugins: ['react'],
       parserOptions: {
         ecmaFeatures: {
           jsx: true,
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -124,16 +125,15 @@ module.exports = {
...
-      ],
+      plugins: ['react'],
       parserOptions: {
         ecmaFeatures: {
           jsx: true,
         },
       },
       rules: {
         'react/display-name': 2,
+        'react/jsx-inline-conditional': 2,
```
holic, line 136: Because this is helps identify+fix a certain class of rendering bugs and isn't strictly stylistic, I enabled this by default. Let me know if I should change this!

ljharb: Adding a rule to the recommended config is a breaking change, so we'll basically never be doing that. Please remove it.

index.js
```diff
@@ -158,9 +158,7 @@ module.exports = {
       },
     },
     all: {
-      plugins: [
-        'react',
-      ],
+      plugins: ['react'],
```
ljharb, line 161: please revert unrelated formatting changes; if perhaps you're running prettier on repos that do not use it, please don't do that either :-)

tests/lib/rules/jsx-inline-conditional.js
```diff
@@ -0,0 +1,76 @@
...
+      },
+    },
+    {
+      code: '<div>{possiblyNull ?? <SomeComponent />}</div>',
+      parser: parsers.TYPESCRIPT_ESLINT,
+    },
+    {
+      code: '<div>{possiblyNull ?? <SomeComponent />}</div>',
+      parser: parsers['@TYPESCRIPT_ESLINT'],
+    },
```
ljharb, line 53: these should not be needed, since the parsers.all helper should automatically run every test case in every parser.
```suggestion
```

tests/lib/rules/jsx-inline-conditional.js
```diff
@@ -0,0 +1,76 @@
...
+      code: '<div>{someCondition ? <div>{anotherCondition ? <SomeComponent /> : null}</div> : null}</div>',
+    },
+    {
+      code: '<div>{someCondition && someOtherCondition ? <SomeComponent /> : null}</div>',
+    },
+    {
+      code: '<div>{possiblyNull ?? <SomeComponent />}</div>',
+      parserOptions: {
+        ecmaVersion: 2020,
+      },
```
ljharb, line 44:
```suggestion
      code: '<div>{possiblyNull ?? <SomeComponent />}</div>',
      features: ['nullish coalescing'],
```
and then we'll need to add that feature to the `parsers.all` helper

ljharb (changes requested):
Doesn't this duplicate the recently added `jsx-no-leaked-render` rule?

ljharb:
I think #3203 may have covered this?

PR #3319, opened 2022-07-04
[Docs] `jsx-no-target-blank`: Fix link to link-type-noreferrer

Luccasoli:
I updated the link to the "noreferrer" explanation instead of a link to the "noopener" explanation, since the paragraph where this link is located only talks about the "noreferrer".


ljharb:
What's wrong with https://html.spec.whatwg.org/multipage/links.html#link-type-noopener ?

Luccasoli:
Hi @ljharb, what's up? 

Sorry for the PR without context..., I suggested this change because this paragraph I edited talks about the "noreferrer" configuration instead of "noopener", so I thought this link would be more appropriate, what do you think?

ljharb:
Let's include both links.

ljharb (approved):

PR #3320, opened 2022-07-04
[Fix] `jsx-key`: catch key errors inside conditional statements

### LINKED ISSUE #3117: Eslint does not work if the jsx key is missed inconditions
Eslint does not work if the jsx key is missed in conditions

Screenshot:
image

Code:
```tsx
const Test = (): React.ReactNode => {
  const list = [1, 2, 3, 4, 5];

  return (
    <div>
      {list.map(item => {
        if (item < 2) {
          /* The eslint error didn't work: */
          return <div>{item}</div>;
        }

        return <div />;
      })}
    </div>
  );
};

```

TildaDares:
Fixes #3117

### DIFF
--- a/lib/rules/jsx-key.js
+++ b/lib/rules/jsx-key.js
@@ -86,8 +86,28 @@ module.exports = {
       }
     }
 
-    function getReturnStatement(body) {
-      return body.filter((item) => item.type === 'ReturnStatement')[0];
+    function getReturnStatements(node) {
+      const returnStatements = arguments[1] || [];
+      if (node.type === 'IfStatement') {
+        if (node.consequent) {
+          getReturnStatements(node.consequent, returnStatements);
+        }
+        if (node.alternate) {
+          getReturnStatements(node.alternate, returnStatements);
+        }
+      } else {
+        node.body.forEach((item) => {
+          if (item.type === 'IfStatement') {
+            getReturnStatements(item, returnStatements);
+          }
+
+          if (item.type === 'ReturnStatement') {
+            returnStatements.push(item);
+          }
+        });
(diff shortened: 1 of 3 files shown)


lib/rules/jsx-key.js
```diff
@@ -86,8 +86,25 @@ module.exports = {
...
-    function getReturnStatement(body) {
-      return body.filter((item) => item.type === 'ReturnStatement')[0];
+    function getReturnStatements(node) {
+      const returnStatements = arguments[1] || [];
+      if (node.type === 'IfStatement') {
+        if (node.consequent) getReturnStatements(node.consequent, returnStatements);
+        if (node.alternate) getReturnStatements(node.alternate, returnStatements);
+      } else {
+        const body = node.body;
+        body.forEach((item) => {
```
ljharb, line 96:
```suggestion
        node.body.forEach((item) => {
```

lib/rules/jsx-key.js
```diff
@@ -189,10 +206,12 @@ module.exports = {
 
         if (isFn || isArrFn) {
           if (fn.body.type === 'BlockStatement') {
-            const returnStatement = getReturnStatement(fn.body.body);
-            if (returnStatement && returnStatement.argument) {
-              checkIteratorElement(returnStatement.argument);
-            }
+            const returnStatements = getReturnStatements(fn.body);
+            returnStatements.forEach((returnStatement) => {
```
ljharb, line 210:
```suggestion
            getReturnStatements(fn.body).forEach((returnStatement) => {
```

ljharb (approved):
Looks great!

PR #3321, opened 2022-07-05
[Fix] `display-name`: Accept forwardRef and Memo nesting in newer React versions

### LINKED ISSUE #2954: display-name has false positive on nested use of memo and forwardRef
Since #2399, `eslint-plugin-react` knows not to complain about code like this:

```tsx
export const ComponentWithMemo = React.memo(function Component({ world }) {
 return <div>Hello {world}</div>
})

export const ComponentWithForwardRef = React.forwardRef(function Component({ world }) {
  return <div>Hello {world}</div>
})
```

In that pull request, nested use of `memo` and `forwardRef` like so...

```tsx
export const ComponentWithForwardRef = React.memo(React.forwardRef(function Component({ world }) {
  return <div>Hello {world}</div>
}))
```

...was intentionally treated as a `display-name` violation due to [a React bug causing such components to be anonymous](https://github.com/facebook/react/issues/16722). This issue was [solved in November 2019](https://github.com/facebook/react/pull/17274), so the `display-name` should now accept this.

TildaDares:
Fixes #2954

### DIFF
--- a/lib/rules/display-name.js
+++ b/lib/rules/display-name.js
@@ -11,6 +11,7 @@ const Components = require('../util/Components');
 const astUtil = require('../util/ast');
 const componentUtil = require('../util/componentUtil');
 const docsUrl = require('../util/docsUrl');
+const testReactVersion = require('../util/version').testReactVersion;
 const propsUtil = require('../util/props');
 const report = require('../util/report');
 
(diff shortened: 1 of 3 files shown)


lib/rules/display-name.js
```diff
@@ -58,11 +59,27 @@ module.exports = {
...
     }
 
+    /**
+     * Checks if React.forwardRef is nested inside React.memo
+     * @param {ASTNode} node The AST node being checked.
+     * @returns {Boolean} True if React.forwardRef is nested inside React.memo, false if not.
+     */
+    function isNestedMemo(node) {
+      return node.type === 'CallExpression'
+        && node.arguments[0].type === 'CallExpression'
```
ljharb, line 69: what if it's a call expression with no arguments?

lib/rules/display-name.js
```diff
@@ -58,11 +59,27 @@ module.exports = {
...
+        && utils.isPragmaComponentWrapper(node);
+    }
+
     /**
      * Reports missing display name for a given component
      * @param {Object} component The component to process
      */
     function reportMissingDisplayName(component) {
+      if (isNestedMemo(component.node)
+        && testReactVersion(context, '0.14.10 || 15.7.0 || >= 16.12.0')) {
```
ljharb, line 79:
```suggestion
      if (
        isNestedMemo(component.node)
        && testReactVersion(context, '^0.14.10 || ^15.7.0 || >= 16.12.0')
      ) {
```

ljharb (changes requested):
Let's also add some test cases for v0.14.11, v15.7.1, and v16.12.1

lib/rules/display-name.js
```diff
@@ -65,8 +65,11 @@ module.exports = {
      * @returns {Boolean} True if React.forwardRef is nested inside React.memo, false if not.
      */
     function isNestedMemo(node) {
+      const argumentIsCallExpression = node.arguments && node.arguments[0]
+        ? node.arguments[0].type === 'CallExpression' : false;
```
ljharb, line 69:
```suggestion
      const argumentIsCallExpression = node.arguments && node.arguments[0] && node.arguments[0].type === 'CallExpression';
```

ljharb (approved):

PR #3326, opened 2022-07-08
[Fix] `forbid-prop-types`: Ignore objects that are not of type React.PropTypes

### LINKED ISSUE #2722: forbid-prop-types warning shown on Yup validation
Hey there, a project I work on uses both prop-types and [Yup](https://github.com/jquense/yup) which have a similar syntax.

The following where I'm declaring a Yup validation object:
```js
const formValidation = Yup.object().shape({
    name: Yup.string(),
    customer_ids: Yup.array()
});
```

Gives me a warning on the line `customer_ids: Yup.array()` like so:
```
Prop type `array` is forbidden eslint(react/forbid-prop-types)
```

I believe this is a bug as this is not a prop type declaration.

TildaDares:
Fixes #2722

### DIFF
--- a/lib/rules/forbid-prop-types.js
+++ b/lib/rules/forbid-prop-types.js
@@ -60,6 +60,27 @@ module.exports = {
     const configuration = context.options[0] || {};
     const checkContextTypes = configuration.checkContextTypes || false;
     const checkChildContextTypes = configuration.checkChildContextTypes || false;
+    let propTypesPackageName = null;
+    let reactPackageName = null;
+    let isForeignPropTypesPackage = false;
+
+    function isPropTypesPackage(node) {
+      return (
+        node.type === 'Identifier'
+        && (
+          node.name === null
+          || node.name === propTypesPackageName
+          || !isForeignPropTypesPackage
+        )
+      ) || (
+        node.type === 'MemberExpression'
+        && (
+          node.object.name === null
+          || node.object.name === reactPackageName
+          || !isForeignPropTypesPackage
+        )
+      );
+    }
 
     function isForbidden(type) {
       const forbid = configuration.forbid || DEFAULTS;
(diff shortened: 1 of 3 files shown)


lib/rules/forbid-prop-types.js
```diff
@@ -181,6 +181,14 @@ module.exports = {
       },
 
       CallExpression(node) {
+        if (
+          node.callee.object
+          && !propsUtil.isPropTypesDeclaration(node.callee)
+          && node.callee.object.name !== 'PropTypes'
```
ljharb, line 187: can we ensure this is just `React.PropTypes`, or that `PropTypes` is imported from `prop-types` or `React`? otherwise if someone names anything `PropTypes` it'd bypass this rule.

ljharb (commented):
Looks great! Just one question.

lib/rules/forbid-prop-types.js
```diff
@@ -60,6 +60,19 @@ module.exports = {
...
+    let reactPackageName = null;
+
+    function isPropTypesPackage(node) {
+      return (
+        node.type === 'Identifier'
+        && node.name === propTypesPackageName
+      ) || (
+        node.type === 'MemberExpression'
+        && node.property.name === 'PropTypes'
+        && node.object.name === reactPackageName
```
ljharb, line 73:
```suggestion
        && (node.object.name === null || node.object.name === reactPackageName)
```

lib/rules/forbid-prop-types.js
ljharb: Let's add two test cases - one that imports react as something that's not React, and one that imports prop-types as something that's not PropTypes - and that might allow us to remove this line?

ljharb: The first one has `import { any } from "prop-types"; `?

ljharb: ohh hm, right. in that case, i think we should be setting the propTypesPackageName to the empty string maybe? or, maybe we need to track a list of all the named imports from `prop-types` as well?

ljharb: i'm not really sure.

Basically, the common pattern is `import PropTypes from 'prop-types'; … PropTypes.something`, or `import React from 'react'; … React.PropTypes.something`. However, if someone does `import { something } from 'prop-types'`, then `something` should still be able to be forbidden.

ljharb:
@TildaDares i've rebased this, and added two "invalid" test cases, which seem to be failing.

ljharb (approved):
Looks great!

ljharb:
I added a missing test case, and removed the "import declaration node with no specifiers" - both because i believe it can't ever be hit, and also because it won't crash in that case anyways. This should be ready to merge once tests pass.

Issue #2962, opened 2021-04-09
False negative when target and rel attributes are dynamic


just-boris:
I have this code example

```js
<a
    href={href}
    target={isExternal ? '_blank' : undefined}
    rel={isExternal ? 'noopener noreferrer' : undefined}
/>
```

Both target and rel attribute are dynamically defined, and always both present when needed. However, the rule reports it as an error:

```
Using target="_blank" without rel="noreferrer" is a security risk
``` 

Plugin version: 7.23.2

ljharb:
It’s not possible to reliably handle dynamic cases like this statically. I agree this specific usage is safe.

Either use an override comment, or, conditionally render one of two anchor tags, both statically defined.

just-boris:
Yeah, this is exactly what I did, added `eslint-disable-next-line` with a link to this issue.

I will keep this open and let you decide if you want to address this limitation anyhow. Perhaps, the rule could be more conservative and warn only when it is super confident there is an issue

ljharb:
Conservative means whatever’s safest; warning here is safest. That said, false positives are harmful.

If this case can be handled, and it’s generalizable beyond this exact case, I’m happy to accept a PR.

V2dha:
For now, `target` attribute can be assigned using ternary operator provided one of the consequent or alternate is assigned as `_blank`. Similar, thing can be added for the `rel` attribute from which the value can be extracted from either of the consequent or alternate which contains `Literal` type of value. What do you think @ljharb?

ljharb:
I think that makes sense - any place we currently require a static string should also allow a ternary where both options are an acceptable static string.

PR #3332, opened 2022-07-13
[Fix] False negative when rel attribute assigned using ternary

V2dha:
Fixes [#2962](https://github.com/jsx-eslint/eslint-plugin-react/issues/2962)

### DIFF
--- a/lib/rules/jsx-no-target-blank.js
+++ b/lib/rules/jsx-no-target-blank.js
@@ -74,7 +74,12 @@ function getStringFromValue(value) {
       if (value.expression.type === 'TemplateLiteral') {
         return value.expression.quasis[0].value.cooked;
       }
-      return value.expression && value.expression.value;
+      const expr = value.expression;
+      return expr && (
+        expr.type === 'ConditionalExpression'
+          ? [expr.consequent.value, expr.alternate.value]
+          : expr.value
+      );
     }
   }
   return null;
@@ -88,12 +93,15 @@ function hasSecureRel(node, allowReferrer, warnOnSpreadAttributes, spreadAttribu
 
   const relAttribute = node.attributes[relIndex];
   const value = getStringFromValue(relAttribute.value);
-  const tags = value && typeof value === 'string' && value.toLowerCase().split(' ');
-  const noreferrer = tags && tags.indexOf('noreferrer') >= 0;
-  if (noreferrer) {
-    return true;
-  }
-  return allowReferrer && tags && tags.indexOf('noopener') >= 0;
+  return [].concat(value).every((item) => {
+    const tags = typeof item === 'string' ? item.toLowerCase().split(' ') : false;
+    const noreferrer = tags && tags.indexOf('noreferrer') >= 0;
+    if (noreferrer) {
+      return true;
+    }
+    const noopener = tags && tags.indexOf('noopener') >= 0;
+    return allowReferrer && noopener;
+  });
 }
 
 const messages = {
--- a/tests/lib/rules/jsx-no-target-blank.js
+++ b/tests/lib/rules/jsx-no-target-blank.js
@@ -28,6 +28,7 @@ const parserOptions = {
 
 const ruleTester = new RuleTester({ parserOptions });
 const defaultErrors = [{ messageId: 'noTargetBlankWithoutNoreferrer' }];
+const allowReferrerErrors = [{ messageId: 'noTargetBlankWithoutNoopener' }];
 
 ruleTester.run('jsx-no-target-blank', rule, {
   valid: parsers.all([
@@ -141,6 +142,19 @@ ruleTester.run('jsx-no-target-blank', rule, {
     {
       code: '<a href target="_blank"/>',
     },
+    {
+      code: '<a href={href} target={isExternal ? "_blank" : undefined} rel="noopener noreferrer" />',
+    },
+    {
+      code: '<a href={href} target={isExternal ? undefined : "_blank"} rel={isExternal ? "noreferrer" : "noopener noreferrer"} />',
+    },
+    {
+      code: '<a href={href} target={isExternal ? undefined : "_blank"} rel={isExternal ? "noreferrer noopener" : "noreferrer"} />',
+    },
+    {
+      code: '<a href={href} target="_blank" rel={isExternal ? "noreferrer" : "noopener"} />',
+      options: [{ allowReferrer: true }],
+    },
   ]),
   invalid: parsers.all([
     {
@@ -251,13 +265,13 @@ ruleTester.run('jsx-no-target-blank', rule, {
       code: '<a href="https://example.com/20" target="_blank" rel></a>',
       output: '<a href="https://example.com/20" target="_blank" rel="noopener"></a>',
       options: [{ allowReferrer: true }],
-      errors: [{ messageId: 'noTargetBlankWithoutNoopener' }],
+      errors: allowReferrerErrors,
     },
     {
       code: '<a href="https://example.com/20" target="_blank"></a>',
       output: '<a href="https://example.com/20" target="_blank" rel="noopener"></a>',
       options: [{ allowReferrer: true }],
-      errors: [{ messageId: 'noTargetBlankWithoutNoopener' }],
+      errors: allowReferrerErrors,
     },
     {
       code: '<a target="_blank" href={ dynamicLink }></a>',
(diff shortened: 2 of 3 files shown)


lib/rules/jsx-no-target-blank.js
```diff
@@ -74,6 +74,19 @@ function getStringFromValue(value) {
       if (value.expression.type === 'TemplateLiteral') {
         return value.expression.quasis[0].value.cooked;
       }
+      const expr = value.expression;
+      if (expr.type === 'ConditionalExpression') {
+        if (expr.consequent.type === 'Literal'
+        && ((expr.consequent.value && expr.consequent.value.toLowerCase() === 'noreferrer')
+        || (expr.consequent.value && expr.consequent.value.toLowerCase() === 'noopener noreferrer'))) {
```
ljharb, line 81:
```suggestion
        if (
          expr.consequent.type === 'Literal'
          && (
            (expr.consequent.value && expr.consequent.value.toLowerCase() === 'noreferrer')
            || (expr.consequent.value && expr.consequent.value.toLowerCase() === 'noopener noreferrer')
          )
        ) {
```

also are you sure the consequent value will always be a string?

also, what if the value is `noreferrer noopener`?

V2dha: For `noreferrer noopener`, I can add another case. Also if the consequent's value is a string it should match one of the three cases.

ljharb: right, but if it's not a string, `.toLowerCase()` will crash

V2dha: that's why both the conditions should be true similar to in the function `attributeValuePossiblyBlank(attribute)`

ljharb: what if they're a number, or an object or array?

it's a completely acceptable answer to say "all the parsers we know about always make this a string", and maybe that was the justification for the other function - but it seems like it might be safer to have `typeof expr.consequent.value === 'string' && `?

V2dha: That makes sense! I'll update it

lib/rules/jsx-no-target-blank.js
```diff
@@ -74,6 +74,29 @@ function getStringFromValue(value) {
...
       }
+      const expr = value.expression;
+      if (expr.type === 'ConditionalExpression') {
+        if (
+          typeof expr.consequent.value === 'string'
+          && (
+            (expr.consequent.value && expr.consequent.value.toLowerCase() === 'noreferrer')
+            || (expr.consequent.value && expr.consequent.value.toLowerCase() === 'noopener noreferrer')
+            || (expr.consequent.value && expr.consequent.value.toLowerCase() === 'noreferrer noopener')
+          )
```
ljharb, line 85: seems worth putting this logic into a helper function, so it can be used in both places?

ljharb (approved):
Looks great!

lib/rules/jsx-no-target-blank.js
```diff
@@ -74,6 +74,29 @@ function getStringFromValue(value) {
...
+        ) {
+          return expr.consequent.value;
+        }
+        if (
+          typeof expr.alternate.value === 'string'
+          && (
+            (expr.alternate.value && expr.alternate.value.toLowerCase() === 'noreferrer')
+            || (expr.alternate.value && expr.alternate.value.toLowerCase() === 'noopener noreferrer')
+            || (expr.consequent.value && expr.consequent.value.toLowerCase() === 'noreferrer noopener')
```
ljharb, line 94: should this be checking the alternate, instead of the consequent?

lib/rules/jsx-no-target-blank.js
```diff
@@ -74,6 +74,29 @@ function getStringFromValue(value) {
       if (value.expression.type === 'TemplateLiteral') {
         return value.expression.quasis[0].value.cooked;
       }
+      const expr = value.expression;
+      if (expr.type === 'ConditionalExpression') {
+        if (
+          typeof expr.consequent.value === 'string'
```
ljharb, line 80: it kind of seems here like if the consequent is valid, that the alternate can be invalid?

can we add some tests like `condition ? 3 : 'noopener noreferrer'` and `condition ? 'noopener noreferrer' : '3'` etc?

tests/lib/rules/jsx-no-target-blank.js
```diff
@@ -141,6 +141,21 @@ ruleTester.run('jsx-no-target-blank', rule, {
...
+    },
+    {
+      code: '<a href={href} target={isExternal ? "_blank" : undefined} rel={isExternal ? undefined : "noopener noreferrer"} />',
+    },
+    {
+      code: '<a href={href} target={isExternal ? undefined : "_blank"} rel={isExternal ? "noreferrer" : "noopener noreferrer"} />',
+    },
+    {
+      code: '<a href={href} target="_blank" rel={isExternal ? "noopener" : "noopener noreferrer"} />',
+    },
```
ljharb, line 158: these are valid because both branches are valid.

tests/lib/rules/jsx-no-target-blank.js
```diff
@@ -141,6 +141,21 @@ ruleTester.run('jsx-no-target-blank', rule, {
...
       code: '<a href target="_blank"/>',
     },
+    {
+      code: '<a href={href} target={isExternal ? "_blank" : undefined} rel="noopener noreferrer" />',
+    },
+    {
+      code: '<a href={href} target={isExternal ? "_blank" : undefined} rel={isExternal ? "noopener noreferrer" : undefined} />',
+    },
+    {
+      code: '<a href={href} target={isExternal ? "_blank" : undefined} rel={isExternal ? undefined : "noopener noreferrer"} />',
```
ljharb, line 151: these can be valid all `undefined` branches can be ignored, and all other branches are valid.

ljharb (changes requested):

lib/rules/jsx-no-target-blank.js
V2dha: Yes, the condition first checks the consequent's value if it's valid then it returns that value, otherwise, it checks the alternate's value if it is valid then it returns that value similar to as in the function `attributeValuePossiblyBlank(attribute)`. 
I thought it was workable since both of the values were not going to be assigned.

V2dha: So should I change the condition for both the branches to be true and then return one of them since both would be true?

V2dha: How about instead of checking for different types of combinations that can be assigned to `rel`  we can just check if it's string type and return it to the `hasSecureRel` function since the other two conditions are doing the same and I think we have to return the string value since that's what the function is doing when called to get the `rel` attributes value. But we have to figure out how to return both the consequent and alternate values?

ljharb: Maybe it could return an array of strings - in every case? That way it'd be an array of 1 string for most nodes, but for a ternary it'd be an array of 2 strings.

lib/rules/jsx-no-target-blank.js
```diff
@@ -88,6 +110,17 @@ function hasSecureRel(node, allowReferrer, warnOnSpreadAttributes, spreadAttribu
...
+  if (Array.isArray(value)) {
+    for (let index = 0; index < value.length; index++) {
+      const tags = value[index] && typeof value[index] === 'string' && value[index].toLowerCase().split(' ');
+      const noreferrer = tags && tags.indexOf('noreferrer') >= 0;
+      const noopener = tags && tags.indexOf('noopener') >= 0;
+      if (!noreferrer) {
+        return allowReferrer && noopener;
+      }
+    }
+    return true;
```
ljharb, line 122:
```suggestion
    return [].concat(value).filter(Boolean).some((item) => {
      const tags = typeof item === 'string' && item.toLowerCase().split(' ');
      const noreferrer = tags && tags.indexOf('noreferrer') >= 0;
      const noopener = tags && tags.indexOf('noopener') >= 0;
      if (!noreferrer) {
        return allowReferrer && noopener;
      }
      return false;
    });
```
also the [].concat means you don't have to check if it's an array; you can just use this code for both the single and the multi case.

V2dha: instead of `some(item)`, I think having `every(item)` would be workable since every value should be true; But I don't why the test cases are failing because of that. If `some(item)` is kept, then as soon it would get first true value it won't check the rest and there would give no errors for the cases like `rel={isExternal ? "noreferrer" : "3"}`

ljharb (approved):

ljharb:
@V2dha i've rebased this, and there's still 48 test failures - all things that expect an error but now get none. I think the some/every change may be related.

tests/lib/rules/jsx-no-target-blank.js
```diff
@@ -390,14 +390,5 @@ ruleTester.run('jsx-no-target-blank', rule, {
       code: '<a href={href} target="_blank" rel={isExternal ? 3 : "noopener noreferrer"} />',
       errors: defaultErrors,
     },
-    {
-      code: '<a href={href} target="_blank" rel={isExternal ? "noopener noreferrer" : "3"} />',
```
ljharb: These cases should still be here and be invalid.

lib/rules/jsx-no-target-blank.js
ljharb: I agree with you that every is the proper choice

ljharb (approved):

PR #3338, opened 2022-07-15
[New] `forbid-dom-props`: add `disallowedFor` option

### LINKED ISSUE #2946: Rule option "disallowedFor" for component/dom element prop
Hi all 👋 

First off thank you so much for this plugin, it has helped me many times, and saved me countless hours of debugging

I would like to put forward an idea to extend the following two rules
- react/forbid-component-props
- react/forbid-dom-props

The former accepts an `allowedFor` option, and I would like to introduce its counterpart `dissalowedFor`.
It would be particularly interesting for the `react/forbid-dom-props` rule as the Web platform constantly evolves, and even though it is rare, they sometimes deprecate old attributes.

To give you an example, the `<form>` tag used to receive an `accept` argument, however it is now deprecated and that attribute should be set on an `<input type="file">`.

There are no ways at the moment in `eslint-plugin-react` to express that. I would be happy to implement that directly but wanted to create this issue first to gather feedback from the community and settled on the proper syntax, naming and behavioura

# Syntax

```
{
  "propName": "accept"
  "disallowedFor": [form],
  "message": "Avoid using the accept attribute on <form>"
}
```

~To ensure backwards compatibility I believe the `allowedFor` option should take precendance over the `disallowedFor` option in the case of the following example, only `allowedFor` would be applied and a warning would be added to the terminal to notify the developer that they have conflicting rules configuration~

Edit: as @ljharb pointed out, the following schema would not be valid as the two options would be mutually exclusive

```
{
  "propName": "accept"
  "disallowedFor": [form],
  "allowedFor": [input]
  "message": "Avoid using the accept attribute on <form>"
}
```

TildaDares:
Fixes #2946

### DIFF
--- a/docs/rules/forbid-dom-props.md
+++ b/docs/rules/forbid-dom-props.md
@@ -43,12 +43,13 @@ Examples of **correct** code for this rule:
 ### `forbid`
 
 An array of strings, with the names of props that are forbidden. The default value of this option `[]`.
-Each array element can either be a string with the property name or object specifying the property name and an optional
-custom message:
+Each array element can either be a string with the property name or object specifying the property name, an optional
+custom message, and a DOM nodes disallowed list (e.g. `<div />`):
 
 ```js
 {
   "propName": "someProp",
+  "disallowedFor": ["DOMNode", "AnotherDOMNode"],
   "message": "Avoid using someProp"
 }
 ```
--- a/lib/rules/forbid-dom-props.js
+++ b/lib/rules/forbid-dom-props.js
@@ -18,6 +18,21 @@ const DEFAULTS = [];
 // Rule Definition
 // ------------------------------------------------------------------------------
 
+/**
+ * @param {Map<string, object>} forbidMap // { disallowList: null | string[], message: null | string }
+ * @param {string} prop
+ * @param {string} tagName
+ * @returns {boolean}
+ */
+function isForbidden(forbidMap, prop, tagName) {
+  const options = forbidMap.get(prop);
+  return options && (
+    typeof tagName === 'undefined'
+    || !options.disallowList
+    || options.disallowList.indexOf(tagName) !== -1
+  );
+}
+
 const messages = {
   propIsForbidden: 'Prop "{{prop}}" is forbidden on DOM Nodes',
 };
@@ -47,6 +62,13 @@ module.exports = {
                 propName: {
                   type: 'string',
                 },
+                disallowedFor: {
+                  type: 'array',
+                  uniqueItems: true,
+                  items: {
+                    type: 'string',
+                  },
+                },
                 message: {
                   type: 'string',
                 },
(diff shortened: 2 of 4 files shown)


docs/rules/forbid-dom-props.md
```diff
@@ -43,12 +43,13 @@ Examples of **correct** code for this rule:
 ### `forbid`
 
 An array of strings, with the names of props that are forbidden. The default value of this option `[]`.
-Each array element can either be a string with the property name or object specifying the property name and an optional
-custom message:
+Each array element can either be a string with the property name or object specifying the property name, an optional
+custom message, and a DOM nodes blacklist(e.g. `<div />`):
```
ljharb, line 47:
```suggestion
custom message, and a DOM nodes disallowed list (e.g. `<div />`):
```

docs/rules/forbid-dom-props.md
```diff
@@ -43,12 +43,13 @@ Examples of **correct** code for this rule:
...
 
 ```js
 {
   "propName": "someProp",
+  "disallowedFor": [DOMNode, AnotherDOMNode],
```
ljharb, line 52:
```suggestion
  "disallowedFor": ["DOMNode", "AnotherDOMNode"],
```
this should work with non-JS configs :-)

lib/rules/forbid-dom-props.js
```diff
@@ -66,13 +73,20 @@ module.exports = {
...
-    function isForbidden(prop) {
-      return forbid.has(prop);
+    function isForbidden(prop, tagName) {
+      const options = forbid.get(prop);
+      const disallowList = options ? options.disallowList : undefined;
+      return typeof disallowList !== 'undefined'
+        && (typeof tagName === 'undefined'
+        || disallowList.length === 0
+        || (disallowList.length > 0
+        && disallowList.indexOf(tagName) !== -1));
```
ljharb, line 89:
```suggestion
      return typeof disallowList !== 'undefined' && (
        typeof tagName === 'undefined'
        || disallowList.indexOf(tagName) !== -1
      );
```
indexOf is free if the array's empty.

ljharb (changes requested):

lib/rules/forbid-dom-props.js
ljharb: ahh right, this is "is forbidden". what about this?
```suggestion
      const disallowList = options ? options.disallowList : [];
      return typeof tagName !== 'undefined' && disallowList.indexOf(tagName) !== -1;
```
we might not even need the tagName check

ljharb: ahhh sorry, i get it. When the disallow list is empty, then everything is forbidden.

ok - so let's try a different approach. (I don't think the tagName should ever be undefined?) Something like the following:
Currently, there's a `forbid` map of prop name to options. I think we should update that code so that instead of it having `options.disallowList`, it has an `options.isAllowed` function.

That function then would be:
```js
(tag) => options.disallowList && options.disallowList.length > 0 ? options.disallowList.indexOf(tag) === -1 : true
```

Would that work?

ljharb: It would contain a predicate function.

ljharb: Neither? Predicate functions are great. What i was thinking is that the confusing logic could be wrapped up in a function when creating the Map, so that the individual check here could be simpler.

ljharb (approved):

PR #3339, opened 2022-07-15
[Docs] `sort-comp`: add class component examples

maurer2:
Hello, I updated the examples for sort-comp, that were still using `createClass` with class component examples.

Cheers

### DIFF
--- a/docs/rules/sort-comp.md
+++ b/docs/rules/sort-comp.md
@@ -24,6 +24,15 @@ var Hello = createReactClass({
 });
 ```
 
+```jsx
+class Hello extends React.Component {
+  render() {
+    return <div>Hello</div>;
+  }
+  static displayName = 'Hello';
+}
+```
+
 Examples of **correct** code for this rule:
 
 ```jsx
@@ -33,6 +42,15 @@ var Hello = createReactClass({
     return <div>Hello</div>;
   }
 });
+```
+
+```jsx
+class Hello extends React.Component {
+  static displayName = 'Hello';
+  render() {
+    return <div>Hello</div>;
+  }
+}
 ```
 
 ## Rule Options
@@ -128,6 +146,16 @@ var Hello = createReactClass({
 });
 ```
 
+```jsx
+class Hello extends React.Component {
+  render() {
+    return <div>Hello</div>;
+  }
+  onClick = this.onClick.bind(this);
+  onClick() {}
+}
+```
+
 Examples of **correct** code for this rule, with the above configuration:
 
 ```jsx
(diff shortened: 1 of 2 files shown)


docs/rules/sort-comp.md
```diff
@@ -16,23 +16,23 @@ The default configuration ensures that the following order must be followed:
...
 
 Examples of **correct** code for this rule:
 
 ```jsx
-var Hello = createReactClass({
-  displayName : 'Hello',
-  render: function() {
+  static displayName = 'Hello';
+  render(){
```
ljharb, line 32:
```suggestion
  render () {
```

docs/rules/sort-comp.md
```diff
@@ -120,23 +120,23 @@ For example, if you want to place your event handlers (`onClick`, `onSubmit`, et
 Examples of **incorrect** code for this rule, with the above configuration:
 
 ```jsx
-var Hello = createReactClass({
-  render: function() {
+  render(){
```
ljharb, line 124:
```suggestion
  render () {
```

docs/rules/sort-comp.md
```diff
@@ -120,23 +120,23 @@ For example, if you want to place your event handlers (`onClick`, `onSubmit`, et
...
     return <div>Hello</div>;
-  },
-  onClick: function() {}
-});
+  }
+  onClick = () => {}
```
ljharb, line 127: functions should never be stored in class fields
```suggestion
  onClick = this.onClick.bind(this);
  onClick() {}
```

docs/rules/sort-comp.md
```diff
@@ -120,23 +120,23 @@ For example, if you want to place your event handlers (`onClick`, `onSubmit`, et
...
 ```
 
 Examples of **correct** code for this rule, with the above configuration:
 
 ```jsx
-var Hello = createReactClass({
-  onClick: function() {},
-  render: function() {
+  onClick = () => {}
```
ljharb, line 135: same here

ljharb (changes requested):
Rather than replacing these examples, let's add them so we have both

maurer2:
Okay, I restored the `createClass` examples and fixed the formatting issues.
Cheers

docs/rules/sort-comp.md
```diff
@@ -24,6 +24,15 @@ var Hello = createReactClass({
 });
 ```
 
+```jsx
+  render () {
```
ljharb, line 29:
```suggestion
  render() {
```
whoops, sorry, i think i spaced it wrong in my original suggestion. same throughout.

maurer2:
Implemented the onClick changes, but I think `onClick = this.onClick.bind(this);` needs to be in the constructor. Public class fields should also work as an alternative.

ljharb (approved):
Thanks!

PR #3347, opened 2022-07-27
[Fix] `display-name`, component detection: fix HOF returning null as Components

### LINKED ISSUE #3346: False positives for react/display-name in HOF returning null
Any of these constructs are incorrectly identified as Components:

```javascript
any = () => () => null
{ any: () => () => null }
any = function() {return function() {return null;};}
{ any: function() {return function() {return null;};} }
```

jxm-math:
Fixes #3346


ljharb (approved):
Thanks!

PR #3349, opened 2022-08-04
[Docs] `jsx-no-useless-fragment`: add more examples of correct code

karlhorky:
Hey @ljharb ! 👋

Just a quick PR to:

1. Add some simpler examples of how to resolve the `react/jsx-no-useless-fragment` problems when they appear (there are not as many simple examples of just removing the fragment when there's a single child)
2. Add an example with an embedded expression and ternary

### DIFF
--- a/docs/rules/jsx-no-useless-fragment.md
+++ b/docs/rules/jsx-no-useless-fragment.md
@@ -27,11 +27,17 @@ Examples of **incorrect** code for this rule:
     <div />
   </>
 </section>
+
+{showFullName ? <>{fullName}</> : <>{firstName}</>}
 ```
 
 Examples of **correct** code for this rule:
 
 ```jsx
+{foo}
+
+<Foo />
+
 <>
   <Foo />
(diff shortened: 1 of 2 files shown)


docs/rules/jsx-no-useless-fragment.md
```diff
@@ -32,6 +32,10 @@ Examples of **incorrect** code for this rule:
 Examples of **correct** code for this rule:
 
 ```jsx
+{foo}
+
+<Foo />
```
ljharb, line 37: i'm confused, why are these needed to make the following fragment useless?

karlhorky: ah sorry if this was unclear, they were intended to be the simplifications for these (the first two examples in the incorrect code):

```jsx
<>{foo}</>

<><Foo /></>
```

ljharb (approved):

karlhorky:
Thanks for the review and merge!

PR #3351, opened 2022-08-08
[Refactor] `jsx-closing-bracket-location`, `jsx-no-bind`: fix eslint issues

caroline223:
@ljharb

### DIFF
--- a/lib/rules/jsx-closing-bracket-location.js
+++ b/lib/rules/jsx-closing-bracket-location.js
@@ -165,7 +165,7 @@ module.exports = {
      * @return {String} The characters used for indentation
      */
     function getIndentation(tokens, expectedLocation, correctColumn) {
-      correctColumn = correctColumn || 0;
+      const newColumn = correctColumn || 0;
       let indentation;
       let spaces = [];
       switch (expectedLocation) {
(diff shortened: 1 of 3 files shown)


ljharb (commented):
I don't see any eslint overrides being removed, or config being changed?

caroline223:
@ljharb May you review this PR for me? Thanks.

lib/rules/jsx-closing-bracket-location.js
```diff
@@ -165,7 +165,7 @@ module.exports = {
      * @return {String} The characters used for indentation
      */
     function getIndentation(tokens, expectedLocation, correctColumn) {
-      correctColumn = correctColumn || 0;
+      const newColumn = correctColumn || 0;
```
caroline223, line 168: @ljharb Here is where I made one change. Would this be plausible?

lib/rules/jsx-no-bind.js
```diff
@@ -111,6 +114,11 @@ module.exports = {
       return null;
     }
 
+    /**
+     * @param {string | number} violationType
+     * @param {string} variableName
```
caroline223, line 119: @ljharb May you review the changes that I have made here?

ljharb: I think this PR needs to be rebased onto latest main, but in general it seems good :-)

ljharb:
Synced with #3356.

PR #3353, opened 2022-08-10
[Fix] `jsx-no-leaked-render`: autofix nested "&&" logical expressions

### LINKED ISSUE #3323: Bug: `jsx-no-leaked-render` faulty autofix
Version: 7.30.1
Config:
```js
'react/jsx-no-leaked-render': ['error', { validStrategies: ['coerce', 'ternary'] }],
```

Before fix:
```tsx
const MyComponent = () => {
  const items = []
  const breakpoint = { phones: true }

  return <div>{items.length > 0 && breakpoint.phones && <span />}</div>
}
```

Post fix:

```tsx
const MyComponent = () => {
  const items = []
  const breakpoint = { phones: true }

  return <div>{!!!!!!!!!!!!!!!!!!!!items.length > 0 && breakpoint.phones && <span />}</div>
}
```

Expected:

```tsx
const MyComponent = () => {
  const items = []
  const breakpoint = { phones: true }

  return <div>{items.length > 0 && !!breakpoint.phones && <span />}</div>
}
```

On top of that, the auto fix is quite slow.

hduprat:
A partial fix to #3323

I welcome suggestions and help if you think it does not improve performance 🙂

### DIFF
--- a/lib/rules/jsx-no-leaked-render.js
+++ b/lib/rules/jsx-no-leaked-render.js
@@ -40,19 +40,30 @@ function getIsCoerceValidNestedLogicalExpression(node) {
   return COERCE_VALID_LEFT_SIDE_EXPRESSIONS.some((validExpression) => validExpression === node.type);
 }
 
+function extractExpressionBetweenLogicalAnds(node) {
+  if (node.type !== 'LogicalExpression') return [node];
+  if (node.operator !== '&&') return [node];
+  return [].concat(
+    extractExpressionBetweenLogicalAnds(node.left),
+    extractExpressionBetweenLogicalAnds(node.right)
+  );
+}
+
 function ruleFixer(context, fixStrategy, fixer, reportedNode, leftNode, rightNode) {
   const sourceCode = context.getSourceCode();
   const rightSideText = sourceCode.getText(rightNode);
 
   if (fixStrategy === COERCE_STRATEGY) {
-    let leftSideText = sourceCode.getText(leftNode);
-    if (isParenthesized(context, leftNode)) {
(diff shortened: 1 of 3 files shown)


Belco90 (changes requested):
LGTM! Could you add the scenarios reported in #3323 as valid tests? So it's clear that those cases have been fixed.

lib/rules/jsx-no-leaked-render.js
```diff
@@ -40,19 +40,28 @@ function getIsCoerceValidNestedLogicalExpression(node) {
   return COERCE_VALID_LEFT_SIDE_EXPRESSIONS.some((validExpression) => validExpression === node.type);
 }
 
+function extractExpressionBetweenLogicalAnds(node) {
+  if (node.type !== 'LogicalExpression') return [node];
+  if (node.operator !== '&&') return [node];
+  const op = [];
+  return op.concat(extractExpressionBetweenLogicalAnds(node.left), extractExpressionBetweenLogicalAnds(node.right));
```
ljharb, line 47:
```suggestion
  return [].concat(
    extractExpressionBetweenLogicalAnds(node.left),
    extractExpressionBetweenLogicalAnds(node.right)
  );
```

ljharb (changes requested):
Looks great, but could we add the test case from the OP of the linked issue?

lib/rules/jsx-no-leaked-render.js
hduprat: @ljharb it's done now

Belco90 (approved):

ljharb (approved):

Issue #2366, opened 2019-07-31
fix for jsx-sort-props doesn't respect comments


7rulnik:
Fix for jsx-sort-props doesn't respect comments.

For example:

```jsx
<foo
  b={1}
  // comment for a
  a={0}
/>
```

will be fixed like this:

```jsx
<foo
  a={0}
  // comment for a
  b={1}
/>
```

But I think it should be like this:

```jsx
<foo
  // comment for a
  a={0}
  b={1}
/>
```

ljharb:
Comment attachment is a very hard problem; some people put comments above, and some below, and many don’t comment at all.

We may be able to handle something here, but we’d be constrained by how Babel implemented comment attachment.

7rulnik:
Yeah, I agree.

About Babel AST:

```
<foo
  a={0}
  // comment for a
  b={1}
/>
```

For this example Babel will keep comment in ast for a prop as trailing comment and for b same comment as leading. I think we can configure it via options. Something like `preserveComment:trailing|leading`. But seems that it will be not so trivial.

But maybe we can implement option just for disabling autofix?

ljharb:
It seems prudent to disable autofixes entirely for props that have a leading or a trailing comment, no option needed.

7rulnik:
Maybe we should disable it entirely for component's that contains props with comments? Otherwise result of autofix will be confusing.

For example group of props before comment sorted in alphabetical order and after comment too. But it doesn't make sense, because it's still not valid.
  
```jsx
<input
  a
  b
  c
  // fofof
  f
  a
  b
  c
/>
```

If so, I can submit PR

ljharb:
In that example I’d expect everything to be sorted except the ones surrounding the comment, and the dev being forced to manually fix those.

7rulnik:
So, something like this? Ok, I will take a look
```jsx
<input
  a
  a
  b
  b
  c
  c
  // fofof
  f
/>
```

Vages:
Have you had a look yet, @7rulnik ? No pressure; I'm just evaluating whether to have a go at this myself, and I'd like to hear if you're stuck at some bump which I'm also likely to hit.

Vages:
And by the way, in your example: Has the `c` directly above the `f` prop been moved because it is connected to the comment, or has it just been moved because that's the correct place to put it when sorting alphabetically?

ljharb:
@Vages if the original was:
```jsx
<input
  a
  b
[...]

PR #3358, opened 2022-08-16
[Fix] `jsx-sort-props`: sorted attributes now respect comments

ROSSROSALES:
Fixes https://github.com/jsx-eslint/eslint-plugin-react/issues/2366

Previously comments were not sorted with rule.
With fix, comments are included during sort.

Special Attribute Conditions:
- if comment found between two attributes, attributes and comment are sorted to end (to be manually sorted if necessary)
- above condition also applies to block comments
- single-line comments on same attribute-line are sorted

### DIFF
--- a/lib/rules/jsx-sort-props.js
+++ b/lib/rules/jsx-sort-props.js
@@ -46,9 +46,26 @@ function isReservedPropName(name, list) {
   return list.indexOf(name) >= 0;
 }
 
+let attributeMap;
+// attributeMap = [endrange, true||false if comment in between nodes exists, it needs to be sorted to end]
+
+function shouldSortToEnd(node) {
+  const attr = attributeMap.get(node);
+  return !!attr && !!attr[1];
+}
+
 function contextCompare(a, b, options) {
   let aProp = propName(a);
   let bProp = propName(b);
+
+  const aSortToEnd = shouldSortToEnd(a);
+  const bSortToEnd = shouldSortToEnd(b);
+  if (aSortToEnd && !bSortToEnd) {
+    return 1;
+  }
+  if (!aSortToEnd && bSortToEnd) {
+    return -1;
+  }
 
   if (options.reservedFirst) {
     const aIsReserved = isReservedPropName(aProp, options.reservedList);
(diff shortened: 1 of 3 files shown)


ljharb:
Looks like there's older eslint versions where `sourceCode.getCommentsAfter` doesn't exist, so we'll need to handle that both in the rule, and in the test cases

ROSSROSALES (commented):
Thank you for the review!
I really appreciate your time. I have learned many of my own faults with redundant code, naming variables, and conditional statements. I hope in the future to improve and reduce these nuances.

ROSSROSALES:
> Looks like there's older eslint versions where `sourceCode.getCommentsAfter` doesn't exist, so we'll need to handle that both in the rule, and in the test cases

For this change:
Would we need to modify the code to identify which ESLint version is being used?
Then create another function using older deprecated methods of ESLint?

I am unsure other ways of handling errors for older ESLint versions.

ROSSROSALES:
@ljharb 
I decided to go with latter of "work with newer ESLint versions". Other useful methods have been deprecated as well and would throw same previous error.

After changes,
I am getting a new error:
```AssertionError: Did not specify errors for an invalid test of jsx-sort-props```

I tried to search issue online and through other rule using older ESLint, but I had no luck. 
How is this error handled?

**Comment by ljharb** [comment]
The `invalid` list needs to be changed from an array, to a `[].concat()` call.

I'm thinking about whether or not we need an option for this, or whether it should be the default behavior. Either way, there aren't any tests for the `commentbetween` option - so, let's either add those, or remove the code handling that option?

**Comment by ROSSROSALES** [comment]
> The `invalid` list needs to be changed from an array, to a `[].concat()` call.
> 
> I'm thinking about whether or not we need an option for this, or whether it should be the default behavior. Either way, there aren't any tests for the `commentbetween` option - so, let's either add those, or remove the code handling that option?

I created the `commentbetween` as an option defaulted to true because of how I initially understood `sort((a, b) => contextCompare(a, b, options)))` . I agree with you that it should be the default behaviour of the rule. I will make sure to edit and remove the option so that the default behaviour reflects it. Please let me know in the next pull request if I made appropriate changes.

File: tests/lib/rules/jsx-sort-props.js [hunk 1]
```diff
@@ -821,5 +822,231 @@ ruleTester.run('jsx-sort-props', rule, {
...
+        {
+          messageId: 'sortPropsByAlpha',
+          line: 2,
+        },
+        {
+          messageId: 'sortPropsByAlpha',
+          line: 2,
+        },
+      ],
+    } : [],
```
ljharb, line 1050:
```suggestion
    } : []
```
now that it's a function call, a trailing function argument comma doesn't work in the older nodes we support

lib/rules/jsx-sort-props.js
```diff
@@ -118,32 +135,81 @@ function contextCompare(a, b, options) {
...
   let groupCount = 0;
+  function addtoSortableAttributeGroups(attribute) {
+    sortableAttributeGroups[groupCount - 1].push(attribute);
+  }
+
   for (let i = 0; i < attributes.length; i++) {
+    const attribute = attributes[i];
+    const nextAttribute = attributes[i + 1];
+    const attributeline = attribute.loc.start.line;
+    let comment = []
```
ljharb, line 154:
```suggestion
    let comment = [];
```

lib/rules/jsx-sort-props.js
```diff
@@ -118,32 +135,81 @@ function contextCompare(a, b, options) {
...
+    try {
+      comment = sourceCode.getCommentsAfter(attribute)
+    } catch (e) {
+    } finally {
+    }
     const lastAttr = attributes[i - 1];
```
ljharb, line 160:
```suggestion
    let comment = [];
    try {
      comment = sourceCode.getCommentsAfter(attribute);
    } catch (e) { /**/  }
    const lastAttr = attributes[i - 1];
```

lib/rules/jsx-sort-props.js
```diff
@@ -118,32 +135,81 @@ function contextCompare(a, b, options) {
...
+        attributeMap.set(attribute, [attribute.range[1], false]);
+        addtoSortableAttributeGroups(attribute);
+      } else {
+        const firstComment = comment[0];
+        const commentline = firstComment.loc.start.line;
+        if (comment.length === 1) {
+          if (attributeline + 1 === commentline && nextAttribute) {
+            attributeMap.set(attribute, [nextAttribute.range[1], true]);
+            addtoSortableAttributeGroups(attribute);
+            i++;
```
ljharb, line 184:
```suggestion
            i += 1;
```

lib/rules/jsx-sort-props.js
```diff
@@ -118,32 +135,81 @@ function contextCompare(a, b, options) {
...
+          } else if (attributeline === commentline) {
+            if (firstComment.type === 'Block') {
+              attributeMap.set(attribute, [nextAttribute.range[1], true]);
+              i++;
```
ljharb, line 188:
```suggestion
              i += 1;
```

lib/rules/jsx-sort-props.js
```diff
@@ -118,32 +135,81 @@ function contextCompare(a, b, options) {
...
+          const commentNextAttribute = sourceCode.getCommentsAfter(nextAttribute);
+          attributeMap.set(attribute, [nextAttribute.range[1], true]);
+          if (
+            commentNextAttribute.length === 1
+            && nextAttribute.loc.start.line === commentNextAttribute[0].loc.start.line
+          ) {
+            attributeMap.set(attribute, [commentNextAttribute[0].range[1], true]);
+          }
+          addtoSortableAttributeGroups(attribute);
+          i++;
```
ljharb, line 204:
```suggestion
          i += 1;
```

ljharb (commented):
adding some lint fixes

lib/rules/jsx-sort-props.js
```diff
@@ -118,32 +135,81 @@ function contextCompare(a, b, options) {
...
   for (let i = 0; i < attributes.length; i++) {
+    const attribute = attributes[i];
+    const nextAttribute = attributes[i + 1];
+    const attributeline = attribute.loc.start.line;
+    let comment = []
+    try {
+      comment = sourceCode.getCommentsAfter(attribute)
+    } catch (e) {
+    } finally {
+    }
```
ljharb, line 159:
```suggestion
    let comment = [];
    try {
      comment = sourceCode.getCommentsAfter(attribute);
    } catch (e) { /**/ }
```

ljharb (approved):

ROSSROSALES:
Thank you for reviewing the changes! @ljharb

PR #3369, opened 2022-08-25
[Fix] `jsx-key` not detecting missing keys in `Array.from`'s mapping function

### LINKED ISSUE #3368: react/jsx-key doesn't detect Array.from
[`react/jsx-key`](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-key.md) doesn't detect a specific case where an iterable is [mapped using `Array.from`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#parameters):

```
const mySet = new Set()

// ...

// Should warn for missing key property, but doesn't
Array.from(mySet, () => <div />)
```

sjarva:
Closes #3368 

Now the `jsx-key` rule will catch these invalid cases:

```
Array.from([1, 2 ,3], function(x) { return <App /> });
Array.from([1, 2 ,3], (x => { return <App /> }));
Array.from([1, 2 ,3], (x => <App />));
```

and will allow these valid cases:

```
Array.from([1, 2, 3], function(x) { return <App key={x} /> });
Array.from([1, 2, 3], (x => <App key={x} />));
Array.from([1, 2, 3], (x => {return <App key={x} />}));
Array.from([1, 2, 3], someFn);
Array.from([1, 2, 3]);
```

In addition, I did some refactoring to keep the code DRY (the checking for keys in `Array.prototype.map` and `Array.from` had many repetitions).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ## Unreleased
 
+### Fixed
+* [`jsx-key`]: fix detecting missing key in `Array.from`'s mapping function ([#3369][] @sjarva)
+
+[#3369]: https://github.com/jsx-eslint/eslint-plugin-react/pull/3369
+
 ## [7.31.0] - 2022.08.24
 
 ### Added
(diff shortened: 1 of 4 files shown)


ljharb (approved):
Looks great, thanks!

PR #3370, opened 2022-08-26
[Fix] `jsx-no-leaked-render` autofixed JSXElements away in coerce strategy

### LINKED ISSUE #3354: jsx-no-leaked-render strategies question config question
Hello,

Is there a way to make **coerce** to be the preferred strategy, and to not adjust the code when there's a ternary ?

I have code that looks like
`{ foo ? <thing1> : <anotherthing2>}`

with coerce enabled as the rule, its converting the code to

`{ !!foo && <thing1>}` and i lose the else. 

Is there a way to make coerce apply only when there's no ternary?

For instance, I would like the coerce rule to apply only when I do
`{ foo && <thing> }`

When I make ternary the default it converts all of the 
`{ foo && <thing> }` to `{ foo ? <thing>: null }`

Is it possible to have ternary and coerce as valid strategies without autofixing?

`{ foo ? <thing1> : <anotherthing2>}` 
`{ !!foo && <thing1>}`

Thanks for your time!

sjarva:
Closes #3354 

Now, when the `jsx-no-leaked-render` has `{validStrategies: ['coerce']}`, the following code is considered as valid and therefore is not autofixed:

```jsx

{ foo ? <Component1 /> : <Component2 />}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ### Fixed
 * [`jsx-key`]: fix detecting missing key in `Array.from`'s mapping function ([#3369][] @sjarva)
+* [`jsx-no-leaked-render`]: coerce strategy now allows a ternary ([#3370][], @sjarva)
 
+[#3370]: https://github.com/jsx-eslint/eslint-plugin-react/pull/3370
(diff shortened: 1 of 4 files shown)


ljharb (approved):
Seems great!

sjarva:
Thanks once again for a quick review! 🤗

PR #3375, opened 2022-08-30
[Docs] Replacing legacy ref string with ref variable

Luccasoli:
Just rewriting `ref` props using `ref` variables.


ljharb:
I think it’d be better to *add* these examples instead of changing them, since both patterns are checked.

ljharb (approved):
Actually this is fine as-is

PR #3379, opened 2022-09-02
Allows using undefined or empty string for fixer

gkiely:
jsx-no-leaked-render ternary fixes the error with `null`.

Some codebases disable the use of `null` and only use `undefined`, alternatively some developers may choose to use an empty string.

This commit allows configuring that.


gkiely:
Decided to write my own plugin, as my needs are specific/niche.

ljharb:
I'm a bit confused. If you had such a rule in your codebase (for no good reason; null's a perfectly fine value), then wouldn't *that* rule handle autofixing nulls to undefineds?

Issue #3380, opened 2022-09-02
react/no-unknown-property throws on valid properties


AhmadMayo:
I recently upgraded eslint-plugin-react from v7.30.1 to v7.31.2, and when I tried to build, the rule started throwing on valid properties like key, viewBox, fill, d, and other SVG related properties - and none of the in kebab case, or contains a colon in their name.

eslint version: 8.23.0
nextjs version: 12.2.5

![image](https://user-images.githubusercontent.com/15085304/188234070-ec9124a9-1d7d-4f6b-96a9-9e6522a52fcb.png)

mjchang:
Also getting this for other svg properties as well
- stroke
- strokeWidth
- strokeLinecap
- strokeLinejoin
- transform
- clipRule
- x1
- x2
- y1
- y2

AhmedBHameed:
I confirm the issue.

AhmadMayo:
@mjchang @AhmedBHameed when i commented this line from my eslint config, the issue seized to occur.
```
  extends: ["plugin:@next/next/recommended"],
```
Do you use next too? can you confirm if this resolves the issue for you too? if so, then the problem is related to next's plugin

mjchang:
I don't use next. I use airbnb's eslint which includes this.

peter-resmed:
![image](https://user-images.githubusercontent.com/109127976/188237711-ba5e892e-cc5e-4785-a0e5-9624bf306da5.png)
same issue

AhmedBHameed:
@AhmadMayo 

I do use nextjs.I have default project extends
```json
{
	"extends": ["next/core-web-vitals"]
}
```

Which also causing eslint to complain.

I guess removing my line means ignoring eslint completely? However, the build run successfully when removing the line from `.eslintrc.json`

I reverted the update from v7.30.1 to v7.29.4 and the app return back to live.

I would say that might be a bug with the new version maybe?

AhmadMayo:
@AhmedBHameed maybe. It's occurring randomly for me now. reverting back to version 7.29 didn't resolve the issue for me.

AhmedBHameed:
@AhmadMayo 

Did you revert back your `yarn.lock` & `package.json`? Also did you delete both `node_modules` and `.next` before installing a fresh deps with `yarn`?

Try this and then build again.
That is what i did.

AhmadMayo:
@AhmedBHameed same errors. I'll turn off the rule until I figure out what's going on.

[...]

PR #3381, opened 2022-09-02
[Fix] no-unknown-property: No unknown property doesn't recognize svg properties

AhmadMayo:
fixes https://github.com/jsx-eslint/eslint-plugin-react/issues/3380

I don't know if the added attributes are exhaustive or not. I sure hope MDN covered all possible attributes.

### DIFF
--- a/lib/rules/no-unknown-property.js
+++ b/lib/rules/no-unknown-property.js
@@ -29,6 +29,9 @@ const DOM_ATTRIBUTE_NAMES = {
 const ATTRIBUTE_TAGS_MAP = {
   // image is required for SVG support, all other tags are HTML.
   crossOrigin: ['script', 'img', 'video', 'audio', 'link', 'image'],
+  fill: ['svg'],
+  property: ['meta'],
+  viewBox: ['svg'],
 };
 
 const SVGDOM_ATTRIBUTE_NAMES = {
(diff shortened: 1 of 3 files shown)


lib/rules/no-unknown-property.js
```diff
@@ -29,6 +29,9 @@ const DOM_ATTRIBUTE_NAMES = {
 const ATTRIBUTE_TAGS_MAP = {
   // image is required for SVG support, all other tags are HTML.
   crossOrigin: ['script', 'img', 'video', 'audio', 'link', 'image'],
+  fill: ['svg'],
+  property: ['meta'],
+  viewBox: ['svg'],
```
ljharb, line 34: i added these as well.

ljharb (approved):
Looks great!

lib/rules/no-unknown-property.js
```diff
@@ -29,6 +29,9 @@ const DOM_ATTRIBUTE_NAMES = {
 const ATTRIBUTE_TAGS_MAP = {
   // image is required for SVG support, all other tags are HTML.
   crossOrigin: ['script', 'img', 'video', 'audio', 'link', 'image'],
+  fill: ['svg'],
```
shepmaster, line 32: `fill` appears to be OK [on a number of other tags](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill), such as `path`

ljharb: This should be handled by #3385

PR #3398, opened 2022-09-05
[Fix] no-unknown-property: onError and onLoad on <picture> and <iframe> elements

maiis:
This PR adds:

- allows `onError` and `onLoad` property on `<picture>` element
- allows `onError` and `onLoad` property on `<iframe>` element

PR #3406, opened 2022-09-05
[Fix] `no-unknown-property`: Mark onLoad/onError as supported on iframes

akx:


ljharb:
I combined this with #3398. (Duplicate of #3398)

Issue #3401, opened 2022-09-05
react/no-unknown-property stopped supporting `imageSrcSet` and `imagSizes` on <link /> tag


terrymun:
Version 7.31.6 will cause a linting error to be thrown when `imageSrcSet` or `imageSizes` are being used in a JSX `<link />` element, even though this attribute is perfectly valid: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-imagesrcset

Combing through the rules reveal that these attributes are missing: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/lib/rules/no-unknown-property.js

It is also available in the [`LinkHTMLAttributes` type](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L2283):

```typescript
interface LinkHTMLAttributes<T> extends HTMLAttributes<T> {
  as?: string | undefined;
  crossOrigin?: string | undefined;
  href?: string | undefined;
  hrefLang?: string | undefined;
  integrity?: string | undefined;
  media?: string | undefined;
  imageSrcSet?: string | undefined;
  imageSizes?: string | undefined;
  referrerPolicy?: HTMLAttributeReferrerPolicy | undefined;
  rel?: string | undefined;
  sizes?: string | undefined;
  type?: string | undefined;
  charSet?: string | undefined;
}
```

terrymun:
PR has been created here: #3407

PR #3407, opened 2022-09-05
[Fix] `no-unknown-property`: allow `imageSrcSet` and `imageSizes` attributes on `<link>`

terrymun:
This pull requests allows `imageSrcSet` and `imageSizes` on the `<link />` element, as these properties are actually valid and allowed:

* MDN Docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-imagesrcset
* React typings: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L2276

This is a fix for #3401


ljharb (approved):
Thanks!

terrymun:
@ljharb No... thank _you_ for looking into the PR so quickly. Sincerely appreciate the response time :)

PR #3413, opened 2022-09-06
Don't add markdownlint config to NPM package

jorrit:
Perhaps a `"files"` key in package.json is a better idea than expanding the `"ignore"` list again. But I'll leave that to the maintainer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ## Unreleased
 
+### Changed
+
+* [Meta] npmignore markdownlint config ([#3413][] @jorrit)
+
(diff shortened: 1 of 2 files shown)


ljharb:
The files key is dangerous and should never be used; expanding the ignore list is proper.

jorrit:
Alright, I didn't know that. Do you have some reference where I can read more about the dangers?

ljharb:
No - it's just the way it behaves. If you forget to add a file to "npmignore", the worst consequence is "a file nothing cares about ends up in the package" - iow, harmless, and at worst inconvenient. If you forget to add a file to "files", the worst consequence is "downstream users are broken", which can easily cause millions of dollars of person-hours of work.

ljharb (approved):

PR #3417, opened 2022-09-07
[New] `forbid-component-props`: add `disallowedFor` option

### LINKED ISSUE #2946: Rule option "disallowedFor" for component/dom element prop

### LINKED ISSUE #3422: Add Rule option "disallowedFor" for component prop
Suppose we have a component that accept a prop named string.

``` javascript
function Comp(props) {
    return <div>{props.string}</div>;
}

function App() {
    return <Comp string="213"  />;
}

```

As needs change, we need to replace string prop by another named number.

``` javascript

function Comp(props) {
    return <div>{props.number || props.string}</div>;
}
Comp.propTypes = {
    number: PropTypes.number,
    /**
     * @deprecated use number instead
     */
    string: PropTypes.string
};
Comp.defaultProps = {
    string: '123',
    number: 123
};
```

The string prop will works in a period of time, but it's better to replace it with number prop.
This rule option can be use to avoid using deprecated prop for Comp.

``` json
{
"react/forbid-component-props": [
    "error", { "forbid": [{"propName": "string", "disallowedList": ["Comp"]}] }
  ]
}

```
image

jacketwpbb:
Fixes #3422.

### DIFF
--- a/docs/rules/forbid-component-props.md
+++ b/docs/rules/forbid-component-props.md
@@ -50,8 +50,18 @@ custom message, and a component allowlist:
 ```js
 {
   "propName": "someProp",
-  "allowedFor": [SomeComponent, AnotherComponent],
-  "message": "Avoid using someProp"
+  "allowedFor": ["SomeComponent", "AnotherComponent"],
+  "message": "Avoid using someProp except SomeComponent and AnotherComponent"
+}
+```
+
+Use `disallowedFor` as an exclusion list to warn on props for specific components. `disallowedFor` must have at least one item.
+
+```js
+{
+  "propName": "someProp",
+  "disallowedFor": ["SomeComponent", "AnotherComponent"],
+  "message": "Avoid using someProp for SomeComponent and AnotherComponent"
 }
 ```
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 * [`display-name`]: add `checkContextObjects` option ([#3529][] @JulesBlm)
 * [`jsx-first-prop-new-line`]: add `multiprop` option ([#3533][] @haydncomley)
 * [`no-deprecated`]: add React 18 deprecations ([#3548][] @sergei-startsev)
+* [`forbid-component-props`]: add `disallowedFor` option ([#3417][] @jacketwpbb)
 
 ### Fixed
 * [`no-array-index-key`]: consider flatMap ([#3530][] @k-yle)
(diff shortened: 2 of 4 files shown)


docs/rules/forbid-component-props.md
```diff
@@ -50,8 +50,18 @@ custom message, and a component whitelist:
...
 {
   "propName": "someProp",
-  "allowedFor": [SomeComponent, AnotherComponent],
-  "message": "Avoid using someProp"
+  "allowedFor": ["SomeComponent", "AnotherComponent"],
+  "message": "Avoid using someProp except SomeComponent and AnotherComponent"
+}
+```
+
+Use `disallowedFor` as a blacklist to lint props for specific components. `disallowedFor` should have at least one item.
```
sjarva, line 58: Let's use inclusive language and keep the terminology that this rule uses: replace "blacklist" with "disallowlist" or "blocklist"

(I'm leaving this as a comment and not a "request changes" because I'm not sure if the community members are allowed to request changes to others' PRs.) Totally up to @ljharb whether you require this too or not.

ljharb:
#2946 was already resolved in #3338.

There isn't any consensus on doing anything further, so in the future, please do not send PRs for changes you're not sure are wanted.

In the meantime, let's leave this open, and please file a *new* issue so we can discuss _if_ this is desired.

jacketwpbb:
Sorry，it was my first time to make a pull request, I have created a new issue for discussion.

ljharb (approved):

PR #3429, opened 2022-09-10
[New] support new config system

jjangga0214:
Supporting eslint's new config system of eslint.

Note that the legacy config system always has `require()`d plugins and sharable configs, while the new system is ESM.

Thus [conditional export](https://nodejs.org/api/packages.html#conditional-exports) is great to keep compatibility.

`plugin:` protocol(e.g. `plugin:react/recommended`) is not valid any more. Thus the new plugin doesn't have `configs`.

Sharable configs for the new config system are now exported as `eslint-plugin-react/all`, `eslint-plugin-react/recommended`, `eslint-plugin-react/jsx-runtime`


ljharb:
um, what? The new system is ESM? Why can’t it be CJS?

jjangga0214:
@ljharb Your point is right.
Though eslint recommends ESM for the new config system, it supports CJS as well.

So, in conclusion,

```
if ( require()d ){
  the user is either using legacy or new config system. 
} else if( import()ed ) {
  the user is using new config system
}
```

For supporting the new config system in CJS, I patched .mjs to .js in the new commit.

And `eslint-plugin-react/new` would be for users who want to use CJS for the new config system.

ljharb:
Native ESM doesn't have any benefits over CJS (or transpiled ESM), and lacks the majority of the relevant tooling capabilities, so I'd strongly prefer to avoid using it entirely - let's keep everything in CJS.

What is the benefit of the new config system? Why does an eslint user care what config system this plugin is using?

jjangga0214:
> What is the benefit of the new config system?

You may want to read these if you haven't.

- https://eslint.org/blog/2022/08/new-config-system-part-1/
- https://eslint.org/blog/2022/08/new-config-system-part-2/
- https://eslint.org/blog/2022/08/new-config-system-part-3/

>

jjangga0214:
스크린샷 2022-09-11 오후 7 52 54

I added a new dependency named [globals](https://www.npmjs.com/package/globals).
This requires higher `engines` limitation.
How would we do?

1. Increase `engines`.
1. Or embed a part of `globals`. This would not be a hard task.

ljharb:
That issue benefits shared configs, not plugins.

configs/recommended.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const globals = require('globals');
+const react = require('../index');
+const legacy = require('../legacy');
+
+module.exports = {
+  files: ['**/*.{js,jsx,ts,tsx}'],
+  globals: Object.assign({}, globals.serviceworker, globals.browser),
```
ljharb, line 9: We don’t specify this now, why would we add it? Environment-specific things should be only in the top level. For example, most people don’t use service workers (and shouldn’t), so it would be harmful to include here; and some react codebases are server-only.

jjangga0214: > things only in the top level.

jjangga0214: > things only in the top level.

The new system is all about a concept like `overrides` by **chaining**.
So there is no such a 'top' level, actually.
Compared to the legacy, configs aggressively specify their preference, and override each other, then finally the user has to override by their preference at the end of the chain.

For example, this snippet from the official blog post shows the design intention.

스크린샷 2022-09-12 오후 12 14 11

The second config's `semi` rule is overridden by the third. Not only rules, but everything can be overridden.

A sharable`customConfig` is just dropped as-is at the first place of the array.
If `eslint-plugin-react/all` or `eslint-plugin-react/recommended` are inserted as-is like that way, and if the object does not have `files` property preconfigured, the `files` would be `['**/*.js']` by default in the new system, which users might think unintuitive.

So the user would have to almost always destruct the config.

```js
import react from `eslint-plugin-react/recommended`

export default [
  {
    files: ['**/*.{js,jsx,ts,tsx}'],
    ...react
  },
  {
  // ... 
  }
]
```

Instead of,


```js
import react from `eslint-plugin-react/recommended`

export default [
  react,
  {
  // ... 
  }
]
```

This is not a big deal, but the point is it might become a boilerplate in many cases.

How do you think?

IMHO,
- `globals`: I respect your opinion. I'd remove it.
- `files`: I think it's good to be the default. If you still think differently, I'd remove it.

ljharb: The top level means the top-level application - the end user.

ljharb: So this comment was about `globals`, but personally I’d remove files too. I think this is a bad design choice, and end users should be the only ones to ever put in filepaths, because they’re the only ones that can know how their codebase is organized.

ljharb: Also, why would that files glob be unintuitive for the user? It’s the default behavior of eslint already, for years.

jjangga0214: > Also, why would that files glob be unintuitive for the user? It’s the default behavior of eslint already, for years.
Because by default, if we don't preconfigure `files`, it would be `['**/*.js']`, while majority of users use react with .jsx or .tsx. That's why I thought it'd be unintuitive default for users. 
Do you still want to remove `files`?

ljharb: I see what you're saying, but yes, I think this should not include `files`.

jjangga0214:
Okay, I rewrote the config without `files` and `globals`, and README as well. 
Would you please check them out?

configs/recommended.js
jjangga0214: Okay.

package.json
```diff
@@ -3,7 +3,17 @@
...
   "author": "Yannick Croissant <yannick.croissant+npm@gmail.com>",
   "description": "React specific linting rules for ESLint",
-  "main": "index.js",
+  "main": "legacy.js",
+  "exports": {
+    ".": {
+      "import": "./index.js",
+      "require": "./legacy.js"
+    },
+    "./new": "./index.js",
```
ljharb, line 12: i'm a bit confused still. why is this different entrypoint needed?

Why can't the current "main" be used as-is in the new config system?

jjangga0214: The current "main" (which is renamed to legacy.js in this PR), does not follow the new config spec.
In the legacy system, a plugin can have preset(s) in the `configs` property.
However, plugin: protocol (e.g. `plugin:react/recommended`) is not valid any more in the new system. 
Thus, the new plugin(a new entrypoint index.js) doesn't have `configs`, following the spec.
This does not mean "the current main" cannot be used in the new system.
However, because it has a non-standard field `configs`, people might try to depend on it to load shareable configs, which is also not a standard way.
Eslint officially recommends people t o import shareable config directly, and this is an intention to establish the intended convention.

package.json
```diff
@@ -3,7 +3,17 @@
   "version": "7.31.8",
   "author": "Yannick Croissant <yannick.croissant+npm@gmail.com>",
   "description": "React specific linting rules for ESLint",
-  "main": "index.js",
+  "main": "legacy.js",
+  "exports": {
```
ljharb, line 7: Adding "exports" is always a breaking change, so this is a nonstarter.

package.json
ljharb: I'm comfortable with the risk of people choosing to do that. Since the current plugin Just Works in the new config system, it seems like the only change that might be beneficial is adding new entry points for the configs to be accessed directly?

ljharb (approved):
I fixed up the code samples, added the missing mjs/cjs/mtsx/mjsx extensions (native ESM should always be .mjs, not .js, so it's critical to allow that), converted the CJS files you added to ESM, and decided to put the configs under a top-level `configs` directory instead, and refactored `index.js` to pull from those.

I'm dismayed that eslint's new config system removes the env key; users shouldn't have to import globals themselves, especially if they'd have to depend directly on the globals package.

README.md
```diff
@@ -109,6 +109,174 @@ Enable the rules that you would like to use.
...
+}
+```
+
+**Note**: These configurations will import `eslint-plugin-react` and enable JSX in [parser options](https://eslint.org/docs/user-guide/configuring/language-options#specifying-parser-options).
+
+## Configuration (new: `eslint.config.js`)
+
+From [`v8.21.0`](https://github.com/eslint/eslint/releases/tag/v8.21.0), eslint announced a new config system.
+In the new system, `.eslintrc*` is no longer used. `eslint.config.js` would be the default config file name.
+In eslint `v8`, the legacy system (`.eslintrc*`) would still be supported, while in eslint `v9`, only the new system would be supported.
```
AriPerkkio, line 148: > while in eslint `v9`, only the new system would be supported

Where is this part coming from? ESLint has not announced the end-of-life for `.eslintrc`.

ljharb: whew, i'm really glad to hear that. let's remove this line unless there's a citation.

AriPerkkio: This same phrase seems to be used in other community plugins as well, https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1886.

I just asked about this on ESLint chat and the maintainers said no such announcement about EOL has been made.

ljharb: That's the same contributor putting it in there, so it may come from the same misunderstanding.

I can't comment on that PR, so it'd probably be helpful if you commented there as well about the incorrect EOL info.

jjangga0214: @AriPerkkio Thanks for fixing my confusion again :)
(He also did the same in another repo.)

PR #3441, opened 2022-09-26
[New] `jsx-no-leaked-render`: add `ignoreAttributes` option

### LINKED ISSUE #3292: False positive for react/jsx-no-leaked-render
Following code triggers `react/jsx-no-leaked-render` although it's perfectly valid:

```
return <MyReactComponent
          isDisabled={foo && bar === 0}
          onClick={handleOnClick} />;
```

with:
```
foo: boolean;
bar: number;
```

aleclarson:
Fixes #3292

When `ignoreAttributes` is true, validation of JSX attribute values is skipped.

## Todo
- [ ] write docs

### DIFF
--- a/docs/rules/jsx-no-leaked-render.md
+++ b/docs/rules/jsx-no-leaked-render.md
@@ -153,6 +153,60 @@ const Component = ({ elements }) => {
 
 The supported options are:
 
+### `ignoreAttributes`
+
+Boolean. When set to `true`, this option ignores all attributes except for `children` during validation, preventing false positives in scenarios where these attributes are used safely or validated internally. Default is `false`.
+
+It can be set like:
+
+```jsonc
+{
+  // ...
+  "react/jsx-no-leaked-render": [<enabled>, { "ignoreAttributes": true }]
+  // ...
+}
+```
+
+Assuming the following options: `{ "ignoreAttributes": true }`
+
+Examples of **incorrect** code for this rule, with the above configuration:
+
+```jsx
+function MyComponent({ value }) {
+  return (
+    <MyChildComponent nonChildrenProp={value && 'default'}>
+      {value && <MyInnerChildComponent />}
+    </MyChildComponent>
+  );
+}
+```
+
+Even with `ignoreAttributes: true`, `children` expressions are still checked. In the example above, `{value && <MyInnerChildComponent />}` will still be flagged.
+
+Nested JSX children within attributes are also still checked:
+
+```jsx
+const Component = ({ enabled }) => {
+  return (
+    <Foo bar={
+      <Something>{enabled && <MuchWow />}</Something>
+    } />
+  )
+}
+```
+
+Here, even though `<Something>…</Something>` is inside an attribute of `<Foo>`, the `{enabled && <MuchWow />}` expression is children of `<Something>`, so it is still flagged.
+
+Examples of **correct** code for this rule, with the above configuration:
+
+```jsx
+const Component = ({ enabled, checked }) => {
+  return <CheckBox checked={enabled && checked} />
+}
+```
+
+With `ignoreAttributes: true`, logical expressions in non-children attributes like `checked` are not flagged.
+
 ### `validStrategies`
 
 An array containing `"coerce"`, `"ternary"`, or both (default: `["ternary", "coerce"]`) - Decide which strategies are considered valid to prevent leaked renders (at least 1 is required). The "coerce" option will transform the conditional of the JSX expression to a boolean. The "ternary" option transforms the binary expression into a ternary expression returning `null` for falsy values. The first option from the array will be the strategy used when autofixing, so the order of the values matters.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 ### Added
 * [`jsx-props-no-multi-spaces`]: improve autofix for multi-line ([#3930][] @justisb)
 * [`jsx-handler-names`]: support namespaced component names ([#3943][] @takuji)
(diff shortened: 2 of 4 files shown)


levrik:
I think `ignoreProps` would be a more suitable name here.

ljharb (commented):
Can you elaborate on the use case here?

Certainly the component you're passing the prop value to might not be directly rendering it - but I'm not sure why that makes it OK to pass a renderable falsy value.

aleclarson:
> Can you elaborate on the use case here?
> 
> Certainly the component you're passing the prop value to might not be directly rendering it - but I'm not sure why that makes it OK to pass a renderable falsy value.

The use case is preferring to avoid false positives. The component receiving the prop is responsible for avoiding leaky render. Many prefer to keep the warning local to where the prop is actually rendered.

ljharb:
I see - so you're saying, you still want to report on leaky renders on DOM elements, but you don't want to do that on props on custom elements so that the custom component has the responsibility to check for it?

In that case, what i'd expect is that this could be `ignoreNonDOMProps`, and it would only suppress warning on props of non-DOM elements - would that meet your use case?

tcl333:
This rule is unusable to me until this PR is landed. I need to handle a very basic usecase of `<Component isFooBar={isFoo && isBar}>`. This rule currently auto-breaks that line by converting it into `<Component isFooBar={isFoo ? isBar : null }>`. My component doesn't take `null` as a prop, nor should it be required to.

l1135677068:
I want to know how to use the ignoreAttributes option?
i use it like belows code, but it throw an error.
```
  "react/jsx-no-leaked-render": ["error", { "validStrategies": ["coerce"], ignoreAttributes: true }]
```
Is there a use case that i can refer to?

ljharb:
@l1135677068 this PR isn't merged yet, and thus can't be released yet either, so you can't use it yet.

ljharb:
I've just rebased this PR; it's ready to go once there's documentation. If someone wants to leave a code review suggestion, I can pull it in and land this.

l1135677068:
> If someone wants to leave a code review suggestion, I can pull it in and land this.

ok, i get it.

docs/rules/jsx-no-leaked-render.md
```diff
@@ -151,6 +151,10 @@ const Component = ({ elements }) => {
 
 The supported options are:
 
+### `ignoreAttributes`
+
+*TODO*
+
```
artemxknpv, line 157: @aleclarson hey, we need it badly, any chance reviving the draft? 🙏 Let's discuss this variant:

````suggestion
### `ignoreAttributes`
Boolean. When set to `true`, this option ignores all attributes except for `children` during validation, preventing false positives in scenarios where these attributes are used safely or validated internally. Default is `false`.

It can be set like:
```json5
{
  // ...
  "react/jsx-no-leaked-render": [<enabled>, { "ignoreAttributes": true }]
  // ...
}
```

Example of incorrect usage with default setting (`ignoreAttributes: false`) and the rule enabled (consider `value` might be undefined):

```jsx
function MyComponent({ value }) {
  return (
    <MyChildComponent nonChildrenProp={value && 'default'}>
      {value && <MyInnerChildComponent />}
   </MyChildComponent>
  );
}
```

This would trigger a warning in both `nonChildrenProp` and `children` props because `value` might be undefined.

By setting `ignoreAttributes` to `true`, the rule will not flag this scenario in `nonChildrenProp`, reducing false positives, **but will keep the warning of `children` being leaked**.
````

Comment (artemxknpv) [reply docs/rules/jsx-no-leaked-render.md:157]: @ljharb cc

**Comment by mikedidomizio** [comment]
I think considering this PR has been open for 2 years without movement from the author (who is active), it's potentially safe to say they have higher priorities at this time.

I would suggest to get this past the finish line, by someone championing it and picking up where the original author left off.  Fork his fork, make the necessary documentation changes, and create a separate pull request and link it to this one.  @artemxknpv you wrote a documentation change suggestion, maybe you'd like to take this on?

It's so close, and it's crazy to me that 2 years have passed on this and all it required were some documentation updates.

I would also suggest that `ignoreAttributes` be replaced with `ignoreProps` or `ignoreNonDOMProps` like @ljharb has suggested.  I think `ignoreProps` is much more concise and simple but that's my opinion.

File: docs/rules/jsx-no-leaked-render.md [hunk 2]
```diff
@@ -151,6 +151,33 @@ const Component = ({ elements }) => {
...
+    <MyChildComponent nonChildrenProp={value && 'default'}>
+      {value && <MyInnerChildComponent />}
+   </MyChildComponent>
+  );
+}
+```
+
+This would trigger a warning in both `nonChildrenProp` and `children` props because `value` might be undefined.
+
+By setting `ignoreAttributes` to `true`, the rule will not flag this scenario in `nonChildrenProp`, reducing false positives, **but will keep the warning of `children` being leaked**.
```
ljharb, line 180:
```suggestion
By setting `ignoreAttributes` to `true`, the rule will not flag this scenario in `nonChildrenProp`, reducing false positives, **but will keep the warning of `children` being leaked**.

```

ljharb (approved):
Seems pretty good.

CruuzAzul:
I agree with the use of `ignoreProps`, which I think is more in line with what the props do. Does anyone have any idea how much more needs to be done on the PR before we can merge? 🤗 Maybe @ljharb

palexandrefernandes:
Heya folks, any idea when this will be released? Not really aware of what the `eslint-plugin-react` release schedule looks like :(

ljharb:
There's no fixed schedule. It will be released when I both have time and when I think sufficient features are in the release to warrant one. In this case, I'm waiting on #3979 which is blocked on eslint-plugin-import.

palexandrefernandes:
Totally understandable, thanks for the answer @ljharb. Will keep an eye on this :heart:

PR #3442, opened 2022-09-26
[NEW]: ensure-matching-remove-event-listener

AndreaPontrandolfo:
Tentative to add a rule that enforce developers to cleanup eventListeners in the useEffect block.


ljharb (commented):
I'm a bit confused how why this needs a full lint rule - manually adding event listeners should be *exceedingly* rare, as the proper approach is generally to add them as jsx props. Obviously `window` is a special case, but generally one puts this inside a single component managed by domain experts, and everyone else just uses the component.

docs/rules/ensure-matching-remove-event-listener.md
```diff
@@ -0,0 +1,43 @@
...
+}, [])
+```
+
+Examples of **correct** code for this rule:
+
+```jsx
+useEffect(() => {
+    window.addEventListener("keydown", handleUserKeyPress);
+    return () => {
+    window.removeEventListener("keydown", handleUserKeyPress);
```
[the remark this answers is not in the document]
AndreaPontrandolfo: Hi, appreciate the review!
I wrote this rule (and actually included this in my own Eslint plugin) because this mistake is pretty frequent at my company. I usually happen to work with people of all different skill grades and not everyone knows about every single nuance of React. 
Eslint isn't only useful to standardize our codebases, it can also prevent mistakes like this and even help juniors learn something with minimal effort, like in this case.
I think this rule has a pretty solid use-case and could be useful to many people but i'm very eager to listen to other people opinion on this, coming from different angles.

AndreaPontrandolfo:
Hi, appreciate the review!
I wrote this rule (and actually included this in my own Eslint plugin) because this mistake is pretty frequent at my company. I usually happen to work with people of all different skill grades and not everyone knows about every single nuance of React.
Eslint isn't only useful to standardize our codebases, it can also prevent mistakes like this and even help juniors learn something with minimal effort, like in this case.
I think this rule has a pretty solid use-case and could be useful to many people but i'm very eager to listen to other people opinion on this, coming from different angles.

ljharb:
That totally makes sense - but it seems like one way to solve that is to force usage of a sanctioned custom component and ban via eslint any usage of `window.addEventListener` or `addEventListener` outside of that component :-)

AndreaPontrandolfo:
> That totally makes sense - but it seems like one way to solve that is to force usage of a sanctioned custom component and ban via eslint any usage of `window.addEventListener` or `addEventListener` outside of that component :-)

That's also an idea, but would be a vastly more complex rule than this, i think. This one is relatively simple and straightforward, and doesn't require any particular setup on part of the developers.

AndreaPontrandolfo:
Hi, @ljharb, i'm not really sure why the CI is failing. What should i do?

ljharb:
The documentation checker seems to be erroring; there's a generation script to run.

docs/rules/ensure-matching-remove-event-listener.md
```diff
@@ -0,0 +1,47 @@
...
+    window.addEventListener("keydown", handleUserKeyPress);
+    return () => {
+        window.removeEventListener("keydown", handleUserKeyPress);
+    };
+}, []);
+```
+
+## When Not To Use It
+
+You don't need this rules if you want to allow developers to not remove eventListeners added in the DOM.
```
ljharb, line 47: what about `element.addEventListener`, added with a ref?

lib/rules/ensure-matching-remove-event-listener.js
```diff
@@ -0,0 +1,117 @@
...
+  requiredCleanup: 'Missing a cleanup function for the addEventListener.',
+  requiredRemoveEventListener: 'Missing a matching removeEventListener.',
+};
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'Enforces that every addEventListener has a matching removeEventListener in the return statement of the same useEffect block',
```
ljharb, line 26:
```suggestion
      description: 'Enforces that every addEventListener has a matching removeEventListener in the return statement of the same useEffect block',
```

lib/rules/ensure-matching-remove-event-listener.js
```diff
@@ -0,0 +1,117 @@
...
+              report(context, messages, messageId, {
+                node,
+              });
+            } else {
+              const messageId = 'requiredRemoveEventListener';
+
+              report(context, messages, messageId, {
+                node,
+              });
+            }
```
ljharb, line 111:
```suggestion
            const messageID = hasReturnStatement ? 'requiredRemoveEventListener' : 'requiredCleanup';
            report(context, messages[messageID], messageID, {
              node,
            });
```

ljharb (changes requested):

lib/rules/ensure-matching-remove-event-listener.js
AndreaPontrandolfo: Resolved with [e685c66](https://github.com/jsx-eslint/eslint-plugin-react/pull/3442/commits/e685c669864a29af3bdc0e9f955f9726a4498e23)

lib/rules/ensure-matching-remove-event-listener.js
AndreaPontrandolfo: Resolved with [f017a31](https://github.com/jsx-eslint/eslint-plugin-react/pull/3442/commits/f017a31ccee4949d6eb3b0101507e663b50bbb59)

docs/rules/ensure-matching-remove-event-listener.md
AndreaPontrandolfo: I'm not sure what you mean, can you expand?

ljharb: I mean that `addEventListener` isn't just on window, it's also on every DOM element, and the same reasoning you've used for "window" would apply to element-based listeners as well.

PR #3444, opened 2022-09-29
[Docs] `no-unstable-nested-components`: Warn about memoized, nested components

eps1lon:
The issue this rule catches is a very common beginner issue that requires understanding of the reconciler. Ideally this issue wouldn't exist but until then let's make sure you start out with a linter that catches these issues.

I adjusted the docs to no longer recommend memoizing nested component and document migration strategies.

/cc @gaearon

### DIFF
--- a/docs/rules/no-unstable-nested-components.md
+++ b/docs/rules/no-unstable-nested-components.md
@@ -2,9 +2,9 @@
 
 💼 This rule is enabled in the following [configs](https://github.com/jsx-eslint/eslint-plugin-react#shareable-configurations): `all`.
 
-Creating components inside components without memoization leads to unstable components. The nested component and all its children are recreated during each re-render. Given stateful children of the nested component will lose their state on each re-render.
+Creating components inside components (nested components) will cause React to throw away the state of those nested components on each re-render of their parent.
 
-React reconciliation performs element type comparison with [reference equality](https://github.com/facebook/react/blob/v16.13.1/packages/react-reconciler/src/ReactChildFiber.js#L407). The reference to the same element changes on each re-render when defining components inside the render block. This leads to complete recreation of the current node and all its children. As a result the virtual DOM has to do extra unnecessary work and [possible bugs are introduced](https://codepen.io/ariperkkio/pen/vYLodLB).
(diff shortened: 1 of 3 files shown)


ljharb (changes requested):
Adding anything to a shared config is a semver-major change, so it's unlikely to land any time soon. I'd love to land the docs updates separately tho!

eps1lon:
> Adding anything to a shared config is a semver-major change, so it's unlikely to land any time soon. I'd love to land the docs updates separately tho!

Gotcha. Let's discuss the docs changes then and I'll file the semver-major separately.

gaearon:
Should we just add this logic to rules of Hooks?

ljharb:
@gaearon if it's hook-related then it definitely makes sense.

docs/rules/no-unstable-nested-components.md
```diff
@@ -2,9 +2,9 @@
...
+Creating components inside components (nested components) will cause React to throw away the state of those nested components on each re-render of their parent.
 
-React reconciliation performs element type comparison with [reference equality](https://github.com/facebook/react/blob/v16.13.1/packages/react-reconciler/src/ReactChildFiber.js#L407). The reference to the same element changes on each re-render when defining components inside the render block. This leads to complete recreation of the current node and all its children. As a result the virtual DOM has to do extra unnecessary work and [possible bugs are introduced](https://codepen.io/ariperkkio/pen/vYLodLB).
+React reconciliation performs element type comparison with [reference equality](https://reactjs.org/docs/reconciliation.html#elements-of-different-types). The reference to the same element changes on each re-render when defining components inside the render block. This leads to complete recreation of the current node and all its children. As a result the virtual DOM has to do extra unnecessary work and [possible bugs are introduced](https://codepen.io/ariperkkio/pen/vYLodLB).
```
AriPerkkio, line 7: > `[possible bugs are introduced](https://codepen.io/ariperkkio/pen/vYLodLB)`

I'll update this demo not to include `useCallback` example at some point.

AriPerkkio:
If I remember right this is the line that allows memoized nested components:

https://github.com/jsx-eslint/eslint-plugin-react/blob/2575bba0ae30b1b0b4013539be3b0dc4eb8114eb/lib/rules/no-unstable-nested-components.js#L435-L436

This was clearly a mistake in the rule's initial design. `useCallback` etc. don't guarantee the component is actually created just once. And moving the component out of the render block should always be the first step to recommend.

ljharb (approved):

PR #3448, opened 2022-10-03
[Fix] `jsx-no-constructed-context-values`: fix false positive for usage in non-components

### LINKED ISSUE #3295: jsx-no-constructed-context-values: class instance false positive
Consider the following code:

```
const app = new App(); // The 'app' new expression (at line 13) passed as the value prop to the Context provider (at line 22) changes every render. To fix this consider wrapping it in a useMemo hook.eslint[react/jsx-no-constructed-context-values](https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules/jsx-no-constructed-context-values.md)

const root = ReactDOM.createRoot(document.getElementById('root'));

root.render(
  <React.StrictMode>
    <AppContext.Provider value={app}>
      <AppView />
    </AppContext.Provider>
  </React.StrictMode>,
);
```

The rule docs explain:
> React Context, and all its child nodes and Consumers are rerendered whenever the value prop changes. Because each Javascript object carries its own identity, things like object expressions ({foo: 'bar'}) or function expressions get a new identity on every run through the component. This makes the context think it has gotten a new object and can cause needless rerenders and unintended consequences.

The rule description only mentions object literals and function expressions and it does not seem to me like React would re-initialize previously initialized class instance on a re-render.
`

golopot:
Fixes #3295

### DIFF
--- a/tests/lib/rules/jsx-no-constructed-context-values.js
+++ b/tests/lib/rules/jsx-no-constructed-context-values.js
@@ -31,13 +31,13 @@ const ruleTester = new RuleTester({ parserOptions });
 ruleTester.run('react-no-constructed-context-values', rule, {
   valid: parsers.all([
     {
-      code: '<Context.Provider value={props}></Context.Provider>',
+      code: 'const Component = () => <Context.Provider value={props}></Context.Provider>',
     },
     {
-      code: '<Context.Provider value={100}></Context.Provider>',
+      code: 'const Component = () => <Context.Provider value={100}></Context.Provider>',
     },
     {
-      code: '<Context.Provider value="Some string"></Context.Provider>',
+      code: 'const Component = () => <Context.Provider value="Some string"></Context.Provider>',
     },
     {
       code: 'function Component() { const foo = useMemo(() => { return {} }, []); return (<Context.Provider value={foo}></Context.Provider>)}',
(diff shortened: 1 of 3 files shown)


tests/lib/rules/jsx-no-constructed-context-values.js
```diff
@@ -31,13 +31,13 @@ const ruleTester = new RuleTester({ parserOptions });
...
-      code: '<Context.Provider value={props}></Context.Provider>',
+      code: 'const Component = () => <Context.Provider value={props}></Context.Provider>',
     },
     {
-      code: '<Context.Provider value={100}></Context.Provider>',
+      code: 'const Component = () => <Context.Provider value={100}></Context.Provider>',
     },
     {
-      code: '<Context.Provider value="Some string"></Context.Provider>',
+      code: 'const Component = () => <Context.Provider value="Some string"></Context.Provider>',
```
ljharb, line 40: presumably all of these existing test cases would still pass with this change?

ljharb (approved):

tests/lib/rules/jsx-no-constructed-context-values.js
golopot: Yes, I change it because otherwise these test cases would pass because they are not in components, not because other any reason which is supposed to be tested.

Issue #3443, opened 2022-09-27
Allow nested destructuring with `react/hook-use-state`


philer-jambit:
State is not always flat and sometimes it can make sense to destructure it directly. Example:

```js
const [{foo, bar, baz}, setFooBarBaz] = useState({foo: "omg", bar: "wtf", baz: "bbq"})
const [[index, value], setValueWithIndex] = useState([0, "hello"])
```

Note that in these cases the setter name can not be symmetrical to the destructured variables.

It would be nice if `react/hook-use-state` permitted this use case (maybe optional?).

Either way, thanks for the great work! :+1:

ljharb:
That definitely seems like a reasonable option to add ("allowDestructuredState"?)

PR #3449, opened 2022-10-04
[New] `hook-use-state`: add `allowDestructuredState` option

metreniuk:
This PR fixes https://github.com/jsx-eslint/eslint-plugin-react/issues/3443 by adding a new option to the rule.
I want to be sure that I'm on the right path, so I could add fixes later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


ljharb (approved):
When viewing the diff ignoring whitespace changes, it looks great - nice and simple. Thanks!

ljharb (approved):
I'm going to hold off on landing semver-minor things until after I've cut the next patch release, but hopefully that won't be long.

PR #3450, opened 2022-10-05
[Perf] improve performance for rules using isCreateElement

golopot:
This change makes `eslint .` run about 4% faster by optimizing `isCreateElement`.

Screenshot from 2022-10-06 02-13-27

Screenshot from 2022-10-06 02-15-39


ljharb (approved):
Makes sense.

PR #3451, opened 2022-10-06
[Perf] improve performance by optimizing getId

golopot:
Screenshot from 2022-10-06 17-16-27

Screenshot from 2022-10-06 17-17-33

Profiler shows it's about 2% faster overall.


ljharb (approved):
We can do this in this case since it's inside an abstraction, but this kind of microoptimization is imo crossing the line to where the readability hit isn't worth the performance gain.

golopot:
Good point.

PR #3452, opened 2022-10-07
[Fix] sort props test cases

ROSSROSALES:
All tests cases currently pass.
Fixed typos in test cases. Concatenated all invalid tests.

### DIFF
--- a/lib/rules/sort-prop-types.js
+++ b/lib/rules/sort-prop-types.js
@@ -8,7 +8,7 @@ const variableUtil = require('../util/variable');
 const propsUtil = require('../util/props');
 const docsUrl = require('../util/docsUrl');
 const propWrapperUtil = require('../util/propWrapper');
-// const propTypesSortUtil = require('../util/propTypesSort');
+const propTypesSortUtil = require('../util/propTypesSort');
 const report = require('../util/report');
 
 // ------------------------------------------------------------------------------
@@ -29,7 +29,7 @@ module.exports = {
       recommended: false,
       url: docsUrl('sort-prop-types'),
     },
-    // fixable: 'code',
+    fixable: 'code',
 
     messages,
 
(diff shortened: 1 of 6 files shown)


lib/rules/sort-prop-types.js
```diff
@@ -106,17 +106,17 @@ module.exports = {
...
+         return propTypesSortUtil.fixPropTypesSort(
+           fixer,
+           context,
+           declarations,
+           ignoreCase,
+           requiredFirst,
+           callbacksLast,
+           sortShapeProp
+         );
+       }
```
ljharb, line 119: ahhh i think we had this commented out specifically because of the comment issues; can we try just commenting out the `output` part of the test cases to avoid this PR changing the rule at all?

ROSSROSALES: Sorry just to clarify, did you want just output of the test cases with the comment, to be commented out?
For example, it is currently like this below. Did you want the test case to only have the `output` to be commented out?

Without the `function fix(fixer)` you mentioned above, the rule will only return a warning without an option for `--fix`

```
    //{
    //  code: `
    //    var First = createReactClass({
    //      propTypes: {
    //        /* z */
    //        z: PropTypes.string,
    //        /* a */
    //        a: PropTypes.any
    //      },
    //      render: function() {
    //        return <div />;
    //      }
    //    });
    //  `,
    //  // Disabled test for comments -- fails
    //  // output: `
    //  //   var First = createReactClass({
    //  //     propTypes: {
    //  //       /* a */
    //  //       a: PropTypes.any,
    //  //       /* z */
    //  //       z: PropTypes.string
    //  //     },
    //  //     render: function() {
    //  //       return <div />;
    //  //     }
    //  //   });
    //  // `,
    //  errors: [
    //    {
    //      messageId: 'propsNotSorted',
    //      line: 7,
    //      column: 13,
    //      type: 'Property',
    //    },
    //  ],
    //},
    //  //       a: PropTypes.any,
    //  //       /* z */
    //  //       z: PropTypes.string
    //  //     },
    //  //     render: function() {
    //  //       return <div />;
    //  //     }
    //  //   });
    //  // `,
    //  errors: [
    //    {
    //      messageId: 'propsNotSorted',
    //      line: 7,
    //      column: 13,
    //      type: 'Property',
    //    },
    //  ],
    //},

ljharb: i think we should be able to just remove one level of comments - ie, leave "output" commented but uncomment the rest of the test case.

that way we wouldn't need to uncomment the fixer yet.

ljharb (approved):

PR #3455, opened 2022-10-10
`no-unknown-property`: allow `allowFullScreen` on `iframe`

almeidx:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-allowfullscreen


ljharb (approved):
Thanks!

ljharb:
I'll be backporting this into a patch release of v7.31.10 shortly, before later publishing a v7.31.11 (which will also include this fix).

ljharb:
Published.

PR #3459, opened 2022-10-11
[Perf] component detection: improve performance by avoiding traversing parents unnecessarily

golopot:
This change makes an `eslint .` with `plugin:react/all` enabled run about 13% faster.

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -109,7 +109,7 @@ class Components {
   set(node, props) {
     const list = Lists.get(this);
     let component = list[getId(node)];
-    while (!component) {
+    while (!component || component.confidence < 1) {
       node = node.parent;
       if (!node) {
         return;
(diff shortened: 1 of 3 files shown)


tests/lib/rules/destructuring-assignment.js
```diff
@@ -459,7 +459,7 @@ ruleTester.run('destructuring-assignment', rule, {
     },
     {
       code: `
-        var Hello = React.createClass({
+        var Hello = createReactClass({
           render: function() {
             return <Text>{this.props.foo}</Text>;
           }
```
golopot, line 465: This should not be reported because `React.createClass` is **not** the correct pragma here. So there should be no component detected thus no report.

lib/util/Components.js
```diff
@@ -871,7 +877,7 @@ function componentRule(rule, context) {
         return;
       }
 
-      node = utils.getParentComponent();
```
golopot, line 874: No need to find parent components here.

ljharb (approved):
Seems legit, thanks

PR #3460, opened 2022-10-12
[actions]: Add npm publish workflow

varunsh-coder:
This PR adds a GitHub Actions workflow to publish to npm registry using one-time password (OTP). 

- Runs on workflow dispatch and expects a tag
- It checks that the tag follows semver pattern and version has not been published before to npm registry
- Checks that required checks have run and passed when the tag was pushed
- Uses `step-security/wait-for-secrets` so OTP can be sent to the workflow

It requires setting two secrets:
1. `NODE_AUTH_TOKEN` - this should be a Publish token for npm registry. The package should have `require two-factor authentication to publish` checked in the package settings. 
2. `SLACK_WEBHOOK_URL` - this is to get notification on Slack when OTP needs to be entered. If this notification is not needed, I can remove this part from the workflow. 

I have also added `harden-runner` to the publish job. 

Examples:
1. Publish to npm registry: https://github.com/harden-runner-canary/eslint-plugin-react/actions/runs/3237008675/jobs/5303545348#step:8:1
2. Tag not in semver pattern (failed as expected): https://github.com/harden-runner-canary/eslint-plugin-react/actions/runs/3235307416/jobs/5299550718#step:5:22
3. Version already published (failed as expected): https://github.com/harden-runner-canary/eslint-plugin-react/actions/runs/3236736891/jobs/5302945983#step:5:22

@ljharb, please let me know if you have any feedback.  

Signed-off-by: Varun Sharma <varunsh@stepsecurity.io>

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,0 +1,125 @@
+name: Publish Package to npm
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Tag to publish"
+        required: true
+
+permissions:
+  contents: read
+
+jobs:
+  check-version:
+    runs-on: ubuntu-latest
+    outputs:
+      is-new-version: ${{ steps.cpv.outputs.is-new-version }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.inputs.tag }}
+
+      - name: Validate semver pattern
+        run: npx semver ${{ inputs.tag }}
+
+      - name: Check package version
+        id: cpv
+        uses: PostHog/check-package-version@v2
+
+      - name: Validate package version
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const isNewVersion = `${{ steps.cpv.outputs.is-new-version }}`;
+            if (isNewVersion === 'true') {
+                console.log(`Version ${context.payload.inputs.tag} has not been published yet`);
+            } else {
+                core.setFailed(`Version ${context.payload.inputs.tag} is already published`);
+            }
+
+  check-status:
+    needs: check-version
+    if: needs.check-version.outputs.is-new-version == 'true'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Verify checks passed
+        uses: actions/github-script@v6
+        with:
+          result-encoding: string
+          retries: 3
+          script: |
+            const ref = context.payload.inputs.tag;
+
+            console.log(`Checking status checks for ${ref}`);
+
+            const { owner, repo } = context.repo;
+            const { default_branch: branch } = context.payload.repository;
+
+            const branch = github.rest.repos.getBranch({ owner, repo, branch });
+
+            const checkSuites = await github.rest.checks.listSuitesForRef({ owner, repo, ref });
+
+            if (checkSuites.some(({ status }) => 'completed')) {
+              core.setFailed(`Some workflows for ${context.payload.inputs.tag} are still in-progress`);
+            }
+
+            const { data: { check_runs: checkRuns } } = await Promise.all(
+              (await branch).data.protection.required_status_checks.checks.map(({ context }) => (
+                github.rest.checks.listForRef({
+                  owner,
+                  repo,
+                  ref,
+                  check_name: context
+                })
+              )
+            )
+
+            checkRuns.forEach(({ name, status, conclusion }) => {
+              if (status !== 'completed' || conclusion !== 'success') {
+                console.log(`${name} check failed`);
+                core.setFailed(`Required status check ${name} did not succeed`);
+              }
+              console.log(`${name} check passed`);
+            });
+
+  publish:
+    needs: [check-status]
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - uses: step-security/harden-runner@v1
+        with:
+          egress-policy: block
+          allowed-endpoints: >
+            github.com:443
+            nodejs.org:443
+            prod.api.stepsecurity.io:443
+            registry.npmjs.org:443
+
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.inputs.tag }}
+
+      - uses: ljharb/actions/node/install@main
+        name: "nvm install lts/* && npm install"
+        with:
+          node-version: "lts/*"
+        env:
+          NPM_CONFIG_LEGACY_PEER_DEPS: true
+
+      - run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NODE_AUTH_TOKEN }}" >> .npmrc
+
+      - run: npm publish --dry-run
+
+      - uses: step-security/wait-for-secrets@v1
+        id: wait-for-secrets
+        with:
+          slack-webhook-url: ${{ secrets.SLACK_WEBHOOK_URL }}
+          secrets: |
+            OTP:
+              name: 'OTP to publish package'
+              description: 'OTP from authenticator app'
+
+      - run: npm publish --access public --otp ${{ steps.wait-for-secrets.outputs.OTP }}


.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,135 @@
...
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.inputs.tag }}
+
+      - name: Match semver pattern
+        uses: actions-ecosystem/action-regex-match@v2
+        id: regex-match
+        with:
+          text: ${{ github.event.inputs.tag }}
+          regex: '^v((([0-9]+)\.([0-9]+)\.([0-9]+)(?:-([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?)(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?)$'
```
ljharb, line 27: i'd prefer to validate with the npm semver module rather than a regex, but that doesn't need to be a blocker

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,135 @@
...
+      - name: Validate tag
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const match = `${{ steps.regex-match.outputs.match }}`;
+            if(match === '') {
+              core.setFailed(`Tag ${context.payload.inputs.tag} does not match semver pattern`);
+            } else {
+              const isNewVersion = `${{ steps.cpv.outputs.is-new-version }}`;
+              if(isNewVersion === 'true') {
```
ljharb, line 42:
```suggestion
            if (match === '') {
              core.setFailed(`Tag ${context.payload.inputs.tag} does not match semver pattern`);
            } else {
              const isNewVersion = `${{ steps.cpv.outputs.is-new-version }}`;
              if (isNewVersion === 'true') {
```

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,135 @@
...
+
+              const check_runs = await github.rest.checks.listForRef({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                ref: context.payload.inputs.tag,
+                check_name: requiredCheck.context
+              });
+
+              for (const check_run of check_runs.data.check_runs) {
+                if (!(check_run.status === 'completed' && check_run.conclusion === 'success')){
```
ljharb, line 89:
```suggestion
                if (check_run.status !== 'completed' || check_run.conclusion !== 'success') {
```

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,135 @@
...
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - uses: step-security/harden-runner@v1
+        with:
+          egress-policy: block
+          allowed-endpoints: >
+            github.com:443
+            hooks.slack.com:443
```
ljharb, line 109: can this be conditionally included only if the SLACK_WEBHOOK_URL secret exists?

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,135 @@
...
+            registry.npmjs.org:443
+
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.inputs.tag }}
+
+      - uses: actions/setup-node@v3
+        with:
+          node-version: "16.x"
+          registry-url: "https://registry.npmjs.org"
```
ljharb, line 120: i'd prefer to use my `ljharb/node/install` action for this, since that's what i'm using everywhere else

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,135 @@
...
+        with:
+          slack-webhook-url: ${{ secrets.SLACK_WEBHOOK_URL }}
+          secrets: |
+            OTP:
+              name: 'OTP to publish package'
+              description: 'OTP from authenticator app'
+
+      - run: npm publish --access public --otp ${{ steps.wait-for-secrets.outputs.OTP }}
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NODE_AUTH_TOKEN }}
```
ljharb, line 135: could we do `npm publish --dry-run` before wait-for-secrets, so we know if the publish will otherwise succeed before asking for the OTP?

ljharb (commented):
Seems great! Just a few comments before we merge it and give it a shot.

.github/workflows/npm-publish.yml
varunsh-coder: Updated to use semver module.

.github/workflows/npm-publish.yml
varunsh-coder: Fixed the formatting issues.

.github/workflows/npm-publish.yml
varunsh-coder: Updated the comparison logic

.github/workflows/npm-publish.yml
varunsh-coder: I cannot think of a way to only add `hooks.slack.com:443` if `SLACK_WEBHOOK_URL` exists. Let me know if you have any ideas.

.github/workflows/npm-publish.yml
varunsh-coder: Made the change. `actions/setup-node` was setting up the `.npmrc` file. So, I had to add a separate step to do that when using `ljharb/node/install`.

.github/workflows/npm-publish.yml
varunsh-coder: done.

varunsh-coder:
Thanks @ljharb for the review! Updated the workflow. 

Here is the latest run: 
https://github.com/harden-runner-canary/eslint-plugin-react/actions/runs/3251993055/jobs/5337708215#step:9:156

Example of when version did not match semver pattern using semver module: 
https://github.com/harden-runner-canary/eslint-plugin-react/actions/runs/3251600036

.github/workflows/npm-publish.yml
```diff
@@ -19,31 +19,30 @@ jobs:
...
-        id: regex-match
-        with:
-          text: ${{ github.event.inputs.tag }}
-          regex: '^v((([0-9]+)\.([0-9]+)\.([0-9]+)(?:-([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?)(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?)$'
-
       - name: Check package version
         id: cpv
         uses: PostHog/check-package-version@v2
 
+      - run: npm install semver
```
ljharb, line 26: i think this can just be `npx semver $version`, and it'll fail if it's not valid? (and the proper version number will be echoed, so that output could be used in the next step)

.github/workflows/npm-publish.yml
ljharb: yeah that might be tricky. maybe just comment it out for now?
```suggestion
            # hooks.slack.com:443
```

.github/workflows/npm-publish.yml
```diff
@@ -19,31 +19,30 @@ jobs:
...
         with:
           script: |
-            const match = `${{ steps.regex-match.outputs.match }}`;
-            if(match === '') {
-              core.setFailed(`Tag ${context.payload.inputs.tag} does not match semver pattern`);
-            } else {
+            const semver = require('semver')
+
+            const tag = `${context.payload.inputs.tag}`;
+            const version = tag.substring(1);
```
ljharb, line 35: `npx semver v1.2.3` also trims off the `v` for you

ljharb (commented):
Thanks, this seems pretty good :-)

.github/workflows/npm-publish.yml
varunsh-coder: Updated to use `npx semver ${{ inputs.tag }}`. Since this fails if not valid, removed few lines from a later step that were doing that explicitly.

.github/workflows/npm-publish.yml
varunsh-coder: `allowed-endpoints` in harden-runner does not support comments. I have created an [issue](https://github.com/step-security/harden-runner/issues/199) to track it. Removed `hooks.slack.com:443` for now.

ljharb (approved):

.github/workflows/npm-publish.yml
ljharb: That's great. Most of my projects already have one, but it's better to be generic.

PR #3464, opened 2022-10-15
fix: `jsx-no-target-blank` support for conditionals

### LINKED ISSUE #3462: react/jsx-no-target-blank not working as expected on conditionals
### Code

```jsx
const openInNewTab = true; // could be true or false

  return (
    <a
      href="https://google.com"
      target={openInNewTab ? "_blank" : "self"}
      rel={openInNewTab ? "noreferrer" : undefined}
    >
      Open Google
    </a>
  );
}
```

### Currently what's happening

If I pass `target='_blank'` based on some condition as shown in the above snippet, and pass `rel='noreferrer'` on same condition. It is still showing me warning/error as per the config. 

### What should happen?

Ideally if the condition is same for both the props it should recognise it and should not give us warning/error.

Please let me know if I am missing something? 

### Reference:

Sample Codesandbox - https://codesandbox.io/s/mystifying-lake-l0420z?file=/src/App.js

akulsr0:
Fixes #3462 

I have added the fix for support of conditionals in `jsx-no-target-blank` rule. 
But not sure about one test case and it's failing. Would need some help on that.

```html
<a href={href} target={isExternal ? "_blank" : undefined} rel={isExternal ? "noopener noreferrer" : undefined} />
```

As per the test case this is expected to be invalid. I am not sure if this should be invalid or not?


lib/rules/jsx-no-target-blank.js
```diff
@@ -75,24 +75,32 @@ function getStringFromValue(value) {
...
-          ? [expr.consequent.value, expr.alternate.value]
-          : expr.value
-      );
+      if (expr && (expr.type === 'ConditionalExpression')) {
+        const relValues = [expr.consequent.value, expr.alternate.value];
+        if (targetValue.type === 'JSXExpressionContainer' && targetValue.expression && targetValue.expression.type === 'ConditionalExpression') {
+          const targetTestCond = targetValue.expression.test.name;
+          const relTestCond = value.expression.test.name;
+          const targetBlankIndex = [targetValue.expression.consequent.value, targetValue.expression.alternate.value].indexOf('_blank');
+          if (targetTestCond === relTestCond) return relValues[targetBlankIndex];
```
ljharb, line 84:
```suggestion
          if (targetTestCond === relTestCond) {
            return relValues[targetBlankIndex];
          }
```

ljharb (approved):
Seems pretty good, thanks

lib/rules/jsx-no-target-blank.js
```diff
@@ -75,24 +75,32 @@ function getStringFromValue(value) {
         return value.expression.quasis[0].value.cooked;
       }
       const expr = value.expression;
-      return expr && (
-        expr.type === 'ConditionalExpression'
-          ? [expr.consequent.value, expr.alternate.value]
-          : expr.value
-      );
+      if (expr && (expr.type === 'ConditionalExpression')) {
```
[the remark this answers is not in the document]
akulsr0: Done.

lib/rules/jsx-no-target-blank.js
akulsr0: Done.

akulsr0:
I have moved the following testcase (which was failing) to valid, since it is a valid scenario.

```html
<a href={href} target={isExternal ? "_blank" : undefined} rel={isExternal ? "noopener noreferrer" : undefined} />
```

Reference: https://stackoverflow.com/questions/50709625/link-with-target-blank-and-rel-noopener-noreferrer-still-vulnerable

ljharb (approved):

PR #3471, opened 2022-10-24
[Fix] `sort-prop-types`: avoid crash when autofixing

ROSSROSALES:
Fixes https://github.com/jsx-eslint/eslint-plugin-react/issues/3470

Now rule respects multiple comments before and after prop.
Fixes issue of no comments before and after prop not being considered during parsing.


ljharb:
@ROSSROSALES i think this needs to be rebased on _latest_, and i'm not sure of the best way to do it. Can you give it a try?

ROSSROSALES:
> 

I'll give it a try to rebase on latest

ljharb (approved):
Seems pretty good if tests pass.

PR #3483, opened 2022-11-08
[Improvement] Issue Template Proposal

ROSSROSALES:
## Issue Template for Bugs

This Proposal is to improve the overall quality of how issues are presented and solved.
Issues presented with a standardized method should:
- improve understanding of problem
  - description and example
- reduce number of unnecessary clarifying questions to ask
  - "why do this?" or "can you show me an example?" or "Have you read X?"
- makes users creating an issue think if it's an issue that has not already been solved, or if this problem has been asked / answered before
  - Users presenting an issue can rethink their issue and possibly solve their problem by going through a checklist/process of submitting an issue through a template.

### Discussion/Thoughts
When creating this template, it is important for the user to have incentive to submit an issue.
With that in mind, we want to make sure the template:

- [x] not overwhelming with too much info to fill in
- [x] is simple to understand
- [x] taken inspiration from other used issue templates
- [x] enough information is filled so that contributors aiming to solve issue would have to ask minimal questions

### Goal / Conclusion
Issues that are presented by users are issues that users have tried to debug before submitting it.
First time / New contributors / Veteran contributors have a better background to tackle the issue.
Improve community contribution and participation through reducing barrier to entry for contributors and reducing maintainer time for minor matters (ie. asking redundant questions across issues to users in different time zones, waiting days for a reply)

There is always room for improvement, please let me know if I am missing any information that may be required or removed.
Thank you!

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,81 @@
+name: Bug Report
+description: File a bug report
+title: "[Bug]: "
+labels: ["bug", "triage"]
+assignees: []
+body:
+  - type: checkboxes
+    attributes:
+      label: Is there an existing issue for this?
+      description: Please search to see if an issue already exists for the bug you encountered.
+      options:
+      - label: I have searched the existing issues and my issue is unique
+        required: true
+      - label: My issue appears in the command-line and not only in the text editor
+        required: true
+  - type: textarea
+    id: Code
+    attributes:
+      label: Description Overview
+      description: A clear and concise description of bug w/ examples
+      placeholder: |
+        Brief description here
+
+        Show example of your code (as text format), add images/videos/gifs to help explain example
+        and/or Link of repo to where issue is occurring
+
+        What is happening? / What is the error?
+
+        What command(s) did you run to reproduce issue?
+      value: |
+        <!--Brief description-->
+
+
+        <!--Show example of your code (as text format), add images/videos/gifs to help explain example-->
+        <!--and/or Link of repo to where issue is occurring-->
+
+
+        <!--What is happening? / What is the error?-->
+
+
+        <!--What command(s) did you run to reproduce issue?-->
+
+    validations:
+      required: true
+  - type: textarea
+    id: expected_behavior
+    attributes:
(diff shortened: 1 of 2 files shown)


ljharb:
This is great! Could we try using the form fields here? https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms

Also, let's try to get the user's eslint config as well in a textarea.

ROSSROSALES:
> This is great! Could we try using the form fields here? https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms
> 
> Also, let's try to get the user's eslint config as well in a textarea.

Thank you for the review Jordan! I will add the forms fields and add ESLint config so that we can have a better understanding of the user's settings.

ROSSROSALES:
New issues template to support Github's beta form fields.
Below is a screenshot of how issues template is displayed to issue contributor
Issues Template as YAML for eslint-react

ljharb:
Looks great! Can the field names be more of markdown and less of an identifier? like "node version" etc

ROSSROSALES:
Screenshot of updated changes
Issues Template as YAML for eslint-react 2 0

ljharb (approved):
Thanks!

ROSSROSALES:
Hey @ljharb, I tried create an issue to test the template. I noticed backticks on the labels.
For example:
https://github.com/jsx-eslint/eslint-plugin-react/blob/b4cab033e954e5785a092f2242639673832d080a/.github/ISSUE_TEMPLATE/bug_report.yml#L64

I think .yml file has trouble reading this which causes this error:
```
YAML syntax error: (): found character that cannot start any token while scanning for the next token at line 64 column 14.
```
Currently the when trying to create an issue, this issue template is inaccessible and defaults to the original blank issue format.
What were the reason for the backticks?

ljharb:
oh whoops. i added them last minute to try to make it format like backticked text does. i'll remove them.

ljharb:
Try now?

ROSSROSALES:
Looks good now! 
Thanks for the prompt reply! I appreciate it.

ljharb:
Thanks for catching it :-)

PR #3485, opened 2022-11-13
Update README.md

rovvi:
The existing code from the docs will result in `Cannot find module 'eslint-plugin-react/configs/recommended'`.  Updated the README page with a working example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,65 +217,42 @@ In the new config system, `plugin:` protocol(e.g. `plugin:react/recommended`) is
 As eslint does not automatically import the preset config (shareable config), you explicitly do it by yourself.
 
 ```js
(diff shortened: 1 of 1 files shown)


ljharb (changes requested):
Why are you just removing parts of the readme that were added in #3429?

ljharb:
Please note, just like *everywhere* on github, you should be reading docs for the tagged version you're using, not just what's on main/master. For npm packages, read the published readme, not the repo's readme.

ljharb:
That assumption is incorrect - on almost every repo on github - whenever there are unpublished changes, as there are here.

ljharb:
This seems fine to close, since the readme on master is correct. Happy to reopen after the next release if there's still something that needs changing.

Issue #3481, opened 2022-11-04
`react/jsx-key` Does not report errors when child is optional


tylercraft:
This does not warn on this use case:

```
data.map(x => somecheck && <Hello>{x}</Hello>);
```

PR #3490, opened 2022-11-15
[Fix] `jsx-key`: detect keys in logical expression and conditional expression

metreniuk:
Fixes https://github.com/jsx-eslint/eslint-plugin-react/issues/3481
I've also updated the docs to stay consistent with the comment from below `In the last example the key is being spread...`

### DIFF
--- a/lib/rules/jsx-key.js
+++ b/lib/rules/jsx-key.js
@@ -149,9 +149,19 @@ module.exports = {
      */
     function checkArrowFunctionWithJSX(node) {
       const isArrFn = node && node.type === 'ArrowFunctionExpression';
-
-      if (isArrFn && (node.body.type === 'JSXElement' || node.body.type === 'JSXFragment')) {
+      const shouldCheckNode = (n) => n && (n.type === 'JSXElement' || n.type === 'JSXFragment');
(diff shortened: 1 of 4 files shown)


lib/rules/jsx-key.js
```diff
@@ -149,10 +149,19 @@ module.exports = {
...
       }
+      if (node.body.type === 'LogicalExpression' && shouldCheckNode(node.body.right)) {
+        checkIteratorElement(node.body.right);
+      }
+      if (node.body.type === 'ConditionalExpression' && shouldCheckNode(node.body.consequent)) {
+        checkIteratorElement(node.body.consequent);
+      }
+      if (node.body.type === 'ConditionalExpression' && shouldCheckNode(node.body.alternate)) {
+        checkIteratorElement(node.body.alternate);
+      }
```
ljharb, line 164:
```suggestion
      if (node.body.type === 'ConditionalExpression') {
        if (shouldCheckNode(node.body.consequent)) {
          checkIteratorElement(node.body.consequent);
        }
        if (shouldCheckNode(node.body.alternate)) {
          checkIteratorElement(node.body.alternate);
        }
      }
```

ljharb (approved):
This looks great!

Issue #2926, opened 2021-02-17
'jsx-newline' add ignore options


RaddishIoW:
I have the `jsx-newline` rule turned on, but it causes problems with any `{/* eslint-disable-next-line <rulename> */}` comments in my code.

If I have the following:
```
{/* eslint-disable-next-line react/forbid-component-props */}
<Button popoverOpen='#settings-popover' style={{ width: 'fit-content' }}>
  <Icon f7='gear' />
</Button>
```
`jsx-newline` marks the `<Button ...>` line as needing a newline before, causing the `ignore-next-line` comment to stop working.

Would it be possible to add options to the `jsx-newline` rule, to allow ignoring?
e.g.
```
"react/jsx-newline": {
    2,
    { "ignore": [ '{/* */}'] }
}
```

ljharb:
I'd say an `ignoreComments` option would be reasonable.

Sreejit7:
is this not added to the plugin yet? we have a similar problem in our repo, and looking for a fix 👀

ljharb:
@Sreejit7 the tests pass, so it should already be fixed. If you're fully updated and still seeing problems, please file a new issue.

jacob-orbiit:
@ljharb was ignore comments added to options? If so, it's not documented: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-newline.md#rule-options

ljharb:
@jacob-orbiit looks like it's not. want to make a PR to document it?

jacob-orbiit:
Actually, I just tried to use it, and eslint throws that it's not recognised.

ljharb:
@jacob-orbiit ahhh no, we didn't end up adding the option, because it wasn't needed - we just fixed the detection.

PR #3493, opened 2022-11-17
[Fix] `jsx-newline`: No newline between comments and jsx elements

### LINKED ISSUE #3478: `jsx-newline` puts line breaks between comments and new JSX elements
I have the `jsx-newline` rule turned on, which when run with the `--fix` option, puts line breaks between any comments (including eslint-disable comments) and JSX elements, thus undoing the whole point of using comments. 

Let's say I have something like this:

```jsx
<Element as="header" marginBottom="micro">
        {/* eslint-disable-next-line react/jsx-sort-props */}
        <Heading as="h6" align="right">
            Transaction value
        </Heading>
<Element>
```

The `jsx-newline` rule puts a line break between the eslint-disable comment and <Heading />, therefore causing the `eslint-disable-{some rule}` to stop working.

Can we handle putting line breaks between comments & new JSX elements?

I saw something similar in #2926 (which was closed earlier as completed) but that doesn't work properly as of `v7.31.10`

justmejulian:
Fixes #3478 and  #2926 and adds a couple more tests to show what was failing.

From what I can gather this should be the default behaviour. If not then then this can be hidden behind an option.

I was not sure about the behaviour of multiple Block comments before a Multiline Component. Decided the comments should stick together, based on sujan-s comment in #3478.

### DIFF
--- a/lib/rules/jsx-newline.js
+++ b/lib/rules/jsx-newline.js
@@ -72,6 +72,15 @@ module.exports = {
     const jsxElementParents = new Set();
     const sourceCode = context.getSourceCode();
 
+    function isBlockCommentInCurlyBraces(element) {
+      const elementRawValue = sourceCode.getText(element);
+      return /^\s*{\/\*/.test(elementRawValue);
+    }
+
+    function isNonBlockComment(element) {
+      return !isBlockCommentInCurlyBraces(element) && (element.type === 'JSXElement' || element.type === 'JSXExpressionContainer');
+    }
+
     return {
       'Program:exit'() {
         jsxElementParents.forEach((parent) => {
@@ -93,7 +102,15 @@ module.exports = {
               // Check adjacent sibling has the proper amount of newlines
               const isWithoutNewLine = !/\n\s*\n/.test(firstAdjacentSibling.value);
 
-              if (allowMultilines && (isMultilined(element) || isMultilined(secondAdjacentSibling))) {
+              if (isBlockCommentInCurlyBraces(element)) return;
+
+              if (
+                allowMultilines
+                && (
+                  isMultilined(element)
+                  || isMultilined(elements.slice(index + 2).find(isNonBlockComment))
+                )
+              ) {
                 if (!isWithoutNewLine) return;
 
                 const regex = /(\n)(?!.*\1)/g;
@@ -115,6 +132,7 @@ module.exports = {
               }
 
               if (isWithoutNewLine === prevent) return;
+
               const messageId = prevent
                 ? 'prevent'
                 : 'require';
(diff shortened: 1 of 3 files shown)


tests/lib/rules/jsx-newline.js
```diff
@@ -105,6 +105,43 @@ new RuleTester({ parserOptions }).run('jsx-newline', rule, {
...
+        <Button popoverOpen='#settings-popover' style={{ width: 'fit-content' }}>
+          {/* fake-eslint-disable-next-line should also work insinde a component */}
+          <Icon f7='gear' />
+        </Button>
+      `,
+    },
+    {
+      code: `
+        <Button popoverOpen='#settings-popover' style={{ width: 'fit-content' }}>
+          {/* should work insinde a component */}
```
ljharb, line 119:
```suggestion
          {/* fake-eslint-disable-next-line should also work inside a component */}
          <Icon f7='gear' />
        </Button>
      `,
    },
    {
      code: `
        <Button popoverOpen='#settings-popover' style={{ width: 'fit-content' }}>
          {/* should work inside a component */}
```

lib/rules/jsx-newline.js
```diff
@@ -72,6 +72,22 @@ module.exports = {
...
+    function findFirstNonBlockComment(index, elements) {
+      if (index >= elements.length) return;
+
+      const element = elements[index];
+      const isElementJSX = element.type === 'JSXElement' || element.type === 'JSXExpressionContainer';
+
+      if (!isBlockCommentInCurlyBraces(element) && isElementJSX) return element;
+
+      return findFirstNonBlockComment(index + 1, elements);
+    }
```
ljharb, line 89: is there a reason to use recursion here rather than `elements.find(predicate)`?

lib/rules/jsx-newline.js
```diff
@@ -93,7 +109,13 @@ module.exports = {
               // Check adjacent sibling has the proper amount of newlines
               const isWithoutNewLine = !/\n\s*\n/.test(firstAdjacentSibling.value);
 
-              if (allowMultilines && (isMultilined(element) || isMultilined(secondAdjacentSibling))) {
+              if (isBlockCommentInCurlyBraces(element)) return;
+
+              if (
+                allowMultilines
+                && (isMultilined(element)
+                  || isMultilined(findFirstNonBlockComment(index + 2, elements)))
```
ljharb, line 117:
```suggestion
                && (
                  isMultilined(element)
                  || isMultilined(findFirstNonBlockComment(index + 2, elements)
                )
              )
```

ljharb (changes requested):
Test cases look great!

tests/lib/rules/jsx-newline.js
justmejulian: oh ups. thanks!

lib/rules/jsx-newline.js
justmejulian: No didn't think of using find. Let me give it a try.

justmejulian: Added a version with `find` please let me no which version you prefer.

ljharb (approved):

PR #3494, opened 2022-11-18
[Fix] `no-unknown-property`: do not check `fbs` elements

brian-ogilvie:
Currently `<fbs>` nodes are tagged with this linter warning:

```
Unknown property 'desc' foundNuclide-ESLint[react/no-unknown-property](https://www.internalfb.com/intern/bunny?q%3Deslint%2Breact%2Fno-unknown-property)
```

`<fbt>` nodes are already exempt from this lint rule, and `<fbs>` should be treated the same. 

Internal WP discussion: https://fb.workplace.com/groups/eslintfeedback/permalink/6064426636924059/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,9 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 ### Fixed
 * configs: avoid legacy config system error ([#3461][] @ljharb)
 * [`sort-prop-types`]: restore autofixing ([#3452][] @ROSSROSALES)
+* [`no-unknown-property`]: do not check `fbs` elements ([#3494][] @brianogilvie)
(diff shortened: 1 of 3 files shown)


ljharb (changes requested):
oof, i had no idea there were two of these. is that an exhaustive list? (i can't view the workplacae discussion, obv)

This is good to go as soon as we can add a test using `fbs`.

brian-ogilvie:
Yeah, `fbs` has basically the same api as `fbt`, but it returns the pure string result instead of a node. This is useful for things that really need to be string for a18y purposes like the an aria-label, etc. 

I'll add a test

ljharb (approved):
Thanks! I'll try to get another patch out in the next day or three.

PR #3502, opened 2022-12-05
[Fix] `jsx-no-leaked-render`:  removing parentheses for conditionals

### LINKED ISSUE #3498: `jsx-no-leaked-render` causes code to be incorrect by removing parentheses
```
// maybeObject: object | undefined
// isFoo: boolean
{maybeObject && (isFoo ? <Aaa /> : <Bbb />)}

// autofixes to:
{!!maybeObject && isFoo ? <Aaa /> : <Bbb />}
```
This has an entirely different meaning. My original code was hiding everything if `maybeObject` was `undefined`. The new code will display `<Bbb />` instead.

To show how absurd this is, in other spot in my codebase, the `else` condition of the ternary was also `maybeObject`. This meant that my new code had a runtime error.

akulsr0:
Fixes #3498


ljharb (approved):
Assuming the test fails prior to your fix, and passes afterwards, this looks great, thanks!

PR #3506, opened 2022-12-15
[Fix] : Added global event listeners

### LINKED ISSUE #3505: [Bug]: `no-unknown-property` is missing some valid global properties (event handlers)

## Brief description
Several global event handler properties that are valid on all HTML elements are being limited to only a few by this rule.

See the "event handler" section in the [MDN documentation on global attributes], and the [HTML specification on global event handlers here](https://html.spec.whatwg.org/multipage/webappapis.html#event-handlers-on-elements,-document-objects,-and-window-objects).

**Show example of your code (as text format), add images/videos/gifs to help explain example
and/or Link of repo to where issue is occurring**

```lsx
<svg>
  <image onLoad={() => console.log("Loaded!")} />
</svg>
```

## What is happening? / What is the error?
In the above example, the error is `"Invalid property 'onLoad' found on tag 'image', but it is only allowed on: script img, link, picture, iframe, object"`.

Enabled the rule and the above code will show an error.

## Brief description
Any global event handler should be able to be defined for any HTML element. (Note that most of these are already in `@types/react`).

**Show example of code (as text format), add images/videos/gifs to help explain expected behavior**

This code should not show an error for defining `onLoad`:

```jsx
<svg>
  <image onLoad={() => console.log("Loaded!")} />
</svg>
```

7.31.11

### eslint version

8.28.0

### node version

16.14.2

axlemax:
Fixes #3505. 

Added some attributes that should be able to be used on any element, along with a way to exclude only specific elements for some properties (such as `onFocus` on a `<body>` element). 

For more info on these global events, see the "event handler" section in the [MDN documentation on global attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes), and the [HTML specification on global event handlers here](https://html.spec.whatwg.org/multipage/webappapis.html#event-handlers-on-elements,-document-objects,-and-window-objects).

### DIFF
--- a/lib/rules/no-unknown-property.js
+++ b/lib/rules/no-unknown-property.js
@@ -69,33 +69,10 @@ const ATTRIBUTE_TAGS_MAP = {
   align: ['applet', 'caption', 'col', 'colgroup', 'hr', 'iframe', 'img', 'table', 'tbody', 'td', 'tfoot', 'th', 'thead', 'tr'], // deprecated, but known
   valign: ['tr', 'td', 'th', 'thead', 'tbody', 'tfoot', 'colgroup', 'col'], // deprecated, but known
   noModule: ['script'],
-  // Media events allowed only on audio and video tags, see https://github.com/facebook/react/blob/256aefbea1449869620fb26f6ec695536ab453f5/CHANGELOG.md#notable-enhancements
-  onAbort: ['audio', 'video'],
-  onCancel: ['dialog'],
-  onCanPlay: ['audio', 'video'],
-  onCanPlayThrough: ['audio', 'video'],
-  onClose: ['dialog'],
-  onDurationChange: ['audio', 'video'],
-  onEmptied: ['audio', 'video'],
+  // Note: Most events can be added to any HTML element, even if not specified in the standard
+  // see https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes
   onEncrypted: ['audio', 'video'],
-  onEnded: ['audio', 'video'],
-  onError: ['audio', 'video', 'img', 'link', 'source', 'script', 'picture', 'iframe'],
-  onLoad: ['script', 'img', 'link', 'picture', 'iframe', 'object'],
-  onLoadedData: ['audio', 'video'],
-  onLoadedMetadata: ['audio', 'video'],
-  onLoadStart: ['audio', 'video'],
-  onPause: ['audio', 'video'],
-  onPlay: ['audio', 'video'],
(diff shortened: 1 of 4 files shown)


ljharb:
I don't think this is a viable approach. For example, `onLoad` on an element that doesn't load anything makes no sense - even if HTML allows it, it's 100% likely to be a bug.

A better change would be allowing `onLoad` on `image` SVG tags (not HTML tags tho).

lib/rules/no-unknown-property.js
```diff
@@ -69,33 +69,10 @@ const ATTRIBUTE_TAGS_MAP = {
   align: ['applet', 'caption', 'col', 'colgroup', 'hr', 'iframe', 'img', 'table', 'tbody', 'td', 'tfoot', 'th', 'thead', 'tr'], // deprecated, but known
   valign: ['tr', 'td', 'th', 'thead', 'tbody', 'tfoot', 'colgroup', 'col'], // deprecated, but known
   noModule: ['script'],
-  // Media events allowed only on audio and video tags, see https://github.com/facebook/react/blob/256aefbea1449869620fb26f6ec695536ab453f5/CHANGELOG.md#notable-enhancements
-  onAbort: ['audio', 'video'],
-  onCancel: ['dialog'],
-  onCanPlay: ['audio', 'video'],
```
ljharb, line 75: eg, `onCanPlay` isn't a defensible attribute on an element that doesn't play things

lib/rules/no-unknown-property.js
```diff
@@ -69,33 +69,10 @@ const ATTRIBUTE_TAGS_MAP = {
...
-  onCanPlayThrough: ['audio', 'video'],
-  onClose: ['dialog'],
-  onDurationChange: ['audio', 'video'],
-  onEmptied: ['audio', 'video'],
+  // Note: Most events can be added to any HTML element, even if not specified in the standard
+  // see https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes
```
ljharb, line 73: i think our standard should be higher than "whatever wacky things HTML allows" in a linter whose goal is to prevent bugs.

ljharb (changes requested):

lib/rules/no-unknown-property.js
axlemax: Yeah that's fair. I can update it to follow a more sensible philosophy. I do agree that the HTML spec is overly permissive with these.

ljharb:
Happy to reopen if there's a viable path forward here.

PR #3510, opened 2022-12-19
[Docs] Correct example for prefer-exact-props rule

### LINKED ISSUE #3509: [Bug]: Error in docs for prefer-exact-props

### Description Overview

The flow example for prefer-exact-props has a typo.
https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/prefer-exact-props.md

```
  props: {|
    bar: string
  }|

  render() {
    return <p>{this.props.bar}</p>;
  }
}
```

The pipe is outside of the brackets instead of inside.

### Expected Behavior

Example uses correct syntax.

7.31.11

### eslint version

n/a

### node version

n/a

smackfu:
Fixes #3509


ljharb (approved):

Dubbeditt (commented):
****

PR #3511, opened 2022-12-24
[Fix] `jsx-no-leaked-render`: invalid fixes in coerce mode

### LINKED ISSUE #3431: `jsx-no-leaked-render` performs redundant props check and invalid fix in coerce mode
# Context

There is a sample `Checkbox` component:

```typescript
const Checkbox = ({
  isChecked = false,
  isIndeterminate = false,
}: {
  isChecked?: boolean;
  isIndeterminate?: boolean;
}) => (
  <input 
    checked={isIndeterminate ? false : isChecked} 
    type="checkbox"
  />
);
```

The eslint config file contains a rule:

```
"react/jsx-no-leaked-render": ["error", { validStrategies: ["coerce"] }],
```

# Expected behavior

Linting the code raises no issues

# Actual behavior 

1. Eslint displays an error for the line:

```typescript
checked={isIndeterminate ? false : isChecked}
```

```
Potential leaked value that might cause unintentionally rendered values or rendering crasheseslint[react/jsx-no-leaked-render](https://github.com/jsx-eslint/eslint-plugin-react/tree/master/docs/rules/jsx-no-leaked-render.md)
```

2. When auto-fix is performed, the code above is replaced with

```typescript
checked={!!isIndeterminate && false} 
```

which causes the logical issues (the `isChecked` is removed in that case).

akulsr0:
Fixes #3431

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 * [`jsx-no-leaked-render`]: Don't report errors on empty strings if React >= v18 ([#3488][] @himanshu007-creator)
 * [`no-invalid-html-attribute`]: convert autofix to suggestion ([#3474][] @himanshu007-creator @ljharb)
 * [`jsx-no-leaked-render`]: fix removing parentheses for conditionals ([#3502][] @akulsr0)
+* [`jsx-no-leaked-render`]: invalid fixes in coerce mode ([#3511][] @akulsr0)
 
 ### Changed
 * [Docs] [`jsx-no-leaked-render`]: Remove mentions of empty strings for React 18 ([#3468][] @karlhorky)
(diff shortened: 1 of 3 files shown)


ljharb (approved):
Thanks!

ljharb:
@akulsr0 seems like tests are failing

akulsr0:
@ljharb I am not sure why tests are failing here, I have tried with different node versions in local, its passing all test cases.

ljharb:
The failures are in eslint < 5.

ljharb:
I’m not sure why the output is different in eslint 3 and 4, but we may have to make the test output condition on eslint version.

akulsr0:
Do we have any example case where output is based on eslint version?

ljharb:
@akulsr0 mostly examples of test cases that are entirely skipped, but https://github.com/jsx-eslint/eslint-plugin-react/blob/12fe9447844ffed58c714c56bcfc89886a2ba5be/tests/lib/rules/sort-prop-types.js#L505-L540 should hopefully be enough for you to figure out how to adapt it?

ljharb:
I tweaked your changes so the test cases still run in eslint 3, and it exposed a bug, which i then fixed. This should hopefully be good to go once tests pass.

Issue #3416, opened 2022-09-06
react/no-unknown-property: allow viewbox on marker


dylang:
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/viewBox

As of `7.31.7`:

```jsx
error: Invalid property 'viewBox' found on tag 'marker', but it is only allowed on: svg (react/no-unknown-property):
  18 |     <Styled.BubbleMeter height={Math.round(radius * 2)} width={Math.round(radius * 2)} role="presentation">
  19 |       <defs>
> 20 |         <marker id={markerId} viewBox="0 0 2 2" refX="1" refY="1" markerWidth="1" markerHeight="1" orient="auto">
     |                               ^
  21 |           <Styled.Cap cx="1" cy="1" r="1" palette={capPalette} strokeWidth="0" />
  22 |         </marker>
  23 |       </defs>
```

PR #3525, opened 2023-01-24
[Fix] no-unknown-property `fill` for `marker`

alexey-koran:
Similar to https://github.com/jsx-eslint/eslint-plugin-react/issues/3416 but for fill of marker

https://developer.mozilla.org/en-US/docs/Web/SVG/Element/marker

```jsx
error  Invalid property 'fill' found on tag 'marker', but it is only allowed on: altGlyph, circle, ellipse, g, line, mask, path, polygon, polyline, rect, svg, text, textPath, tref, tspan, use, animate, animateColor, animateMotion, animateTransform, set  react/no-unknown-property

<marker fill='#000' />
```


ljharb (approved):
Thanks!

PR #3529, opened 2023-02-10
[New] `checkContextObjects` option in `display-name`

JulesBlm:
I'm proposing adding a new rule that checks if the `displayName` property is set for context objects, see [`Context.displayName`](https://reactjs.org/docs/context.html#contextdisplayname) in the React docs.

With multiple Context providers in an app, consistently setting a `displayName` makes it easy to see which is which in the React dev tools.

Some things to consider

* Should it perhaps be part of the `display-name` rule instead of a separate rule?
* Does it need more tests?
* Does this checking approach make sense? (This is my first time writing an ESLint rule, so I wouldn't be surprised if this is not the best way)
* Does it need to check more cases?

### DIFF
--- a/lib/rules/display-name.js
+++ b/lib/rules/display-name.js
@@ -8,6 +8,7 @@
 const values = require('object.values');
 
 const Components = require('../util/Components');
+const isCreateContext = require('../util/isCreateContext');
 const astUtil = require('../util/ast');
 const componentUtil = require('../util/componentUtil');
 const docsUrl = require('../util/docsUrl');
@@ -21,6 +22,7 @@ const report = require('../util/report');
 
 const messages = {
   noDisplayName: 'Component definition is missing display name',
+  noContextDisplayName: 'Context definition is missing display name',
 };
 
 module.exports = {
@@ -40,6 +42,9 @@ module.exports = {
         ignoreTranspilerName: {
           type: 'boolean',
         },
+        checkContextObjects: {
+          type: 'boolean',
+        },
       },
       additionalProperties: false,
     }],
@@ -48,6 +53,9 @@ module.exports = {
   create: Components.detect((context, components, utils) => {
     const config = context.options[0] || {};
     const ignoreTranspilerName = config.ignoreTranspilerName || false;
+    const checkContextObjects = (config.checkContextObjects || false) && testReactVersion(context, '>= 16.3.0');
+
+    const contextObjects = new Map();
 
     /**
      * Mark a prop type as declared
@@ -84,6 +92,16 @@ module.exports = {
 
       report(context, messages.noDisplayName, 'noDisplayName', {
         node: component.node,
+      });
+    }
+
+    /**
+     * Reports missing display name for a given context object
+     * @param {Object} contextObj The context object to process
+     */
+    function reportMissingContextDisplayName(contextObj) {
+      report(context, messages.noContextDisplayName, 'noContextDisplayName', {
+        node: contextObj.node,
       });
     }
 
@@ -144,6 +162,16 @@ module.exports = {
     // --------------------------------------------------------------------------
 
     return {
+      ExpressionStatement(node) {
+        if (checkContextObjects && isCreateContext(node)) {
+          contextObjects.set(node.expression.left.name, { node, hasDisplayName: false });
+        }
+      },
+      VariableDeclarator(node) {
+        if (checkContextObjects && isCreateContext(node)) {
+          contextObjects.set(node.id.name, { node, hasDisplayName: false });
+        }
+      },
       'ClassProperty, PropertyDefinition'(node) {
         if (!propsUtil.isDisplayNameDeclaration(node)) {
           return;
@@ -154,6 +182,14 @@ module.exports = {
       MemberExpression(node) {
         if (!propsUtil.isDisplayNameDeclaration(node.property)) {
           return;
+        }
+        if (
+          checkContextObjects
+          && node.object
+          && node.object.name
+          && contextObjects.has(node.object.name)
+        ) {
+          contextObjects.get(node.object.name).hasDisplayName = true;
         }
         const component = utils.getRelatedComponent(node);
         if (!component) {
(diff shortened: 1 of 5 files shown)


tests/lib/rules/context-display-name.js
```diff
@@ -0,0 +1,111 @@
...
+      `,
+    },
+    {
+      code: `
+        import * as React from 'react';
+
+        const Hello = React.createContext();
+
+        const obj = {};
+        obj.displayName = "False positive";
```
ljharb, line 67: should there be some of these test cases that avoid React.createContext entirely?

ljharb (commented):
Overall, this seems reasonable. Should this maybe be an option on the `display-name` rule?

JulesBlm:
> Overall, this seems reasonable. Should this maybe be an option on the display-name rule?

I think that makes sense, the intention is the same. Main difference is that `createContext` does not create a component directly but a context object containing two components.
I looked at the implementation of `display-name` and it makes heavy use of the `Components` utility, which can't be used for `createContext`. I thought it would be easier and cleaner to make a separate rule for it but I'm not married to the idea of it being a separate rule. What would your preference be?

tests/lib/rules/context-display-name.js
JulesBlm: Just like this?

```javascript
const obj = {};
obj.displayName = "False positive";
```

ljharb: yup, looks good!

JulesBlm: OK added!

ljharb:
while that's a fair point, since the rule is "display name" and not "component display name" i think it's worth figuring out the complexity inside the rule, rather than adding complexity for users.

JulesBlm:
> while that's a fair point, since the rule is "display name" and not "component display name" i think it's worth figuring out the complexity inside the rule, rather than adding complexity for users.

Alright makes sense, I turned it into a `checkContextObjects` option in `display-name`. As far as I can tell, the logic for detecting component isn't of much use for detecting context definitions so I chose not make use of the `Components` utility for it.

ljharb (commented):
There's a few lines missing test coverage; otherwise it looks pretty great!

lib/rules/display-name.js
```diff
@@ -87,6 +124,18 @@ module.exports = {
       });
     }
 
+    /**
+     * Reports missing display name for a given context object
+     * @param {Object} contextObj The context object to process
+     */
+    function reportMissingContextDisplayName(contextObj) {
+      if (testReactVersion(context, '>= 16.3.0')) {
```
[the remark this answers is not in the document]
JulesBlm: Yes sure, I moved it in `checkContextObjects` like this:

```js
const checkContextObjects = (config.checkContextObjects || false) && testReactVersion(context, '>= 16.3.0');
```

JulesBlm:
> There's a few lines missing test coverage; otherwise it looks pretty great!

Thanks! I noticed I forgot to enable the new flag for the new `valid` tests so I added that. I added few more tests too, some formatting fixes and moved the React version check.

I'm a little confused how to increase the test coverage. What would need to be done to meet the threshold? Test the `isCreateContext` util?

lib/rules/display-name.js
```diff
@@ -70,6 +77,36 @@ module.exports = {
...
     }
 
+    /**
+     * Checks if the node is a React.createContext call
+     * @param {ASTNode} node - The AST node being checked.
+     * @returns {Boolean} - True if node is a React.createContext call object literal, false if not.
+     */
+    function isCreateContext(node) {
+      if (!node.init) {
+        return false;
```
ljharb, line 87: this line is what we'd need to hit - when would `node` have a falsy `init` value?

JulesBlm: Initially I put it in here as a guard without much thought. This comment got me thinking that contexts declared like this were not covered by the rule

```

File: lib/rules/display-name.js [hunk 4]
```diff
@@ -70,6 +77,58 @@ module.exports = {
       return node.type === 'CallExpression' && argumentIsCallExpression && utils.isPragmaComponentWrapper(node);
     }
 
+    /**
+     * Checks if the node is a React.createContext call
+     * @param {ASTNode} node - The AST node being checked.
+     * @returns {Boolean} - True if node is a React.createContext call, false if not.
+     */
+    function isCreateContext(node) {
```
ljharb, line 85: this seems like a good idea to add to `Components.js`, so other rules could share it?

ljharb (commented):
Looks good overall

lib/rules/display-name.js
JulesBlm: Initially I put this line in as guard without much thought. Thinking about it, `node.init` can only be falsy with a variable declaration without an initial value eg `let Hello;`. This got me thinking about the following: The rule would not detect contexts in assignments like this

```js
import { createContext } from 'react';

var Hello = {};
Hello = React.createContext(); // or createContext()
```

While unusual, I added a check for this too along with tests for it.

JulesBlm:
Hey, is there anything I can do to move this PR forward?

lib/rules/display-name.js
ljharb: Sure, a separate file is fine too, it just may need to be in the things provided to `Components.detect`

ljharb:
@JulesBlm sorry for the delayed reply; i replied to your comment above and will review today.

ljharb (approved):
This looks great - I'll land this later tonight.

lib/rules/display-name.js
ljharb: Awesome, thanks!

JulesBlm:
Awesome thanks for your review @ljharb!

christianvuerings:
@ljharb https://www.npmjs.com/package/eslint-plugin-react still shows `7.32.2` as the latest version. Did anything go wrong during the npm publish process?
image

ljharb:
huh, maybe so, let me check

ljharb:
ah thanks, published it manualy

christianvuerings:
@ljharb that worked, thank you!

PR #3533, opened 2023-02-15
[New] `jsx-first-prop-new-line`: add `multiprop` option

haydncomley:
The `jsx-first-prop-new-line` rule is great but it just falls short in one place for me. This addition allows the rule to enforce a new line, but only if there are multiple props, else it forces it inline - a hybrid between `multiline` and `multiline-multiprop`.

Before Linting (currently the div element thinks that it is valid, but I want it to throw a linting error)

After Linting (the div is automatically formatted to put the prop on the first line, but the component below is kept multiline)

(I hope I've formatted this okay as I couldn't find a template for PRs)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ### Added
 * [`display-name`]: add `checkContextObjects` option ([#3529][] @JulesBlm)
+* [`jsx-first-prop-new-line`]: add `multiprop` option ([#3533][] @haydncomley)
 
 ### Fixed
 * [`no-array-index-key`]: consider flatMap ([#3530][] @k-yle)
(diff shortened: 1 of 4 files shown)


ljharb:
So, just to clarify, this is `multiline-multiprop` but with an additional restriction that *only* multi prop elements may be multiline?

haydncomley:
> So, just to clarify, this is `multiline-multiprop` but with an additional restriction that _only_ multi prop elements may be multiline?

Yes indeed 🙂

ljharb (approved):

sanketnaik99:
@ljharb Is there an ETA on when this will be released?

mindnektar:
@ljharb just spent some time debugging why i couldn't use this even though it's in the docs. guess i found the answer. i'd love to use this option, so i've got the same question: when is the next release? :)

mindnektar:
by the way: what if the prop itself is multiline? i would like to enforce something like this in such cases:

```
<div
    className={classnames(
        'foo',
        { 'foo--bar': props.bar }
    )}
>
    ...
</div>
```

right now, the `multiprop` option would cause this:

```
<div className={classnames(
    'foo,
    { 'foo--bar': props.bar }
)}
>
```

ljharb:
@mindnektar everywhere on github, you have to look at the docs on the tagged release and not on the default branch.

@sanketnaik99 no, there is no ETA. hopefully soon.

Issue #3537, opened 2023-02-24
[jsx-curly-brace-presence] template literals contains just expressions


taozhou-glean:
we enabled jsx-curly-brace-presence, but I am seeing patterns in some code like following:

```
<Comp label={`${componentLabel}`} />

// or

<Comp>
{`${someText}`}
</Comp>
```

which would be same as:

```
<Comp label={componentLabel} />

// or

<Comp>
{someText}
</Comp>
```

and this seems to be an ideal case to be caught by jsx-curly-brace-presence ;)

=> Found "eslint-plugin-react@7.32.2"
=> Found "eslint@8.34.0"

ljharb:
I wasn't paying enough attention I guess, because these aren't actually the same.

PR #3538, opened 2023-02-24
[Fix] `jsx-curly-brace-presence`: handle single and only expression template literals

taozhou-glean:
fix https://github.com/jsx-eslint/eslint-plugin-react/issues/3537

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ### Fixed
 * [`no-array-index-key`]: consider flatMap ([#3530][] @k-yle)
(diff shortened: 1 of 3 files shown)


ljharb (approved):

Newbie012:
Hi @ljharb, I believe this change is wrong. This is a valid use case for stringifying a primitive (e.g., from number to string). Although you could do it via `String(var)` or `var.toString()`, using `={'${number}`} is a valid use case as well.

If you insist on keeping it as it is, I suggest putting it behind a flag 🙏.

taozhou-glean:
let me revert this now

ljharb:
@Newbie012 yes, you're right, it should be reverted.

PR #3547, opened 2023-03-09
Fixes #3482 - handle formatMessage calls

### LINKED ISSUE #3482: `react/no-unstable-nested-components` does not allow ignoring only object properties
I would like to keep this rule active but disable it for object properties, I have this:

```jsx
function MyComponent() {
  return (
    <p>
      {intl.formatMessage(messages.allSelected, {
        size: selectedEvents.size,
        b: chunks => <b>{chunks}</b>, // <-- report error here without `allowAsProps`
      })}
    </p>
  );
}
```
But adding the `allowAsProps: true` option allows me to do things like this:
```jsx
<Dropdown
  Trigger={({ onClick }) => ( // <-- don't report error here with `allowAsProps`
    <button type="button" onClick={onClick}>
      Bla bla
      <span className="caret" />
    </button>
  )}
>
  {/* ... */}
</Dropdown>
```

AsaAyers:
I handled this similar to `isMap` because `formatMessage` seems like a similar special case. https://formatjs.io/docs/intl#formatmessage

When defining a message to be translated, `defaultMessage` can contain placeholders like `{name}`, but it also uses an HTML-like format: `<a>link text</a>`. Notice there's no href, because it gets replaced using a function that returns JSX that fills in the details. 

```
intl.formatMessage(messages.example, {
  name: 'Asa'
  a: (...chunks) => <a href="/">{...chunks}</a>
})
```

> I'm a bit confused. `Trigger` here is definitely a component, and should be warned on - but `b` isn't a component, because it's got a lowercase name (if it was `B` i'd say it _is_ a component).
https://github.com/jsx-eslint/eslint-plugin-react/issues/3482#issuecomment-1304406131

This plugin seems to define components as any function that returns JSX. By that definition, this IS a component even if it happens to be assigned to a property of an object. Components can either start with a capital letter or be part of an object, so in this example `formatMessage` could internally call `<values.a>link text</values.a>` and use it directly as a component.

### DIFF
--- a/lib/rules/no-unstable-nested-components.js
+++ b/lib/rules/no-unstable-nested-components.js
@@ -322,7 +322,10 @@ module.exports = {
 
   create: Components.detect((context, components, utils) => {
     const allowAsProps = context.options.some((option) => option && option.allowAsProps);
-    const ignoreFunctions = context.options.some((option) => option && option.ignoreFunctions);
+    const ignoreFunctions = context.options.reduce(
+      (ignore, option) => ignore.concat(option && option.ignoreFunctions),
+      []
+    );
 
     /**
      * Check whether given node is declared inside class component's render block
--- a/lib/rules/no-unstable-nested-components.js
+++ b/lib/rules/no-unstable-nested-components.js
@@ -144,31 +144,6 @@ function isMapCall(node) {
 }
 
 /**
- * Check whether given node is part of a formatMessage call
- * ```jsx
- * {intl.formatMessage(messages.testMessage, {
- * a: (...chunks) => <a href="/">{chunks}</a>
- * })}
- * ```
- * @param {ASTNode} node The AST node
- * @returns {Boolean} True if node is directly inside `map` call, false if not
- */
-function isFormatMessageCall(node) {
-  return (
-    node
-    && node.parent
-    && node.parent.type === 'Property'
-    && node.parent.parent
-    && node.parent.parent.type === 'ObjectExpression'
-    && node.parent.parent.parent
-    && node.parent.parent.parent.type === 'CallExpression'
-    && node.parent.parent.parent.callee
-    && node.parent.parent.parent.callee.property
-    && node.parent.parent.parent.callee.property.name === 'formatMessage'
-  );
-}
-
-/**
  * Check whether given node is `ReturnStatement` of a React hook
  * @param {ASTNode} node The AST node
  * @param {Context} context eslint context
@@ -189,6 +164,30 @@ function isReturnStatementOfHook(node, context) {
     && callExpression.callee
     && HOOK_REGEXP.test(callExpression.callee.name)
   );
+}
+
+/**
+ * Check whether given node is declared inside a specific function.
+ *
+ * ```jsx
+ * {intl.formatMessage(messages.testMessage, {
+ *   a: (...chunks) => <a href="/">{chunks}</a>
+ *   // ^ This returns JSX, so technically it's a component
+ * })}
+ * ```
+ * @param {ASTNode} node The AST node
+ * @param {Context} context eslint context
+ * @param {Array<string>} ignoredFunctions
+ * @returns {Boolean} True if component is declared inside a render prop, false if not
+ */
+function isDescendantOfIgnoredFunction(node, context, ignoredFunctions) {
+  const code = context.getSourceCode();
+  return context.getAncestors().some((n) => ignoredFunctions.some((functionName) => {
+    if (n.type === 'CallExpression') {
+      return code.getText(n.callee) === functionName;
+    }
+    return false;
+  }));
 }
 
 /**
@@ -310,6 +309,12 @@ module.exports = {
         allowAsProps: {
           type: 'boolean',
         },
+        ignoreFunctions: {
+          type: 'array',
+          items: {
+            type: 'string',
+          },
+        },
       },
       additionalProperties: false,
     }],
@@ -317,6 +322,7 @@ module.exports = {
 
   create: Components.detect((context, components, utils) => {
     const allowAsProps = context.options.some((option) => option && option.allowAsProps);
+    const ignoreFunctions = context.options.some((option) => option && option.ignoreFunctions);
 
     /**
      * Check whether given node is declared inside class component's render block
@@ -456,7 +462,6 @@ module.exports = {
         // Prevent reporting components created inside Array.map calls
         || isMapCall(node)
         || isMapCall(node.parent)
-        || isFormatMessageCall(node)
 
         // Do not mark components declared inside hooks (or falsy '() => null' clean-up methods)
         || isReturnStatementOfHook(node, context)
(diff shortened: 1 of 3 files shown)


lib/rules/no-unstable-nested-components.js
```diff
@@ -143,6 +143,31 @@ function isMapCall(node) {
...
+    && node.parent
+    && node.parent.type === 'Property'
+    && node.parent.parent
+    && node.parent.parent.type === 'ObjectExpression'
+    && node.parent.parent.parent
+    && node.parent.parent.parent.type === 'CallExpression'
+    && node.parent.parent.parent.callee
+    && node.parent.parent.parent.callee.property
+    && node.parent.parent.parent.callee.property.name === 'formatMessage'
+  );
```
ljharb, line 168: this seems very hardcoded to one specific API from one specific hook. formatMessage could be renamed, and is `useIntl` a core hook, or a third-party one?

AsaAyers: It's a 3rd party library. I'm open to other suggestions on how to solve this, but I wanted to start with something that I know works. It's not a prop, so `allowAsProps` isn't relevant. I couldn't figure out how the `customValidators` work, but they seem to be related to `propTypes`, so that didn't seem to fit either.


I think having an option to just not traverse down into specific functions could cover both use cases. `ignoreFunctions: ['map', 'intl.formatMessage']`. What do you think?

ljharb: `intl.formatMessage` seems safe enough, but `map` is just *way* too broad.

imo the reality is that this is a very strange and unidiomatic API, and it simply conflicts with static analysis. I suspect that there could be changes in the useIntl library that would obviate this problem, but that's ofc not something either of us have control over.

lib/rules/no-unstable-nested-components.js
```diff
@@ -166,6 +166,30 @@ function isReturnStatementOfHook(node, context) {
...
+ * @param {Context} context eslint context
+ * @param {Array<string>} ignoredFunctions
+ * @returns {Boolean} True if component is declared inside a render prop, false if not
+ */
+function isDescendantOfIgnoredFunction(node, context, ignoredFunctions) {
+  const code = context.getSourceCode();
+  return context.getAncestors().some((n) => ignoredFunctions.some((functionName) => {
+    if (n.type === 'CallExpression') {
+      return code.getText(n.callee) === functionName;
+    }
```
AsaAyers, line 188: Is `code.getText(n.callee)` an ok solution here? Are there better ways to handle this check? I figure this could break if you format your code awkwardly. :man_shrugging: Since we're really working around a bad API that we can't change, maybe this is fine. What do you think?

```
intl
  .formatMessage()
```

lib/rules/no-unstable-nested-components.js
AsaAyers: >  it simply conflicts with static analysis

It certainly conflicts with ESLint rules for static analysis. In my test code I included the call to `defineMessages` because I think it's good to have the whole context of how this is used. `defineMessages` doesn't do anything at runtime, it's exclusively there for static analysis/types to mark the places that need to be extracted for localization. It just returns the same object that's passed in. 

https://github.com/AsaAyers/eslint-plugin-react/blob/86e925a4e06fbdb8b37dbdbf42f0e10b5fb6b541/tests/lib/rules/no-unstable-nested-components.js#L231-L256

lib/rules/no-unstable-nested-components.js
AsaAyers: I don't see a cleaner way to get the whole `intl.formatMessage`. The code below works, but only for exactly `object.property()` type calls. This code can't handle handle `ignoreFunctions: [ 'myCustomFunction', 'why.would.anyone.do.this' ]`. Especially since my initial PR targeted exactly and exclusively `intl.formatMessage`, I'm trying to keep this a bit more flexible. 

```
      if (n.callee.type === 'MemberExpression') {
        if (n.callee.object.type === 'Identifier' && n.callee.property.type === 'Identifier') {
          return `${n.callee.object.name}.${n.callee.property.name}` === functionName;
        }
      }
```

image

Considering this is all working around an API that we agree is bad and we can't change, do you think it's ok to stick with `getText`? 

A couple questions that come to mind for me that I don't know how to answer are: 
- Is it expensive to call `context.getSourceCode()`? 
- Should it only be called once somewhere else, or is it better to keep it with the feature that uses it?

ljharb: It's not expensive, and it's fine either way. I generally put it in a variable when it's used multiple times, but i don't think it makes a material difference.

The challenge with the getText approach is that it's very brittle - you'd need a bunch of other linting rules to ensure that a particular function is required/imported with the right name, destructured with the right name or not, etc, in order for any config value to be meaningful.

AsaAyers: I agree that you have a point about it being brittle, but I'm not sure how solid this really needs to be. 

For my codebase we simply have a convention of using `const intl = useIntl()` and `intl.formatMessage()`. I'm fine with this rule not catching `const { formatMessage } = useIntl()`. We use code owners on the lint rules, so if someone adds `formatMessage` to the rule, we'll request they change the source instead and follow the naming convention.

If this can be achieved through types, I'm really looking for target this function. `import('react-intl').IntlFormatters['formatMessage']`. Without types, it seems like a long journey to work through:  `the method formatMessage` on `the object returned` from `the useIntl() import` on ` react-intl module`. Are there existing tools within the repo that would do some parts of that?

ljharb: both eslint itself, and this plugin, have some utility functions to help you track that, but unfortunately nothing super trivial to use.

essentially you'd have to record all requires/imports and the identifier used, and then find all the potential call expressions and store those, and then at program exit, check if any of the candidates you found came from the specifier in question.

AsaAyers: If that's the hurdle that needs to be cleared here, I think I'll take a change in direction. I'd actually like a rule closer to `react-hooks/exhaustive-deps` that checks the stability of the JSX call.

```
export function Component() {
  const SomeComponent = () => <div />;
  //    ^ The 'SomeComponent' function makes the dependencies of useEffect Hook
  //    (at line 41) change on every render. To fix this, wrap the definition of
  //    'SomeComponent' in its own useCallback() Hook.
  //    - react-hooks/exhaustive-deps

  React.useEffect(() => {
    SomeComponent();
  }, [SomeComponent]);

  return (
    <SomeComponent />
  // ^ Idea: "SomeComponent" has an unstable identity, causing it to un-mount
  // and re-mount instead of re-render
  );
}
```

ljharb: I think we might have that already? https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unstable-nested-components.md

AsaAyers: Yes, that's this plugin. But I can't use it because our codebase uses `intl.formatMessage()` and there are too many false positives. It's also written into our coding standards, so the problem isn't going away.


The project I'm working on has an internal ESLint plugin. I have something there that examines each JSXIdentifier and tries to figure out if it's stable. It makes some assumptions, but it's working well enough for what I need. It assumes anything from a hook is stable, anything from an outer scope, or the current scope's parameters are stable. I assume anything inside a `describe/test` function doesn't need to be scanned.

PR #3548, opened 2023-03-09
[New] Align `no-deprecated` rule with React 18 deprecations

sergei-startsev:
The list of deprecated APIs in React 18.0.0:

- react-dom: `ReactDOM.render`
- react-dom: `ReactDOM.hydrate`
- react-dom: `ReactDOM.unmountComponentAtNode`
- react-dom/server: `ReactDOMServer.renderToNodeStream`

Ref: https://github.com/facebook/react/blob/main/CHANGELOG.md#deprecations

Migration guide: https://beta.reactjs.org/blog/2022/03/08/react-18-upgrade-guide

### DIFF
--- a/docs/rules/no-deprecated.md
+++ b/docs/rules/no-deprecated.md
@@ -36,14 +36,28 @@ import React, { PropTypes } from 'react';
 componentWillMount() { }
 componentWillReceiveProps() { }
 componentWillUpdate() { }
+
+// React 18 deprecations
+import { render } from 'react-dom';
+ReactDOM.render(<div></div>, container);
+
+import { hydrate } from 'react-dom';
+ReactDOM.hydrate(<div></div>, container);
+
+import {unmountComponentAtNode} from 'react-dom';
+ReactDOM.unmountComponentAtNode(container);
+
+import { renderToNodeStream } from 'react-dom/server';
+ReactDOMServer.renderToNodeStream(element);
 ```
 
 Examples of **correct** code for this rule:
 
 ```jsx
+// when React < 18
 ReactDOM.render(<MyComponent />, root);
 
-// When [1, {"react": "0.13.0"}]
+// when React is < 0.14
 ReactDOM.findDOMNode(this.refs.foo);
 
 import { PropTypes } from 'prop-types';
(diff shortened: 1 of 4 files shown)


sergei-startsev:
@ljharb / @gaearon for awareness.

tests/lib/rules/no-deprecated.js
```diff
@@ -48,9 +48,9 @@ ruleTester.run('no-deprecated', rule, {
     // Not deprecated
     'var element = React.createElement(\'p\', {}, null);',
     'var clone = React.cloneElement(element);',
-    'ReactDOM.render(element, container);',
-    'ReactDOM.unmountComponentAtNode(container);',
```
ljharb, line 52: we should be able to keep these test cases with the react version set to 17.999.999

docs/rules/no-deprecated.md
```diff
@@ -36,13 +36,24 @@ import React, { PropTypes } from 'react';
...
+ReactDOM.unmountComponentAtNode(container);
+
+import { renderToNodeStream } from 'react-dom/server';
+ReactDOMServer.renderToNodeStream(element);
 ```
 
 ```jsx
-ReactDOM.render(<MyComponent />, root);
```
ljharb, line 44: let's keep this, but with a comment that it only applies to react < 18

tests/lib/rules/no-deprecated.js
sergei-startsev: Added one more test to cover this scenario.

docs/rules/no-deprecated.md
sergei-startsev: Added a comment.

tests/lib/rules/no-deprecated.js
ljharb: I meant, keep the "valid" tests cases, because these remain valid in react < 18.

ljharb: I meant, keep the "valid" cases, because these remain valid in react < 18.

sergei-startsev: You cannot leave it where it was before, React version isn't specified there.

ljharb: right - i mean, edit the existing rule to add the react version, so that "a passing test for ReactDOM.render" remains

sergei-startsev: React version can be specified using eslint `settings.react.version` property which is demonstrated in the added tests. Introducing multiple sources to determine React version doesn't seem like a good idea to me.

sergei-startsev: @ljharb Is there anything else that prevents the PR from being merged? I believe the requested scenario has been covered:
https://github.com/jsx-eslint/eslint-plugin-react/blob/1375bd1d893bdf73ee81e54bc73c5a317828d176/tests/lib/rules/no-deprecated.js#L122-L132

ljharb (approved):

PR #3557, opened 2023-03-28
Add `disallowedFor` to `forbid-component-props` and `allowedFor` to `forbid-dom-props`

cincodenada:
My use case is component props: I am using a component system (Antd) and want to disallow the `required` parameter on `Form.Item` to avoid confusion with a `require` parameter in HOCs that we create.

I found #3422 already existing, so this is a PR to solve that, as well as adding `allowList` to `forbid-dom-props` for symmetry.

I haven't yet figured out what the best way is to raise configuration issues, right now I'm just throwing an error, let me know if there's a better way!

(Motivation details: in Antd `require` only controls styling, and doesn't actually validate, in our HOCs I want it to validate, but don't want that confusion to bleed over into using `require` in the base (non-HOC) `Form.Item`s)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 * [`display-name`]: add `checkContextObjects` option ([#3529][] @JulesBlm)
 * [`jsx-first-prop-new-line`]: add `multiprop` option ([#3533][] @haydncomley)
 * [`no-deprecated`]: add React 18 deprecations ([#3548][] @sergei-startsev)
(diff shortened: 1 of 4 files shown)


cincodenada:
Sigh, okay, so I just realized that this is largely a duplicate of #3416, which I missed because it wasn't tagged on #3422 becuase that issue started as a discussion on #2946.

The discussion on #3416 also states:
> In the future, please do not send PRs for changes you're not sure are wanted.

If there is some sort of policy here, it should be stated in `CONTRIBUTING.md` - there was an existing issue without neither satisfying resolution nor refusal, thus I created a PR.

ljharb:
I'm going to pull your changes into #3417, but i'll drop the forbid-dom-props changes, since that needs to happen in another PR.

PR #3563, opened 2023-04-22
[Docs] Fix code syntax highlighting

nbsp1221:
I fixed the issue with code syntax highlighting that caused a red background line to appear.

Before
image

After
image


ljharb:
These are meant to be json, though, not js

nbsp1221:
@ljharb 

I found `json5`. It seems that by using it, I can fix code syntax highlighting while retaining the fact these meant JSON.

ljharb (approved):

PR #3577, opened 2023-05-14
[New] `hook-use-state`: add option to use useState hook with no value

azat-io:
This is useful if you want to forcibly re-render a component

### DIFF
--- a/lib/rules/hook-use-state.js
+++ b/lib/rules/hook-use-state.js
@@ -41,6 +41,10 @@ module.exports = {
           default: false,
           type: 'boolean',
         },
+        allowNoValue: {
+          default: false,
+          type: 'boolean',
+        },
       },
       additionalProperties: false,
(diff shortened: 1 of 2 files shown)


ljharb:
This seems like a special enough case that it warrants an eslint override comment, and not a whole new option to the rule. What codebases would have more than one occurrence of this pattern?

azat-io:
I would suggest disabling error messages if no value is specified.

It doesn't sound like a naming problem:

```js
const [, forceUpdate] = useState({})
```

ljharb:
This feels like an antipattern - components should never need to rerender unless props or state have changed.

ljharb:
Closing pending further discussion, since I'm not yet convinced this isn't an antipattern. I'd be happy to reopen if there's a reasonable use case where this option would be helpful.

PR #3579, opened 2023-05-16
[New Rule] prefer-use-state-lazy-initialization

patorjk:
Adds a new rule to detect function calls inside useState and recommends that an initializer function be used instead. For example, this code:

`const [value, setValue] = useState(generateTodos());
`

Would trigger the rule into recommending this instead:

`const [value, setValue] = useState(() => generateTodos());
`

More info: [React docs on avoiding recreating initial state](https://react.dev/reference/react/useState#avoiding-recreating-the-initial-state)


ljharb:
I think that rules that are specific to hooks belong in `eslint-plugin-react-hooks`, not here.

patorjk:
> I think that rules that are specific to hooks belong in `eslint-plugin-react-hooks`, not here.

The description for [that repo](https://github.com/facebook/react/tree/main/packages/eslint-plugin-react-hooks) says its for enforcing the [Rules of Hooks](https://legacy.reactjs.org/docs/hooks-rules.html). There are 2 rules of hooks, and thus only 2 eslint rules in that repo. This rule is for finding wasteful function calls. It's oriented around the useState hook, but hooks are now a core part of React. I think not including rules related to them would leave out a huge part of React.

ljharb:
Then it should be named "eslint-plugin-react-rules-of-hooks` or something :-) The readme there says it *enforces* the rules - it doesn't say they're *only* for that. I think it's worth trying to PR it into there before trying here.

patorjk:
> Then it should be named "eslint-plugin-react-rules-of-hooks` or something :-) The readme there says it _enforces_ the rules - it doesn't say they're _only_ for that. I think it's worth trying to PR it into there before trying here.

The more I think about it, it does kind of make sense, especially since this is something their docs talk about. I created a [PR](https://github.com/facebook/react/pull/26822) over there to see what they say.

ternaus:
`react/prefer-use-state-lazy-initialization` is included in [@ternaus/eslint-plugin-react v8.0.0](https://github.com/ternaus/eslint-plugin-react/releases/tag/v8.0.0), an independent ESLint 10 / React 19+ continuation. The rule reports eager calls in `useState` initial values and leaves function-valued lazy initializers alone. The full scope and installation path are in [#4036](https://github.com/jsx-eslint/eslint-plugin-react/issues/4036).

PR #3581, opened 2023-05-30
Reword jsx-no-bind rationale

gpoole:
I'm proposing rewording the rationale for jsx-no-bind to:
* clarify that the performance concerns apply to specific situations; although it's qualified, the statement "is bad for performance" is a general statement and [not strictly accurate](https://react.dev/reference/react/memo#skipping-re-rendering-when-props-are-unchanged)
* refer explicitly to `memo` since I think `memo`ised components are probably the most common component with performance impacts from this rule
* include the fact that this rule also covers any function declared in the render method of a component, including both anonymous or named functions, not just arrow or bound functions, as shown in the examples

### DIFF
--- a/docs/rules/jsx-no-bind.md
+++ b/docs/rules/jsx-no-bind.md
@@ -2,7 +2,7 @@
 
 <!-- end auto-generated rule header -->
 
-A `bind` call or [arrow function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) in a JSX prop will create a brand new function on every single render. This is bad for performance, as it may cause unnecessary re-renders if a brand new function is passed as a prop to a component that uses reference equality check on the prop to determine if it should update.
+Using `bind` on a function or declaring a function in the render method of a component or the body of a functional component, and then passing that function as a prop will mean that the brand new function that is created on every single render will be considered a completely different function. This can affect performance in some situations, as it may cause unnecessary re-renders if a brand new function is passed as a prop to a component that uses reference equality check on the prop to determine if it should update, such as a component wrapped with [`memo`](https://react.dev/reference/react/memo#memo), or if the prop is used in any hook's "dependency array".
 
 Note that this behavior is different for `ref` props, which is a special case in React that **does not** cause re-renders when a brand new function is passed.  See [`ignore-refs`](#ignorerefs) below for more information.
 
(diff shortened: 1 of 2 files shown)


docs/rules/jsx-no-bind.md
```diff
@@ -2,7 +2,7 @@
 
 <!-- end auto-generated rule header -->
 
-A `bind` call or [arrow function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) in a JSX prop will create a brand new function on every single render. This is bad for performance, as it may cause unnecessary re-renders if a brand new function is passed as a prop to a component that uses reference equality check on the prop to determine if it should update.
+Using `bind` on a function or declaring a function in the render method of a component, and then passing that function as a prop will mean a brand new function is created on every single render. This can affect performance in some situations, as it may cause unnecessary re-renders if a brand new function is passed as a prop to a component that uses reference equality check on the prop to determine if it should update, such as a component wrapped with [`memo`](https://react.dev/reference/react/memo#memo).
```
ljharb, line 5:
```suggestion
Using `bind` on a function or declaring a function in the render method of a component or the body of a functional component, and then passing that function as a prop will mean that the brand new function that is created on every single render will be considered a completely different function. This can affect performance in some situations, as it may cause unnecessary re-renders if a brand new function is passed as a prop to a component that uses reference equality check on the prop to determine if it should update, such as a component wrapped with [`memo`](https://react.dev/reference/react/memo#memo), or if the prop is used in any hook's "dependency array".
```

ljharb (changes requested):
Thoughts on this?

Also, could we maybe add some examples as well to showcase the different kinds of functions that apply?

docs/rules/jsx-no-bind.md
gpoole: I think the current correct and incorrect usage examples clearly demonstrate all the cases I'm aware of, but I thought all possibilities should be referenced in the rationale as well, which I think they now are.

I think that calling out the dependency array is a good example of the performance impact but from looking into it it doesn't seem like it's covered by the rule. I think that could create ambiguity about what the rule covers, even it's useful information, what do you think? Am I missing anything with that?

gpoole:
I've also now updated the "when not to use section" to be consistent with the rationale. Sorry for the multiple micro edits!

docs/rules/jsx-no-bind.md
ljharb: I don't think it can - hook dependency arrays is just one of the ways that creating a function inside the render path, and passing it around, can cause issues.

gpoole: Makes sense, well I'm happy to go ahead with that 👍

ljharb (approved):

PR #3582, opened 2023-06-01
[Fix] update configuration URL

gokaygurcan:
Hi, 

When I run `eslint` on some React files, this is what I get:

```bash
$ npm run lint:js

> @gokaygurcan/test@1.0.0 lint:js
> eslint --config .eslintrc --color .

Warning: React version not specified in eslint-plugin-react settings. See https://github.com/jsx-eslint/eslint-plugin-react#configuration .
```

And the given URL resolves the repository, but doesn't scroll to the anchored place because of this change: https://github.com/jsx-eslint/eslint-plugin-react/commit/17858beeedaaf5d0eb4fd1cc292fb34d07f9f659#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R20 

This PR is here to fix the URL and point the correct location.

I hope this makes sense. Thanks in advance for your consideration.

Best,
G.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ npm install eslint eslint-plugin-react --save-dev
 
 It is also possible to install ESLint globally rather than locally (using `npm install -g eslint`). However, this is not recommended, and any plugins or shareable configs that you use must be installed locally in either case.
 
-## Configuration (legacy: `.eslintrc*`)
+## Configuration (legacy: `.eslintrc*`) <a id="configuration"></a>
 
 Use [our preset](#recommended) to get reasonable defaults:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,9 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ### Changed
 * [Docs] [`jsx-newline`], [`no-unsafe`], [`static-property-placement`]: Fix code syntax highlighting ([#3563][] @nbsp1221)
+* [readme] resore configuration URL ([#3582][] @gokaygurcan)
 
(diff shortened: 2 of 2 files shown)


gokaygurcan:
> Error: This pull request must have the “allow edits” checkbox checked.

Well, that's a weird requirement, but here we go.

ljharb:
Thanks! Instead of changing the outputted URL, can we add `<a id="configuration"></a>` in the appropriate place, so that the broken URL works again? [Cool URLs don't change](https://www.w3.org/Provider/Style/URI), and we don't want to be uncool :-)

gokaygurcan:
That's a better idea. I sure can do it like that. Lemme do some clickety-click and get back to you.

README.md
```diff
@@ -17,7 +17,7 @@ npm install eslint eslint-plugin-react --save-dev
 
 It is also possible to install ESLint globally rather than locally (using `npm install -g eslint`). However, this is not recommended, and any plugins or shareable configs that you use must be installed locally in either case.
 
-## Configuration (legacy: `.eslintrc*`)
+## <a id="configuration">Configuration</a> (legacy: `.eslintrc*`)
```
ljharb, line 20:
```suggestion
<a id="configuration"></a>
## Configuration (legacy: `.eslintrc*`)
```

ljharb (approved):
Thanks!

ljharb:
@gokaygurcan 
> Well, that's a weird requirement, but here we go.

it's checked by default, and it allows maintainers to rebase and adjust your PR branch as needed to land it. I've always found it weird that people sometimes *un* check it ¯\\\_(ツ)_/¯

gokaygurcan:
Thanks a lot ^_^

PR #3583, opened 2023-06-01
[Fix] `destructuring-assignment`: remove `useContext` destructuring check

### LINKED ISSUE #3536: [Bug]: `destructuring-assignment` enforces destructuring of modern context

### Description Overview

https://github.com/jsx-eslint/eslint-plugin-react/pull/2797 was recently merged and released. That PR requires any value returned from `useContext` to be destructured if it's an object:

```js
function MyComponent() {
  const foo = useContext(FooContext)
  return <>{foo.val}</> // error
}
```

It seems like there may have been some confusion or miscommunication here. That PR was meant to close https://github.com/jsx-eslint/eslint-plugin-react/issues/2309, but as noted in https://github.com/jsx-eslint/eslint-plugin-react/issues/2309#issuecomment-674361485, the rule is only supposed to warn for destructuring of _legacy_ context (i.e., not `useContext`).

As I understand it, the spirit of the rule is to require destructuring of the plain objects that React creates. That includes `props`, as well as legacy `state` and `context` (which are always stored in objects). But it doesn't seem like it should include modern context, because context can store any (single) value. That value may or may not be an object, and even if it is an object, it's not necessarily a plain object that serves as a container for conceptually separate values. For example, it could be an instance of a class (`const logger = useContext(LoggerContext)`) in which case destructuring doesn't make sense (or, at least, it doesn't seem within the spirit/scope of this rule to enforce that variety of destructuring).

The new behavior also results in an inconsistency between modern state and modern context, even though these are conceptually the same in terms of being able to store any single value:

```js
function MyComponent() {
  const foo = useContext(FooContext)
  const bar = useState({val: 1})
  return (
    <>
      {foo.val} // error
      {bar.val} // no error
    </>
  )
}
```

### Expected Behavior

I would expect this rule not to warn on values returned from `useContext` (or `useState`).

v7.32.2

### eslint version

v8.34.0

### node version

v18.12.0

102:
It seems that the use of modern context is forced to use the destructuring assignment after #2797; the problem was raised in #3536. 

~~This PR introduces additional flag that won't affect the existing behavior when it is unset, but will allow users to upgrade the lib version without having this rule disabled.~~

The approach was reviewed and now it consists of the following:

1. #2797 changes were reverted
2. Additional test cases were added in order to make sure that the value returned from `useContext` could now be used both with destructuring and without it

Fixes #3536

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,11 +20,13 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 * [`display-name`], [`prop-types`]: when checking for a capitalized name, ignore underscores entirely ([#3560][] @ljharb)
 * [`no-unused-state`]: avoid crashing on a class field function with destructured state ([#3568][] @ljharb)
 * [`no-unused-prop-types`]: allow using spread with object expression in jsx ([#3570][] @akulsr0)
+* Revert "[`destructuring-assignment`]: Handle destructuring of useContext in SFC" ([#3583][] [#2797][] @102)
 
 ### Changed
 * [Docs] [`jsx-newline`], [`no-unsafe`], [`static-property-placement`]: Fix code syntax highlighting ([#3563][] @nbsp1221)
 * [readme] resore configuration URL ([#3582][] @gokaygurcan)
 
+[#3583]: https://github.com/jsx-eslint/eslint-plugin-react/pull/3583
 [#3582]: https://github.com/jsx-eslint/eslint-plugin-react/pull/3582
 [#3570]: https://github.com/jsx-eslint/eslint-plugin-react/pull/3570
 [#3568]: https://github.com/jsx-eslint/eslint-plugin-react/issues/3568
--- a/lib/rules/destructuring-assignment.js
+++ b/lib/rules/destructuring-assignment.js
@@ -8,7 +8,6 @@ const Components = require('../util/Components');
 const docsUrl = require('../util/docsUrl');
 const isAssignmentLHS = require('../util/ast').isAssignmentLHS;
 const report = require('../util/report');
-const testReactVersion = require('../util/version').testReactVersion;
 
 const DEFAULT_OPTION = 'always';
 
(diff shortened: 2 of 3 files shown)


ljharb:
I'm not sure we need an option - since the return value of `useContext` isn't always an object, the rule should simply never force destructuring of it.

102:
> I'm not sure we need an option - since the return value of `useContext` isn't always an object, the rule should simply never force destructuring of it.

This will basically reduce the changes to reverting #2797, if I understand it correctly (and I’m absolutely ok with this option, will update the PR shortly)

ljharb:
ah, hmm. cc @Zinyon - what are your thoughts here?

102:
> ah, hmm. cc @Zinyon - what are your thoughts here?

We can also make the option to do the opposite — ignore by default, but enable the `useContext` check using the flag

102:
Hi @ljharb, @Zinyon, any thoughts on the further steps?

Zinyon:
@102 @ljharb Hi!

seems proper options to handle issues like https://github.com/jsx-eslint/eslint-plugin-react/issues/3536.
how about making option to ignore by default ? 
there can be many cases that does not need destructuring like modern context as single value (not object) i think.

ljharb:
@Zinyon it seems like that since useContext can be any kind of value, that it's simply incorrect to force destructuring of it.

Zinyon:
@ljharb i agree on second thought. seems better to remove logic for reporting about destruturing ``useContext`` 
https://github.com/jsx-eslint/eslint-plugin-react/pull/3583/files#diff-6351eca58740d2b829e9e8bf54fde9e65b07419cc99a2e96e2b4fdf960969d1aR242-R243

102:
@ljharb, @Zinyon, the `useContext` checks were removed (and also a few additional test cases were added to make sure that `useContext` destructuring is not being enforced)

Zinyon (approved):

ljharb (approved):
Thanks!

Issue #3589, opened 2023-06-25
[Bug]: semver dep security vulnerability


AviVahl:
### Is there an existing issue for this?

- [X] I have searched the existing issues and my issue is unique
- [X] My issue appears in the command-line and not only in the text editor

### Description Overview

When installing package using npm, audit fails with:
```
$ npm audit
# npm audit report

semver  <7.5.2
Severity: moderate
semver vulnerable to Regular Expression Denial of Service - https://github.com/advisories/GHSA-c2qf-rxjj-qqgw
fix available via `npm audit fix --force`
Will install eslint-plugin-react@7.25.3, which is a breaking change
node_modules/semver
  eslint-plugin-react  7.19.0 || >=7.26.0
  Depends on vulnerable versions of semver
  node_modules/eslint-plugin-react

2 moderate severity vulnerabilities

To address all issues (including breaking changes), run:
  npm audit fix --force
````

Running `npm audit fix --force` downgrades to `eslint-plugin-react@7.25.3` :eyes: 

### Expected Behavior

No security vulnerabilities.

### eslint-plugin-react version

7.32.2

### eslint version

8.43.0

### node version

18.16.1

ljharb:
It’s not a vulnerability here - like most transitive dep CVEs, it’s a false positive -  and we can’t upgrade because v7 drops support for engines we need to support.

1EDExg0ffyXfTEqdIUAYNZGnCeajIxMWd2vaQeP:
The babel team has backported the fix to semver v6, can we use that? https://github.com/babel/babel/pull/15742

ljharb:
I’d really rather not use a fork if we can avoid it.

PR #3596, opened 2023-07-02
Update package.json

CaptThomas:
Hey reviewers, thanks for a great tool. This PR is designed to fix https://github.com/advisories/GHSA-c2qf-rxjj-qqgw. I tested this locally using Windows 11 x64 using node v20.3.1 and npm 9.7.2 and it worked as expected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "object.values": "^1.1.6",
     "prop-types": "^15.8.1",
     "resolve": "^2.0.0-next.4",
-    "semver": "^6.3.0",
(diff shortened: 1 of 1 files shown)


ljharb (changes requested):
Unfortunately, it would be a breaking change.

ljharb:
It’s not a vulnerability here - like most transitive dep CVEs, it’s a false positive -  and we can’t upgrade because v7 drops support for engines we need to support. Duplicate of #3589.

PR #3597, opened 2023-07-07
fix(semver): update semver to latest version to correct vulnerability

blabute:
Snyk reported a vulnerability with this version of semver. See https://security.snyk.io/vuln/SNYK-JS-SEMVER-3247795. This change updates to the latest version to correct the issue.

It also updates the engine to be greater than 10 due to an issue in a check.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "object.values": "^1.1.6",
     "prop-types": "^15.8.1",
     "resolve": "^2.0.0-next.4",
-    "semver": "^6.3.0",
+    "semver": "^7.5.3",
     "string.prototype.matchall": "^4.0.8"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "eslint": "^3 || ^4 || ^5 || ^6 || ^7 || ^8"
   },
   "engines": {
-    "node": ">=4"
+    "node": ">=10"
   },
   "keywords": [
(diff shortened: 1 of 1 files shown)


blabute:
Seeing some other PRs failing due to this vulnerability: https://github.com/jsx-eslint/eslint-plugin-react/pull/3593

blabute:
Looks like https://iojs.org/ is also unavailable at the moment which is causing some checks to fail. Should we just update this url to be https://nodejs.org/ per https://github.com/nodejs/iojs.org/issues/432#issue-343377853?

@ljharb can you help with this?

ljharb:
It’s not a vulnerability here - like most transitive dep CVEs, it’s a false positive - and we can’t upgrade because v7 drops support for engines we need to support.

Duplicate of #3589.

ljharb:
Upgrading the engines would be a breaking change, as well, and that's just not something we'll likely ever do.

blabute:
> Upgrading the engines would be a breaking change, as well, and that's just not something we'll likely ever do.

Thanks for the reply! Node 4 [security support ended in April 2018](https://endoflife.date/nodejs). Are you thinking this package will always support those legacy versions of Node?

ljharb:
Yes, platform support has no bearing on ecosystem support.

Either way, the semver maintainers are backporting the fix to v6, so there’s nothing that needs to be done to address this false positive but wait.

PR #3603, opened 2023-07-19
[actions] Replace `set-output` commands with environment files

xt0rted:
These commands were deprecated awhile back, removing them will also get rid of the warnings in the run log.

- https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
- https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,7 +23,7 @@ jobs:
 
       - name: Get version from tag
         id: tag_name
-        run: echo ::set-output name=current_version::${GITHUB_REF#refs/tags/v}
+        run: echo "current_version=${GITHUB_REF#refs/tags/v}" >> "$GITHUB_OUTPUT"
         shell: bash
 
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -42,7 +42,10 @@ jobs:
           _links="${_links//$'\n'/'%0A'}"
           _links="${_links//$'\r'/'%0D'}"
           # Set output 'links' to $_links
-          echo "::set-output name=links::${_links}"
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
```
ljharb, line 45: are we confident this can't ever generate a string that would be invalid?

xt0rted: This is how they recommend it's done, so I sure hope so. Prior to what's in the docs now they were showing it done like so:

```bash
DELIMITER=$(uuidgen)
echo "links<<$DELIMITER" >> "$GITHUB_OUTPUT"
echo "$_links" >> "$GITHUB_OUTPUT"
echo "$DELIMITER" >> "$GITHUB_OUTPUT"
```

I've been doing it both ways with changelog content & json blobs and haven't had any issues yet.

Here's the PR where this was added https://github.com/github/docs/pull/22463 and the original comment where `uuidgen` was suggested (shows up in another sample on the same page still) https://github.com/github/docs/pull/15354#issuecomment-1055867931.

ljharb: uuidgen seems simpler than this, but fair enough either way

xt0rted: Yea, I'm not sure which way is "better" which is why I have both in my repos now 😆

ljharb (approved):
rebased and updated to use the uuidgen approach; i'll have to wait to test until the next release.

PR #3605, opened 2023-07-20
[Fix] `require-default-props`: fix config schema

### LINKED ISSUE #3604: [Bug]: `require-default-props` JSON schema is incorrect
(coming from https://github.com/Shinigami92/eslint-define-config/pull/213)

The JSON schema for the `require-default-props` rule incorrectly wraps the `enum` in an `allow`, resulting in a JSON schema that matches an object instead of a JSON schema that matches strings.

https://github.com/jsx-eslint/eslint-plugin-react/blob/31282dd27c8488efdb53967271d31c79d3072f74/lib/rules/require-default-props.js#L43-L52

The downstream result of the incorrect schema:
image

[Here’s](https://github.com/jsx-eslint/eslint-plugin-react/blob/31282dd27c8488efdb53967271d31c79d3072f74/lib/rules/jsx-sort-props.js#L370-L373) an example of another rule whose JSON schema is correctly written.

controversial:
Fix #3604 

Removes extraneous “allow” keyword from JSON schema definition. I couldn’t find any references to this keyword in JSON schema, nor documentation on its purpose, but it seems to break the TypeScript definitions generated in projects like [eslint-define-config](https://github.com/Shinigami92/eslint-define-config).

If the “allow” keyword has a purpose here that I’m not aware of, please feel free to close this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ## Unreleased
 
+### Fixed
+* [`require-default-props`]: fix config schema ([#3605][] @controversial)
+
+[#3605]: https://github.com/jsx-eslint/eslint-plugin-react/pull/3605
+
 ## [7.33.0] - 2023.07.19
 
 ### Added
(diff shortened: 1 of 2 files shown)


ljharb (approved):
This does match the intention in #3249 from what I can tell (also looking at the tests). Thanks!

PR #3608, opened 2023-07-21
[Fix] `jsx-curly-brace-presence`: make unwrapping single-expression template literals configurable

### LINKED ISSUE #3607: [Bug]: jsx-curly-brace-presence template literal fix can break things

### Description Overview

#3538 (see #3537) fixes a template literal with a single placeholder to just be that placeholder.

This can change the semantics of the prop:

```typescript
const foo = 123;
<Component foo={`${foo}`} />
```
casts `foo` to a string (and thus passes `"123"`), while the "fixed"
```typescript
const foo = 123;
<Component foo={foo} />
```
passes `foo` as a number `123`; when using TypeScript and `Component` expects `foo: string`, that's a type error. When not using TypeScript, if `Component` behaves differently with a string prop cf. a number prop, this changes that behavior.

This also stands for children:
```jsx
<span>An error occurred: {`${err}`}</span>
```
will be changed to 
```jsx
<span>An error occurred: {err}</span>
```
 which can be a type error if `err` is not a valid React child (and when not using TypeScript, it can cause e.g. `false` not to get rendered as text.

### Expected Behavior

Whether template literals with a single placeholder are unwrapped should be configurable.

v7.33.0

### eslint version

v8.45.0

### node version

v20.3.0

akx:
Fixes #3607 (makes #3538 opt-in)

cc @taozhou-glean

### DIFF
--- a/docs/rules/jsx-curly-brace-presence.md
+++ b/docs/rules/jsx-curly-brace-presence.md
@@ -20,7 +20,7 @@ You can pass in options to enforce the presence of curly braces on JSX props, ch
 
 ```js
 ...
-"react/jsx-curly-brace-presence": [<enabled>, { "props": <string>, "children": <string>, "propElementValues": <string> }]
+"react/jsx-curly-brace-presence": [<enabled>, { "props": <string>, "children": <string>, "propElementValues": <string>, "unwrapTemplateLiterals": <boolean> }]
 ...
 ```
 
@@ -47,7 +47,13 @@ If passed in the option to fix, this is how a style violation will get fixed
 
 - All fixing operations use double quotes.
 
-For examples:
+### `unwrapTemplateLiterals`
+
+- `true`: unwrap template literals that only have a single expression inside of them.
+  Since template literals return strings, this may cause changes in semantics, or type errors.
+- `false` (default): do not unwrap template literals that only have a single expression inside of them.
+
+## Examples
 
 Examples of **incorrect** code for this rule, when configured with `{ props: "always", children: "always" }`:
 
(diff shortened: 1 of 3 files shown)


docs/rules/jsx-curly-brace-presence.md
```diff
@@ -20,7 +20,7 @@ You can pass in options to enforce the presence of curly braces on JSX props, ch
 
 ```js
 ...
-"react/jsx-curly-brace-presence": [<enabled>, { "props": <string>, "children": <string>, "propElementValues": <string> }]
+"react/jsx-curly-brace-presence": [<enabled>, { "props": <string>, "children": <string>, "propElementValues": <string>, "unwrapTemplateLiterals": <boolean> }]
```
taozhou-glean, line 23: maybe `unwrapSingleExpressionOnlyTemplateLiterals` just to be more explicit ?

taozhou-glean (commented):
lgtm, defer to owner to approve

docs/rules/jsx-curly-brace-presence.md
```diff
@@ -47,7 +47,13 @@ If passed in the option to fix, this is how a style violation will get fixed
 
 - All fixing operations use double quotes.
 
-For examples:
+### `unwrapTemplateLiterals`
+
+- `true`: unwrap template literals that only have a single expression inside of them.
+  Since template literals return strings, this may cause changes in semantics, or type errors.
```
ljharb, line 53: this makes it an unsafe autofix, which means it can't ever exist.

ljharb (changes requested):
I think the proper fix here is just to revert #3538, but keep the test cases (without the autofixes).

PR #3611, opened 2023-07-23
fix: revert single template literal expression which might be intended usage

taozhou-glean:
This reverts https://github.com/jsx-eslint/eslint-plugin-react/pull/3538, while keeping the test case there for explanation and future breakage prevention.

The previous commit broke for intended type casting usage: https://github.com/jsx-eslint/eslint-plugin-react/issues/3607

Sorry for adding the change, did not think enough before making it, apparently this is definitely legit use case for casting other type to string, we might be able to keep this rule and limit it to string single template literal with typechecker, but just a bit overkill at this moment.

### DIFF
--- a/lib/rules/jsx-curly-brace-presence.js
+++ b/lib/rules/jsx-curly-brace-presence.js
@@ -131,10 +131,6 @@ module.exports = {
       return containsLineTerminators(text) && text.trim() === '';
     }
 
-    function isSingleExpressionTemplateLiteral(child) {
-      return child.type === 'TemplateLiteral' && child.expressions.length === 1 && child.quasis.map((quasis) => quasis.value.raw).join('') === '';
-    }
-
     function wrapNonHTMLEntities(text) {
       const HTML_ENTITY = '<HTML_ENTITY>';
       const withCurlyBraces = text.split(HTML_ENTITY_REGEX()).map((word) => (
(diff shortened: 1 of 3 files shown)


lib/rules/jsx-curly-brace-presence.js
```diff
@@ -131,10 +131,6 @@ module.exports = {
       return containsLineTerminators(text) && text.trim() === '';
     }
 
-    function isSingleExpressionTemplateLiteral(child) {
-      return child.type === 'TemplateLiteral' && child.expressions.length === 1 && child.quasis.map((quasis) => quasis.value.raw).join('') === '';
```
taozhou-glean, line 135: we might be able to limit this rule to `string` single template literal with typechecker, but just a bit overkill at this moment

ljharb:
This has nothing to do with TS; JS also can use propTypes or runtime type checks - it's precisely as critical a use case in JS as it is in TS.

Honest mistake, though - I missed this too when reviewing your initial PR.

ljharb (approved):

taozhou-glean:
> This has nothing to do with TS; JS also can use propTypes or runtime type checks - it's precisely as critical a use case in JS as it is in TS.
> 
> Honest mistake, though - I missed this too when reviewing your initial PR.

right, updated the description.

tests/lib/rules/jsx-curly-brace-presence.js
```diff
@@ -469,13 +469,16 @@ ruleTester.run('jsx-curly-brace-presence', rule, {
...
-      options: [{ props: 'never' }],
+      code: '<App label={`${label}`} />',
+      output: '<App label={label} />',
+      options: ['never'],
     },
     {
-      code: '<App>{`${label}${suffix}`}</App>',
-      options: [{ children: 'never' }],
+      code: '<App>{`${label}`}</App>',
+      output: '<App>{label}</App>',
```
taozhou-glean, line 480: any particular reason why we keep `output` ?

ljharb: no, actually this line should be removed

akx:
Thanks, I didn't have the bandwidth over the weekend to do this in #3608 👍

lucapollani:
Will this fix be released soon?

ljharb:
I'm sure it will.

PR #3614, opened 2023-08-05
[Fix] `no-deprecated`: false positive on commonjs import in `no-deprecated` rule

### LINKED ISSUE #3606: [Bug]: false positive on commonjs imports for no-deprecated

### Description Overview

I'm getting a `no-deprecated` eslint error for the following:

```js
const { renderToString } = require('react-dom/server');
```

None for ESM imports though:

```js
import { renderToString } from 'react-dom/server';
```

### Expected Behavior

This should not trigger a `no-deprecated` linting error:

```js
const { renderToString } = require('react-dom/server');
```

v7.33.0

### eslint version

v8.45.0

### node version

v18.12.1

akulsr0:
Fixes #3606

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ## Unreleased
 
+### Fixed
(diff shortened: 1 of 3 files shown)


ljharb (approved):
thanks! i ended up combining the check with the previous iteration.

Issue #3578, opened 2023-05-15
Feature request: `sort-prop-types` should work on TS Props types


tylerlaprade:
### Is there an existing issue for this?

- [X] I have searched the existing issues and my issue is unique
- [X] My issue appears in the command-line and not only in the text editor

### Description Overview

It currently only works on Class components, which are not in favor in the codebases in which I work. The rule doesn't trigger any warnings on properties on a functional component.

### Expected Behavior

It should behave the same on functional components as it does on class components.

### eslint-plugin-react version

v7.32.2

### eslint version

8.40.0

### node version

v16.17.1

kshirish:
Facing the same issue. @ljharb 

```
import React from 'react';
import PropTypes from 'prop-types';
import classNames from 'classnames';

const variantMap = {
  filled: {
    primary: 'ui-text-white ui-bg-primary',
    secondary: 'ui-text-white ui-bg-secondary',
  },
  outlined: {
    primary: 'ui-border ui-border-primary ui-text-primary',
    secondary: 'ui-border ui-border-secondary ui-text-secondary',
  },
  text: { primary: 'ui-text-primary', secondary: 'ui-text-secondary' },
};

const sizeMap = {
  sm: 'ui-px-2 ui-py-1 ui-text-xs',
  md: 'ui-px-3 ui-py-2 ui-text-sm',
  lg: 'ui-px-4 ui-py-3 ui-text-base',
};

const Button = ({
  variant = 'filled',
  color = 'primary',
  size = 'md',
  children,
  disabled = false,
  className,
  ...restProps
}) => {
  const variantClass = variantMap[variant][color];
  const sizeClass = sizeMap[size];

  return (
    <button
      className={classNames(
        'ui-rounded',
        variantClass,
        sizeClass,
        {
          'ui-pointer-events-none': disabled,
        },
        className,
      )}
      type="button"
      {...restProps}
    >
      {children}
    </button>
  );
};

Button.propTypes = {
  variant: PropTypes.oneOf(['filled', 'outlined', 'text']),
  color: PropTypes.oneOf(['primary', 'secondary']),
  size: PropTypes.oneOf(['sm', 'md', 'lg']),
  disabled: PropTypes.bool,
  className: PropTypes.string,
  children: PropTypes.oneOfType([
    PropTypes.arrayOf(PropTypes.node),
    PropTypes.node,
  ]).isRequired,
};

Button.displayName = 'Button';

export default Button;

```


[...]

PR #3615, opened 2023-08-09
[New] `sort-prop-types`: give errors on TS types

akulsr0:
This PR is an attempt to solve #3578

### DIFF
--- a/lib/rules/sort-prop-types.js
+++ b/lib/rules/sort-prop-types.js
@@ -52,6 +52,9 @@ module.exports = {
         sortShapeProp: {
           type: 'boolean',
         },
+        checkTypes: {
+          type: 'boolean',
+        },
       },
       additionalProperties: false,
     }],
@@ -64,8 +67,14 @@ module.exports = {
     const ignoreCase = configuration.ignoreCase || false;
     const noSortAlphabetically = configuration.noSortAlphabetically || false;
     const sortShapeProp = configuration.sortShapeProp || false;
+    const checkTypes = configuration.checkTypes || false;
+
+    const typeAnnotations = new Map();
 
     function getKey(node) {
+      if (node.type === 'ObjectTypeProperty') {
+        return context.getSourceCode().getFirstToken(node).value;
+      }
       if (node.key && node.key.value) {
         return node.key.value;
       }
@@ -215,7 +224,32 @@ module.exports = {
       }
     }
 
-    return {
+    function handleFunctionComponent(node) {
+      const firstArg = node.params[0].typeAnnotation && node.params[0].typeAnnotation.typeAnnotation;
+      if (firstArg && firstArg.type === 'TSTypeReference') {
+        const propType = typeAnnotations.get(firstArg.typeName.name)
+          && typeAnnotations.get(firstArg.typeName.name)[0];
+        if (propType && propType.members) {
+          checkSorted(propType.members);
+        }
+      } else if (firstArg && firstArg.type === 'TSTypeLiteral') {
+        if (firstArg.members) {
+          checkSorted(firstArg.members);
+        }
+      } else if (firstArg && firstArg.type === 'GenericTypeAnnotation') {
+        const propType = typeAnnotations.get(firstArg.id.name)
+          && typeAnnotations.get(firstArg.id.name)[0];
+        if (propType && propType.properties) {
+          checkSorted(propType.properties);
+        }
+      } else if (firstArg && firstArg.type === 'ObjectTypeAnnotation') {
+        if (firstArg.properties) {
+          checkSorted(firstArg.properties);
+        }
+      }
+    }
+
+    return Object.assign({
       CallExpression(node) {
         if (!sortShapeProp || !isShapeProp(node) || !(node.arguments && node.arguments[0])) {
           return;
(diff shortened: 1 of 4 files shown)


ljharb:
Seems like there's some uncovered added lines; can you add tests that hit them?

ljharb:
Looks like there's still a few failing tests.

akulsr0:
Yep, tried to debug this. It seems for `node 8/9` and `eslint 6`, the `variableUtil.findVariableByName` util is unable to grab the type node. Not sure why this is happening

akulsr0:
@ljharb I have fixed the previous failing tests.

lib/rules/sort-prop-types.js
```diff
@@ -215,6 +217,22 @@ module.exports = {
       }
     }
 
+    function handleFunctionComponent(node) {
+      const firstArg = node.params[0].typeAnnotation && node.params[0].typeAnnotation.typeAnnotation;
+      if (firstArg && firstArg.type === 'TSTypeReference') {
+        const propType = variableUtil.findVariableByName(context, firstArg.typeName.name)
+        || (typeAnnotations.get(firstArg.typeName.name)
+          && typeAnnotations.get(firstArg.typeName.name)[0]);
```
ljharb, line 225:
```suggestion
          || (
            typeAnnotations.get(firstArg.typeName.name)
            && typeAnnotations.get(firstArg.typeName.name)[0]
          );
```
looks like || fallback still isn't covered by a test case?

akulsr0: I am not sure why it comes as uncovered, this fallback is essentially the fix for previous test case failure which was happening for `node 8/9` and `eslint 6`

ljharb: ah, perhaps it's codecov maxing out on what it can merge 🤔

akulsr0: Fixed it, all checks are passing now.

ljharb:
ok, I rebased this, but I think this needs to go behind an option, otherwise it would be a breaking change.

lib/rules/sort-prop-types.js
```diff
@@ -52,6 +52,9 @@ module.exports = {
         sortShapeProp: {
           type: 'boolean',
         },
+        allowTypescript: {
```
ljharb, line 55: let's call it `checkTypes`?

tests/lib/rules/sort-prop-types.js
```diff
@@ -480,6 +480,19 @@ ruleTester.run('sort-prop-types', rule, {
...
+      code: `
+        type Props = {
+          zzz: string;
+          aaa: string;
+        }
+        function Foo(props: Props) {
+          return null;
+        }
+      `,
+      features: ['ts', 'no-babel'],
```
ljharb, line 494:
```suggestion
      features: ['types'],
```
this should work for flow too, no?

lib/rules/sort-prop-types.js
akulsr0: ok sure

tests/lib/rules/sort-prop-types.js
akulsr0: `'no-babel'` would be required.

ljharb: why? babel is what parses flow.

lib/rules/sort-prop-types.js
```diff
@@ -261,6 +285,29 @@ module.exports = {
...
+          const currentNode = [].concat(
+            typeAnnotations.get(node.id.name) || [],
+            node.typeAnnotation
+          );
+          typeAnnotations.set(node.id.name, currentNode);
+        }
+      },
+
+      FunctionDeclaration: handleFunctionComponent,
+      ArrowFunctionExpression: handleFunctionComponent,
```
ljharb, line 310: maybe:
```suggestion
      FunctionDeclaration: checkTypes && handleFunctionComponent,
      ArrowFunctionExpression: checkTypes && handleFunctionComponent,
```

lib/rules/sort-prop-types.js
```diff
@@ -215,6 +221,24 @@ module.exports = {
       }
     }
 
+    function handleFunctionComponent(node) {
+      if (!checkTypes) {
+        return;
+      }
```
ljharb, line 227: instead of:
```suggestion
```

tests/lib/rules/sort-prop-types.js
akulsr0: The error type for both case is different (`TSPropertySignature` and `ObjectTypeProperty`), which I have to mention in test case.

akulsr0: Oops, my bad mentioning the error type isn't mandatory. Fixed it.

lib/rules/sort-prop-types.js
akulsr0: Won't this give an error if `checkTypes` is falsy?

akulsr0: Both (`checkTypes && handleFunctionComponent` and `checkTypes ? handleFunctionComponent : undefined`) of the above gives error, is following fine?
```js
FunctionDeclaration: checkTypes ? handleFunctionComponent : () => {},
ArrowFunctionExpression: checkTypes ? handleFunctionComponent : () => {},
```

akulsr0: Doing the `Object.assign` way

akulsr0: @ljharb this was done!

lib/rules/sort-prop-types.js
```diff
@@ -261,6 +295,39 @@ module.exports = {
...
+        if (node.right.type === 'ObjectTypeAnnotation') {
+          const currentNode = [].concat(
+            typeAnnotations.get(node.id.name) || [],
+            node.right
+          );
+          typeAnnotations.set(node.id.name, currentNode);
+        }
+      },
+
+      TSTypeAliasDeclaration(node) {
```
ljharb, line 319: i think all of these can also be moved to the "checkTypes" branch below?

tests/lib/rules/sort-prop-types.js
ljharb, line 494:
```suggestion
      features: ['types'],
```

akulsr0: this was done

lib/rules/sort-prop-types.js
akulsr0: this was done

ljharb (approved):

PR #3623, opened 2023-08-25
[Fix] `jsx-no-leaked-render`: preserve RHS parens for multiline jsx element while fixing

### LINKED ISSUE #3551: [Bug]: Excessive code changes done by --fix react/jsx-no-leaked-render

### Description Overview

When using `'react/jsx-no-leaked-render': ['error', { validStrategies: ['coerce', 'ternary'] }],` and running `eslint --fix` the resulting code changes go beyond what's required for the actual fix by also removing parentheses, causing excessive code diffs.

Screenshot 2023-03-17 at 9 03 07 PM

I would expect the second hunk to be a single line change, just like the first hunk. Running this fix across a large codebase is unfeasible because the changes cannot be human-verified.

`eslint --fix` with `'react/jsx-no-leaked-render': ['error', { validStrategies: ['coerce', 'ternary'] }],`

### Expected Behavior

I'd expect no parentheses would be removed and no code-style to change. Other than the `!!` required to fix the issue, nothing else at all should change in the code.

Screenshot 2023-03-17 at 9 09 34 PM

7.21.5

### eslint version

8.35.0

### node version

16.13.1

akulsr0:
Fixes #3551

### DIFF
--- a/lib/rules/jsx-no-leaked-render.js
+++ b/lib/rules/jsx-no-leaked-render.js
@@ -82,6 +82,15 @@ function ruleFixer(context, fixStrategy, fixer, reportedNode, leftNode, rightNod
     if (rightNode.type === 'ConditionalExpression') {
       return fixer.replaceText(reportedNode, `${newText} && (${rightSideText})`);
     }
+    if (rightNode.type === 'JSXElement') {
+      const rightSideTextLines = rightSideText.split('\n');
+      if (rightSideTextLines.length > 1) {
+        const rightSideTextLastLine = rightSideTextLines[rightSideTextLines.length - 1];
+        const indentSpacesStart = ' '.repeat(rightSideTextLastLine.search(/\S/));
+        const indentSpacesClose = ' '.repeat(rightSideTextLastLine.search(/\S/) - 2);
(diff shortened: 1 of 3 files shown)


lib/rules/jsx-no-leaked-render.js
```diff
@@ -82,6 +82,15 @@ function ruleFixer(context, fixStrategy, fixer, reportedNode, leftNode, rightNod
     if (rightNode.type === 'ConditionalExpression') {
       return fixer.replaceText(reportedNode, `${newText} && (${rightSideText})`);
     }
+    if (rightNode.type === 'JSXElement') {
+      const rightSideTextLines = rightSideText.split('\n');
+      if (rightSideTextLines.length > 1) {
+        const rightSideTextLastLine = rightSideTextLines[rightSideTextLines.length - 1];
+        const indentSpacesStart = Array(rightSideTextLastLine.search(/\S/)).fill(' ').join('');
```
ljharb, line 89: let’s use string repeat here instead of array/fill/join, esp since Array() should never be used.

ljharb (approved):
Looks great, thanks!

PR #3629, opened 2023-09-04
[Fix] `@typescript-eslint` v6, v7, v8 use typeArguments with fallback to typeParameters

### LINKED ISSUE #3602: [Bug]: `@typescript-eslint` v6 throws deprecation warnings

### Description Overview

Using `@typescript-eslint/parser` 6.x throws a warning on the command line saying `DeprecationWarning: The 'typeParameters' property is deprecated on CallExpression nodes. Use 'typeArguments' instead`.

My project is an ejected CRA config; this prints out on the command line every time that I run `eslint src` on my project after I upgraded my `@typescript-eslint` dependencies to use the latest v6.

See https://typescript-eslint.io/linting/troubleshooting/#the-key-property-is-deprecated-on-type-nodes-use-key-instead-warnings for more info.

### Expected Behavior

A warning should not be printed for deprecations.

v7.32.2

### eslint version

v8.45.0

### node version

v18.16.1

HenryBrown0:
Follows the [guide](https://typescript-eslint.io/linting/troubleshooting/#the-key-property-is-deprecated-on-type-nodes-use-key-instead-warnings) from `@typescript-eslint` to upgrade to `v6`, by adding a small utility function `getTypeArguments` which returns the `typeArguments` or falls back to `typeParameters`.

Closes #3602

### DIFF
--- a/lib/rules/boolean-prop-naming.js
+++ b/lib/rules/boolean-prop-naming.js
@@ -259,14 +259,16 @@ module.exports = {
         return;
       }
 
-      const annotationTypeParams = component.node.parent.id.typeAnnotation.typeAnnotation.typeParameters;
+      const annotationTypeArguments = propsUtil.getTypeArguments(
+        component.node.parent.id.typeAnnotation.typeAnnotation
+      );
       if (
-        annotationTypeParams && (
-          annotationTypeParams.type === 'TSTypeParameterInstantiation'
-          || annotationTypeParams.type === 'TypeParameterInstantiation'
+        annotationTypeArguments && (
+          annotationTypeArguments.type === 'TSTypeParameterInstantiation'
+          || annotationTypeArguments.type === 'TypeParameterInstantiation'
         )
       ) {
-        return annotationTypeParams.params.find(
+        return annotationTypeArguments.params.find(
           (param) => param.type === 'TSTypeReference' || param.type === 'GenericTypeAnnotation'
         );
       }
--- a/lib/rules/jsx-first-prop-new-line.js
+++ b/lib/rules/jsx-first-prop-new-line.js
@@ -7,6 +7,7 @@
 
 const docsUrl = require('../util/docsUrl');
 const report = require('../util/report');
+const propsUtil = require('../util/props');
 
 // ------------------------------------------------------------------------------
 // Rule Definition
@@ -55,7 +56,8 @@ module.exports = {
               report(context, messages.propOnNewLine, 'propOnNewLine', {
                 node: decl,
                 fix(fixer) {
-                  return fixer.replaceTextRange([(node.typeParameters || node.name).range[1], decl.range[0]], '\n');
+                  const nodeTypeArguments = propsUtil.getTypeArguments(node);
+                  return fixer.replaceTextRange([(nodeTypeArguments || node.name).range[1], decl.range[0]], '\n');
                 },
               });
             }
--- a/lib/rules/jsx-max-depth.js
+++ b/lib/rules/jsx-max-depth.js
@@ -91,7 +91,7 @@ module.exports = {
     function findJSXElementOrFragment(startNode, name, previousReferences) {
       function find(refs, prevRefs) {
         for (let i = refs.length - 1; i >= 0; i--) {
-          if (has(refs[i], 'writeExpr')) {
+          if (typeof refs[i].writeExpr !== 'undefined') {
             const writeExpr = refs[i].writeExpr;
 
             return (jsxUtil.isJSX(writeExpr)
--- a/lib/util/propTypes.js
+++ b/lib/util/propTypes.js
@@ -639,8 +639,8 @@ module.exports = function propTypesInstructions(context, components, utils) {
         typeName = node.typeName.name;
         const leftMostName = getLeftMostTypeName(node.typeName);
         const shouldTraverseTypeParams = genericReactTypesImport.has(leftMostName);
-        const nodeTypeParams = node.typeParameters;
-        if (shouldTraverseTypeParams && nodeTypeParams && nodeTypeParams.length !== 0) {
+        const nodeTypeArguments = propsUtil.getTypeArguments(node);
+        if (shouldTraverseTypeParams && nodeTypeArguments && nodeTypeArguments.length !== 0) {
           // All react Generic types are derived from:
           // type PropsWithChildren<P> = P & { children?: ReactNode | undefined }
           // So we should construct an optional children prop
@@ -662,7 +662,7 @@ module.exports = function propTypesInstructions(context, components, utils) {
           const idx = genericTypeParamIndexWherePropsArePresent[
             leftMostName !== rightMostName ? rightMostName : importedName
           ];
-          const nextNode = nodeTypeParams.params[idx];
+          const nextNode = nodeTypeArguments.params[idx];
           this.visitTSNode(nextNode);
           return;
         }
@@ -759,10 +759,10 @@ module.exports = function propTypesInstructions(context, components, utils) {
 
     convertReturnTypeToPropTypes(node, rootNode) {
       // ReturnType<T> should always have one parameter
-      const nodeTypeParams = node.typeParameters;
-      if (nodeTypeParams) {
-        if (nodeTypeParams.params.length === 1) {
-          let returnType = nodeTypeParams.params[0];
+      const nodeTypeArguments = propsUtil.getTypeArguments(node);
+      if (nodeTypeArguments) {
+        if (nodeTypeArguments.params.length === 1) {
+          let returnType = nodeTypeArguments.params[0];
           // This line is trying to handle typescript-eslint-parser
           // typescript-eslint-parser TSTypeQuery is wrapped by TSTypeReference
           if (astUtil.isTSTypeReference(returnType)) {
@@ -794,9 +794,9 @@ module.exports = function propTypesInstructions(context, components, utils) {
                     case 'ObjectExpression':
                       iterateProperties(context, res.properties, (key, value, propNode) => {
                         if (propNode && astUtil.isCallExpression(propNode.argument)) {
-                          const propNodeTypeParams = propNode.argument.typeParameters;
-                          if (propNodeTypeParams) {
-                            this.visitTSNode(propNodeTypeParams);
+                          const propNodeTypeArguments = propsUtil.getTypeArguments(propNode.argument);
+                          if (propNodeTypeArguments) {
+                            this.visitTSNode(propNodeTypeArguments);
                           } else {
                             // Ignore this CallExpression return value since it doesn't have any typeParameters to let us know it's types.
                             this.shouldIgnorePropTypes = true;
@@ -816,8 +816,8 @@ module.exports = function propTypesInstructions(context, components, utils) {
                       });
                       break;
                     case 'CallExpression':
-                      if (res.typeParameters) {
-                        this.visitTSNode(res.typeParameters);
+                      if (propsUtil.getTypeArguments(res)) {
+                        this.visitTSNode(propsUtil.getTypeArguments(res));
                       } else {
                         // Ignore this CallExpression return value since it doesn't have any typeParameters to let us know it's types.
                         this.shouldIgnorePropTypes = true;
@@ -1002,9 +1002,9 @@ module.exports = function propTypesInstructions(context, components, utils) {
         break;
       case 'GenericTypeAnnotation':
         if (propTypes.id.name === '$ReadOnly') {
-          const propTypeParams = propTypes.typeParameters;
+          const propTypeArguments = propsUtil.getTypeArguments(propTypes);
           ignorePropsValidation = declarePropTypesForObjectTypeAnnotation(
-            propTypeParams.params[0],
+            propTypeArguments.params[0],
             declaredPropTypes
           );
         } else {
@@ -1041,11 +1041,16 @@ module.exports = function propTypesInstructions(context, components, utils) {
       return;
     }
 
+    let propTypesArguments = null;
+    if (node.parent) {
+      propTypesArguments = propsUtil.getTypeArguments(node.parent);
+    }
+
     if (
       node.parent
       && node.parent.callee
-      && node.parent.typeParameters
-      && node.parent.typeParameters.params
+      && propTypesArguments
+      && propTypesArguments.params
       && (
         node.parent.callee.name === 'forwardRef' || (
           node.parent.callee.object
@@ -1055,9 +1060,8 @@ module.exports = function propTypesInstructions(context, components, utils) {
         )
       )
     ) {
-      const propTypesParams = node.parent.typeParameters;
       const declaredPropTypes = {};
-      const obj = new DeclarePropTypesForTSTypeAnnotation(propTypesParams.params[1], declaredPropTypes, rootNode);
+      const obj = new DeclarePropTypesForTSTypeAnnotation(propTypesArguments.params[1], declaredPropTypes, rootNode);
       components.set(node, {
         declaredPropTypes: obj.declaredPropTypes,
         ignorePropsValidation: obj.shouldIgnorePropTypes,
@@ -1103,7 +1107,7 @@ module.exports = function propTypesInstructions(context, components, utils) {
       if (
         annotation
         && annotation.type !== 'TSTypeReference'
-        && annotation.typeParameters == null
+        && propsUtil.getTypeArguments(annotation) == null
       ) {
         return;
       }
--- a/lib/util/propTypes.js
+++ b/lib/util/propTypes.js
@@ -41,7 +41,8 @@ function isFunctionType(node) {
  */
 function isSuperTypeParameterPropsDeclaration(node) {
   if (node && (node.type === 'ClassDeclaration' || node.type === 'ClassExpression')) {
-    if (node.superTypeParameters && node.superTypeParameters.params.length > 0) {
+    const parameters = propsUtil.getSuperTypeArguments(node);
+    if (parameters && parameters.params.length > 0) {
       return true;
     }
   }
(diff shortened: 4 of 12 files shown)


lib/util/propTypes.js
```diff
@@ -1011,9 +1017,9 @@ module.exports = function propTypesInstructions(context, components, utils) {
         )
       )
     ) {
-      const propTypes = node.parent.typeParameters.params[1];
+      const propTypes = node.parent.typeArguments || node.parent.typeParameters;
```
ljharb, line 1020: i'd probably call this typeParams, since propTypes is the second param of it.

ljharb (commented):
Seems good, altho there's an uncovered line.

lib/rules/boolean-prop-naming.js
```diff
@@ -240,21 +240,25 @@ module.exports = {
...
-        && component.node.parent.id.type === 'Identifier'
-        && component.node.parent.id.typeAnnotation
-        && component.node.parent.id.typeAnnotation.typeAnnotation
-        && component.node.parent.id.typeAnnotation.typeAnnotation.typeParameters
-        && (
-          component.node.parent.id.typeAnnotation.typeAnnotation.typeParameters.type === 'TSTypeParameterInstantiation'
-          || component.node.parent.id.typeAnnotation.typeAnnotation.typeParameters.type === 'TypeParameterInstantiation'
-        )
+        !component.node.parent
+        || component.node.parent.type !== 'VariableDeclarator'
```
ljharb, line 244: must we invert the condition? that makes for a noisier diff.

HenryBrown0: I want to exit early here. I could invert the expression with `!(...statement...)` to have a smaller diff, but that feels dirty. Thoughts?

ljharb: it'd probably be easier to read the diff with a single commit that inverts the condition, but changes nothing else; and then the rest of this PR on top of it?

HenryBrown0: Ah I understand, this commit is very close to doing that; https://github.com/jsx-eslint/eslint-plugin-react/pull/3629/commits/0f45d976eb6f97b2cad0e1f72f2a802ea3aab657

HenryBrown0: If it's helpful, I can open smaller prs for each of the files touched?

ljharb: sure, that works too - or maybe one PR containing solely refactors (that don't change any behavior). then i can land that, and this can be rebased on top of it.

henkkasoft:
Hopefully this is fixed and merged soon.

ljharb:
@HenryBrown0 any update on those smaller PRs?

HenryBrown0:
> @HenryBrown0 any update on those smaller PRs?

Sorry I got side tracked, I'll try getting some tests written in the next few days

ljharb:
Can we add some tests that would have failed without these typeArguments checks? That may require adding tests that run on typescript-eslint v6.

ljharb:
Fair, I'll take that (and rebase this) assuming tests pass :-) thanks!

osdiab:
looking forward to this!

ljharb (approved):

ljharb:
@HenryBrown0 unfortunately a number of tests are failing. any idea why?

HenryBrown0:
> @HenryBrown0 unfortunately a number of tests are failing. any idea why?

Unit tests are failing on master for me, has a dependency changed?

It appears only minor version of Node.js 6 are failing, e.g. 18.6.8 and 19.6.8. I suspect this is a red herring as the matrix build reports using different node versions;
- 18.5.9 uses 18.19.0 and passes
- 18.6.8 uses 18.19.0 and fails
Not sure why this is happening

ljharb:
Thanks, in that case I'll check out master and report back.

ljharb:
@HenryBrown0 tests seem to be passing on master https://github.com/jsx-eslint/eslint-plugin-react/actions (i checked on my fork before pushing to this repo, so i expect these to pass).

tests/helpers/parsers.js
```diff
@@ -37,7 +37,7 @@ const parsers = {
   BABEL_ESLINT: path.join(__dirname, NODE_MODULES, 'babel-eslint'),
   '@BABEL_ESLINT': path.join(__dirname, NODE_MODULES, '@babel/eslint-parser'),
   TYPESCRIPT_ESLINT: path.join(__dirname, NODE_MODULES, 'typescript-eslint-parser'),
-  '@TYPESCRIPT_ESLINT': path.join(__dirname, NODE_MODULES, '@typescript-eslint/parser'),
+  '@TYPESCRIPT_ESLINT': path.join(__dirname, NODE_MODULES, '@typescript-eslint/parser/dist'),
```
HenryBrown0, line 40: `@typescript-eslint/parser` removed the `main` field in `v6`, which pointed to `dist/index.js`. Not sure how tests were passing before, as this would have been wrong for all of v6.

HenryBrown0: v6 supports node versions `^16.0.0 || >=18.0.0`, but even if I'm running this locally with node 18 it still fails to find the export

ljharb: interesting, indeed node 18 should work - line 15 in https://unpkg.com/browse/@typescript-eslint/parser@6.19.0/package.json points there.

lib/rules/jsx-first-prop-new-line.js
```diff
@@ -54,7 +54,7 @@ module.exports = {
               report(context, messages.propOnNewLine, 'propOnNewLine', {
                 node: decl,
                 fix(fixer) {
-                  return fixer.replaceTextRange([(node.typeParameters || node.name).range[1], decl.range[0]], '\n');
+                  return fixer.replaceTextRange([(node.typeArguments || node.typeParameters || node.name).range[1], decl.range[0]], '\n');
```
aslilac, line 57: having `node.typeParameters` in the middle here will trigger the deprecation warning if it should ever fall back to `node.name`. is that anything to be concerned about?

ljharb: good point - so in v6, `typeArguments` will be conditionally present, but `typeParameters` unconditionally a warning getter?

JoshuaKGoldberg:
👋 @HenryBrown0 do you still have time to work on this? We're at v7 in typescript-eslint (https://typescript-eslint.io/blog/announcing-typescript-eslint-v7) and working on v8 - which will remove the old properties altogether.

Even if there are still some cases that log for deprecated property access (https://github.com/jsx-eslint/eslint-plugin-react/pull/3629/files#r1511579466) I personally would very much like to see this PR merged. That way the `v8` branch of typescript-eslint can use the impacted rules. Want any help? ❤️

package.json
JoshuaKGoldberg, line 1: Once https://github.com/jsx-eslint/eslint-plugin-react/pull/3629/files#r1454015312 is fixed, we get this lovely crash on any rule that uses `@typescript-eslint/parser`:

```plaintext
TypeError: import_typescript.default.isTokenKind is not a function
      at forEachToken (/Users/josh/repos/eslint-plugin-react/node_modules/ts-api-utils/lib/index.cjs:260:35)
      at Object.forEachComment (/Users/josh/repos/eslint-plugin-react/node_modules/ts-api-utils/lib/index.cjs:309:10)
      at convertComments (/Users/josh/repos/eslint-plugin-react/node_modules/@typescript-eslint/typescript-estree/dist/convert-comments.js:40:13)
      at astConverter (/Users/josh/repos/eslint-plugin-react/node_modules/@typescript-eslint/typescript-estree/dist/ast-converter.js:56:66)
      at parseAndGenerateServices (/Users/josh/repos/eslint-plugin-react/node_modules/@typescript-eslint/typescript-estree/dist/parser.js:170:66)
      at Object.parseForESLint (/Users/josh/repos/eslint-plugin-react/node_modules/@typescript-eslint/parser/dist/parser.js:93:80)
      at Object.parseForESLint (/Users/josh/repos/eslint-plugin-react/node_modules/eslint/lib/rule-tester/rule-tester.js:314:36)
      at parse (/Users/josh/repos/eslint-plugin-react/node_modules/eslint/lib/linter/linter.js:834:22)
      at Linter._verifyWithoutProcessors (/Users/josh/repos/eslint-plugin-react/node_modules/eslint/lib/linter/linter.js:1312:33)
      at Linter.verify (/Users/josh/repos/eslint-plugin-react/node_modules/eslint/lib/linter/linter.js:1454:57)
      at runRuleForItem (/Users/josh/repos/eslint-plugin-react/node_modules/eslint/lib/rule-tester/rule-tester.js:833:35)
      ...
```

This is because TypeScript is on an old version that doesn't support the default export:

https://github.com/jsx-eslint/eslint-plugin-react/blob/4467db503e38b9356517cf6926d11be544ccf4b1/package.json#L77

I tried updating the devDependency to `^4.0.2` and then the error becamse:

```ts
TypeError: ts.getModifierFlags is not a function
      at Object.hasStaticModifierFlag (/Users/josh/repos/eslint-plugin-react/node_modules/typescript-estree/lib/node-utils.js:403:21)
      at convert (/Users/josh/repos/eslint-plugin-react/node_modules/typescript-estree/lib/convert.js:996:29)
      at convertChild (/Users/josh/repos/eslint-plugin-react/node_modules/typescript-estree/lib/convert.js:77:12)
      ...
```

...from the deprecated `typescript-estree` package. Which, along with `typescript-eslint-parser`, don't support the latest TypeScript versions. So they can't be used in environments that have `typescript@4` (roughly - I haven't checked the exact range support).

https://github.com/jsx-eslint/eslint-plugin-react/blob/2e6b5578cf9d8df2a4d35b9f5f3d851cf07ff483/package.json#L78

When I set `const tsOld = false;` in `parsers.js`, then all tests pass with `typescript@4` and `@typescript-eslint/*@6`.

@ljharb I can vaguely guess the right solution is to change around when the `typescript-eslint-parser` tests should be run, but I'm not clear on what the exact change should be. Is this something you can do?

ljharb: I'll take a look - basically, we do some hijinks in our CI to install the right version combinations of things so it'll work.

pzimmermaninmo:
@HenryBrown0 I found something very similar and wanted to get your thoughts

I tested this PR with my codebase and it in fact fixed the original error `DeprecationWarning: The 'typeParameters' property is deprecated on CallExpression nodes. Use 'typeArguments' instead`.

However, I am now getting a very similar error where `superTypeParameters` is deprecated and `superTypeArguments` should be used instead.  Have you seen similar?  Does it sound like something that should be fixed here too?

HenryBrown0:
> @HenryBrown0 I found something very similar and wanted to get your thoughts
> 
> I tested this PR with my codebase and it in fact fixed the original error `DeprecationWarning: The 'typeParameters' property is deprecated on CallExpression nodes. Use 'typeArguments' instead`.
> 
> However, I am now getting a very similar error where `superTypeParameters` is deprecated and `superTypeArguments` should be used instead. Have you seen similar? Does it sound like something that should be fixed here too?

This is a question for @ljharb, if it's a deprecated warning I imagine the preference would be to get this in first and a pr after for `superTypeParameters`

pzimmermaninmo:
> > @HenryBrown0 I found something very similar and wanted to get your thoughts
> > I tested this PR with my codebase and it in fact fixed the original error `DeprecationWarning: The 'typeParameters' property is deprecated on CallExpression nodes. Use 'typeArguments' instead`.
> > However, I am now getting a very similar error where `superTypeParameters` is deprecated and `superTypeArguments` should be used instead. Have you seen similar? Does it sound like something that should be fixed here too?
> 
> This is a question for @ljharb, if it's a deprecated warning I imagine the preference would be to get this in first and a pr after for `superTypeParameters`

Definitely a bit of scope creep for sure.  I just figured I'd mention especially since its coming from the same `propTypes.js` file.

ljharb:
It'd be fine to include it in this PR, or a follow-on; but since this one still has lots of failing tests we probably shouldn't increase complexity yet :-)

hampustagerud:
@ljharb I read in https://github.com/jsx-eslint/eslint-plugin-react/issues/3602#issuecomment-2070961147 that you're open to pulling in additional changes to this PR, rather than opening another PR, so I'm here to submit my [branch](https://github.com/hampustagerud/eslint-plugin-react/tree/support-%40typescript-eslint/parser-v6-8) which updates the workflows to pass the tests. In order to get the workflows to run properly I had to disable legacy peer dependency installation for jobs running v6+ of the parser, since it requires it's peer dependencies (typescript in this case) to be installed correctly.

I have also added v7 and v8 to the test matrix (you can check the [runs](https://github.com/hampustagerud/eslint-plugin-react/actions) for more details), since that is kind of blocking our Eslint 9 upgrade right now 🙂 https://github.com/jsx-eslint/eslint-plugin-react/pull/3629#issuecomment-2116022056 turned out to be an issue, but only very minor changes required in the code 👍

hampustagerud:
I have now rebased my branch so it's up to date with the latest changes in master 👍 

Please let me know if there's more I can help out with to land this @ljharb! 🙂

ljharb:
I've pulled in your changes; let's see how the tests fare :-)

lib/rules/jsx-max-depth.js
```diff
@@ -91,7 +91,7 @@ module.exports = {
     function findJSXElementOrFragment(startNode, name, previousReferences) {
       function find(refs, prevRefs) {
         for (let i = refs.length - 1; i >= 0; i--) {
-          if (has(refs[i], 'writeExpr')) {
+          if (has(refs[i], 'writeExpr') && refs[i].writeExpr) {
```
ljharb, line 94: when is the property present but falsy?

package.json
```diff
@@ -54,7 +54,7 @@
     "@types/eslint": "=7.2.10",
     "@types/estree": "0.0.52",
     "@types/node": "^4.9.5",
-    "@typescript-eslint/parser": "^2.34.0 || ^3.10.1 || ^4.0.0 || ^5.0.0",
+    "@typescript-eslint/parser": "^2.34.0 || ^3.10.1 || ^4.0.0 || ^5.0.0 || ^6.0.0 || ^7.0.0 || ^8.0.0",
```
ljharb, line 57:
```suggestion
    "@typescript-eslint/parser": "^2.34.0 || ^3.10.1 || ^4 || ^5 || ^6.20 || ^7.14.1 || ^8.4",
```

tests/helpers/parsers.js
ljharb: any update here?

ljharb (commented):
Overall LGTM!

tests/helpers/parsers.js
hampustagerud: Short version: I added another commit which replaces the path joining with `require.resolve` 🙂 If it needs to be `path.join` there needs to be some edge case handling here.

Longer version: I could get this working using both `/dist` and `dist/index.js` but not without any of those. It kind of makes sense why this doesn't work with newer versions of `@typescript-eslint/parser` (with an exports field) if you look at the [resolution logic node uses](https://nodejs.org/api/modules.html#all-together). It seems like the exports field are only taken into consideration when code is required as node modules, which node doesn't do if the import path is a file system path (starts with `./`, `../` or `/`). If a file system path is provided, it will only try loading it as a file or directory. When loading as a directory only the `main` field is taken into account, not the `exports` field.

So when just providing `path.join(__dirname, NODE_MODULES, '@typescript-eslint/parser` as the path it will get an absolute path to the directory so node will only look for the main field. I tried to find some context behind why the `path.join` pattern was used in the [commit introducing it](https://github.com/jsx-eslint/eslint-plugin-react/blob/9fbd037193d9f5c0980e11e80825e726d61c8ab3/tests/helpers/parsers.js) but could not find a PR or anything. Anyway, replacing those with `require.resolve('<parser lib>')` gives an absolute path but kicks all of the resolving work over to node and makes it go the "node modules route" instead which resolves the exports field 👍 It's supported on all node versions, so no issues in the workflows from using this instead of manually constructing absolute paths 🙂

lib/rules/jsx-max-depth.js
hampustagerud: This is the output of keys/values of the refs:

|  v7  |  v8  |
| --- | ---- |
|   Screenshot 2024-09-06 at 23 33 42   |   Screenshot 2024-09-06 at 23 33 55   |

The source code doesn't look a whole lot different, so maybe there's some changes between how the code is built before publishing 🤔

package.json
hampustagerud: I get an issue with Node v19 here, it only allows installing up to v7.2.0 since then the engines field is introduced which doesn't list 19 🙂

lib/rules/jsx-max-depth.js
ljharb: I'm confused, what value does it hold in the v8 image in the bottom?

package.json
ljharb: let's tweak the workflow then, so it installs v6 < 20, and v7 on node 20, and v8 on node 22?

lib/rules/jsx-max-depth.js
hampustagerud: Sorry, that was the terminal color scheme 🤦  They are `undefined`

Screenshot 2024-09-07 at 00 12 50

package.json
hampustagerud: Disabled v7 on Node 19 which was enough to get [all steps green](https://github.com/hampustagerud/eslint-plugin-react/actions/runs/10745877419) 👍

lib/rules/jsx-max-depth.js
ljharb: so it's not about truthy/falsy, just undefined? in that case instead of checking `has` can we just check if `typeof refs[i].writeExpr === 'undefined'`?

hampustagerud: We sure can 🙂 I updated the commit and no issues in the CI 👍

tests/helpers/parsers.js
ljharb: ok that makes sense.

package.json
ljharb: @hampustagerud i'm not sure i see this commit pushed up to your branch - either my suggested change, or anything about v7 on node 19

hampustagerud: I added it to the commits already in the branch

- [fa6c54007f24131c02d65f725c566c4228da6b66](https://github.com/hampustagerud/eslint-plugin-react/commit/fa6c54007f24131c02d65f725c566c4228da6b66)
- [9bdbcd1931df5787e863e1ccdd4cd1a85bbc88a4](https://github.com/hampustagerud/eslint-plugin-react/commit/9bdbcd1931df5787e863e1ccdd4cd1a85bbc88a4) 

If you re-pick my commits it should show up, or I can reset to what was there and add it in new commits if that's easier 🙂

PR #3638, opened 2023-10-03
[New] `no-invalid-html-attribute`: add support for `apple-touch-startup-image` `rel` attributes in `link` tags

thomashockaday:
Currently this attribute fails on the latest version:

```
“apple-touch-startup-image” is never a valid “rel” attribute value  react/no-invalid-html-attribute
```

According to [the docs](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-invalid-html-attribute.md) the accepted values for this rule come from MDN, and MDN states that this value is [non-standard](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel) but not disallowed, the same as `apple-touch-icon`.

If I am incorrect here please let me know.

To test this I've added a generic example as seen in [Apple's own documentation](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html#//apple_ref/doc/uid/TP40002051-CH3-SW6) as well as a longer example more like something that a [PWA asset generator](https://github.com/elegantapp/pwa-asset-generator) would give you (which is recommended by [web.dev](https://web.dev/learn/pwa/enhancements/)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,14 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ### Added
 * [`sort-prop-types`]: give errors on TS types ([#3615][] @akulsr0)
+* [`no-invalid-html-attribute`]: add support for `apple-touch-startup-image` `rel` attributes in `link` tags ([#3638][] @thomashockaday)
 
 ### Fixed
 * [`jsx-no-leaked-render`]: preserve RHS parens for multiline jsx elements while fixing ([#3623][] @akulsr0)
 * [`jsx-key`]: detect conditional returns ([#3630][] @yialo)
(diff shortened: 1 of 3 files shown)


ljharb (approved):
Thanks!

ljharb:
The failing tests are unrelated to this PR; I'll fix them on master after landing this.

GabrielF9:
Hey guys, sorry to be bothering, I don't know how the releases works here. Just pinging to know if we have a prevision when this fix will become to a release.

ljharb:
There’s no timeline; it’ll be included in the next release.

GabrielF9:
Thanks!

PR #3645, opened 2023-10-20
[New] add requireDataLowercase option to no-unknown-property rule

### LINKED ISSUE #3643: [Bug]: no-unknown-property gives false-positives on data-* attributes, but React gives warnings on them

### Description Overview

There was an https://github.com/jsx-eslint/eslint-plugin-react/issues/3395 issue raised, which resulted in the rule being updated to support passing data-* attributes with any casing, but even the sandbox the reporter provided included a warning from react: "Warning: React does not recognize the `data-testID` prop on a DOM element.". This is still an issue right now with react 18, as seen in this sandbox.

https://codesandbox.io/s/adoring-robinson-2z75gh

### Expected Behavior

Expected behaviour: data-* attributes are reported, when they have any capital letters. 

Actual behaviour: data-* attributes with uppercase letters are not reported by rule, react dumps warnings into console.

v7.33.2

### eslint version

v8.51.0

### node version

v20

HermanBilous:
Fixes #3643

### DIFF
--- a/docs/rules/no-unknown-property.md
+++ b/docs/rules/no-unknown-property.md
@@ -51,12 +51,13 @@ var AtomPanel = <atom-panel class="foo"></atom-panel>;
 
 ```js
 ...
-"react/no-unknown-property": [<enabled>, { ignore: <ignore> }]
+"react/no-unknown-property": [<enabled>, { ignore: <ignore>, requireDataLowercase: <requireDataLowercase> }]
 ...
 ```
 
 - `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
 - `ignore`: optional array of property and attribute names to ignore during validation.
+- `requireDataLowercase`: optional (default: `false`), require data-\* attributes to contain only lowercase characters. React will issue a warning when data-\* attributes contain uppercase characters. In order to catch such attributes, set the `requireDataLowercase` option to `true`.
 
 If you are using a library that passes something as a prop to JSX elements, it is recommended to add those props to the ignored properties.
 
--- a/lib/rules/no-unknown-property.js
+++ b/lib/rules/no-unknown-property.js
@@ -16,6 +16,7 @@ const report = require('../util/report');
 
 const DEFAULTS = {
   ignore: [],
+  requireDataLowercase: false,
 };
 
 const DOM_ATTRIBUTE_NAMES = {
@@ -430,6 +431,16 @@ function isValidDataAttribute(name) {
 }
 
 /**
+ * Checks if an attribute name has at least one uppercase characters
+ *
+ * @param {String} name
+ * @returns {boolean} Result
+ */
+function hasUpperCaseCharacter(name) {
+  return name.toLowerCase() !== name;
+}
+
+/**
  * Checks if an attribute name is a standard aria attribute by compering it to a list
  * of standard aria property names
  *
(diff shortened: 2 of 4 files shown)


lib/rules/no-unknown-property.js
```diff
@@ -16,6 +16,7 @@ const report = require('../util/report');
 
 const DEFAULTS = {
   ignore: [],
+  allowDataUpperCase: true,
```
ljharb, line 19: let’s call this “requireDataLowercase” and default it to false.

tests/lib/rules/no-unknown-property.js
```diff
@@ -573,6 +577,18 @@ ruleTester.run('no-unknown-property', rule, {
         },
       ],
     },
+    {
+      code: '<div data-UPPERCASE="invalid" />',
+      errors: [
+        {
+          messageId: 'unknownProp',
```
ljharb, line 584: i think this case should have its own error message, explaining the same thing react’s does - it’s not exactly unknown, it’s valid in html.

ljharb (changes requested):

tests/lib/rules/no-unknown-property.js
```diff
@@ -99,6 +99,10 @@ ruleTester.run('no-unknown-property', rule, {
     { code: '<div data-index-number="1234"></div>;' },
     { code: '<div data-e2e-id="5678"></div>;' },
     { code: '<div data-testID="bar" data-under_sCoRe="bar" />;' },
+    {
+      code: '<div data-testID="bar" data-under_sCoRe="bar" />;',
```
ljharb, line 103: can we add this same test case as invalid also, with the option set to true?

docs/rules/no-unknown-property.md
```diff
@@ -51,12 +51,13 @@ var AtomPanel = <atom-panel class="foo"></atom-panel>;
...
 ```js
 ...
-"react/no-unknown-property": [<enabled>, { ignore: <ignore> }]
+"react/no-unknown-property": [<enabled>, { ignore: <ignore>, requireDataLowercase: <requireDataLowercase> }]
 ...
 ```
 
 - `enabled`: for enabling the rule. 0=off, 1=warn, 2=error. Defaults to 0.
 - `ignore`: optional array of property and attribute names to ignore during validation.
+- `requireDataLowercase`: optional (default: `false`), require data-\* attributes to contain only lowercase characters. React will issue a warning when data-\* attributes contain uppercase characters. In order to catch such attributes, set `requireDataLowercase` option to `true`.
```
ljharb, line 60:
```suggestion
- `requireDataLowercase`: optional (default: `false`), require data-\* attributes to contain only lowercase characters. React will issue a warning when data-\* attributes contain uppercase characters. In order to catch such attributes, set the `requireDataLowercase` option to `true`.
```

lib/rules/no-unknown-property.js
```diff
@@ -526,6 +542,12 @@ module.exports = {
       return (context.options[0] && context.options[0].ignore) || DEFAULTS.ignore;
     }
 
+    function getRequireDataLowercase() {
+      return (context.options[0] && context.options[0].requireDataLowercase !== 'undefined')
```
ljharb, line 546:
```suggestion
      return (context.options[0] && typeof context.options[0].requireDataLowercase !== 'undefined')
```

ljharb (approved):

venn-eng:
This is great, thank you! Can't wait to use it :)

PR #3652, opened 2023-11-13
Add displaystyle as known property on math elements

lounsbrough:
The `math` element supports a property called `displaystyle` which is not currently recognized by the `no-unknown-property` rule.

https://developer.mozilla.org/en-US/docs/Web/MathML/Global_attributes/displaystyle


ljharb (approved):
Thanks!

PR #3654, opened 2023-11-15
[New] `prefer-read-only-props`, `prop-types`, component detection: allow components to be async functions

### LINKED ISSUE #3653: [Bug]: `react/prefer-read-only-props` doesn't work with async components

### Description Overview

The code below does not trigger the rule. But if I remove the `async` keyword and the `getData()` the rule now errors correctly. I'm using nextjs app dir with server components.

```ts
type Props = {
  isInverted: boolean
}

const MyAsyncComponent = async ({ isInverted }: Props) => {
  const data = await getData();

  return (
    // some jsx
  )
}
```

### Expected Behavior

Shouldn't the rule errors whether the component is async or not? Maybe I misunderstood something in the docs.

v7.33.2

### eslint version

v8.53.0

### node version

v18.x

pnodet:
React introduced server components, now async functions can be components.

Closes #3653

### DIFF
--- a/lib/util/Components.js
+++ b/lib/util/Components.js
@@ -855,7 +855,7 @@ function componentRule(rule, context) {
     },
 
     FunctionExpression(node) {
-      if (node.async) {
+      if (node.async && node.generator) {
         components.add(node, 0);
         return;
       }
@@ -868,7 +868,7 @@ function componentRule(rule, context) {
     },
 
     FunctionDeclaration(node) {
-      if (node.async) {
+      if (node.async && node.generator) {
         components.add(node, 0);
         return;
       }
@@ -881,11 +881,6 @@ function componentRule(rule, context) {
     },
 
     ArrowFunctionExpression(node) {
-      if (node.async) {
-        components.add(node, 0);
-        return;
-      }
-
       const component = utils.getStatelessComponent(node);
       if (!component) {
         return;
--- a/tests/lib/rules/prop-types.js
+++ b/tests/lib/rules/prop-types.js
@@ -1571,25 +1571,28 @@ ruleTester.run('prop-types', rule, {
       options: [{ skipUndeclared: false }],
     },
     {
-      // Async functions can't be components.
+      // Async generator functions can't be components.
       code: `
-        var Hello = async function(props) {
+        var Hello = async function* (props) {
+          yield null;
           return <div>Hello {props.name}</div>;
         }
       `,
     },
     {
-      // Async functions can't be components.
+      // Async generator functions can't be components.
       code: `
-        async function Hello(props) {
+        async function* Hello(props) {
+          yield null;
           return <div>Hello {props.name}</div>;
         }
       `,
     },
     {
-      // Async functions can't be components.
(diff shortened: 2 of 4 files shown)


lib/util/Components.js
```diff
@@ -855,11 +855,6 @@ function componentRule(rule, context) {
     },
 
     FunctionExpression(node) {
-      if (node.async) {
-        components.add(node, 0);
-        return;
```
ljharb, line 860: instead of just allowing any async functions - are there any static aspects of async components that can help ensure we're not going to just grab a bunch of non-component async functions?

for example, i assume an async generator isn't allowed to be a component, so we should still check for that?

pnodet:
@ljharb 
I am not familiar with the AST, but we could indeed make a stricter check.

One thing that comes to my mind is checking the return statement of the function, since it should return null or a JSX Element to be valid, but I have no idea how feasible this is.

As you mentioned, async generator functions (`async function*`) are not valid as React components. This is a clear exclusion criterion.

ljharb:
As long as the proper test cases are in the PR, it's ok if they're failing; I can try to bring it over the finish line.

ljharb:
Note that react components as of v16 can return almost *anything* - including an array of renderable things - so checking the return statement is no longer a disqualifier.

pnodet:
> Note that react components as of v16 can return almost anything - including an array of renderable things - so checking the return statement is no longer a disqualifier.

Okay, I updated the PR to allow only async generators.

tests/lib/rules/prop-types.js
```diff
@@ -1570,30 +1570,6 @@ ruleTester.run('prop-types', rule, {
       `,
       options: [{ skipUndeclared: false }],
     },
-    {
-      // Async functions can't be components.
```
ljharb, line 1574: let's update these to be async generators?

pnodet:
Hello @ljharb

I see that 1 check is failing, if I am not mistaken due to a lower code coverage percentage.
I'am not sure what test should be added for this to be honest, but I can try to implement some if you have suggestions

lib/util/Components.js
```diff
@@ -881,7 +881,7 @@ function componentRule(rule, context) {
     },
 
     ArrowFunctionExpression(node) {
-      if (node.async) {
+      if (node.async && node.generator) {
```
ljharb, line 884:
```suggestion
      if (node.async) {
```
arrows can't be generators, so i think this change will make coverage pass

pnodet:
```diff
-     if (node.async && node.generator) {
+     if (node.async) {
```

This would indeed make the coverage pass, but the if statement is an early return to prevent async function to be considered as components. Which is what we want to achieve. I went with removing the early return and it seems to pass all checks!

ljharb (approved):

pnodet:
Hello @ljharb Hope you're doing well!
I was wondering when do you think the next release including this changes would be? 
Is there ongoing work you're waiting for and where I could help?

ljharb:
@pnodet i don't have a timeline, but I'll try to do it soon.

PR #3667, opened 2023-12-15
[New] `forward-ref-uses-ref`: add rule for checking ref parameter is added

### LINKED ISSUE #3158: rule to ensure forwardRef includes a `ref` argument.
Howdy, not sure if this is a question or suggestion for a new rule, but i've not managed to find an existing rule to cover it.

But i'd like to find a rule that flags the following as an error, because the `ref` argument is not present.

```
import { forwardRef } from 'react';

const Button = forwardRef(props => <button {...props} />)
```
Whether or not the ref is used is already handled by no-unused-vars.

NotWoods:
Closes #3158

Adds a new rule to require that a second parameter is used with `forwardRef`. Offers suggestions to add the parameter or remove the `forwardRef` wrapper.


configs/recommended.js
```diff
@@ -6,6 +6,7 @@ module.exports = Object.assign({}, all, {
   languageOptions: all.languageOptions,
   rules: {
     'react/display-name': 2,
+    'react/forward-ref-uses-ref': 2,
```
ljharb, line 9: adding a new rule to the recommended config is a breaking change.
```suggestion
```

lib/rules/forward-ref-uses-ref.js
```diff
@@ -0,0 +1,99 @@
...
+const report = require('../util/report');
+const getMessageData = require('../util/message');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+const messages = {
+  missingRefParameter:
+    'forwardRef is used with this component but no ref parameter is set',
```
ljharb, line 18:
```suggestion
  missingRefParameter: 'forwardRef is used with this component but no ref parameter is set',
```

lib/rules/forward-ref-uses-ref.js
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * @param {ASTNode} node
+     * @returns {boolean} If the node represents a function call `forwardRef()` or `React.forwardRef()`.
+     */
+    function isForwardRefCall(node) {
+      return (
+        node.type === 'CallExpression'
+        && (isForwardRefIdentifier(node.callee)
+          || (node.callee.type === 'MemberExpression'
+            && isForwardRefIdentifier(node.callee.property)))
```
ljharb, line 56:
```suggestion
          || (node.callee.type === 'MemberExpression' && isForwardRefIdentifier(node.callee.property))
        )
```

tests/lib/rules/forward-ref-uses-ref.js
```diff
@@ -0,0 +1,232 @@
...
+  },
+});
+
+const tests = {
+  valid: parsers.all([
+    {
+      code: `
+        import { forwardRef } from 'react'
+        forwardRef((props, ref) => {
+          return null
```
ljharb, line 33:
```suggestion
          return null;
```
let's consistently use semis in all these test cases

tests/lib/rules/forward-ref-uses-ref.js
```diff
@@ -0,0 +1,232 @@
...
+      code: `
+        import { forwardRef } from 'react'
+        forwardRef((props) => {
+          return null
+        })
+      `,
+      errors: [
+        {
+          message:
+            'forwardRef is used with this component but no ref parameter is set',
```
ljharb, line 109:
```suggestion
          message: 'forwardRef is used with this component but no ref parameter is set',
```

tests/lib/rules/forward-ref-uses-ref.js
```diff
@@ -0,0 +1,232 @@
...
+          suggestions: [
+            {
+              messageId: 'addRefParameter',
+              output: `
+        import { forwardRef } from 'react'
+        forwardRef((props, ref) => {
+          return null
+        })
+      `,
```
ljharb, line 118:
```suggestion
             import { forwardRef } from 'react';
             forwardRef((props, ref) => {
               return null;
             });
           `,
```

NotWoods: The indentation here is set up to match with the input, should I indent the input in lines 100-105 as well?

ljharb: ah, good call, this is fine as is

ljharb (commented):
There’s a few lines that are uncovered; it’d be great to add tests that cover them.

lib/rules/forward-ref-uses-ref.js
```diff
@@ -0,0 +1,94 @@
...
+                getMessageData('addRefParameter', messages.addRefParameter),
+                {
+                  * fix(fixer) {
+                    const param = node.params[0];
+                    // If using shorthand arrow function syntax, add parentheses around the new parameter pair
+                    const shouldAddParentheses = node.type === 'ArrowFunctionExpression' && !isParenthesized(context, param);
+                    if (shouldAddParentheses) {
+                      yield fixer.insertTextBefore(param, '(');
+                    }
+                    yield fixer.insertTextAfter(param, `, ref${shouldAddParentheses ? ')' : ''}`);
```
ljharb, line 78: i would really prefer not to use generators. i think this would be simpler?
```suggestion
                  fix(fixer) {
                    const param = node.params[0];
                    // If using shorthand arrow function syntax, add parentheses around the new parameter pair
                    const shouldAddParentheses = node.type === 'ArrowFunctionExpression' && !isParenthesized(context, param);
                    return [].concat(
                      shouldAddParentheses ? fixer.insertTextBefore(param, '(') : [],
                      fixer.insertTextAfter(param, `, ref${shouldAddParentheses ? ')' : ''}`),
                    };
```

ljharb (changes requested):

tests/lib/rules/forward-ref-uses-ref.js
```diff
@@ -0,0 +1,267 @@
+/**
+ * @fileoverview Ensure symmetric naming of setState hook value and setter variables
+ * @author Duncan Beevers
```
ljharb, line 3: is this your name?

tests/lib/rules/forward-ref-uses-ref.js
```diff
@@ -0,0 +1,267 @@
+/**
+ * @fileoverview Ensure symmetric naming of setState hook value and setter variables
```
ljharb, line 2: also i suspect this is the wrong description :-)

tests/lib/rules/forward-ref-uses-ref.js
NotWoods: Whoops, this got copy-pasted and missed

tests/lib/rules/forward-ref-uses-ref.js
NotWoods: Somewhere along the line git messed something up here. I wrote a description initially, I promise :)

NotWoods: Ah it's because this is the test file 🤦

ljharb:
(there's still some uncovered lines)

NotWoods:
I'll work on this more in January; currently busy with holidays. Will make sure to get the last few lines covered and see if I can figure out why ESLint 6 & 7 are failing.

ljharb (commented):
There's still a few uncovered added lines.

NotWoods:
I don't see the uncovered lines in Codecov, can you help point them out to me? 
image

ljharb:
@NotWoods they're right in the "files changed" tab - see https://github.com/jsx-eslint/eslint-plugin-react/pull/3667/files#diff-25114be5f5c5bf7d89a5cc5dfb905ca0d6aafda4a7cb7b49d20df89ceea52bf4R72 for an example

ljharb (approved):
rebased; LGTM assuming tests pass

PR #3673, opened 2024-01-08
[jsx-no-script-url]: use shared settings to specify link components

### LINKED ISSUE #2865: Make jsx-no-script-url use shared settings to specify link components
Why do we have to specify Link components in `react/jsx-no-script-url` settings when we already have that in shared config? Can't shared config be used here, as it is in `react/jsx-no-target-blank`?

burtek:
Fixes #2865

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,12 +63,14 @@ You should also specify settings that will be shared across all the plugin rules
     "formComponents": [
       // Components used as alternatives to <form> for forms, eg. <Form endpoint={ url } />
       "CustomForm",
-      {"name": "Form", "formAttribute": "endpoint"}
+      {"name": "SimpleForm", "formAttribute": "endpoint"},
+      {"name": "Form", "formAttribute": ["registerEndpoint", "loginEndpoint"]}, // allows specifying multiple properties if necessary
     ],
     "linkComponents": [
       // Components used as alternatives to <a> for linking, eg. <Link to={ url } />
       "Hyperlink",
-      {"name": "Link", "linkAttribute": "to"}
+      {"name": "MyLink", "linkAttribute": "to"},
+      {"name": "Link", "linkAttribute": ["to", "href"]}, // allows specifying multiple properties if necessary
     ]
   }
 }
--- a/docs/rules/jsx-no-script-url.md
+++ b/docs/rules/jsx-no-script-url.md
@@ -23,7 +23,13 @@ Examples of **correct** code for this rule:
 <a href={"javascript:"}></a>
 ```
 
+This rule takes the `linkComponents` setting into account.
+
 ## Rule Options
+
+This rule accepts array option (optional) and object option (optional).
+
+### Array option (default `[]`)
 
 ```json
 {
@@ -45,11 +51,11 @@ Examples of **correct** code for this rule:
 
 Allows you to indicate a specific list of properties used by a custom component to be checked.
 
-### name
+#### name
 
 Component name.
 
-### props
+#### props
 
 List of properties that should be validated.
 
@@ -59,4 +65,38 @@ Examples of **incorrect** code for this rule, when configured with the above opt
 <Link to="javascript:void(0)"></Link>
 <Foo href="javascript:void(0)"></Foo>
 <Foo to="javascript:void(0)"></Foo>
+```
+
+### Object option
+
+#### includeFromSettings (default `false`)
+
+Indicates if the `linkComponents` config in [global shared settings](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/README.md#configuration) should also be taken into account. If enabled, components and properties defined in settings will be added to the list provided in first option (if provided):
+
+```json
+{
+  "react/jsx-no-script-url": [
+    "error",
+    [
+      {
+        "name": "Link",
+        "props": ["to"]
+      },
+      {
+        "name": "Foo",
+        "props": ["href", "to"]
+      }
+    ],
+    { "includeFromSettings": true }
+  ]
+}
+```
+
+If only global settings should be used for this rule, the array option can be omitted:
+
+```jsonc
+{
+  // same as ["error", [], { "includeFromSettings": true }]
+  "react/jsx-no-script-url": ["error", { "includeFromSettings": true }]
+}
 ```
--- a/lib/rules/jsx-no-script-url.js
+++ b/lib/rules/jsx-no-script-url.js
@@ -5,7 +5,9 @@
 
 'use strict';
 
+const includes = require('array-includes');
 const docsUrl = require('../util/docsUrl');
+const linkComponentsUtil = require('../util/linkComponents');
 const report = require('../util/report');
 
 // ------------------------------------------------------------------------------
@@ -21,26 +23,20 @@ function hasJavaScriptProtocol(attr) {
     && isJavaScriptProtocol.test(attr.value.value);
 }
 
-function shouldVerifyElement(node, config) {
-  const name = node.name && node.name.name;
-  return name === 'a' || config.find((i) => i.name === name);
-}
-
 function shouldVerifyProp(node, config) {
   const name = node.name && node.name.name;
   const parentName = node.parent.name && node.parent.name.name;
 
-  if (parentName === 'a' && name === 'href') {
-    return true;
-  }
+  if (!name || !parentName || !config.has(parentName)) return false;
 
-  const el = config.find((i) => i.name === parentName);
-  if (!el) {
-    return false;
-  }
+  const attributes = config.get(parentName);
+  return includes(attributes, name);
+}
 
-  const props = el.props || [];
-  return node.name && props.indexOf(name) !== -1;
+function parseLegacyOption(config, option) {
+  option.forEach((opt) => {
+    config.set(opt.name, opt.props);
+  });
 }
 
 const messages = {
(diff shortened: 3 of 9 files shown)


burtek:
I'm probably gonna need to change `getLinkComponents` util to allow multiple props for single component. You can't really give it `MyComponent` with two different props at this time, while `jsx-no-script-url` allows that. Am thinking about changing a string prop `linkAttribute` to a string array prop `linkAttributes` (and internally mapping `linkAttribute` to `linkAttributes: [linkAttribute]` for backwards compatibility).

docs/rules/jsx-no-script-url.md
```diff
@@ -44,6 +46,11 @@ Examples of **correct** code for this rule:
 ```
 
 Allows you to indicate a specific list of properties used by a custom component to be checked.
+This will override anything passed to `linkComponents` setting.
+
+NOTE: This rule now takes into account `linkComponents` setting and it should be used as primary source of link components.
+The rule still allows passing link components as rule option, but it is meant only as backwards-compatibility feature.
+New setups should only use `linkComponents` setting.
```
ljharb, line 53:
```suggestion
NOTE: This rule now takes into account the `linkComponents` setting, which should be used as the sole source of truth for link components.
The rule still allows passing link components as rule option for backwards compatibility, but this option is deprecated.
```

lib/rules/jsx-no-script-url.js
```diff
@@ -21,26 +22,23 @@ function hasJavaScriptProtocol(attr) {
...
 
-  const props = el.props || [];
-  return node.name && props.indexOf(name) !== -1;
+function parseLegacyOption(option) {
+  const config = linkComponentsUtil.getLinkComponents({}) // get defaults
+  option.forEach(function(opt) {
+    opt.props.forEach(function(prop) { // FIXME: only last prop will work at the moment
+      config.set(opt.name, prop)
+    })
+  })
```
ljharb, line 40:
```suggestion
  const config = linkComponentsUtil.getLinkComponents({}); // get defaults
  option.forEach((opt) => {
    opt.props.forEach((prop) => { // FIXME: only last prop will work at the moment
      config.set(opt.name, prop);
    });
  });
```

lib/rules/jsx-no-script-url.js
```diff
@@ -77,16 +75,17 @@ module.exports = {
         },
         required: ['name', 'props'],
         additionalProperties: false,
+        deprecated: true, // ?
```
ljharb, line 78: sounds good to me. does eslint actually check this?

lib/rules/jsx-no-script-url.js
```diff
@@ -77,16 +75,17 @@ module.exports = {
...
       },
     }],
   },
 
   create(context) {
-    const config = context.options[0] || [];
+    const linkComponents = context.options[0] ? parseLegacyOption(context.options[0]) : linkComponentsUtil.getLinkComponents(context);
```
ljharb, line 84: i think that perhaps this should just always start with the setting, and concatenate the option when present?

lib/rules/jsx-no-script-url.js
burtek: [yes](https://eslint.org/docs/latest/extend/custom-rules#rule-structure)

README.md
```diff
@@ -63,12 +63,14 @@ You should also specify settings that will be shared across all the plugin rules
     "formComponents": [
       // Components used as alternatives to <form> for forms, eg. <Form endpoint={ url } />
       "CustomForm",
-      {"name": "Form", "formAttribute": "endpoint"}
+      {"name": "Form", "formAttributes": ["registerEnpoint", "loginEnpoint"]}, // allows specifying multiple properties if necessary
```
ljharb, line 66:
```suggestion
      {"name": "Form", "formAttributes": ["registerEndpoint", "loginEndpoint"]}, // allows specifying multiple properties if necessary
```

README.md
```diff
@@ -63,12 +63,14 @@ You should also specify settings that will be shared across all the plugin rules
...
+      {"name": "SimpleForm", "formAttribute": "endpoint"}, // deprecated backwards-compatible version
```
ljharb, line 67: i'm not sure we need a new option name here - can we just overload `formAttribute` to be either a string, or an array of strings?

burtek: We can actually. Not a bad idea.

lib/rules/jsx-no-script-url.js
```diff
@@ -21,26 +22,20 @@ function hasJavaScriptProtocol(attr) {
...
-    return true;
-  }
+  if (!name || !parentName || !config.has(parentName)) return false;
 
-  const el = config.find((i) => i.name === parentName);
-  if (!el) {
-    return false;
-  }
+  const attributes = config.get(parentName)
+  return attributes.includes(name);
```
ljharb, line 32:
```suggestion
  return includes(attributes, name);
```

lib/rules/jsx-no-target-blank.js
```diff
@@ -48,16 +48,16 @@ function attributeValuePossiblyBlank(attribute) {
...
   const foundExternalLink = linkIndex !== -1 && ((attr) => attr.value && attr.value.type === 'Literal' && /^(?:\w+:|\/\/)/.test(attr.value.value))(
     node.attributes[linkIndex]);
   return foundExternalLink || (warnOnSpreadAttributes && linkIndex < spreadAttributeIndex);
 }
 
-function hasDynamicLink(node, linkAttribute) {
+function hasDynamicLink(node, linkAttributes) {
   const dynamicLinkIndex = findLastIndex(node.attributes, (attr) => attr.name
-    && attr.name.name === linkAttribute
+    && linkAttributes.includes(attr.name.name)
```
ljharb, line 60:
```suggestion
    && includes(linkAttributes, attr.name.name)
```

lib/rules/jsx-no-target-blank.js
```diff
@@ -48,16 +48,16 @@ function attributeValuePossiblyBlank(attribute) {
   return false;
 }
 
-function hasExternalLink(node, linkAttribute, warnOnSpreadAttributes, spreadAttributeIndex) {
-  const linkIndex = findLastIndex(node.attributes, (attr) => attr.name && attr.name.name === linkAttribute);
+function hasExternalLink(node, linkAttributes, warnOnSpreadAttributes, spreadAttributeIndex) {
+  const linkIndex = findLastIndex(node.attributes, (attr) => attr.name && linkAttributes.includes(attr.name.name));
```
ljharb, line 52:
```suggestion
  const linkIndex = findLastIndex(node.attributes, (attr) => attr.name && includes(linkAttributes, attr.name.name));
```

lib/util/linkComponents.js
```diff
@@ -37,9 +37,9 @@ function getLinkComponents(context) {
   );
   return new Map(map(iterFrom(linkComponents), (value) => {
     if (typeof value === 'string') {
-      return [value, DEFAULT_LINK_ATTRIBUTE];
+      return [value, [DEFAULT_LINK_ATTRIBUTE]];
     }
-    return [value.name, value.linkAttribute];
+    return [value.name, value.linkAttributes || [value.linkAttribute]];
```
ljharb, line 42:
```suggestion
    return [value.name, [].concat(value.linkAttribute)];
```

lib/util/linkComponents.js
```diff
@@ -24,9 +24,9 @@ function getFormComponents(context) {
   );
   return new Map(map(iterFrom(formComponents), (value) => {
     if (typeof value === 'string') {
-      return [value, DEFAULT_FORM_ATTRIBUTE];
+      return [value, [DEFAULT_FORM_ATTRIBUTE]];
     }
-    return [value.name, value.formAttribute];
+    return [value.name, value.formAttributes || [value.formAttribute]];
```
ljharb, line 29:
```suggestion
    return [value.name, [].concat(value.formAttribute)];
```

lib/rules/jsx-no-target-blank.js
```diff
@@ -265,11 +266,11 @@ module.exports = {
...
           }
 
-          const formAttribute = formComponents.get(node.name.name);
+          const formAttributes = formComponents.get(node.name.name);
 
           if (
-            hasExternalLink(node, formAttribute)
-            || (enforceDynamicLinks === 'always' && hasDynamicLink(node, formAttribute))
+            hasExternalLink(node, formAttributes)
+            || (enforceDynamicLinks === 'always' && hasDynamicLink(node, formAttributes))
```
ljharb, line 273: seems like we need code coverage for this line; any chance you could add it?

burtek: Am planning to but I'm confused as I kinda think the test case for it is already there... Will take another look later.

burtek: @ljharb should be good now

ljharb (approved):
This looks good.

I've split the PR into two commits - the first one is objectively semver-minor, and has the "multiple properties" feature.

The second, though, I'm a bit uncertain if it's semver-minor or semver-major. Could there be someone relying on `linkAttributes` *not* applying to this rule?

burtek:
I don't know. Myself I have this plugin configured in multiple projects (both my own and at work) and we always had to duplicate the setup for this rule.

That being said, that's why the first implementation had the rule option override global settings - that way it wouldn't be a breaking change.

ljharb:
hmm.

even for the case where the rule option wasn't passed, though, someone might have the link components setting and expect the rule NOT to warn on them, no?

burtek:
Yeah, you might be right actually.

I find with this PR being marked as semver-major

ljharb:
Unfortunately that'd delay it from landing probably indefinitely.

How about instead, we modify the second commit so that an *additional* option is required, a boolean which defaults to `false`, that enables hooking in the linkComponents (whether the existing option is provided or not)?

burtek:
I can do that

burtek:
@ljharb how about now? :)

ljharb (approved):

PR #3674, opened 2024-01-09
[New] `jsx-filename-extension`: add `ignoreFilesWithoutCode` option to allow empty files

### LINKED ISSUE #3644: [jsx-filename-extension]: option to allow empty files
Given the config

```yaml
rules:
  react/jsx-filename-extension: [2, {allow: as-needed]}]
```

a file that is empty or contains only comments triggers the rule, for example a `.jsx` file with:

```js
// Put your components here
```

I think a option to allow empty files when `as-needed` is specified would be ideal for this use case, not sure whether it should be the default, but I think it could be.

burtek:
Fixes #3644

### DIFF
--- a/lib/rules/jsx-filename-extension.js
+++ b/lib/rules/jsx-filename-extension.js
@@ -16,6 +16,7 @@ const report = require('../util/report');
 const DEFAULTS = {
   allow: 'always',
   extensions: ['.jsx'],
+  ignoreFilesWithoutCode: false,
 };
 
 // ------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


burtek:
The failed tests are some glitch i runner
image
image

lib/rules/jsx-filename-extension.js
```diff
@@ -97,6 +103,9 @@ module.exports = {
         }
 
         if (isAllowedExtension && allow === 'as-needed') {
+          if (ignoreFilesWithoutCode && node.body.length === 0) {
```
ljharb, line 106: does `node.body` include EmptyStatements, comment nodes, etc?

ljharb (approved):
LGTM overall

lib/rules/jsx-filename-extension.js
burtek: No, it doesn't

Issue #3234, opened 2022-03-05
`jsx-boolean-value`: enhance configuration to handle `false` literals


alexmatson-da:
The `react/jsx-boolean-value` rule appears to only work with boolean props that have been assigned `true` literals. However, I have a codebase that sometimes contains (as an example) `<MyComponent someBooleanProp={false} ... />`

Is it possible to add a configuration option for this rule to flag these instances as props that should be removed completely (maybe even with autofix)? 

If not, is there perhaps another rule that can help me achieve this?

Thanks!

ljharb:
Why would you want them removed completely? False and undefined are distinctly different, and it’s not possible to know from the consumer side whether the component is sloppily equating the two or not.

alexmatson-da:
While I'm sure they exist, off the top of my head I can't picture a use-case where a component intentionally exhibits different behavior depending on if an optional boolean prop passed into it is `undefined` rather than `false`.

In our case, optional boolean props get coerced into boolean values (`!!boolProp`), so `boolProp={false}` is really redundant. 

We also use Typescript, so if we wanted a component with a _required_ boolean prop, then it'd be typed as `boolProp: boolean` instead of `boolProp?: boolean`.

In the situations where that doesn't work (i.e., where the prop is required rather than optional), then I think it should be possible to pass an array of prop-names to be treated as exceptions from the rule, which is something `react/jsx-boolean-value` also does

ljharb:
Any time a boolean prop is required, omitting it is an error. Typescript would enforce this as well.

An eslint rule that allowed an optional false to be omitted also might not make sense, because sometimes a missing boolean defaults to true.

In other words, the assumption you wan to rely on is just not possible to rely on, and isn’t even necessarily the majority case.

ljharb:
I don't think we'll pursue this, for the reasons indicated.

tylerlaprade:
For what it's worth, I had exactly this issue and came here searching for the same fix as well. Is it possible to reconsider?

ljharb:
@tylerlaprade if the prop is a required prop, and is marked as a boolean (via propTypes and/or TS prop types), why would you need a rule here?

tylerlaprade:
@ljharb In my project, we have agreed to use `isFoo?: boolean | undefined` for all boolean props and treat `false` the same as `undefined` so that we can use simpler shorthand syntax. This rule helps me with half of my required fixes for coding style:
```
[...]

PR #3675, opened 2024-01-09
[New] `jsx-boolean-value`: add `assumeUndefinedIsFalse` option

developer-bandi:
Fixes https://github.com/jsx-eslint/eslint-plugin-react/issues/3234

A new option has been added referring to the comments on this issue.

I think there are some parts that need explanation, so I'll list them below.

1. As you can see in the test case, if you remove both properties and values, a space remains, but that space can be modified with other rules such as jsx-props-no-multi-spaces, so it was left behind.

2. While adding new options, I noticed that there was some issue with the options in the error message. After setting never or always in the second option, if the modification is made not only to the first option but also to the second option, an error message containing the properties of the second option is displayed as shown below. If it is confirmed to be an issue, a new mr will be created after working on the current issue. An example is below:

```javascript
{
      code: '<App foo={true} bar={true} baz />;',
      output: '<App foo bar baz={true} />;',
      options: ['always', { never: ['foo', 'bar'] }],
      errors: [
        {
          messageId: 'omitBoolean',
          data: { exceptionsMessage: ' for the following props: `foo`, `bar`' },
        },
        {
          messageId: 'omitBoolean',
          data: { exceptionsMessage: ' for the following props: `foo`, `bar`' },
        },
        {
          messageId: 'setBoolean',
          data: { exceptionsMessage: ' for the following props: `foo`, `bar`' },
          // setBoolean attribute is not foo or bar but baz
        }
      ],
    },
```

### DIFF
--- a/docs/rules/jsx-boolean-value.md
+++ b/docs/rules/jsx-boolean-value.md
@@ -14,7 +14,11 @@ This rule will enforce one or the other to keep consistency in your code.
 
 This rule takes two arguments. If the first argument is `"always"` then it warns whenever an attribute is missing its value. If `"never"` then it warns if an attribute has a `true` value. The default value of this option is `"never"`.
 
-The second argument is optional: if provided, it must be an object with a `"never"` property (if the first argument is `"always"`), or an `"always"` property (if the first argument is `"never"`). This property’s value must be an array of strings representing prop names.
+The second argument is optional. If provided, it must be an object. These properties are supported:
+
+First, the `"never"` and `"always"` properties are one set. The two properties cannot be set together. `"never"` must be used when the first argument is `"always"` and `"always"`  must be used when the first argument is `"never"`. This property’s value must be an array of strings representing prop names.
+
+When the first argument is `"never"`, a boolean `"assumeUndefinedIsFalse"` may be provided, which defaults to `false`. When `true`, an absent boolean prop will be treated as if it were explicitly set to `false`.
 
 Examples of **incorrect** code for this rule, when configured with `"never"`, or with `"always", { "never": ["personal"] }`:
 
@@ -38,6 +42,18 @@ Examples of **correct** code for this rule, when configured with `"always"`, or 
 
 ```jsx
 var Hello = <Hello personal={true} />;
+```
+
+Examples of **incorrect** code for this rule, when configured with `"never", { "assumeUndefinedIsFalse": true }`, or with  `"always", { "never": ["personal"], "assumeUndefinedIsFalse": true }`:
+
+```jsx
+var Hello = <Hello personal={false} />;
+```
+
+Examples of **correct** code for this rule, when configured with `"never", { "assumeUndefinedIsFalse": true }`, or with  `"always", { "never": ["personal"], "assumeUndefinedIsFalse": true }`:
+
+```jsx
+var Hello = <Hello />;
 ```
 
 ## When Not To Use It
--- a/lib/rules/jsx-boolean-value.js
+++ b/lib/rules/jsx-boolean-value.js
@@ -39,7 +39,7 @@ function getErrorData(exceptions) {
  * @param {Set<string>} exceptions
  * @param {string} propName
  * @returns {boolean} propName
-*/
+ */
 function isAlways(configuration, exceptions, propName) {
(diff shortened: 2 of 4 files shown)


ljharb:
Yes, I think the second thing is a separate issue that should be fixed separately.

docs/rules/jsx-boolean-value.md
```diff
@@ -14,7 +14,11 @@ This rule will enforce one or the other to keep consistency in your code.
 
 This rule takes two arguments. If the first argument is `"always"` then it warns whenever an attribute is missing its value. If `"never"` then it warns if an attribute has a `true` value. The default value of this option is `"never"`.
 
-The second argument is optional: if provided, it must be an object with a `"never"` property (if the first argument is `"always"`), or an `"always"` property (if the first argument is `"never"`). This property’s value must be an array of strings representing prop names.
+The second argument is optional. if provided, it must be an object. Three property is available.
```
ljharb, line 17:
```suggestion
The second argument is optional. if provided, it must be an object. These properties are supported:
```

lib/rules/jsx-boolean-value.js
```diff
@@ -60,12 +60,25 @@ function isNever(configuration, exceptions, propName) {
...
+/**
+ * @param {object} configObject
+ * @returns {object} configObject
+ */
+function configObjectDefaultSetting(configObject) {
+  const defaultObject = Object.assign({}, configObject);
+  if (defaultObject.assumeUndefinedIsFalse === undefined) {
+    defaultObject.assumeUndefinedIsFalse = false;
+  }
+  return defaultObject;
```
ljharb, line 72:
```suggestion
  return Object.assign({}, configObject, {
    assumeUndefinedIsFalse: ‼configObject.assumeUndefinedIsFalse,
  });
```

lib/rules/jsx-boolean-value.js
```diff
@@ -106,16 +123,21 @@ module.exports = {
...
+              type: 'boolean',
+            },
           },
-        }],
+        },
+        ],
         additionalItems: false,
-      }],
+      },
+      ],
```
ljharb, line 134: making these changes would force a lot more whitespace diffs; please revert them (if you have autoformatting enabled in your editor, please disable it; eslint is the only autoformatter that should be used in this project)

lib/rules/jsx-boolean-value.js
```diff
@@ -139,6 +161,7 @@ module.exports = {
         }
         if (
           isNever(configuration, exceptions, propName)
+          && configObject.assumeUndefinedIsFalse === false
```
ljharb, line 164:
```suggestion
          && !configObject.assumeUndefinedIsFalse
```
with this change, you could probably just get rid of the entire `configObjectDefaultSetting` function?

ljharb (changes requested):

docs/rules/jsx-boolean-value.md
```diff
@@ -14,7 +14,11 @@ This rule will enforce one or the other to keep consistency in your code.
...
+
+First, the `"never"` and `"always"` properties are one set. The two properties cannot be set together. `"never"` must be used when the first argument is `"always"` and `"always"`  must be used when the first argument is `"never"`. This property’s value must be an array of strings representing prop names.
+
+The other is `"assumeUndefinedIsFalse"`. The default value of this option is false. if the first argument is  `"never"`  or the property of the second argument contains  `"never"` , the corresponding property cannot have false.
```
ljharb, line 21:
```suggestion
When the first argument is `"never"`, a boolean `"assumeUndefinedIsFalse"` may be provided, which defaults to `false`. When `true`, an absent boolean prop will be treated as if it were explicitly set to `false`.
```

lib/rules/jsx-boolean-value.js
developer-bandi: Yes, I removed configObjectDefaultSetting by applying the above code.

ljharb (approved):

lib/rules/jsx-boolean-value.js
developer-bandi: Yes I fixed it by changing the IDE settings.

PR #3677, opened 2024-01-11
[New] `jsx-one-expression-per-line`: add `non-jsx` option to allow non-JSX children in one line

### LINKED ISSUE #3669: New option: `jsx-one-expression-per-line` `non-jsx`
Examples of **incorrect** code for this rule:
```jsx
<Hello
  foo="bar
>Children
</Hello>
```
```jsx
<Hello
  foo="bar
>
  Children
</Hello>
```

burtek:
Fixes #3669

### DIFF
--- a/lib/rules/jsx-one-expression-per-line.js
+++ b/lib/rules/jsx-one-expression-per-line.js
@@ -38,7 +38,7 @@ module.exports = {
         type: 'object',
         properties: {
           allow: {
-            enum: ['none', 'literal', 'single-child'],
+            enum: ['none', 'literal', 'single-child', 'non-jsx'],
           },
         },
         default: optionDefaults,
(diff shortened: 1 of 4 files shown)


lib/rules/jsx-one-expression-per-line.js
```diff
@@ -65,6 +65,11 @@ module.exports = {
         return;
       }
 
+      if (options.allow === 'non-jsx'
+        && !children.find((child) => (child.type === 'JSXFragment' || child.type === 'JSXElement'))) {
```
ljharb, line 69:
```suggestion
      if (
        options.allow === 'non-jsx'
        && !children.find((child) => (child.type === 'JSXFragment' || child.type === 'JSXElement'))
      ) {
```

can we add some test coverage for something like `{(<div />)}` and `{(() => <div />)()}`?

ljharb (changes requested):
Thanks, this looks great!

lib/rules/jsx-one-expression-per-line.js
burtek: Sure, once I put my kid to sleep!

ljharb: No rush :-)

ljharb (approved):

PR #3678, opened 2024-01-11
`jsx-newline`: support for checking newline at the start and end of children

### LINKED ISSUE #3663: [Bug]:

### Description Overview

`react/jsx-newline` with `{ "prevent": true }` does not remove empty lines for surrounding tags.

Currently `react/jsx-newline` does not detect any lint errors for the following JSX:
```
<div>

  <span>
    hi
  </span>

</div>
```

### Expected Behavior

The current detection feels like unintended behavior, as I would expect all newlines within JSX to be removed as follows:

```
<div>
  <span>
    hi
  </span>
</div>
```

If there is a different eslint rule that checks for this please let me know!

v7.33.2

### eslint version

v8.55.0

### node version

v18.6.0

burtek:
Fixes #3663

### DIFF
--- a/lib/rules/jsx-newline.js
+++ b/lib/rules/jsx-newline.js
@@ -46,6 +46,10 @@ module.exports = {
             default: false,
             type: 'boolean',
           },
+          checkStartEnd: {
+            default: false,
+            type: 'boolean',
+          },
         },
         additionalProperties: false,
         if: {


lib/rules/jsx-newline.js
```diff
@@ -46,6 +46,10 @@ module.exports = {
             default: false,
             type: 'boolean',
           },
+          checkStartEnd: {
+            default: false,
+            type: 'boolean',
+          },
```
burtek, line 52: @ljharb this makes sense but I wonder if we shouldn't also include option for banning start/end newlines while forcing empty lines between children :thinking: 

So this might actually become an enum of `force` | `prevent` rather than boolean. What do you think, what should I go with?

burtek: Or we might go with:

`undefined` and `false` = do nothing
`true` = behaves same as between children (as per `prevent`)
`force` = forces an empty line
`prevent` = prevents empty line

And it will still take `allowMultilines` into account

ljharb:
`prevent: true` or false obv need to continue working.

I like the idea of effectively deprecating the boolean config, and instead allowing an enum of `'ignore'` (`false`), `'siblings'` (`true`), `'adjacent'` (lexically adjacent, ie everything) - that way if someone wants different options we can easily add them later, but for now we're only adding `adjacent` to cover the likely reading of the existing docs. Thoughts?

burtek:
Kinda feel like deprecating (while still supporting) old config, but making new config be:

```jsonc
{
    between: 'force' | 'force-one' | 'prevent' | 'ignore' // between children
    adjecent: 'force' | 'force-one' | 'prevent' | 'ignore' // before first child and after last child
}
```

`force` - must be at least 1 empty line
`force-one` - must be exactly 1 empty line
`prevent` - must be 0 empty lines
`ignore` - ignore case

Obviously if both are `ignore` then rule does nothing.

Those properties might probably be called something else (better?).

ljharb:
"between" seems weird for a parent/child relationship. Are you just trying to think of possible combinations, or do you have use cases?

A linter saying "adjacent" usually means lexically, not like "siblings" in terms of the AST graph.

burtek:
So I want to distinguish all four options I saw in different projects:

```jsx
<parent>
    <childA />
    <childB />
</parent>
```

```jsx
<parent>
    <childA />
    
    <childB />
</parent>
```

```jsx
<parent>

    <childA />
    
    <childB />
    
</parent>
```

```jsx
<parent>

    <childA />
    <childB />
    
</parent>
```

Easiest way would be allow configuring the rule for newlines between siblings separately from newlines before first and after last sibling. `between` and `adjecent` are names that came to my mind, but as non-native speaker I'm not sure what those names should actually be.

ljharb:
OK, let's aim for those 4 options, with an eye for future extensibility.

That means two keys, one for siblings and one for directly inside of a parent, each of which a "must have a newline" value, and a "must not have a newline" value. I suppose we could also have "ignore", too.

What about `betweenSiblings` and `aroundChildren`?

burtek:
Sounds good

PR #3695, opened 2024-02-21
[Fix] add a few jsdoc type annotations to work around TS inference bugs for consumers

bradzacher:
Figured I'd split this out of #3694

This PR adds a few JSDoc type annotations to hint to TS what the types are. This helps work-around issues like https://github.com/typescript-eslint/typescript-eslint/issues/8522 / https://github.com/microsoft/TypeScript/issues/57460 where TS would infer the wrong type from the JS and would break strictly-typed config files.

### DIFF
--- a/configs/all.js
+++ b/configs/all.js
@@ -9,6 +9,10 @@ function filterRules(rules, predicate) {
   return fromEntries(entries(rules).filter((entry) => predicate(entry[1])));
 }
 
+/**
+ * @param {object} rules - rules object mapping rule name to rule module
+ * @returns {Record<string, 2>}
+ */
 function configureAsError(rules) {
   return fromEntries(Object.keys(rules).map((key) => [`react/${key}`, 2]));
 }
@@ -20,6 +24,12 @@ const deprecatedRules = filterRules(allRules, (rule) => rule.meta.deprecated);
 
 module.exports = {
   plugins: {
+    /**
+     * @type {{
+     *   deprecatedRules: Record<string, import('eslint').Rule.RuleModule>,
+     *   rules: Record<string, import('eslint').Rule.RuleModule>,
+     * }}
+     */
     react: {
       deprecatedRules,
       rules: allRules,
--- a/lib/rules/index.js
+++ b/lib/rules/index.js
@@ -2,6 +2,7 @@
 
 /* eslint global-require: 0 */
 
+/** @type {Record<string, import('eslint').Rule.RuleModule>} */
 module.exports = {
   'boolean-prop-naming': require('./boolean-prop-naming'),
   'button-has-type': require('./button-has-type'),


configs/all.js
```diff
@@ -9,6 +9,10 @@ function filterRules(rules, predicate) {
   return fromEntries(entries(rules).filter((entry) => predicate(entry[1])));
 }
 
+/**
+ * @param {object} rules - rules object mapping rule name to rule module
+ * @returns {Record<string, 2>}
```
ljharb, line 14: can these types be a bit better than `object` and `string`?

bradzacher: Currently this function return type is inferred as one of two things depending on if there are types for `object.fromentries` installed:
- `any` if it's not
- `{ [x: string]: number; [x: number]: number; [x: symbol]: number; }` if it is

Using `Record<string, 2>` (which is the same as `{ [x: string]: 2 }`) is a marked improvement as it *always* has this type now.

Can we do better? Technically yes - this could be defined generically if we want.
I'd have to research how to do this with JSDoc types but it's possible to define this in TS with generics [playground](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXzD0SwHNk4BBAZwFEYYcYAeAFXhAA8MRVgr4qGGFlQkAfAAoYyCCCoAueACVwjYKwA08NAGtUOAO6oxASkUrCMdQAM4UTAHoAJAG8WAX2taATGIDcAFABhKiC8HBUMhjwALwERKTkINR0DDASLvAB8PAA5ABGODiyUKgAtAAODBVlqFAAtiIkuYouDgBU8AB0PfDtDu4a2Xn5yBgYeGUAFlBUZRgAnhUgLfBtnT1dfQNDOblgU+A6IMBlcACOyFgRZXgHpSQgtzBnIFDAeBALq+vdvf2DYa5UCIKBRSrVOb1KAYA4QnA1RbLKg-Dp-LYA3Z5UCCaSYchNMqzKikVD1HgYVEbf47ALuEyBAIRKKBBwOeAAPQA-EEQmEAG5QCDIBBxZkQDAAbVydkchWKb3KVQRtQaTVyAF1WeycpyeUA)
```ts
function configureAsError<T extends string>(rules: Record<T, unknown>): Record<`react/${T}`, 2> { ... }
```

Depends on how strict and correct you would like the types to be really!

bradzacher: This would be the most correct JSDoc that would strictly type the resulting objects:

```js
/**
 * @template {string} Key
 * @template Value
 * @param {Record<Key, Value>} rules
 * @param {(value: Value) => boolean} predicate
 * @returns {Record<Key, Value>}
 */
function filterRules(rules, predicate) {
  return fromEntries(entries(rules).filter((entry) => predicate(entry[1])));
}

/**
 * @template {string} T
 * @param {Record<T, unknown>} rules - rules object mapping rule name to rule module
 * @returns {Record<`react/${T}`, 2>}
 */
function configureAsError(rules) {
  return fromEntries(Object.keys(rules).map((key) => [`react/${key}`, 2]));
}
```

However this causes lint errors.
The JSDoc parser used by the `valid-jsdoc` lint rule isn't permissive enough and crashes on the JSDoc comment if I try to do some better types.

```
/Users/bjz/git/eslint-plugin-react/configs/all.js
   8:1  error  JSDoc syntax error  valid-jsdoc
  19:1  error  JSDoc syntax error  valid-jsdoc
```

ljharb: If we're running tsc on the jsdoc comments, we don't need the valid-jsdoc rule anymore, i think?

bradzacher: Note I didn't add tsc here - just the annotations.
Which was why I added ones that were as simple as possible cos they weren't for repo validation.

If we want to go down the path of using JSDoc types everywhere in this project that's a whole other topic :)

ljharb (approved):

josephharkins:
hey @bradzacher , I think I am getting this error still.

Here is the start of the TS error:
`'reactPlugin.configs.flat' is possibly 'undefined'.ts(18048)
Argument of type 'ReactFlatConfig' is not assignable to parameter of type 'ConfigWithExtends'.
  Types of property 'plugins' are incompatible.
    Type '{ react: { deprecatedRules: Partial<{ 'boolean-prop-naming': RuleModule; 'button-has-type': RuleModule; 'checked-requires-onchange-or-readonly': RuleModule; ... 99 more ...; 'void-dom-elements-no-children': RuleModule; }>; rules: { ...; }; configs: { ...; } & { ...; }; }; }' is not assignable to type 'Plugins'.`

And here is my config:

`// @ts-check

import eslint from "@eslint/js"
import tselint from "typescript-eslint"
import eslintConfigPrettier from "eslint-config-prettier"
import reactPlugin from "eslint-plugin-react"

export default tselint.config(
  eslint.configs.recommended,
  ...tselint.configs.recommended,
  ...tselint.configs.stylistic,
  reactPlugin.configs.flat.recommended,
  reactPlugin.configs.flat["jsx-runtime"],
  eslintConfigPrettier,
  {
    ignores: [
      "**/node_modules/**",
      "**/dist/**",
      "**/build/**",
      "**/.turbo/**",
      "**/coverage/**",
    ],
    rules: {
      "no-console": "off",
      "@typescript-eslint/no-explicit-any": "off",
    },
  }
)
`

bradzacher:
Please don't comment on merged PRs from 9 months ago. 

Instead raise an issue reporting your bug with a repro so it can be assessed and triaged appropriately.

PR #3700, opened 2024-03-04
[Fix] `jsx-no-leaked-render`: prevent wrongly adding parens

### LINKED ISSUE #3698: [Bug]: jsx-no-leaked-render causes code to be incorrect by removing parentheses

### Description Overview

I'm encountering an issue with a code transformation performed by the _jsx-no-leaked-renderer_.
Original condition: `connection && (hasError || hasErrorUpdate)`.
Converted condition: `(!!connection && hasError) || hasErrorUpdate`. 
These conditions are not identical and produce different results for various inputs. For example, it will return different values for false, any, and true (respectively). In the first case it will return false, in the second it will return true.

Despite a [reported issue](https://github.com/jsx-eslint/eslint-plugin-react/issues/3498) and a pull request aimed at fixing this problem over a year ago, I'm still experiencing the same issue. 
Am I missing something, or this is a bug?

### Expected Behavior

I want to preserve original condition, or if it is converted, to be identical as original.

v18.2.0

### eslint version

v8.43.0

### node version

v18.17.1

developer-bandi:
Closes #3698

In jsx-no-leaked-render rule, `connection && (hasError || hasErrorUpdate)` code is changed to `!!connection && hasError || hasErrorUpdate` but two code is not equal. 

so if right exprssion's(`(hasError || hasErrorUpdate)`) type is LogicalExpression, add parentheses both side

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ## Unreleased
 
+### Fixed
+* [`jsx-no-leaked-render`]: prevent wrongly adding parens ([#3700][] @developer-bandi)
+
+[#3700]: https://github.com/jsx-eslint/eslint-plugin-react/pull/3700
+
 ## [7.34.0] - 2024.03.03
 
 ### Added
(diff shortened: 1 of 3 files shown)


ljharb (approved):
Thanks!

Issue #3285, opened 2022-05-09
react/boolean-prop-naming interface does not work


DavidShefcik:
The `react/boolean-prop-naming` rule does not work when using an interface with TypeScript but it works with a `type`

This shows an error like it should:
```typescript
type Props = {
  done: boolean; // Shows error
}
```

This doesn't show an error like it should:
```typescript
interface Props {
  done: boolean; // No error
}
```

My usage of the rule:
```
"react/boolean-prop-naming": [
      "error",
      {
        "rule": "^(is|has)[A-Z]([A-Za-z0-9]?)+",
        "message": "It is better if your prop ({{ propName }}) starts with `is` or `has`"
      }
    ],
```

Versions:
- eslint - `8.15.0`
- eslint-plugin-react `7.29.4`

AgentRBY:
Any updates?

ljharb:
@AgentRBY nope, if there were updates, they'd be posted in the issue, or there'd be a linked PR. (that's true everywhere on github)

Gergely-H:
It does not work with inline-defined prop types either. I would not suggest anyone use inline types, but it might help locate this problem.

Example for getting no error even with @DavidShefcik's rule enabled:
```typescript
const TestComponent: FC<{
  isEnabled: boolean;
  enabled: boolean;
}> = ({ isEnabled, enabled }) => {
//...
};
```

leadq:
when declaring Prop's properties, using "interface" instead of "type" keyword  breaks the rule. I also experienced the same issue

ofelipescherer:
It seems like now at `"eslint": "^8.47.0"` with `"eslint-plugin-react": "^7.33.1"` is not working this rule (react/boolean-prop-naming) neither as type or interface

ljharb:
@ofelipescherer can you provide a repro? what code and config isn't working as you expect on the command line?

fescherer:
> > Yes, of course. I created a [sandbox demo](https://codesandbox.io/s/test-react-plugin-eslint-rule-boolean-hyvtq9?file=/src/App.tsx).
> > The prop `enabled` should give me an error, right? I want to give me as suggestion `isEnabled`.
> 
[...]

PR #3701, opened 2024-03-05
[Fix] `boolean-prop-naming`: detect TS interfaces

developer-bandi:
related to #3285

in this issue I found four error case. The test case is below:
```
interface Props {
  enabled: boolean
}
const Hello = (props: Props) => <div />;
```

```
const Hello = (props: {enabled:boolean}) => <div />;
```

```
type Props = {
  enabled: boolean
}
const Hello = (props: Props & BaseProps) => <div />;
```

```
type Props = {
  enabled: boolean
}
const Hello: React.FC<Props> = (props) => <div />;
```
Currently, PR is working on the first case fix that corrects the interface error, which is the main content raised in the issue.

I think the code needs to be modified so that all test cases succeed, but it seems like there is many cases to work on at once in this PR, so I am wondering whether this should be done in one PR, or if it can be done in a different PR for each test case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ### Fixed
 * [`jsx-no-leaked-render`]: prevent wrongly adding parens ([#3700][] @developer-bandi)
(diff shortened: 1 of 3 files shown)


ljharb (approved):
Thanks, seems straightforward!

PR #3704, opened 2024-03-07
[Fix] `boolean-prop-naming`: literalType error fix

developer-bandi:
related #3285 As mentioned in #3701, the rule is modified to apply even when the type of props is directly specified.

```
const Hello = (props: {enabled:boolean}) => <div />;
```


ljharb (approved):
thanks!

PR #3705, opened 2024-03-08
[Fix] `boolean-prop-naming`: literal TSIntersectionType not allow error fix

developer-bandi:
fix literal TSIntersectionType not allow error like below test case
```
      type Props = {
        enabled: boolean
      }
      type BaseProps = {
        semi: boolean
      }
      
      const Hello = (props: Props & BaseProps) => <div />;`
```
and left one error test case messioned in #3701 is exsit in test code 
```
type Props = {
  enabled: boolean
}
const Hello: React.FC<Props> = (props) => <div />;
```

so I think this pr is merged, the issue #3285 can be closed.


ljharb (approved):
Thanks! restored the flatMap behavior using a dep :-)

Issue #3712, opened 2024-03-14
[Bug]: `no-unknown-property` false-positive without `data-` prefix


silverwind:
### Is there an existing issue for this?

- [X] I have searched the existing issues and my issue is unique
- [X] My issue appears in the command-line and not only in the text editor

### Description Overview

```jsx
<Item dataKey="key"/>
```
```yaml
react/no-unknown-property: [2, requireDataLowercase: true}]
```
Raises false positive:
```
React does not recognize data-* props with uppercase characters on a DOM element. Found 'dataKey', use 'datakey' instead
```
### Expected Behavior

No error because `dataKey` doesn't start with `data-`. The rule docs specifically mentions that it's only checking [`data-*`](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unknown-property.md#rule-options) attributes, e.g. with the `-`.

### eslint-plugin-react version

v7.34.0

### eslint version

v8.57.0

### node version

v21.7.1

PR #3713, opened 2024-03-15
[Fix] `no-unknown-property`: only match `data-*` attributes containing `-`

silverwind:
This fixed behaviour with the `requireDataLowercase` option enabled:

1. For standard HTML elements, a attribute named `dataNotAnDataAttribute` now raises `unknownProp` instead of `dataLowercaseRequired`.
2. For React components, a attribute named `dataNotAnDataAttribute` no longer raise an error.

Also, I removed one useless capture group in this regexp.

Fixes: https://github.com/jsx-eslint/eslint-plugin-react/issues/3712


ljharb (approved):
Thanks!

Issue #3717, opened 2024-03-15
[Bug]: `boolean-prop-naming` rule throws error


jlarmstrongiv:
### Is there an existing issue for this?

- [X] I have searched the existing issues and my issue is unique
- [X] My issue appears in the command-line and not only in the text editor

### Description Overview

[`eslint-plugin-react`](https://togithub.com/jsx-eslint/eslint-plugin-react) breaks in a patch update.

I wasn’t able to narrow it down, but @developer-bandi and @ljharb made changes to the `boolean-prop-naming` in the [current release](https://github.com/jsx-eslint/eslint-plugin-react/releases/tag/v7.34.1).  While the fixes are fantastic, it seems like something also broke in the process.

From https://kitchen-sink.trpc.io/react-hook-form

```tsx
import { zodResolver } from "@hookform/resolvers/zod";
import { useId } from "react";
import type {
  FieldValues,
  SubmitHandler,
  UseFormProps,
  UseFormReturn,
} from "react-hook-form";
import { FormProvider, useForm, useFormContext } from "react-hook-form";
import type { z } from "zod";
import { errorMap } from "zod-validation-error";

// reference https://kitchen-sink.trpc.io/react-hook-form?file=feature%2Freact-hook-form%2FForm.tsx#content

export type UseZodForm<TInput extends FieldValues> = UseFormReturn<TInput> & {
  /**
   * A unique ID for this form.
   */
  id: string;
};
export function useZodForm<TSchema extends z.ZodType>(
  props: Omit<UseFormProps<TSchema["_input"]>, "resolver"> & {
    schema: TSchema;
  },
): UseZodForm<TSchema["_input"]> {
  const form = useForm<TSchema["_input"]>({
    mode: "onTouched",
    ...props,
    resolver: zodResolver(
      props.schema,
      {
        async: true,
        errorMap,
      },
      {
        // This makes it so we can use `.transform()`s on the schema without same transform getting applied again when it reaches the server
        raw: true,
      },
    ),
  }) as UseZodForm<TSchema["_input"]>;

  form.id = useId();

  return form;
}

export type AnyZodForm = UseZodForm<any>;

export function Form<TInput extends FieldValues>(
  props: Omit<React.ComponentProps<"form">, "id" | "onSubmit"> & {
    readonly form: UseZodForm<TInput>;
    readonly handleSubmit: SubmitHandler<TInput>;
  },
): JSX.Element {
[...]

PR #3718, opened 2024-03-16
[Fix] `boolean-prop-naming`: avoid a crash with a non-TSTypeReference type

developer-bandi:
related #3717

I think cause of bug  is no concider below case
```
type Props = {
  enabled: boolean
}
        
const Hello = (props: Props & {
  semi: boolean
}) => <div />;
```
To explain in detail, if the type is not `TSTypeReference` (literal type, etc.), a reference error appears to have occurred because there is no typeName property.

Therefore, if it is not the corresponding type, the type has been modified to be used as is rather than taken from objectTypeAnnotations.


ljharb (approved):
Thanks!

PR #3724, opened 2024-04-03
[New] add `jsx-props-no-spread-multi`

SimonSchick:
Please see notes in code.

I think this rule has value in catching some niche bugs, I personally only saw this ~3-4 times throughout my career.


configs/recommended.js
```diff
@@ -11,6 +11,7 @@ module.exports = Object.assign({}, all, {
     'react/jsx-no-duplicate-props': 2,
     'react/jsx-no-target-blank': 2,
     'react/jsx-no-undef': 2,
+    'react/jsx-props-no-spread-multi': 2,
```
SimonSchick, line 14: Probably bold to suggest this but I have a hard time imaging use-cases for spreading the same expression twice.

docs/rules/jsx-props-no-spread-multi.md
```diff
@@ -0,0 +1,27 @@
...
+
+Enforces that any unique express is only spread once. Generally spreading the same expression twice is an indicator of a mistake since any attribute between the spreads may be overridden when
+the intent was not to. Even when that is not the case this will lead to unnecessary computations to be performed.
+
+## Rule Details
+
+Examples of **incorrect** code for this rule:
+
+```jsx
+<App {...props} myAttr='1' {...props} />
```
SimonSchick, line 15: TODO: Should I add more expression examples eg. function calls?

lib/rules/jsx-props-no-spread-multi.js
```diff
@@ -0,0 +1,60 @@
...
+    return {
+      JSXOpeningElement(node) {
+        const spreads = node.attributes.filter((attr) => attr.type === 'JSXSpreadAttribute');
+        if (spreads.length < 2) {
+          return;
+        }
+        const hashes = new Set();
+        for (const spread of spreads) {
+          // TODO: Deep compare ast function?
+          const hash = JSON.stringify(spread, propertyFilter);
```
SimonSchick, line 49: This is admittedly somewhat of a bandaid.
I used a various of `deepEqual` before to compare all possible matches (which is also O(n^2)).
This performs pretty well and generally has been generating small 'hashes' however I am worried of scenarios where developers create excessively large/syntactically complex spreads.

In these cases we could perform a simple md5/sha hashing though this may be overkill.

lib/types.d.ts
```diff
@@ -11,11 +11,16 @@ declare global {
...
   type JSXElement = ASTNode;
   type JSXFragment = ASTNode;
+  type JSXOpeningElement = ASTNode;
   type JSXSpreadAttribute = ASTNode;
 
-  type Context = eslint.Rule.RuleContext
+  type Context = eslint.Rule.RuleContext;
 
-  type TypeDeclarationBuilder = (annotation: ASTNode, parentName: string, seen: Set<typeof annotation>) => object;
+  type TypeDeclarationBuilder = (
```
SimonSchick, line 19: This was auto-formatted, I assume this is ok?

lib/rules/jsx-props-no-spread-multi.js
```diff
@@ -0,0 +1,60 @@
...
+      url: docsUrl('jsx-props-no-spread-multi'),
+    },
+    messages,
+  },
+
+  create(context) {
+    return {
+      JSXOpeningElement(node) {
+        const spreads = node.attributes.filter((attr) => attr.type === 'JSXSpreadAttribute');
+        if (spreads.length < 2) {
```
SimonSchick, line 43: Would be nice to have some stats on how many spreads exists on average per component, though this probably covers ~99.99% of all component instantiations.

configs/recommended.js
ljharb: Adding anything to the recommended config is a breaking change, so we can’t do that.

ljharb (commented):
Adding a new rule has a cost, and if you’ve only seen it 3-4 times (i never have), i don’t think it’ll be that useful. (Generally it’s better to discuss a new rule in an issue first so it doesn’t feel like work is wasted in the event it’s not merged)

SimonSchick:
No hard feelings if it's not merged, I understand it's more of a niche thing, worst case I will port it over to my own plugin :)

lib/rules/jsx-props-no-spread-multi.js
SimonSchick: Optimized it a little by only hashing the argument rather than the full node.

SimonSchick:
@ljharb do you have any suggestions where I could sample this rule against a larger set of JSX code in the wild? Github search hasn't been particularly helpful due to the lack of ast node search.

I ran this rule again some internal code bases and found 2 more violations which is still not statistically significant but an indicator that this may be more wide-spread than I thought.

SimonSchick:
Any interest/thoughts on this? If not I will close this PR eow and move into a separate library.

ljharb:
@SimonSchick we use https://www.npmjs.com/package/eslint-remote-tester to test things weekly, that might be worth looking into?

docs/rules/jsx-props-no-spread-multi.md
ljharb: More examples, and more tests, are always good.

lib/rules/jsx-props-no-spread-multi.js
ljharb: Can't the identifier being spread be trivially determinable?

lib/types.d.ts
ljharb: nbd, i reverted it. i'd suggest disabling autoformatting in your editor :-)

lib/rules/jsx-props-no-spread-multi.js
SimonSchick: The goal here is to also identify complex ast nodes to be identified, (eg. function calls & what not), I'd assume in 99% of the cases this will just be an identifier so the overhead isn't large.

If you think this may be overkill I can remove this functionality.

ljharb: I think that since it can’t ever be complete, it won’t add much value. Also two function calls aren’t guaranteed to return similar objects.

SimonSchick: One could argue that if you are spreading equivalent function calls (pure or not) you are __probably__ doing something sketchy, though I can see how a plain identifier is the most probable scenario and everything beyond that is too exotic to handle.

I will update this tomorrow (probably).

SimonSchick:
Updated, but CI seems to be wonky, I rebased ontop of upstream (this repo).

ljharb (approved):
LGTM. I'm going to hold off merging this until after I cut a release with all the pending patch changes, though.

SimonSchick:
@ljharb ty for your time and including this!

PR #3725, opened 2024-04-03
autofix-unused-prop-type

Veveue:


ljharb:
It seems very dangerous to me to assume that it's safe to remove just because it's unused. Autofixing has a very high bar for when it's safe, and "removal" is generally not done with autofixes.

Veveue:
> It seems very dangerous to me to assume that it's safe to remove just because it's unused. Autofixing has a very high bar for when it's safe, and "removal" is generally not done with autofixes.

I want to automatically fix many unnecessary imports caused by copying and pasting code.

ljharb:
Yes, but your desire to do less work doesn't mean the autofix is a safe thing to enable.

You could certainly implement this as a _suggestion_, though, if you like?

Veveue:
I understand. I will handle this issue locally. Thank you for your suggestion!

PR #3726, opened 2024-04-06
feat: support eslint ^9

li-jia-nan:
feat: support eslint ^9


Standard8:
I'm not a maintainer here, but this doesn't seem to be enough - see the comments on issue #3699

li-jia-nan:
Sorry, I didn't notice that there was already an issue here, I closed this PR.

Standard8:
Oh, sorry, I don't think you have to close it, but it does seem there's a bit more work than just bumping the compatible versions.

ljharb:
I believe this was obsoleted by #3759.

PR #3727, opened 2024-04-07
[Feat]: provide compatibility with eslint v9

antongolub:
relates #3699
superseeds #3726

### DIFF
--- a/lib/rules/destructuring-assignment.js
+++ b/lib/rules/destructuring-assignment.js
@@ -4,6 +4,7 @@
 
 'use strict';
 
+const astUtil = require('../util/ast');
 const Components = require('../util/Components');
 const docsUrl = require('../util/docsUrl');
 const isAssignmentLHS = require('../util/ast').isAssignmentLHS;
@@ -101,7 +102,7 @@ module.exports = {
     function handleStatelessComponent(node) {
       const params = evalParams(node.params);
 
-      const SFCComponent = components.get((context.sourceCode || context).getScope(node).block);
+      const SFCComponent = components.get(astUtil.getScope(context, node).block);
       if (!SFCComponent) {
         return;
       }
@@ -119,7 +120,7 @@ module.exports = {
     }
 
     function handleStatelessComponentExit(node) {
-      const SFCComponent = components.get((context.sourceCode || context).getScope(node).block);
+      const SFCComponent = components.get(astUtil.getScope(context, node).block);
       if (SFCComponent) {
         sfcParams.pop();
       }
@@ -191,7 +192,7 @@ module.exports = {
       'FunctionExpression:exit': handleStatelessComponentExit,
 
       MemberExpression(node) {
-        let scope = (context.sourceCode || context).getScope(node);
+        let scope = astUtil.getScope(context, node);
         let SFCComponent = components.get(scope.block);
         while (!SFCComponent && scope.upper && scope.upper !== scope) {
           SFCComponent = components.get(scope.upper.block);
@@ -209,7 +210,7 @@ module.exports = {
 
       VariableDeclarator(node) {
         const classComponent = utils.getParentComponent(node);
-        const SFCComponent = components.get((context.sourceCode || context).getScope(node).block);
+        const SFCComponent = components.get(astUtil.getScope(context, node).block);
 
         const destructuring = (node.init && node.id && node.id.type === 'ObjectPattern');
         // let {foo} = props;
@@ -247,7 +248,7 @@ module.exports = {
           && destructureInSignature === 'always'
           && node.init.name === 'props'
         ) {
-          const scope = (context.sourceCode || context).getScope(node);
+          const scope = astUtil.getScope(context, node);
           const scopeSetProps = scope.set.get('props');
           const propsRefs = scopeSetProps && scopeSetProps.references;
           if (!propsRefs) {
--- a/lib/rules/destructuring-assignment.js
+++ b/lib/rules/destructuring-assignment.js
@@ -101,7 +101,7 @@ module.exports = {
     function handleStatelessComponent(node) {
       const params = evalParams(node.params);
 
-      const SFCComponent = components.get(context.getScope(node).block);
+      const SFCComponent = components.get((context.sourceCode || context).getScope(node).block);
       if (!SFCComponent) {
         return;
       }
@@ -119,7 +119,7 @@ module.exports = {
     }
 
     function handleStatelessComponentExit(node) {
-      const SFCComponent = components.get(context.getScope(node).block);
+      const SFCComponent = components.get((context.sourceCode || context).getScope(node).block);
       if (SFCComponent) {
         sfcParams.pop();
       }
@@ -191,7 +191,7 @@ module.exports = {
       'FunctionExpression:exit': handleStatelessComponentExit,
 
       MemberExpression(node) {
-        let scope = context.getScope(node);
+        let scope = (context.sourceCode || context).getScope(node);
         let SFCComponent = components.get(scope.block);
         while (!SFCComponent && scope.upper && scope.upper !== scope) {
           SFCComponent = components.get(scope.upper.block);
@@ -209,7 +209,7 @@ module.exports = {
 
       VariableDeclarator(node) {
         const classComponent = utils.getParentComponent(node);
-        const SFCComponent = components.get(context.getScope(node).block);
+        const SFCComponent = components.get((context.sourceCode || context).getScope(node).block);
 
         const destructuring = (node.init && node.id && node.id.type === 'ObjectPattern');
         // let {foo} = props;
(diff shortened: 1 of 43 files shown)


antongolub:
Oh... One more.
```
ESLint: 9.0.0

TypeError: context.getScope is not a function
Occurred while linting /Users/antongolub/projects/gh/forks/eslint-plugin-react/configs/all.js:3
Rule: "import/no-amd"
    at CallExpression (/Users/antongolub/projects/gh/forks/eslint-plugin-react/node_modules/eslint-plugin-import/lib/rules/no-amd.js:26:25)
    at ruleErrorHandler (/gh/forks/eslint-plugin-react/node_modules/eslint/lib/linter/linter.js:1145:48)
    at /gh/forks/eslint-plugin-react/node_modules/eslint/lib/linter/safe-emitter.js:45:58
```

antongolub:
Awaits https://github.com/import-js/eslint-plugin-import/pull/2996

lib/rules/destructuring-assignment.js
```diff
@@ -247,7 +248,8 @@ module.exports = {
           && destructureInSignature === 'always'
           && node.init.name === 'props'
         ) {
-          const scopeSetProps = context.getScope().set.get('props');
+          const scope = astUtil.getScope(context, node);
```
ljharb, line 251: will this behave differently, since it wasn't passing the node before?

lib/rules/jsx-no-constructed-context-values.js
```diff
@@ -179,7 +180,7 @@ module.exports = {
         }
 
         const valueExpression = valueNode.expression;
-        const invocationScope = context.getScope();
+        const invocationScope = astUtil.getScope(context, node);
```
ljharb, line 183: same here

lib/rules/jsx-no-undef.js
```diff
@@ -49,7 +50,7 @@ module.exports = {
      * @returns {void}
      */
     function checkIdentifierInJSX(node) {
-      let scope = context.getScope();
+      let scope = astUtil.getScope(context, node);
```
ljharb, line 53: and here

lib/rules/no-access-state-in-setstate.js
```diff
@@ -134,7 +136,7 @@ module.exports = {
             if (current.type === 'VariableDeclarator') {
               vars.push({
                 node,
-                scope: context.getScope(),
+                scope: astUtil.getScope(context, node),
```
ljharb, line 139: and here

lib/rules/no-access-state-in-setstate.js
```diff
@@ -155,10 +157,11 @@ module.exports = {
           current.parent.value === current
           || current.parent.object === current
         ) {
+          const scope = astUtil.getScope(context, node);
           while (current.type !== 'Program') {
             if (isFirstArgumentInSetStateCall(current, node)) {
               vars
-                .filter((v) => v.scope === context.getScope() && v.variableName === node.name)
+                .filter((v) => v.scope === scope && v.variableName === node.name)
```
ljharb, line 164: and here

lib/rules/no-access-state-in-setstate.js
```diff
@@ -176,7 +179,7 @@ module.exports = {
           if (property && property.key && property.key.name === 'state' && isDerivedFromThis) {
             vars.push({
               node: property.key,
-              scope: context.getScope(),
+              scope: astUtil.getScope(context, node),
```
ljharb, line 182: and here

ljharb (changes requested):
It'd be very convenient if this was rebased so there was one commit that added and used `astUtil.getScope`, and the rest was done in their own atomic commits :-)

lib/rules/destructuring-assignment.js
antongolub: 1. I just _want to believe_ it works similar enough as far [the migration guide](https://eslint.org/docs/latest/use/migrate-to-9.0.0) suggest this replacement.
2. I can rearrange the commits stack later, but I want to fix the tests first.

lib/util/ast.js
```diff
@@ -201,6 +201,16 @@ function getFirstNodeInLine(context, node) {
...
 }
 
+/**
+ * Get the scope of a context bound node
+ * @param {Context} context
+ * @param {ASTNode} node
+ * @return {Scope}
+ */
+function getScope(context, node) {
+  return (context.sourceCode || context).getScope(node);
```
ljharb, line 211: let’s rework this so it checks for context.sourceCode.getScope specifically, so it doesn’t break on versions of eslint with .sourceCode but not .sourceCode.getScope

antongolub: Done.

lib/rules/destructuring-assignment.js
antongolub: The worst part: 
> Additionally, methods Linter#defineRule(), Linter#defineRules(), Linter#defineParser(), and Linter#getRules() are no longer available.

lib/rules/forbid-prop-types.js
```diff
@@ -162,7 +162,7 @@ module.exports = {
           checkProperties(node.properties);
           break;
         case 'Identifier': {
-          const propTypesObject = variableUtil.findVariableByName(context, node.name);
+          const propTypesObject = variableUtil.findVariableByName(context, node);
```
ljharb, line 165: should this take `node, node.name`? or is that the default behavior

lib/rules/jsx-sort-default-props.js
```diff
@@ -87,11 +87,12 @@ module.exports = {
...
      * Find a variable by name in the current scope.
-     * @param  {string} name Name of the variable to look for.
+     * @param {ASTNode} node The node to check.
      * @returns {ASTNode|null} Return null if the variable could not be found, ASTNode otherwise.
      */
-    function findVariableByName(name) {
-      const variable = variableUtil.variablesInScope(context).find((item) => item.name === name);
+    function findVariableByName(node) {
+      const name = node.name;
+      const variable = variableUtil.variablesInScope(context, node).find((item) => item.name === name);
```
ljharb, line 95: it seems like `variableUtil.variablesInScope(context, node, node.name)` should do this logic for us. am i misreading?

lib/rules/require-render-return.js
```diff
@@ -60,7 +60,7 @@ module.exports = {
 
     return {
       ReturnStatement(node) {
-        const ancestors = context.getAncestors(node).reverse();
+        const ancestors = (context.sourceCode || context).getAncestors(node).reverse();
```
ljharb, line 63: let's put `getAncestors` in a util as well

lib/util/Components.js
```diff
@@ -612,10 +614,11 @@ function componentRule(rule, context) {
     /**
      * Get the parent stateless component node from the current scope
      *
+     * @param {ASTNode} _node
      * @returns {ASTNode} component node, null if we are not in a component
      */
-    getParentStatelessComponent() {
-      let scope = context.getScope();
+    getParentStatelessComponent(_node) {
+      let scope = (context.sourceCode || context).getScope(_node);
```
ljharb, line 621: → util

lib/util/propTypes.js
```diff
@@ -368,10 +368,15 @@ module.exports = function propTypesInstructions(context, components, utils) {
       node
       && node.type === 'Identifier'
     ) {
-      const scope = context.getScope();
-      const identVariable = scope.variableScope.variables.find(
-        (variable) => variable.name === node.name
-      );
+      const scope = astUtil.getScope(context, node);
+      const allVars = [];
+      let vars = (scope.childScopes[0] || scope).variableScope;
```
ljharb, line 373: can you elaborate on the need for this? did eslint 9 just straight up remove this functionality?

lib/rules/forbid-prop-types.js
antongolub: This is a temporary (I hope) workaroud for this case:
```ts
      'Program:exit'() {
        jsxElements.forEach((node) => {
          const openingEl = node.openingElement;
          const elName = elementType(openingEl);

          if (fragmentNames.has(elName) || refersToReactFragment(node, elName)) {
//...
const elementType = require('jsx-ast-utils/elementType');
//...
function refersToReactFragment(node, name) {
      const variableInit = variableUtil.findVariableByName(context, node, name)
```

lib/rules/jsx-sort-default-props.js
antongolub: `variableUtil.variablesInScope(context, node)` returns a set of vars visible from the specified node, and then we pick the one by name, as I understand.

lib/rules/require-render-return.js
antongolub: Sure. Np.

lib/rules/jsx-sort-default-props.js
ljharb: right - but providing the third argument to `variablesInScope` picks it by name already, doesn't it?

ljharb:
I believe this was obsoleted by #3759.

PR #3729, opened 2024-04-08
[New] `async-server-action`: Add rule to require that server actions be async

jorgezreik:
Adds a new rule to require that a server actions (functions with the `'use server'` directive) be async. Autofixes server actions that aren't async by adding the async keyword. Added to recommended as it prevents errors, and if a user is not using React Server Components then the rule is benign.

> Note: It is my first time contributing so let me know if I missed anything!


ljharb (changes requested):
I see some examples using single and double quotes - great! - but none using backticks. Can we add some? (whether it's supported or not)

jorgezreik:
> I see some examples using single and double quotes - great! - but none using backticks. Can we add some? (whether it's supported or not)

I knew I'd miss a few things 😅 Thanks for the feedback, really appreciate it! Taking a look now.

configs/recommended.js
```diff
@@ -5,6 +5,7 @@ const all = require('./all');
 module.exports = Object.assign({}, all, {
   languageOptions: all.languageOptions,
   rules: {
+    'react/async-server-action': 2,
```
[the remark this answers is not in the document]
jorgezreik: Did not know that, great point. Reverting.

CHANGELOG.md
```diff
@@ -1,22 +1,30 @@
...
-* [`boolean-prop-naming`]: detect TS interfaces ([#3701][] @developer-bandi)
-* [`boolean-prop-naming`]: literalType error fix ([#3704][] @developer-bandi)
-* [`boolean-prop-naming`]: allow TSIntersectionType ([#3705][] @developer-bandi)
-* [`no-unknown-property`]: support `popover`, `popovertarget`, `popovertargetaction` attributes ([#3707][] @ljharb)
-* [`no-unknown-property`]: only match `data-*` attributes containing `-` ([#3713][] @silverwind)
-* [`checked-requires-onchange-or-readonly`]: correct options that were behaving opposite ([#3715][] @jaesoekjjang)
+
+- [`jsx-no-leaked-render`]: prevent wrongly adding parens ([#3700][] @developer-bandi)
+- [`boolean-prop-naming`]: detect TS interfaces ([#3701][] @developer-bandi)
+- [`boolean-prop-naming`]: literalType error fix ([#3704][] @developer-bandi)
```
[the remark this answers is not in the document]
jorgezreik: 😱 I hadn't noticed, thanks! Reverting.

docs/rules/async-server-action.md
```diff
@@ -0,0 +1,57 @@
+# Require functions with the `use server` directive to be async (`react/async-server-action`)
+
+💼 This rule is enabled in the ☑️ `recommended` [config](https://github.com/jsx-eslint/eslint-plugin-react/#shareable-configs).
+
+🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
+
+<!-- end auto-generated rule header -->
+
+Require Server Actions (functions with the `use server` directive) to be async, as mandated by the `use server` [spec](https://react.dev/reference/react/use-server).
```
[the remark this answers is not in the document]
jorgezreik: So the feature is still in canary. I understand not wanting to ship a rule for a canary feature, and if that's the final policy then I'm happy to wait until it's out of canary to merge this in.

That being said, Server Components are definitely seeing widespread use, mainly through Vercel, with less stable implementations elsewhere. Additionally, the React team confirmed that they're officially shipping the feature with React 19 (when that comes out is anyone's guess).

Because of this, I think adding this rule (as an optional rule outside of the recommended config) would be very helpful to users using server components, while not impacting users who are not yet using them.

jorgezreik:
Quick question – the coverage report seems to not have been updated to the PR HEAD. Is there any way to force that to update and reflect the coverage of the latest commit (should be full change coverage now)?

lib/rules/async-server-action.js
```diff
@@ -0,0 +1,85 @@
...
+  const functionBody = node.body.body;
+  if (functionBody.length === 0) return false;
+
+  const potentialDirectiveStatement = functionBody[0];
+  if (potentialDirectiveStatement.type !== 'ExpressionStatement') return false;
+
+  const potentialDirectiveExpression = potentialDirectiveStatement.expression;
+  if (potentialDirectiveExpression.type !== 'Literal') return false;
+
+  return potentialDirectiveExpression.value === 'use server';
```
ljharb, line 36: is there a way to use an AST selector string here, like `FunctionDeclaration BlockStatement ExpressionStatement Literal[value="use server"]`, and then this function wouldn't be required at all?

jorgezreik: Great suggestion! I simplified the rule a ton with a selector.

lib/rules/async-server-action.js
```diff
@@ -0,0 +1,52 @@
...
+  create(context) {
+    let currentFunction = null;
+
+    return {
+      ':function'(node) {
+        currentFunction = node;
+      },
+      ':function[async=false]>BlockStatement>:first-child[expression.value="use server"]'() {
+        report(context, messages.asyncServerAction, 'asyncServerAction', {
+          node: currentFunction,
```
ljharb, line 44:
```suggestion
      ':function[async=false]>BlockStatement>:first-child[expression.value="use server"]'(node) {
        report(context, messages.asyncServerAction, 'asyncServerAction', {
          node: node.parentNode.parentNode,
```
or something like this? then you wouldn't need the `:function` listener at all.

lib/rules/async-server-action.js
```diff
@@ -0,0 +1,52 @@
...
+          fix(fixer) {
+            return fixer.insertTextBefore(currentFunction, 'async ');
+          },
```
ljharb, line 47: i'm not 100% convinced this should be an autofix (as opposed to a suggestion). If there is _any_ situation where the proper fix is to remove the "use server" pragma, instead of adding `async`, then an autofix wouldn't be appropriate.

jorgezreik: After thinking about it for a while I do think a suggestion is better. 99% of the time it should be better to add async, but I can imagine edge cases. Also a suggestion would do a better job of informing why async is required.

lib/rules/async-server-action.js
jorgezreik: Yeah that works for sure. I felt the listener was clearer but just chaining up parents is definitely simpler. Changed.

lib/rules/async-server-action.js
jorgezreik: Changed to suggest.

lib/rules/async-server-action.js
```diff
@@ -0,0 +1,57 @@
...
+        const functionName = currentFunction.id ? `\`${currentFunction.id.name}\`` : 'this function';
+        report(context, messages.asyncServerAction, 'asyncServerAction', {
+          node: currentFunction,
+          data: {
+            functionName,
+          },
+          suggest: [{
+            desc: messages.suggestAsync,
+            fix(fixer) {
+              return fixer.insertTextBefore(currentFunction, 'async ');
```
ljharb, line 50: can we also ensure that this won't apply to generators?

jorgezreik: Is there a reason this can't be the case? I'd imagine that if someone added the `"use server"` directive to a generator function it should still be async.

ljharb: can components even *be* generator functions?

also, async generators vs generators is a very different semantic shift than async functions vs functions.

jorgezreik: Changed!

jorgezreik:
A semi-related question that came to mind that I'd like advice on:

Other ESLint rules like `eslint/require-await` and `typescript-eslint/require-await` report async functions that do not have an await operator as incorrect. Any ideas on how we can make rules like these not report on `"use server"`? Adding options to these rules to ignore functions with certain directives would do the trick, but an option like that is React-specific enough that it's unlikely to get merged.

ljharb:
The `require-await` rules are harmful and bad and should never have existed in the first place. An `async function` without any `await`s is perfectly acceptable (unlike a generator without a `yield`).

As such, the solution there is to just disable those horrifically foolish rules.

ljharb (approved):
Rebased. I'd love to see lots more test cases, and I'm not going to merge this until the feature is available in a non-canary full release, so I'll mark it as a draft until then.

jorgezreik:
> Rebased. I'd love to see lots more test cases, and I'm not going to merge this until the feature is available in a non-canary full release, so I'll mark it as a draft until then.

Makes sense to me! When the full release rolls around we can spruce up the test cases and get this merged :)

jorgezreik:
Hi! React 19 [launched as stable](https://react.dev/blog/2024/12/05/react-19) today.

Are we still interested in getting this merged? If so we can fix merge conflicts and spruce up tests to do so!

ljharb:
@jorgezreik yep, let's do those things :-) mark it as ready for review when you think it's so

PR #3731, opened 2024-04-09
Add JSDoc type annotations

y-hsgw:
I have added type annotations and fixed associated type errors.

### DIFF
--- a/lib/rules/button-has-type.js
+++ b/lib/rules/button-has-type.js
@@ -28,6 +28,7 @@ const messages = {
   forbiddenValue: '"{{value}}" is an invalid value for button type attribute',
 };
 
+/** @type {import('eslint').Rule.RuleModule} */
 module.exports = {
   meta: {
     docs: {
@@ -149,14 +150,19 @@ module.exports = {
         }
 
         const props = node.arguments[1].properties;
-        const typeProp = props.find((prop) => prop.key && prop.key.name === 'type');
+        const typeProp = props.find((prop) => (
+          'key' in prop
+          && prop.key
+          && 'name' in prop.key
+          && prop.key.name === 'type'
+        ));
 
         if (!typeProp) {
           reportMissing(node);
           return;
         }
 
-        checkExpression(node, typeProp.value);
+        checkExpression(node, 'value' in typeProp ? typeProp.value : undefined);
       },
     };
   },
--- a/lib/rules/checked-requires-onchange-or-readonly.js
+++ b/lib/rules/checked-requires-onchange-or-readonly.js
@@ -24,7 +24,7 @@ const defaultOptions = {
 };
 
 /**
- * @param {string[]} properties
+ * @param {object[]} properties
  * @param {string} keyName
  * @returns {Set<string>}
  */
@@ -41,6 +41,7 @@ function extractTargetProps(properties, keyName) {
   );
 }
 
+/** @type {import('eslint').Rule.RuleModule} */
 module.exports = {
   meta: {
     docs: {
--- a/lib/rules/forbid-elements.js
+++ b/lib/rules/forbid-elements.js
@@ -20,6 +20,7 @@ const messages = {
   forbiddenElement_message: '<{{element}}> is forbidden, {{message}}',
 };
 
+/** @type {import('eslint').Rule.RuleModule} */
 module.exports = {
   meta: {
     docs: {
@@ -105,13 +106,11 @@ module.exports = {
           return;
         }
 
-        const argType = argument.type;
-
-        if (argType === 'Identifier' && /^[A-Z_]/.test(argument.name)) {
+        if (argument.type === 'Identifier' && /^[A-Z_]/.test(argument.name)) {
           reportIfForbidden(argument.name, argument);
-        } else if (argType === 'Literal' && /^[a-z][^.]*$/.test(argument.value)) {
+        } else if (argument.type === 'Literal' && /^[a-z][^.]*$/.test(String(argument.value))) {
           reportIfForbidden(argument.value, argument);
-        } else if (argType === 'MemberExpression') {
+        } else if (argument.type === 'MemberExpression') {
           reportIfForbidden(getText(context, argument), argument);
         }
       },
--- a/lib/rules/forbid-foreign-prop-types.js
+++ b/lib/rules/forbid-foreign-prop-types.js
@@ -13,6 +13,7 @@ const messages = {
   forbiddenPropType: 'Using propTypes from another component is not safe because they may be removed in production builds',
 };
 
+/** @type {import('eslint').Rule.RuleModule} */
 module.exports = {
   meta: {
     docs: {
@@ -108,7 +109,9 @@ module.exports = {
             && !ast.isAssignmentLHS(node)
             && !isAllowedAssignment(node)
           )) || (
+            // @ts-expect-error The JSXText type is not present in the estree type definitions
             (node.property.type === 'Literal' || node.property.type === 'JSXText')
+            && 'value' in node.property
             && node.property.value === 'propTypes'
             && !ast.isAssignmentLHS(node)
             && !isAllowedAssignment(node)
@@ -121,7 +124,11 @@ module.exports = {
       },
 
       ObjectPattern(node) {
-        const propTypesNode = node.properties.find((property) => property.type === 'Property' && property.key.name === 'propTypes');
+        const propTypesNode = node.properties.find((property) => (
+          property.type === 'Property'
+          && 'name' in property.key
+          && property.key.name === 'propTypes'
+        ));
 
         if (propTypesNode) {
           report(context, messages.forbiddenPropType, 'forbiddenPropType', {
--- a/lib/rules/forbid-prop-types.js
+++ b/lib/rules/forbid-prop-types.js
@@ -26,6 +26,7 @@ const messages = {
   forbiddenPropType: 'Prop type "{{target}}" is forbidden',
 };
 
+/** @type {import('eslint').Rule.RuleModule} */
 module.exports = {
   meta: {
     docs: {
@@ -192,7 +193,9 @@ module.exports = {
           }
           if (node.specifiers.length >= 1) {
             const propTypesSpecifier = node.specifiers.find((specifier) => (
-              specifier.imported && specifier.imported.name === 'PropTypes'
+              'imported' in specifier
+              && specifier.imported
+              && specifier.imported.name === 'PropTypes'
             ));
             if (propTypesSpecifier) {
               propTypesPackageName = propTypesSpecifier.local.name;
@@ -228,12 +231,13 @@ module.exports = {
           return;
         }
 
-        checkNode(node.parent.right);
+        checkNode('right' in node.parent && node.parent.right);
       },
 
       CallExpression(node) {
         if (
-          node.callee.object
+          node.callee.type === 'MemberExpression'
+          && node.callee.object
           && !isPropTypesPackage(node.callee.object)
           && !propsUtil.isPropTypesDeclaration(node.callee)
         ) {
@@ -242,9 +246,12 @@ module.exports = {
 
         if (
           node.arguments.length > 0
-          && (node.callee.name === 'shape' || astUtil.getPropertyName(node.callee) === 'shape')
+          && (
+            ('name' in node.callee && node.callee.name === 'shape')
+            || astUtil.getPropertyName(node.callee) === 'shape'
+          )
         ) {
-          checkProperties(node.arguments[0].properties);
+          checkProperties('properties' in node.arguments[0] && node.arguments[0].properties);
         }
       },
 
@@ -267,7 +274,7 @@ module.exports = {
 
       ObjectExpression(node) {
         node.properties.forEach((property) => {
-          if (!property.key) {
+          if (!('key' in property) || !property.key) {
             return;
           }
 
--- a/lib/rules/jsx-closing-bracket-location.js
+++ b/lib/rules/jsx-closing-bracket-location.js
@@ -20,6 +20,7 @@ const messages = {
   bracketLocation: 'The closing bracket must be {{location}}{{details}}',
 };
 
+/** @type {import('eslint').Rule.RuleModule} */
 module.exports = {
   meta: {
     docs: {
--- a/lib/rules/jsx-curly-spacing.js
+++ b/lib/rules/jsx-curly-spacing.js
@@ -35,6 +35,7 @@ const messages = {
   spaceNeededBefore: 'A space is required before \'{{token}}\'',
 };
 
+/** @type {import('eslint').Rule.RuleModule} */
 module.exports = {
   meta: {
     docs: {
--- a/lib/rules/jsx-equals-spacing.js
+++ b/lib/rules/jsx-equals-spacing.js
@@ -20,6 +20,7 @@ const messages = {
   needSpaceAfter: 'A space is required after \'=\'',
 };
 
+/** @type {import('eslint').Rule.RuleModule} */
 module.exports = {
   meta: {
     docs: {
--- a/lib/rules/jsx-fragments.js
+++ b/lib/rules/jsx-fragments.js
@@ -22,12 +22,12 @@ function replaceNode(source, node, text) {
 }
 
 const messages = {
-  fragmentsNotSupported: 'Fragments are only supported starting from React v16.2. '
-    + 'Please disable the `react/jsx-fragments` rule in `eslint` settings or upgrade your version of React.',
+  fragmentsNotSupported: 'Fragments are only supported starting from React v16.2. Please disable the `react/jsx-fragments` rule in `eslint` settings or upgrade your version of React.',
   preferPragma: 'Prefer {{react}}.{{fragment}} over fragment shorthand',
   preferFragment: 'Prefer fragment shorthand over {{react}}.{{fragment}}',
 };
 
+/** @type {import('eslint').Rule.RuleModule} */
 module.exports = {
   meta: {
     docs: {
@@ -170,7 +170,7 @@ module.exports = {
       ImportDeclaration(node) {
         if (node.source && node.source.value === 'react') {
           node.specifiers.forEach((spec) => {
-            if (spec.imported && spec.imported.name === fragmentPragma) {
+            if ('imported' in spec && spec.imported && spec.imported.name === fragmentPragma) {
               if (spec.local) {
                 fragmentNames.add(spec.local.name);
               }
--- a/lib/rules/jsx-indent-props.js
+++ b/lib/rules/jsx-indent-props.js
@@ -45,6 +45,7 @@ const messages = {
   wrongIndent: 'Expected indentation of {{needed}} {{type}} {{characters}} but found {{gotten}}.',
 };
 
+/** @type {import('eslint').Rule.RuleModule} */
 module.exports = {
   meta: {
     docs: {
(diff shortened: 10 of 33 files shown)


lib/rules/button-has-type.js
```diff
@@ -149,9 +150,11 @@ module.exports = {
         }
 
         const props = node.arguments[1].properties;
-        const typeProp = props.find((prop) => prop.key && prop.key.name === 'type');
+        const typeProp = props.find(
+          (prop) => prop.type === "Property" && prop.key && prop.key.type === "PrivateIdentifier" && prop.key.name === 'type'
+        );
 
-        if (!typeProp) {
+        if (typeProp.type !== "Property") {
```
ljharb, line 157: these changes aren't simple annotation changes. can they be extracted at least to a different commit, if not to a different PR?

lib/rules/checked-requires-onchange-or-readonly.js
```diff
@@ -24,7 +24,7 @@ const defaultOptions = {
 };
 
 /**
- * @param {string[]} properties
+ * @param {Array} properties
```
ljharb, line 27: `Array` is less specific than `string[]`

lib/rules/checked-requires-onchange-or-readonly.js
```diff
@@ -41,6 +41,7 @@ function extractTargetProps(properties, keyName) {
   );
 }
 
+/** @type { import('eslint').Rule.RuleModule } */
```
ljharb, line 44: curly braces should never have spaces directly inside them
```suggestion
/** @type {import('eslint').Rule.RuleModule} */
```

lib/rules/forbid-elements.js
```diff
@@ -103,13 +104,11 @@ module.exports = {
...
-        const argType = argument.type;
-
-        if (argType === 'Identifier' && /^[A-Z_]/.test(argument.name)) {
+        if (argument.type === 'Identifier' && /^[A-Z_]/.test(argument.name)) {
           reportIfForbidden(argument.name, argument);
-        } else if (argType === 'Literal' && /^[a-z][^.]*$/.test(argument.value)) {
+        } else if (argument.type === 'Literal' && /^[a-z][^.]*$/.test(argument.value.toString())) {
           reportIfForbidden(argument.value, argument);
-        } else if (argType === 'MemberExpression') {
+        } else if (argument.type === 'MemberExpression') {
```
ljharb, line 111: the original only has one observable lookup; this has 3. why is this an improvement?

also, `String(x)`, _never_ `x.toString()`

lib/rules/forbid-foreign-prop-types.js
```diff
@@ -108,7 +109,7 @@ module.exports = {
             && !ast.isAssignmentLHS(node)
             && !isAllowedAssignment(node)
           )) || (
-            (node.property.type === 'Literal' || node.property.type === 'JSXText')
+            node.property.type === 'Literal'
             && node.property.value === 'propTypes'
```
ljharb, line 113: this is meaningfully different; it will stop catching JSXText nodes.

ljharb (changes requested):
this isn't just adding type annotations and a lot of the changes are potentially breaking.

lib/rules/button-has-type.js
y-hsgw: I see. My apologies 🙇‍♂️. 
Would it be acceptable to create a PR that adds annotations only to rules where type changes are unnecessary?
Or would it be better to add annotations to all rules listed below, and if changes are needed, ignore the entire file (@ts-nocheck)?

ljharb: yes, that would be fine.

any non-type changes that are needed must be accompanied by tests - if the ostensible type error can't be reproduced, then the type system is wrong.

y-hsgw:
I have addressed this in pull request #3732 and will proceed to close this one. I'll handle the areas requiring type changes separately.

ljharb:
I would prefer to update this PR with those changes rather than abandoning it.

lib/rules/checked-requires-onchange-or-readonly.js
y-hsgw: fixed in 9018532 .

lib/rules/forbid-elements.js
y-hsgw: > the original only has one observable lookup; this has 3. why is this an improvement?

The reason is because the following error message occurs during the execution of the type-check script.
`Property 'name' does not exist on type 'ClassExpression | ArrayExpression | ArrowFunctionExpression | AssignmentExpression | ... 23 more ... | SpreadElement'.  Property 'name' does not exist on type 'ClassExpression'.`

> also, String(x), never x.toString()

thanks! Fixed in 4ae24a2

lib/rules/checked-requires-onchange-or-readonly.js
```diff
@@ -24,7 +24,7 @@ const defaultOptions = {
 };
 
 /**
- * @param {string[]} properties
+ * @param {Object[]} properties
```
ljharb, line 27:
```suggestion
 * @param {object[]} properties
```
let's use TS-jsdoc since that's the one that matters

lib/rules/forbid-foreign-prop-types.js
```diff
@@ -121,7 +123,9 @@ module.exports = {
       },
 
       ObjectPattern(node) {
-        const propTypesNode = node.properties.find((property) => property.type === 'Property' && property.key.name === 'propTypes');
+        const propTypesNode = node.properties.find(
+          (property) => property.type === 'Property' && property.key.type === 'Identifier' && property.key.name === 'propTypes'
+        );
```
ljharb, line 128: this change doesn't seem needed

lib/rules/forbid-prop-types.js
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
ljharb, line 1: please remove this

lib/rules/jsx-closing-bracket-location.js
```diff
@@ -195,7 +196,9 @@ module.exports = {
      */
     function getTokensLocations(node) {
       const sourceCode = context.getSourceCode();
+      // @ts-ignore
       const opening = sourceCode.getFirstToken(node).loc.start;
+      // @ts-ignore
```
ljharb, line 201: please use ts-expect-error, with a comment, instead of ts-ignore.

lib/rules/jsx-curly-spacing.js
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
ljharb, line 1:
```suggestion
```

lib/rules/jsx-equals-spacing.js
```diff
@@ -61,7 +62,9 @@ module.exports = {
 
           const sourceCode = context.getSourceCode();
           const equalToken = sourceCode.getTokenAfter(attrNode.name);
+          // @ts-ignore
           const spacedBefore = sourceCode.isSpaceBetweenTokens(attrNode.name, equalToken);
+          // @ts-ignore
```
ljharb, line 67: ts-expect-error

lib/rules/button-has-type.js
```diff
@@ -149,9 +150,11 @@ module.exports = {
         }
 
         const props = node.arguments[1].properties;
-        const typeProp = props.find((prop) => prop.key && prop.key.name === 'type');
+        const typeProp = props.find(
+          (prop) => prop.type === 'Property' && prop.key && prop.key.type === 'PrivateIdentifier' && prop.key.name === 'type'
+        );
 
-        if (!typeProp) {
+        if (typeProp.type !== 'Property') {
```
ljharb, line 157: test?

lib/rules/jsx-fragments.js
```diff
@@ -171,7 +172,7 @@ module.exports = {
       ImportDeclaration(node) {
         if (node.source && node.source.value === 'react') {
           node.specifiers.forEach((spec) => {
-            if (spec.imported && spec.imported.name === fragmentPragma) {
+            if (spec.type === 'ImportSpecifier' && spec.imported && spec.imported.name === fragmentPragma) {
```
ljharb, line 175: why is this needed? what are the possible `type` values of the objects contained an ImportDeclaration node's `specifiers` array - is it more than just `ImportSpecifier`?

lib/rules/jsx-indent-props.js
```diff
@@ -140,6 +141,7 @@ module.exports = {
      * @return {Number} Indent
      */
     function getNodeIndent(node) {
+      // @ts-ignore
```
ljharb, line 144: ts-expect-error

lib/rules/jsx-indent.js
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
ljharb, line 1:
```suggestion
```

lib/rules/jsx-no-bind.js
```diff
@@ -122,8 +123,8 @@ module.exports = {
       blockVariableNameSets[blockStart][violationType].add(variableName);
     }
 
-    function getBlockStatementAncestors(node) {
-      return context.getAncestors(node).filter(
+    function getBlockStatementAncestors() {
+      return context.getAncestors().filter(
```
ljharb, line 127: i'm pretty sure this change will cause eslint 9 to break. can you confirm?

lib/rules/jsx-no-bind.js
```diff
@@ -168,13 +169,15 @@ module.exports = {
...
           return;
         }
-        const blockAncestors = getBlockStatementAncestors(node);
+        const blockAncestors = getBlockStatementAncestors();
         const variableViolationType = getNodeViolationType(node.init);
 
         if (
           blockAncestors.length > 0
           && variableViolationType
+          && node.parent.type === 'VariableDeclaration'
```
ljharb, line 178: we're already inside VariableDeclarator, this shouldn't be needed

lib/rules/jsx-sort-default-props.js
```diff
@@ -172,6 +173,14 @@ module.exports = {
           return;
         }
 
+        if (
+          node.parent.type !== 'BinaryExpression'
+          && node.parent.type !== 'AssignmentExpression'
+          && node.parent.type !== 'LogicalExpression'
+        ) {
```
ljharb, line 180: tests?

lib/rules/jsx-space-before-closing.js
```diff
@@ -63,6 +64,21 @@ module.exports = {
...
+          && closingSlash.type !== 'Keyword'
+          && closingSlash.type !== 'Punctuator'
+          && closingSlash.type !== 'JSXIdentifier'
+          && closingSlash.type !== 'JSXText'
+          && closingSlash.type !== 'Numeric'
+          && closingSlash.type !== 'String'
+          && closingSlash.type !== 'RegularExpression'
+        ) {
+          return;
+        }
```
ljharb, line 80: tests?

ljharb (changes requested):
i only went like 20% through, but this should help you refine the entire thing

lib/rules/checked-requires-onchange-or-readonly.js
y-hsgw: Thanks! Fixed in 844d483

lib/rules/forbid-foreign-prop-types.js
y-hsgw: Sorry🙇‍♂️  Fixed in https://github.com/jsx-eslint/eslint-plugin-react/commit/844d483855c56a6741cf56b0e3c9be2458d45e68 .

lib/rules/jsx-no-bind.js
y-hsgw: `Property 'kind' does not exist on type 'Node'.`
This error occurs.

lib/rules/jsx-fragments.js
y-hsgw: > what are the possible type values of the objects contained an ImportDeclaration node's specifiers array - is it more than just ImportSpecifier?

`ImportSpecifier | ImportDefaultSpecifier | ImportNamespaceSpecifier` .

I added it because only ImportSpecifier has the imported property.

lib/rules/jsx-no-bind.js
y-hsgw: Sorry. I do not understand it well.
Should I add v9 to eslint in package.json as shown below and see if I get a compile error?
```
  "devDependencies": {
    "eslint": "^3 || ^4 || ^5 || ^6 || ^7 || ^8 || ^9",
    ...
  },
  "peerDependencies": {
    "eslint": "^3 || ^4 || ^5 || ^6 || ^7 || ^8 || ^9"
  },
```

ljharb: No, adding eslint 9 is a hard task. What i mean is, what’s wrong with passing the node here?

lib/rules/jsx-no-bind.js
ljharb: if this is just for narrowing, we can use type casts to tell TS what kind of thing it is.

lib/rules/jsx-equals-spacing.js
y-hsgw: Thanks! Fixed in [eb44ba7](https://github.com/jsx-eslint/eslint-plugin-react/pull/3731/commits/eb44ba71aba5ed0d36518b3c658b7cab5e59c14f) .

lib/rules/jsx-indent-props.js
y-hsgw: Thanks! Fixed in [eb44ba7](https://github.com/jsx-eslint/eslint-plugin-react/pull/3731/commits/eb44ba71aba5ed0d36518b3c658b7cab5e59c14f) .

lib/rules/jsx-no-bind.js
y-hsgw: When looking at the type information for context.getAncestors, it seems that the argument is unnecessary.
``` ts
(method) Rule.RuleContext.getAncestors(): Node[]
```

lib/rules/jsx-no-bind.js
y-hsgw: In my understanding, since it's a JavaScript file, I don't think type assertions can be used.

lib/rules/jsx-closing-bracket-location.js
y-hsgw: Thanks! Fixed in [eb44ba7](https://github.com/jsx-eslint/eslint-plugin-react/pull/3731/commits/eb44ba71aba5ed0d36518b3c658b7cab5e59c14f) .

lib/rules/jsx-closing-bracket-location.js
```diff
@@ -196,9 +196,11 @@ module.exports = {
      */
     function getTokensLocations(node) {
       const sourceCode = context.getSourceCode();
-      // @ts-ignore
+      // The types differ between ASTNode and ESTree.Node.
+      // @ts-expect-error
```
ljharb, line 200:
```suggestion
      // @ts-expect-error The types differ between ASTNode and ESTree.Node.
```

y-hsgw: Fixed in a5ac70f.

lib/rules/forbid-prop-types.js
y-hsgw: Fixed in[25e6afd](https://github.com/jsx-eslint/eslint-plugin-react/pull/3731/commits/25e6afd2a02e62af7d806fa6c2e12de46a97dcad).

lib/rules/jsx-indent.js
y-hsgw: Fixed in[25e6afd](https://github.com/jsx-eslint/eslint-plugin-react/pull/3731/commits/25e6afd2a02e62af7d806fa6c2e12de46a97dcad).

lib/rules/button-has-type.js
```diff
@@ -149,9 +150,11 @@ module.exports = {
         }
 
         const props = node.arguments[1].properties;
```
ljharb, line 152: is the type of `node` properly narrowed to a CallExpression node here? if not, let's do that as an annotation on the visitor function rather than needing to add all these runtime changes.

y-hsgw: I attempted to make corrections in 040d77c, but is my understanding correct?

lib/rules/jsx-sort-default-props.js
y-hsgw: My apologies for the confusion. I mistakenly performed unnecessary type narrowing, so I removed it in commit d59ac20. Since the existing test cases seem to be unaffected, I did not add any additional tests.

lib/rules/jsx-space-before-closing.js
y-hsgw: My apologies for the confusion. I mistakenly performed unnecessary type narrowing, so I removed it in commit https://github.com/jsx-eslint/eslint-plugin-react/commit/d59ac204a929314b6c24956f7c3b31dd036a28fa. Since the existing test cases seem to be unaffected, I did not add any additional tests.

lib/rules/jsx-fragments.js
ljharb: would `'imported' in spec` narrow properly?

lib/rules/jsx-no-bind.js
ljharb: they can, by wrapping the value in parens and preceding it with a `@type comment.

y-hsgw: I didn't know that method! Thank you!  Fixed in e2c3497 .

lib/rules/jsx-fragments.js
y-hsgw: Yes, it will be narrowed down. Alternatively, it might be better to do it as follows.

```js
 if (spec.type === 'ImportSpecifier' && spec.imported.name === fragmentPragma) { ...
```

lib/rules/button-has-type.js
ljharb: That's an acceptable narrowing :-) thanks

ljharb (changes requested):
Please rebase this PR, and extract any changes that aren't about whitespace/parens, or adding/editing comments, to separate commits - and then all the "type annotations" changes can be in a single commit i can review separately. Thanks!

y-hsgw:
I have rebased and cleaned up the commits. Please check.

ljharb:
I've gone ahead and updated this to remove all the breaking changes. The way to safely resolve TypeScript's inability to understand that an `x.foo` check is identical to `'foo' in x && x.foo` semantically is not to hardcode AST node types (that a custom parser might not match) but to add `'foo' in x` and let TS do the narrowing.

ljharb (approved):

lib/rules/no-is-mounted.js
```diff
@@ -38,7 +39,10 @@ module.exports = {
         if (callee.type !== 'MemberExpression') {
           return;
         }
-        if (callee.object.type !== 'ThisExpression' || callee.property.name !== 'isMounted') {
+        if (
+          callee.object.type !== 'ThisExpression'
+          && (!('name' in callee.property) || callee.property.name !== 'isMounted')
+        ) {
```
voxpelli, line 45: This caused a regression, see https://github.com/jsx-eslint/eslint-plugin-react/issues/3819

It should always return when `callee.object.type` is not `'ThisExpression'` – no matter if `'name'` is missing in `callee.property` or not.

The functionality before this PR would be equivalent to this type checker compliant check:

```js
        if (
          callee.object.type !== 'ThisExpression' ||
          !('name' in callee.property) ||
          callee.property.name !== 'isMounted'
        ) {
```

voxpelli: PR with said fix: https://github.com/jsx-eslint/eslint-plugin-react/pull/3821

PR #3734, opened 2024-04-10
Add type annotation for button-has-type rule

y-hsgw:
This PR adds JSDoc type annotations to the button-has-type rule.


y-hsgw:
I will close this as it is planned to be addressed in pull request #3731.

Issue #3735, opened 2024-04-14
[Bug]: peer dependency error with eslint@9


johnhunter:
### Is there an existing issue for this?

- [X] I have searched the existing issues and my issue is unique
- [X] My issue appears in the command-line and not only in the text editor

### Description Overview

We support `eslint@9` but the `peerDependencies` do not include it. This causes npm import failure when using current npm versions.

**Steps to repro:**
1. Update to `eslint@9`
2. Install `eslint-plugin-react@7.34.1`
3. Examine the console output

**Expected:**
eslint-plugin-react install ok

**Actual:**
eslint-plugin-react install fails with

```
npm ERR! Could not resolve dependency:
npm ERR! peer eslint@"^3 || ^4 || ^5 || ^6 || ^7 || ^8" from eslint-plugin-react@7.34.1
npm ERR! node_modules/eslint-plugin-react
npm ERR!   dev eslint-plugin-react@"7.34.1" from the root project
```

Environment:
node: v20.12.2
npm: 10.5.0

The fail will downgrade to a warning if npm legacy-peer-deps is used, but we should address the root cause.

### Expected Behavior

eslint-plugin-react should install ok

### eslint-plugin-react version

v7.34.1

### eslint version

v9.0.0

### node version

v20.12.2

ljharb:
Please wait to file PRs until a maintainer tells you it's desired. Your PR is a duplicate of #3726, and this issue is a duplicate of #3699.

PR #3736, opened 2024-04-14
Add eslint@9 to peer deps

johnhunter:
resolved issue https://github.com/jsx-eslint/eslint-plugin-react/issues/3735


ljharb:
This is a duplicate of #3726, and does not actually test on eslint 9 (or else tons of tests would fail).

PR #3743, opened 2024-04-28
feat: Add eslint v9 support

MatiPl01:
## Description

This PR adds eslint v9 support. It also adjusts respective tests to use v9-style config.

### Fixing tests

I got most of tests working by simply changing the config used in tests or removing duplicate test cases. Below is the list of rules which tests still are failing and fixing them will require a bit more effort. I will continue fixing tests in my free time.

- [x] jsx-no-constructed-context-values
- [x] no-object-type-as-default-prop
- [x] default-props-match-prop-types
- [x] destructuring-assignment
- [x] no-set-state
- [x] display-name
- [x] no-string-refs
- [x] jsx-no-undef
- [x] no-this-in-sfc
- [ ] no-typos
- [x] function-component-definition
- [x] no-unstable-nested-components
- [ ] hook-use-state
- [x] no-unused-prop-types
- [ ] jsx-closing-bracket-location
- [x] jsx-uses-react
- [x] jsx-uses-vars
- [x] no-access-state-in-setstate
- [x] prefer-stateless-function
- [ ] prop-types
- [x] require-default-props
- [x] jsx-first-prop-new-line
- [x] no-danger-with-children
- [x] no-direct-mutation-state
- [x] static-property-placement
- [ ] no-invalid-html-attribute
- [x] void-dom-elements-no-children
- [x] no-multi-comp

### Fixing eslint config

After upgrading `eslint` to v9, the `.eslintrc` config used in the `eslint-plugin-react` repo is no longer valid. We should either migrate from `.eslintrc` config to `eslint.config.js`, or use old version of `eslint` just for linting in this repository.

Because `.eslintrc` config is no longer valid, CI shows that all tests are failing. I ran test files directly with the following command while fixing test issues:

```sh
./node_modules/mocha/bin/_mocha tests/lib/rules/<rule-name>.js
```


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "eslint-plugin-react",
-  "version": "7.34.0",
+  "version": "7.34.1",
```
ljharb, line 3: PRs must never bump versions, and this would be a minor anyways, not a patch

ljharb (changes requested):
We can’t support anything we’re not testing, and this doesn’t add v9 to tests.

package.json
```diff
@@ -57,7 +57,7 @@
     "@typescript-eslint/parser": "^2.34.0 || ^3.10.1 || ^4.0.0 || ^5.0.0",
     "aud": "^2.0.4",
     "babel-eslint": "^8 || ^9 || ^10.1.0",
-    "eslint": "^3 || ^4 || ^5 || ^6 || ^7 || ^8",
+    "eslint": "^3 || ^4 || ^5 || ^6 || ^7 || ^8 | ^9",
```
ljharb, line 60:
```suggestion
    "eslint": "^3 || ^4 || ^5 || ^6 || ^7 || ^8 || ^9",
```

package.json
MatiPl01: Yea, but I don't know why version on `master` is 7.34.0, whilst the last released version is 7.34.1. Is there a different branch which is more up to date than the `master` branch?

ljharb:
Support is definitely intended :-) #3726 is another attempt.

package.json
ljharb: Should be fixed on a rebase.

MatiPl01:
@ljharb

I fixed most of failing tests and adjusted them to eslint v9 style. There are 5 remaining rules which tests still don't pass. I can get back to fixing them later.

I wanted to ask you, if you have an idea how to handle linting in this repo? I saw that `.eslintrc` extends `airbnb` configs which still don't support eslint v9, so we would have to somehow use eslint v9 and v8 at the same time, v8 for linting code in this repo and v9 in tests.

I wonder if creating yarn workspaces with different eslint versions would be a valid solution for you and if it will solve the problem.

tests/lib/rules/static-property-placement.js
```diff
@@ -19,14 +19,16 @@ const PROPERTY_ASSIGNMENT = 'property assignment';
 const RuleTester = require('eslint').RuleTester;
 const rule = require('../../../lib/rules/static-property-placement');
 
-const parsers = require('../../helpers/parsers-old');
+const parsers = require('../../helpers/parsers');
 
 const ruleTesterConfig = {
-  parserOptions: {
+  languageOptions: {
```
ljharb, line 25: this seems like it will break on older eslint versions (and eslintrc config), which we still need to retain support for

MatiPl01: This change was made only in tests and is necessary after upgrading to eslint v9. I don't know if we should keep separate tests for older eslint versions. Maybe we should somehow duplicate all tests to test rules separately with eslint v9 and with older versions.

MatiPl01: The alternative approach is to use eslint v8 as a dev dependency and test rules with eslint v8 as it was done before, and add v9 to peerDependencies but don't test rules with the latest eslint. I saw this approach in the [eslint-plugin-perfectionist](https://github.com/azat-io/eslint-plugin-perfectionist/tree/main) repo, where eslint v9 support was added but tests are still written in v8. This plugin dropped support for eslint v7 and older as the new API eslint API was introduced in v8.

ljharb:
Don't worry about the linting yet; the import plugin has to come first for that.

However, we already have to retain compatibility for every eslint version we currently support, so we can't switch things over to just work on v9, or just on flat config.

If we used workspaces, we'd only use npm, but I don't think that's necessary - the way we support multiple eslint versions now is by manually altering the deps in CI, and we'd do that here as well.

MatiPl01:
> Don't worry about the linting yet; the import plugin has to come first for that.
> 
> However, we already have to retain compatibility for every eslint version we currently support, so we can't switch things over to just work on v9, or just on flat config.
> 
> If we used workspaces, we'd only use npm, but I don't think that's necessary - the way we support multiple eslint versions now is by manually altering the deps in CI, and we'd do that here as well.

I see, thanks for clarification. I still wonder what will be the best way to ensure that tests work in all eslint versions. v9 moved `parserOptions` to the new `languageOptions` property so we can't simply add v9 to the list of tested eslint versions on CI, because all tests will fail.

ljharb:
We'll probably need a helper function that detects the eslint version and adjusts appropriately.

The smallest change for now would be to leave the tests largely as-is, and have the helper change them to the v9 format when applicable - later, we'd probably want to invert that, but that can wait.

ljharb:
I think that #3759 obsoletes this, but if not please reopen it and rebase it.

PR #3748, opened 2024-05-06
[New] `no-danger`: add `customComponentNames` option

### LINKED ISSUE #3434: feat: no-danger find props on all components (not just native)
Unless I am mistaken, it seems today that the rule `react/no-danger` has specific logic to only find usages on native elements. See the `isDOMComponent` check here:

https://github.com/jsx-eslint/eslint-plugin-react/blob/8c9ec8a3929403ac34a22c11cb9bd2575ba766bd/lib/rules/no-danger.js#L63

We think it could be useful to support non-native components as well, especially in cases like `styled-components` wrappers, or components that want to offer this prop to be forwarded internally to a native element.

This could be a new default behavior, or an opt-in behavior via a new rule option.

akulsr0:
Closes #3434

### DIFF
--- a/docs/rules/no-danger.md
+++ b/docs/rules/no-danger.md
@@ -24,6 +24,20 @@ var React = require('react');
 var Hello = <div>Hello World</div>;
 ```
 
+## Rule Options
+
+```js
+...
+"react/no-danger": [<enabled>, {
+  "customComponentNames": Array<string>,
+}]
+...
+```
+
+### customComponentNames
+
+Defaults to `[]`, if you want to enable this rule for all custom components you can pass `customComponentNames` as `['*']`, or else you can pass specific components name to the array.
+
 ## When Not To Use It
 
 If you are certain the content passed to dangerouslySetInnerHTML is sanitized HTML you can disable this rule.
--- a/lib/rules/no-danger.js
+++ b/lib/rules/no-danger.js
@@ -7,6 +7,7 @@
 
 const has = require('hasown');
 const fromEntries = require('object.fromentries/polyfill')();
+const minimatch = require('minimatch');
 
 const docsUrl = require('../util/docsUrl');
 const jsxUtil = require('../util/jsx');
@@ -55,13 +56,32 @@ module.exports = {
 
     messages,
 
-    schema: [],
+    schema: [{
+      type: 'object',
+      properties: {
+        customComponentNames: {
+          items: {
+            type: 'string',
+          },
+          minItems: 0,
+          type: 'array',
+          uniqueItems: true,
+        },
+      },
+    }],
   },
 
   create(context) {
+    const configuration = context.options[0] || {};
+    const customComponentNames = configuration.customComponentNames || [];
+
     return {
       JSXAttribute(node) {
-        if (jsxUtil.isDOMComponent(node.parent) && isDangerous(node.name.name)) {
+        const functionName = node.parent.name.name;
+
+        const enableCheckingCustomComponent = customComponentNames.some((name) => minimatch(functionName, name));
+
+        if ((enableCheckingCustomComponent || jsxUtil.isDOMComponent(node.parent)) && isDangerous(node.name.name)) {
           report(context, messages.dangerousProp, 'dangerousProp', {
             node,
             data: {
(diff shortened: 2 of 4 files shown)


docs/rules/no-danger.md
```diff
@@ -24,6 +24,25 @@ var React = require('react');
...
+...
+```
+
+### checkCustomComponents
+
+Defaults to `false`, when `enabled` allows the rule to check for custom components.
+
+### customComponentNames
+
+If you want to enable this rule for a specific set of custom components, you can pass `checkCustomComponents` as `true` and `customComponentNames` as the array of name of your custom components.
```
ljharb, line 44: i wonder if it'd be simpler to have just an array that defaults to `[]`, but can be set to `['*']` if you want to check everything?

lib/rules/no-danger.js
```diff
@@ -55,13 +55,33 @@ module.exports = {
...
+    const configuration = context.options[0] || {};
+    const customComponentNames = configuration.customComponentNames || [];
+
     return {
       JSXAttribute(node) {
-        if (jsxUtil.isDOMComponent(node.parent) && isDangerous(node.name.name)) {
+        const functionName = node.parent.name.name;
+
+        const enableCheckingCustomComponent = customComponentNames.length > 0
+          && (customComponentNames[0] === '*' || customComponentNames.some((name) => functionName === name));
```
ljharb, line 82: i didn't intend for `*` to be a special value here; i think we should use existing globbing logic to allow `Foo*` or `*Foo` or `*` etc.

lib/rules/no-danger.js
```diff
@@ -55,13 +56,33 @@ module.exports = {
...
 
-    schema: [],
+    schema: [{
+      type: 'object',
+      properties: {
+        customComponentNames: {
+          items: {
+            type: 'string',
+          },
+          minItems: 1,
```
ljharb, line 66:
```suggestion
          minItems: 0,
```
so that a config can reset the value to the default

lib/rules/no-danger.js
```diff
@@ -55,13 +56,33 @@ module.exports = {
...
+    const configuration = context.options[0] || {};
+    const customComponentNames = configuration.customComponentNames || [];
+
     return {
       JSXAttribute(node) {
-        if (jsxUtil.isDOMComponent(node.parent) && isDangerous(node.name.name)) {
+        const functionName = node.parent.name.name;
+
+        const enableCheckingCustomComponent = customComponentNames.length > 0
+          && customComponentNames.some((name) => minimatch(functionName, name));
```
ljharb, line 83:
```suggestion
        const enableCheckingCustomComponent = customComponentNames.some((name) => minimatch(functionName, name));
```
.some already handles empty lists :-)

tests/lib/rules/no-danger.js
```diff
@@ -32,6 +32,26 @@ ruleTester.run('no-danger', rule, {
...
+      `,
+      options: [{ customComponentNames: ['Home'] }],
+    },
+    {
+      code: `
+        function App() {
+          return <TextMUI dangerouslySetInnerHTML={{ __html: "<span>hello</span>" }} />;
+        }
+      `,
+      options: [{ customComponentNames: ['MUI*'] }],
```
ljharb, line 53:
```suggestion
      options: [{ customComponentNames: ['*MUI'] }],
```

ljharb: aha, ok thanks. can you add the same exact test case with the matching pattern, as well?

akulsr0: I have added that test case [here](https://github.com/jsx-eslint/eslint-plugin-react/pull/3748/files#diff-4b8ada5e2441fe2af247dda8f99e5c6d99df1b6cbe622f21c8dc6d7cef18ceadR90-R103)

ljharb: the test is good, but i'm suggesting having *identical* cases where the only difference is the glob pattern

akulsr0: Added same testcase with opposite pattern in invalid cases.

ljharb (approved):

akulsr0:
Let me know if anything is pending here else can we proceed towards merging it?

ljharb:
it's good to land; i want to cut a patch release before landing new semver-minor things, is all :-)

ljharb (approved):
LGTM, but i'm going to hold off landing new semver-minors for a bit.

akulsr0:
Bumping this up

PR #3749, opened 2024-05-06
[Fix] prop-types: `className` missing in prop validation

### LINKED ISSUE #3284: Typescript "'className' is missing in props validation" despite being defined
# Description

I am observing `'className' is missing in props validation` despite it being correctly specified in the type definition.

- `React.DetailedHTMLProps<React.InputHTMLAttributes<HTMLInputElement>, HTMLInputElement>`
- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/90f13b6b5e4ee2db30044f960327e17ed3ef3338/types/react/index.d.ts#L1854
    - `React.HTMLAttributes<HTMLInputElement>.className?: string | undefined`

## Versions

```
    "@types/react": "^18.0.0",
    "eslint": "^8.8.0",
    "eslint-plugin-react": "^7.28.0",
    "react": "^18.0.0",
    "typescript": "^4.5.5"
```

# Example
```typescript
import React from 'react'

const classNames = (...classes: (false | null | undefined | string)[]) =>
  classes.filter(Boolean).join(' ')


export const Input = (
  props: React.DetailedHTMLProps<
    React.InputHTMLAttributes<HTMLInputElement>,
    HTMLInputElement
  >
) => (
  <input
    {...props}
    className={classNames(
      'class-a',
      'class-b',
      // eslint-disable-next-line react/prop-types
      props.className
    )}
  />
)
```

## Expected behaviour

The `eslint-disable-next-line` should not be necessary.
I can also workaround this by adding in `& { className?: string }` into the type definition of props.
```typescript
  props: React.DetailedHTMLProps<
    React.InputHTMLAttributes<HTMLInputElement>,
    HTMLInputElement
  > & { className?: string }

```

akulsr0:
Closes #3284

### DIFF
--- a/lib/util/propTypes.js
+++ b/lib/util/propTypes.js
@@ -645,6 +645,17 @@ module.exports = function propTypesInstructions(context, components, utils) {
           this.shouldSpecifyOptionalChildrenProps = true;
 
           const rightMostName = getRightMostTypeName(node.typeName);
+          if (
+            leftMostName === 'React'
+            && (
+              rightMostName === 'HTMLAttributes'
+              || rightMostName === 'HTMLElement'
+              || rightMostName === 'HTMLProps'
+            )
+          ) {
+            this.shouldSpecifyClassNameProp = true;
+          }
+
           const importedName = localToImportedMap[rightMostName];
           const idx = genericTypeParamIndexWherePropsArePresent[
             leftMostName !== rightMostName ? rightMostName : importedName
@@ -835,6 +846,14 @@ module.exports = function propTypesInstructions(context, components, utils) {
           isRequired: false,
         };
       }
+      if (this.shouldSpecifyClassNameProp) {
+        this.declaredPropTypes.className = {
+          fullName: 'className',
+          name: 'className',
+          isRequired: false,
+        };
+      }
+
       this.foundDeclaredPropertiesList.forEach((tsInterfaceBody) => {
         if (tsInterfaceBody && (tsInterfaceBody.type === 'TSPropertySignature' || tsInterfaceBody.type === 'TSMethodSignature')) {
           let accessor = 'name';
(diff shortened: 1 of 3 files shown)


lib/util/propTypes.js
```diff
@@ -634,6 +634,8 @@ module.exports = function propTypesInstructions(context, components, utils) {
           // So we should construct an optional children prop
           this.shouldSpecifyOptionalChildrenProps = true;
 
+          this.shouldSpecifyClassNameProp = true;
```
ljharb, line 637: why do we need a class property for this?

akulsr0: Since react generic type would have `className` too

ljharb: what i mean is, when is this not always true?

akulsr0: So whenever the type is coming from react import reference, then we can have `className` too (similar to children).
When its not true is: type annotation, type literal, etc...

ljharb: `className` is just a prop, though, like any other. I'm a bit confused why it merits special treatment.

akulsr0: `className` is a react specific prop for HTML elements. `class` is a vanilla HTML attribute.

ljharb: Sure, that's all true - but HTML elements don't use propTypes, so we're only talking about custom components here.

ljharb: @akulsr0 ok so, for custom components, `className` isn't special and is just like any other prop. For HTML elements, it is special and doesn't need a propType.

What are the next steps?

akulsr0: I guess, should wrap this line inside an if condition for the later case.

akulsr0: Added the condition, bumping up!

ljharb (approved):

PR #3750, opened 2024-05-08
[New] add `no-render-return-undefined`: disallow components rendering undefined

### LINKED ISSUE #3020: React 18: Warn when components render `undefined`
Upstream: https://github.com/facebook/react/pull/21869

In React 18 (currently in alpha), components may render `undefined`, and React will render nothing to the DOM instead of throwing an error. This allows components that don't render anything useful (such as optional rendering or components that rely on side effects from hooks or lifecycles) to return `undefined` instead of explicitly returning `null`. However, accidentally rendering nothing in a component could still cause surprises (which is why `require-return-return` exists, though it's only for class components). Additionally, a team may want to have a code style guideline of explicitly returning `undefined` to prevent components from accidentally doing nothing. The React team seems to be recommending that users use linters to catch this issue in React 18 (since React itself will no longer warn), though it's up to debate whether this is something we'd want to enforce by default, as not rendering anything is valid as of React 18.

### Add functionality to `require-render-return`
We could extend this rule to include function components that return `undefined` and class components that explicitly return `undefined`. Documentation should be updated to reflect that this rule affects more components and may not be necessary in React 18. This would technically be a breaking change for React 18 users, as it's potentially valid for them to return `undefined`, and the rule is recommended.
### Add new `no-render-return` rule
Rather than put all functionality in an existing rule that could affect existing configs, we could create a separate rule such as `no-render-return` that covers all these use cases. We may want to deprecated and stop recommending `require-render-return` in this case, as it would have a subset of this rule's functionality.

akulsr0:
Closes #3020


akulsr0:
@ljharb Let me know your thoughts on this!

docs/rules/no-render-return-undefined.md
```diff
@@ -0,0 +1,62 @@
+# Disallow returning undefined from react components (`react/no-render-return-undefined`)
+
+💼 This rule is enabled in the ☑️ `recommended` [config](https://github.com/jsx-eslint/eslint-plugin-react/#shareable-configs).
```
ljharb, line 3: i think the docs need to be updated, run `npm run update:eslint-docs`

docs/rules/no-render-return-undefined.md
```diff
@@ -0,0 +1,62 @@
...
+
+<!-- end auto-generated rule header -->
+
+> In React 18, components may render undefined, and React will render nothing to the DOM instead of throwing an error. However, accidentally rendering nothing in a component could still cause surprises. This rule will warn if the `return` statement in a React Component returns undefined.
```
ljharb, line 7:
```suggestion
> Starting in React 18, components may render `undefined`, and React will render nothing to the DOM instead of throwing an error. However, accidentally rendering nothing in a component could still cause surprises. This rule will warn if the `return` statement in a React Component returns `undefined`.
```

docs/rules/no-render-return-undefined.md
```diff
@@ -0,0 +1,62 @@
...
+
+Issue: [#3020](https://github.com/jsx-eslint/eslint-plugin-react/issues/3020)
```
ljharb, line 9: i'm not sure linking to the issue is needed

docs/rules/no-render-return-undefined.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Rule Details
+
+This rule will warn if the `return` statement in a React component returns undefined.
```
ljharb, line 13:
```suggestion
This rule will warn if the `return` statement in a React component returns `undefined`.
```

docs/rules/no-render-return-undefined.md
```diff
@@ -0,0 +1,62 @@
...
+
+Examples of **incorrect** code for this rule:
```
ljharb, line 15: let's add one that uses `void`

docs/rules/no-render-return-undefined.md
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// OR
+
+function getUI() {
+  if (condition) return <h1>Hello</h1>;
+}
+function App() {
+  return getUI();
+}
```
ljharb, line 36: this won't be possible across files, only if it's in the same file, so unless the function provides return type information (ie, TS) i don't think this is a check we can reliably do.

lib/rules/no-render-return-undefined.js
```diff
@@ -0,0 +1,154 @@
...
+'use strict';
+
+const astUtil = require('../util/ast');
+const docsUrl = require('../util/docsUrl');
+const isFirstLetterCapitalized = require('../util/isFirstLetterCapitalized');
+const report = require('../util/report');
+const variableUtil = require('../util/variable');
+
+const messages = {
+  returnsUndefined: "Don't return undefined from react components",
```
ljharb, line 15:
```suggestion
  returnsUndefined: "Don't return `undefined` from react components",
```

lib/rules/no-render-return-undefined.js
```diff
@@ -0,0 +1,154 @@
...
+};
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Disallow returning undefined from react components',
```
ljharb, line 22:
```suggestion
      description: 'Disallow returning `undefined` from react components',
```

lib/rules/no-render-return-undefined.js
```diff
@@ -0,0 +1,154 @@
...
+      category: 'Best Practices',
+      recommended: false,
+      url: docsUrl('no-render-return-undefined'),
+    },
+    messages,
+    schema: [],
+  },
+
+  create(context) {
+    function getReturnValue(returnNode) {
```
ljharb, line 32: can this take `context` and be defined at module level instead?

lib/rules/no-render-return-undefined.js
```diff
@@ -0,0 +1,154 @@
...
+        return value;
+      }
+
+      switch (returnNode.type) {
+        case 'LogicalExpression': {
+          return getReturnValue(returnNode.right);
+        }
+        case 'ConditionalExpression': {
+          const possibleReturnValue = [getReturnValue(returnNode.consequent), getReturnValue(returnNode.alternate)];
+          const returnsUndefined = possibleReturnValue.some((val) => val === undefined);
```
ljharb, line 64:
```suggestion
          const returnsUndefined = possibleReturnValue.some((val) => typeof val === 'undefined');
```

lib/rules/no-render-return-undefined.js
```diff
@@ -0,0 +1,154 @@
...
+          if (returnsUndefined) return undefined;
```
ljharb, line 65:
```suggestion
          if (returnsUndefined) return;
```

lib/rules/no-render-return-undefined.js
```diff
@@ -0,0 +1,154 @@
...
+      const returnIdentifierName = returnNode && returnNode.name;
+
+      const returnIdentifierValue = getReturnValue(returnNode);
+
+      const returnsArrayHavingUndefined = Array.isArray(returnIdentifierValue)
+          && returnIdentifierValue.some((el) => el && el.type === 'Identifier' && el.name === 'undefined');
+
+      return !returnStatement
+          || returnIdentifierName === 'undefined'
+          || returnIdentifierValue === undefined
```
ljharb, line 114:
```suggestion
          || typeof returnIdentifierValue === 'undefined'
```

package.json
```diff
@@ -15,7 +15,7 @@
     "test": "npm run unit-test",
     "posttest": "aud --production",
     "type-check": "tsc",
-    "unit-test": "istanbul cover node_modules/mocha/bin/_mocha tests/lib/**/*.js tests/util/**/*.js tests/index.js",
+    "unit-test": "istanbul cover node_modules/mocha/bin/_mocha tests/lib/rules/no-render-return-undefined.js",
```
ljharb, line 18: commenting here to ensure this doesn't get merged. fwiw you can `npx mocha tests/lib/rules/no-render-return-undefined.js` to run just that one file locally :-)

ljharb (changes requested):

akulsr0:
Bumping this up!

lib/rules/index.js
```diff
@@ -105,4 +105,5 @@ module.exports = {
   'static-property-placement': require('./static-property-placement'),
   'style-prop-object': require('./style-prop-object'),
   'void-dom-elements-no-children': require('./void-dom-elements-no-children'),
+  'no-render-return-undefined': require('./no-render-return-undefined'),
```
ljharb, line 108: let's sort this alphabetically

ljharb (changes requested):
There's still some unaddressed comments on the docs.

docs/rules/no-render-return-undefined.md
akulsr0: Is it possible to continue without supporting this?

lib/rules/index.js
akulsr0: Done

docs/rules/no-render-return-undefined.md
akulsr0: @ljharb Can we do something here? 
Rest all changes, I've made

ljharb: Yes, I don't think we necessarily have to support this - I'm saying that the docs should be updated to reflect the caveats.

ljharb: Yes, I don't think we necessarily have to support this - I'm saying that the docs should be updated to reflect the caveats.

ljharb (changes requested):

PR #3756, opened 2024-05-11
:sparkles: [enhancement] new install commands added in the README.md

AngelDarco:
I just added new installation commands, pnpm, yarn and bun, to the main README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,19 @@ React specific linting rules for `eslint`
 ## Installation
 
 ```sh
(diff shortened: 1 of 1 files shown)


ljharb:
Duplicate of #3755.

ljharb (changes requested):
I don’t see the value here. npm is the standard, and the default, and if you want to use an alternative you just have to know how to translate npm commands.

PR #3762, opened 2024-05-30
[Fix] Null check `rootNode` before calling `getScope` with it

crnhrv:
I'm getting this error while linting some valid React/TS code after upgrading to the latest version (`7.34.2`):

<details>
<summary>Stack trace</summary>

```

  ESLint: 8.57.0
  
  TypeError: Missing required argument: node.
  Occurred while linting ./OptionsTable.tsx:599
  Rule: "react/display-name"
      at SourceCode.getScope (./node_modules/eslint/lib/source-code/source-code.js:773:19)
      at getScope (./node_modules/eslint-plugin-react/lib/util/eslint.js:19:23)
      at resolveValueForIdentifierNode (./node_modules/eslint-plugin-react/lib/util/propTypes.js:378:21)
      at buildReactDeclarationTypes (./node_modules/eslint-plugin-react/lib/util/propTypes.js:420:5)
      at ./node_modules/eslint-plugin-react/lib/util/propTypes.js:788:39
      at iterateProperties (./node_modules/eslint-plugin-react/lib/util/propTypes.js:71:7)
      at ./node_modules/eslint-plugin-react/lib/util/propTypes.js:773:23
      at DeclarePropTypesForTSTypeAnnotation.convertReturnTypeToPropTypes (./node_modules/eslint-plugin-react/lib/util/propTypes.js:762:34)
      at DeclarePropTypesForTSTypeAnnotation.searchDeclarationByName (./node_modules/eslint-plugin-react/lib/util/propTypes.js:667:14)
```

</details>

I don't really understand enough about the codebase to say whether this is the correct fix, or if this really should be passing in `node` instead of `rootNode` to `getScope` or something like that. This resolves the immediate issue for me at least and seemed reasonable to check the value is valid before it tries to call something that requires it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ## Unreleased
 
+### Fixed
+* [`prop-types`]: null-check rootNode before calling getScope ([#3762][] @crnhrv)
+
+[#3762]: https://github.com/jsx-eslint/eslint-plugin-react/pull/3762
+
 ## [7.34.2] - 2024.05.24
 
 ### Fixed
--- a/lib/util/propTypes.js
+++ b/lib/util/propTypes.js
@@ -372,7 +372,8 @@ module.exports = function propTypesInstructions(context, components, utils) {
    */
   function resolveValueForIdentifierNode(node, rootNode, callback) {
     if (
-      node
+      rootNode
+      && node
       && node.type === 'Identifier'
     ) {
       const scope = getScope(context, rootNode);
--- a/tests/lib/rules/prop-types.js
+++ b/tests/lib/rules/prop-types.js
@@ -3363,6 +3363,20 @@ ruleTester.run('prop-types', rule, {
     },
     {
       code: `
+        import React from "react";
+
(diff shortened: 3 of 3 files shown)


ljharb:
Thanks - we'd need a regression test in order to land this tho.

crnhrv:
@ljharb I've attempted to add some regression tests by trying to find a minimal reproduction. I think the issue relates to using `ReturnType<typeof object>` for the type parameter for a component, but I don't really know the details.

I get the expected failure without the nullcheck:

ishare-1717158109

crnhrv:
I see that code coverage is failing on some indirect changes related to code beneath a `node.type === 'TSParenthesizedType'` conditional. Not sure why that would happen, but is this code actually necessary since `TSParenthesizedType` [seems to no longer be generated as a node type](https://github.com/typescript-eslint/typescript-eslint/pull/3340)?

ljharb:
@crnhrv that PR seems to be in v5 of the parser, but we still support down to v2

ljharb (approved):
wonderful, thanks! don't worry about the coverage.

crnhrv:
Cool, thanks for the help. Will leave it with you then

PR #3768, opened 2024-06-13
[Fix] `no-object-type-as-default-prop`: enable rule when multiple parameters

### LINKED ISSUE #3767: [Bug]: `react/no-object-type-as-default-prop` is disabled when using a `ref` prop

### Brief description

The `react/no-object-type-as-default-prop` does not run when a component has a `ref` prop as a second argument.

### Show example of your code (as text format), add images/videos/gifs to help explain example

Both of these components should error because an object type is used as a default prop.
However, the second component (with a ref) does not error.

```jsx
const ErrorIsShown = ({ foo = {} }) => null;
const ErrorIsAbsent = ({ foo = {} }, ref) => null;
```

image

Link to repro: https://github.com/JulienR1/eslint-react-no-object-type-as-default-prop-repro

```bash
npm run lint
# OR
eslint . --ext js,jsx --report-unused-disable-directives --max-warnings 0
```

```js
// .eslintrc.cjs
module.exports = {
  root: true,
  env: { browser: true, es2020: true },
  extends: ["plugin:react/recommended"],
  ignorePatterns: ["dist", ".eslintrc.cjs"],
  parserOptions: { ecmaVersion: "latest", sourceType: "module" },
  rules: {
    "react/no-object-type-as-default-prop": "error",
    "react/prop-types": "off",
  },
};
```

### Brief description

Components with `ref` props should also generate errors when using default object type props.

image

v7.34.2

### eslint version

v8.57.0

### node version

v20.12.2

JulienR1:
Fixes #3767 by allowing components with more than one parameter to be checked by `no-object-type-as-default-prop`.


ljharb (approved):
Seems pretty straightforward, nice!

RafaelJohn9 (commented):
👍nice

PR #3772, opened 2024-06-21
[New] `jsx-handler-names`: support ignoring component names

### LINKED ISSUE #3722: `jsx-handler-names` support skipping prop naming checks
My project has many components from other dependency packages, like:
```tsx
<ComponentFromOtherLibraryFoo customPropNameFoo={handleSomething} />;
<ComponentFromOtherLibraryBar customPropNameBar={handleSomething} />;
```
Every time I come across them, I need to add an extra line of boring comments:
```tsx
// eslint-disable-next-line jsx-handler-names
<ComponentFromOtherLibraryFoo customPropNameFoo={handleSomething} />;
// eslint-disable-next-line jsx-handler-names
<ComponentFromOtherLibraryBar customPropNameBar={handleSomething} />;
```

To be honest, I don't care about the names of these PROPS(even the components in my own project, I just want it to check my handler function's name)

To make matters worse, if the prop's name is not camelCase (e.g., all lowercase letters: "foofoo"), the option "eventHandlerPropPrefix" is not valid either, because it only applies to CamelCase naming:
```tsx
/* with config: 
{
  eventHandlerPrefix: '(on|handle)',
  eventHandlerPropPrefix: '(on|foo)',
}
*/
// fine
<ComponentFromOtherLibraryFoo fooPropName={handleSomething} />;

// not work, still has error
<ComponentFromOtherLibraryFoo foo={handleSomething} />;
```
So, it would be nice if there were an option like "checkPropName". For those who don't care about prop naming, simply turn it off.
like:
```tsx
{
  eventHandlerPrefix: '(on|handle)',
  eventHandlerPropPrefix: '(on|foo)',
  // ...
  checkPropName: false,
}
```

Anyway, I want to confirm if this option is necessary (and if anyone else has experienced this), and if so, I can create a PR to do it.

akulsr0:
Closes #3722

### DIFF
--- a/docs/rules/jsx-handler-names.md
+++ b/docs/rules/jsx-handler-names.md
@@ -34,7 +34,8 @@ Examples of **correct** code for this rule:
   "eventHandlerPrefix": <eventHandlerPrefix>,
   "eventHandlerPropPrefix": <eventHandlerPropPrefix>,
   "checkLocalVariables": <boolean>,
-  "checkInlineFunction": <boolean>
+  "checkInlineFunction": <boolean>,
+  "ignoreComponentNames": Array<string>
 }]
 ...
 ```
@@ -43,6 +44,7 @@ Examples of **correct** code for this rule:
 - `eventHandlerPropPrefix`: Prefix for props that are used as event handlers. Defaults to `on`
 - `checkLocalVariables`: Determines whether event handlers stored as local variables are checked. Defaults to `false`
 - `checkInlineFunction`: Determines whether event handlers set as inline functions are checked. Defaults to `false`
+- `ignoreComponentNames`: Array of glob strings, when matched with component name, ignores the rule on that component. Defaults to `[]`
 
 ## When Not To Use It
 
--- a/tests/lib/rules/jsx-handler-names.js
+++ b/tests/lib/rules/jsx-handler-names.js
@@ -181,6 +181,24 @@ ruleTester.run('jsx-handler-names', rule, {
       code: '<TestComponent someProp={props.onChange} />',
       options: [{ eventHandlerPropPrefix: false }],
     },
+    {
+      code: '<ComponentFromOtherLibraryBar customPropNameBar={handleSomething} />;',
+      options: [{ checkLocalVariables: true, ignoreComponentNames: ['ComponentFromOtherLibraryBar'] }],
+    },
+    {
+      code: `
+        function App() {
+          return (
+            <div>
+              <MyLibInput customPropNameBar={handleSomething} />;
+              <MyLibCheckbox customPropNameBar={handleSomething} />;
+              <MyLibButtom customPropNameBar={handleSomething} />;
+            </div>
+          )
+        }
+      `,
+      options: [{ checkLocalVariables: true, ignoreComponentNames: ['MyLib*'] }],
+    },
   ]),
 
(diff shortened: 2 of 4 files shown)


akulsr0:
Bump up

docs/rules/jsx-handler-names.md
```diff
@@ -43,6 +44,7 @@ Examples of **correct** code for this rule:
 - `eventHandlerPropPrefix`: Prefix for props that are used as event handlers. Defaults to `on`
 - `checkLocalVariables`: Determines whether event handlers stored as local variables are checked. Defaults to `false`
 - `checkInlineFunction`: Determines whether event handlers set as inline functions are checked. Defaults to `false`
+- `ignoreComponentNames`: Array of regex, when matched with component name, ignores the rule on that component. Defaults to `[]`
```
ljharb, line 47: it's a bad idea for eslint plugins to support regex strings from config; we can support glob strings, though.

ljharb (changes requested):

docs/rules/jsx-handler-names.md
akulsr0: Curious to know the rationale behind it.
Unrelated to this, I recently contributed to [eslint/no-restricted-exports](https://github.com/eslint/eslint/pull/18431), where regex was preferred over glob string.

ljharb: regexes are a CVE magnet, every time they exist. eslint shouldn't have accepted a pattern string either, imo.

akulsr0: Updated to use glob strings

tests/lib/rules/jsx-handler-names.js
```diff
@@ -181,6 +181,24 @@ ruleTester.run('jsx-handler-names', rule, {
...
+          return (
+            <div>
+              <MyLibInput customPropNameBar={handleSomething} />;
+              <MyLibCheckbox customPropNameBar={handleSomething} />;
+              <MyLibButtom customPropNameBar={handleSomething} />;
+            </div>
+          )
+        }
+      `,
+      options: [{ checkLocalVariables: true, ignoreComponentNames: ['MyLib**'] }],
```
ljharb, line 200:
```suggestion
      options: [{ checkLocalVariables: true, ignoreComponentNames: ['MyLib*'] }],
```
one star should be enough

tests/lib/rules/jsx-handler-names.js
```diff
@@ -369,5 +387,33 @@ ruleTester.run('jsx-handler-names', rule, {
...
+          return (
+            <div>
+              <MyLibInput customPropNameBar={handleInput} />;
+              <MyLibCheckbox customPropNameBar={handleCheckbox} />;
+              <MyLibButtom customPropNameBar={handleButton} />;
+            </div>
+          )
+        }
+      `,
+      options: [{ checkLocalVariables: true, ignoreComponentNames: ['MyLibrary**'] }],
```
ljharb, line 402:
```suggestion
      options: [{ checkLocalVariables: true, ignoreComponentNames: ['MyLibrary*'] }],
```

ljharb (approved):
Looks great, thanks. I'm going to hold off landing semver-minors for a little bit, but the next one will include this PR.

PR #3774, opened 2024-06-27
[New]`forbid-component-props`: add `propNamePattern` option

### LINKED ISSUE #3681: Ban props containing an "-" for components
I'm looking for a way to forbid passing props that contains a dash `-` for my components. I only want this to affect custom components, meaning that regular HTML components should be allowed to take `<div aria-xxx="asdf">` but not `<MyComp aria-label="asdf">`. I want to do this because in TypeScript, props beginning with a kebab case like `aria-xxx` are not type checked. https://github.com/microsoft/TypeScript/issues/32447

We plan on instead creating custom aria props like `ariaLabel` for components that can take them.

The closest thing I could find was https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/forbid-component-props.md but it doesn't allow me to specify a regex that could pick up every kebab-case named prop which is what I'm looking for.

I could contribute with an addition to this rule if deemed okay.

akulsr0:
Add `propNamePattern` option to allow / disallow prop name patterns

Closes #3681

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 * support eslint v9 ([#3759][] @mdjermanovic)
 * export flat configs from plugin root and fix flat config crash ([#3694][] @bradzacher @mdjermanovic)
 * add [`jsx-props-no-spread-multi`] ([#3724][] @SimonSchick)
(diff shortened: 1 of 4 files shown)


akulsr0:
Can we rerun test for failing checks? Not sure of cause of failure

ljharb:
it's a real failure; i think it's that destructuring syntax doesn't work in function signatures in node 4.

ljharb (approved):

Issue #3763, opened 2024-06-03
[Bug]: react/boolean-prop-naming not working with TS & declaration


mario-iliev:
### Is there an existing issue for this?

- [X] I have searched the existing issues and my issue is unique
- [X] My issue appears in the command-line and not only in the text editor

### Description Overview

There is no indication in IDE that the prop is not correct.
Terminal is throwing:
```
ESLint: 8.57.0

TypeError: Cannot read properties of undefined (reading 'properties')
```

The affected component code:
`const Component = ({ color, withBackground }: ComponentProps & { withBackground?: boolean }) => (...);` 

I managed to fix the problem by moving the props definition in a separate type:
```
type ExtendedComponentProps = ComponentProps & { withBackground?: boolean };

const Component = ({ color, withBackground }: ExtendedComponentProps) => (...);
```

By doing this, the IDE is properly suggesting that "withBackground" is not covering the rules:

```    
'react/boolean-prop-naming': [
      'warn',
      {
        rule: '^(is|has|are|can|should|show|hide)[A-Z]([A-Za-z0-9]?)+',
      },
]
```

Packages versions:
```
 "eslint": "8.57.0",
 "eslint-plugin-react": "7.34.2",
 "@typescript-eslint/eslint-plugin": "7.11.0",
 "@typescript-eslint/parser": "7.11.0",
 "typescript": "5.4.5"
```

### Expected Behavior

Properly validate the props.

### eslint-plugin-react version

v7.34.2

### eslint version

v8.57.0

### node version

v18.18.0

ljharb:
I believe this is fixed but not yet released. See #3733.

ljharb:
Released in [v7.34.3](https://github.com/jsx-eslint/eslint-plugin-react/releases/tag/v7.34.3). Happy to reopen if it's not fixed.

tylerlaprade:
I'm getting a similar error with `sort-prop-types`.
#### Config:
```json
"react/sort-prop-types": [
      "error",
      {
        "callbacksLast": true,
        "requiredFirst": true,
        "sortShapeProp": true,
        "noSortAlphabetically": true,
        "checkTypes": true
      }
    ],
```

#### Error:
```bash
$ eslint "{src,tests}/**/*.{js,ts,tsx}" --fix

Oops! Something went wrong! :(

ESLint: 8.57.0

TypeError: Cannot read properties of undefined (reading 'typeAnnotation')
[...]

PR #3779, opened 2024-07-08
[Fix] `sort-prop-types`: Check for undefined before accessing `node.typeAnnotation.typeAnnotation`

tylerlaprade:
Fixes the bug reported in [this comment](https://github.com/jsx-eslint/eslint-plugin-react/issues/3763#issuecomment-2215163106)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 ### Fixed
 
 * [`prop-types`]: fix `className` missing in prop validation false negative ([#3749] @akulsr0)
+* [`sort-prop-types`]: Check for undefined before accessing `node.typeAnnotation.typeAnnotation` ([#3779] @tylerlaprade)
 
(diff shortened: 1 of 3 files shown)


ljharb (changes requested):
We'll need a regression test before this fix can land.

ljharb:
@tylerlaprade can you share the code that errored in your linked comment?

tylerlaprade:
@ljharb, in a file with a `.tsx` extension, the following repros:

```typescriptreact
function Foo() {
  return <></>;
}
```

ljharb:
@tylerlaprade hmm, can you confirm what version of the TS eslint parser you're using? it doesn't reproduce for me.

ljharb (approved):

tylerlaprade:
Thanks for adding the test case! It looks like some snapshots need to be updated, but when I `npm test` locally, I don't see any failures.

ljharb:
we don't use snapshots :-) it's just that some of the test jobs don't support fragments. i switched it to a div instead.

PR #3784, opened 2024-07-18
[Fix] `sort-prop-types`: single line type ending without semicolon

### LINKED ISSUE #3783: [Bug]: `sort-prop-types`'s autofix doesn't move the semicolon, producing broken code, on single-line Props types

### Description Overview

`sort-prop-types`, with `checkTypes: true`, breaks the code when run on single-line Props with autofix due to the semicolon staying at the end instead of moving.

image

The props were rearranged without the semicolon dividing them

`eslint --fix`

### Expected Behavior

Semicolon be moved along with the moved prop when running autofix

```tsx
type CustomProps = { onChange: (event: { target: { name: string; value: string } }) => void; name: string };
```

should become

```tsx
type CustomProps = { name: string; onChange: (event: { target: { name: string; value: string } }) => void };
```

Instead, it is fixing to

```tsx
type CustomProps = { name: string onChange: (event: { target: { name: string; value: string } }) => void; };
```

v7.34.4

### eslint version

v8.57.0

### node version

v20.14.10

akulsr0:
Partially fixes #3783

### DIFF
--- a/lib/util/propTypesSort.js
+++ b/lib/util/propTypesSort.js
@@ -126,6 +126,7 @@ const commentnodeMap = new WeakMap(); // all nodes reference WeakMap for start a
  * @param {Boolean=} callbacksLast whether or not to sort callbacks after everything else.
  * @param {Boolean=} noSortAlphabetically whether or not to disable alphabetical sorting of the elements.
  * @param {Boolean=} sortShapeProp whether or not to sort propTypes defined in PropTypes.shape.
+ * @param {Boolean=} checkTypes whether or not sorting of prop type definitions are checked.
  * @returns {Object|*|{range, text}} the sort order of the two elements.
  */
 function fixPropTypesSort(
@@ -136,7 +137,8 @@ function fixPropTypesSort(
   requiredFirst,
   callbacksLast,
   noSortAlphabetically,
-  sortShapeProp
+  sortShapeProp,
+  checkTypes
 ) {
   function sortInSource(allNodes, source) {
     const originalSource = source;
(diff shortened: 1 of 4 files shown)


lib/util/propTypesSort.js
```diff
@@ -197,7 +199,8 @@ function fixPropTypesSort(
             sortedAttrText = attrSource.slice(sortedAttr.range[0], sortedAttr.range[1]);
           }
         }
-        return `${acc.slice(0, commentnodeMap.get(attr).start)}${sortedAttrText}${acc.slice(commentnodeMap.get(attr).end)}`;
+        const sortedAttrTextVal = !checkTypes || sortedAttrText.endsWith(';') ? sortedAttrText : `${sortedAttrText};`;
```
ljharb, line 202:
```suggestion
        const sortedAttrTextVal = checkTypes ? sortedAttrText.replace(/;?^/, ';') : sortedAttrText;
```
seems a bit simpler

tests/lib/rules/sort-prop-types.js
```diff
@@ -2325,16 +2325,16 @@ ruleTester.run('sort-prop-types', rule, {
     {
       code: `
         const Foo = (props: {
-          zzz: string,
-          aaa: string,
+          zzz: string;
+          aaa: string;
```
ljharb, line 2329: let's not change existing test cases - we need to be able to handle any mix of commas and semicolons, since TS types can have both

akulsr0:
@ljharb Failing test looks like a fluke, can we rerun failing test once?

ljharb:
The failures are due to https://github.com/nodejs/node/issues/53902. Once that's resolved, I'll rerun them.

ljharb (approved):

Issue #3554, opened 2023-03-19
Rule to disallow blank lines between tags


sspenst:
I'm looking for a rule to disallow blank spaces between tags. For example, this should be disallowed:

```
<div>
  <span>1</span>

  <span>2</span>
</div>
```

and this should be allowed:

```
<div>
  <span>1</span>
  <span>2</span>
</div>
```

I found [vue/padding-line-between-tags](https://eslint.vuejs.org/rules/padding-line-between-tags.html) which does what I'm looking for, but it would be nice if it existed here as well.

etherealm13:
I have raised a PR for adding this rule: https://github.com/jsx-eslint/eslint-plugin-react/pull/3789.

ljharb:
@etherealm13 i'll take a look, but the "question" label and the lack of a "help wanted" label means it's not yet decided to accept the rule.

PR #3789, opened 2024-07-22
new rule: padding-lines-between-tags #3554

etherealm13:
#3554 Rule to disallow blank lines between tags 
Based on https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/padding-line-between-tags.js

### DIFF
--- a/lib/rules/padding-lines-between-tags.js
+++ b/lib/rules/padding-lines-between-tags.js
@@ -8,6 +8,8 @@
 'use strict';
 
 const docsUrl = require('../util/docsUrl');
+const report = require('../util/report');
+const eslintUtil = require('../util/eslint');
 
 /**
  * Split the source code into multiple lines based on the line delimiters.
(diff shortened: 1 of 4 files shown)


etherealm13:
@ljharb Let me know your thoughts.

lib/rules/padding-lines-between-tags.js
```diff
@@ -0,0 +1,189 @@
...
+        },
+      });
+    }
+  }
+}
+
+function checkNewLine(context, configureList) {
+  const firstConsistentBlankLines = new Map();
+
+  const reverseConfigureList = [...configureList].reverse();
```
ljharb, line 93:
```suggestion
  const reverseConfigureList = [].concat(configureList).reverse();
```
the iterator protocol is slow and best avoided

ljharb (changes requested):
A few changes, but in particular, it would need thorough tests.

Issue #3686, opened 2024-02-08
Rule enhancement: allow regex in forbid-component-props allowedFor / disallowedFor


bschlenk:
### Is there an existing issue for this?

- [X] I have searched the existing issues and my issue is unique
- [X] My issue appears in the command-line and not only in the text editor

### Description Overview

My team uses `react/forbid-component-props` to disallow className and style props, but we maintain a list of legacy components that took these props before we introduced this rule. We now also have a bunch of icon components in the codebase that all have names like `Icon24Something`. We like to pass style or className props to these to override css properties in some cases, so it'd be nice to be able to allow any component that matches `/^Icon\d+/`.

Does this sound like a change you'd be happy to accept? I'd be okay sending a PR. Otherwise I'm also happy to fork the rule internally for our use case, but figured I'd check here to contribute back to the community.

### Expected Behavior

I'd expect to be able to pass a regex to `allowedFor`.

### eslint-plugin-react version

v7.32.2

### eslint version

v8.36.0

### node version

v18.17.1

bschlenk:
What I'd add is (in addition to `allowedFor` and `disallowedFor`) `allowedForRegex` and `disallowedForRegex` options that take an optional string. Then you can check in both the arrays and test the regex against the string. This would be a backwards compatible change.

ljharb:
Allowing regex strings in eslint config is an extremely bad idea (it's a CVE farm), but allowing globs would be tenable.

bschlenk:
What do you mean by CVE farm?

Looks like this package already depends on minimatch, I can rework my approach.

bschlenk:
I think I get it. It isn't a real issue (the user of this package owns their config so why would they ddos themself with a bad regex) but you'd rather not deal with automated security reports against your package. If that's the case, seems unfortunate but understandable.

bschlenk:
Alright, unless you think globs would be a useful addition here, I'm going to close this one out. Globs aren't really meant for generic string matching IMO, more for paths. I'm fine with modifying this rule in our repo or figuring out a different api for our icons that doesn't involve className or style.

ljharb:
If globs can help you solve your use case, I think it's useful, but I agree they're more for paths than strings.

ljharb:
You can do that with globs - `*end or `start*`

figmatom:
[...]

PR #3805, opened 2024-08-22
[New] `forbid-component-props`: add `allowedForPatterns` and `disallowedForPatterns` options

Efimenko:
Initially, issue was described [here](https://github.com/jsx-eslint/eslint-plugin-react/issues/3686), but it was closed by [another PR](https://github.com/jsx-eslint/eslint-plugin-react/pull/3774) (although this PR implements different case).
So, this PR just reimplements `Regex` implementation with `globs` as was suggested in [one](https://github.com/jsx-eslint/eslint-plugin-react/issues/3686#issuecomment-1951461748) of the comments.

### DIFF
--- a/lib/rules/forbid-component-props.js
+++ b/lib/rules/forbid-component-props.js
@@ -52,6 +52,11 @@ module.exports = {
                     uniqueItems: true,
                     items: { type: 'string' },
                   },
+                  allowedForPatterns: {
+                    type: 'array',
+                    uniqueItems: true,
+                    items: { type: 'string' },
+                  },
                   message: { type: 'string' },
                 },
                 additionalProperties: false,
@@ -66,17 +71,30 @@ module.exports = {
                     minItems: 1,
                     items: { type: 'string' },
                   },
+                  disallowedForPatterns: {
+                    type: 'array',
+                    uniqueItems: true,
+                    minItems: 1,
+                    items: { type: 'string' },
+                  },
                   message: { type: 'string' },
                 },
-                required: ['disallowedFor'],
+                anyOf: [
+                  { required: ['disallowedFor'] },
+                  { required: ['disallowedForPatterns'] },
+                ],
                 additionalProperties: false,
               },
-
               {
                 type: 'object',
                 properties: {
                   propNamePattern: { type: 'string' },
                   allowedFor: {
+                    type: 'array',
+                    uniqueItems: true,
+                    items: { type: 'string' },
+                  },
+                  allowedForPatterns: {
                     type: 'array',
                     uniqueItems: true,
                     items: { type: 'string' },
@@ -95,9 +113,18 @@ module.exports = {
                     minItems: 1,
                     items: { type: 'string' },
                   },
+                  disallowedForPatterns: {
+                    type: 'array',
+                    uniqueItems: true,
+                    minItems: 1,
+                    items: { type: 'string' },
+                  },
                   message: { type: 'string' },
                 },
-                required: ['disallowedFor'],
+                anyOf: [
+                  { required: ['disallowedFor'] },
+                  { required: ['disallowedForPatterns'] },
+                ],
                 additionalProperties: false,
               },
             ],
@@ -114,8 +141,10 @@ module.exports = {
       const propPattern = value.propNamePattern;
       const prop = propName || propPattern;
       const options = {
-        allowList: typeof value === 'string' ? [] : (value.allowedFor || []),
-        disallowList: typeof value === 'string' ? [] : (value.disallowedFor || []),
+        allowList: [].concat(value.allowedFor || []),
+        allowPatternList: [].concat(value.allowedForPatterns || []),
+        disallowList: [].concat(value.disallowedFor || []),
+        disallowPatternList: [].concat(value.disallowedForPatterns || []),
         message: typeof value === 'string' ? null : value.message,
         isPattern: !!value.propNamePattern,
       };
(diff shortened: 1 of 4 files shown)


Efimenko:
@ljharb what do you think?

lib/rules/forbid-component-props.js
```diff
@@ -115,7 +142,9 @@ module.exports = {
       const prop = propName || propPattern;
       const options = {
         allowList: typeof value === 'string' ? [] : (value.allowedFor || []),
+        allowPatternList: typeof value === 'string' ? [] : value.allowedForPatterns || [],
         disallowList: typeof value === 'string' ? [] : (value.disallowedFor || []),
+        disallowPatternList: typeof value === 'string' ? [] : value.disallowedForPatterns || [],
```
ljharb, line 147:
```suggestion
        allowPatternList: [].concat(value.allowedForPatterns || []),
        disallowList: [].concat(value.disallowedFor || []),
        disallowPatternList: [].concat(value.disallowedForPatterns || []),
```
altho i'm not sure why any of these would be able to be strings - doesn't the schema prevent that?

lib/rules/forbid-component-props.js
```diff
@@ -140,10 +169,48 @@ module.exports = {
         return false;
       }
 
+      function checkIsTagForbiddenByAllowList() {
+        return options.allowList.indexOf(tagName) === -1;
+      }
```
ljharb, line 174: this seems overkill to make into a function, especially if it's used once. let's just inline it.

lib/rules/forbid-component-props.js
```diff
@@ -140,10 +169,48 @@ module.exports = {
...
+
+      function checkIsTagForbiddenByAllowPatternList() {
+        if (options.allowPatternList.length === 0) {
+          return true;
+        }
+
+        return options.allowPatternList.every(
+          (pattern) => !minimatch(tagName, pattern)
+        );
+      }
```
ljharb, line 184: kind of same here? i'd rather checkIsTagForbiddenByAllowOptions be mutiline with some early returns.

lib/rules/forbid-component-props.js
```diff
@@ -140,10 +169,48 @@ module.exports = {
...
+        }
+
+        return options.disallowPatternList.some(
+          (pattern) => minimatch(tagName, pattern)
+        );
+      }
+
+      function checkIsTagForbiddenByDisallowOptions() {
+        return checkIsTagForbiddenByDisallowList() || checkIsTagForbiddenByDisallowPatternList();
+      }
```
ljharb, line 206: same feedback on these three functions

ljharb (changes requested):

lib/rules/forbid-component-props.js
Efimenko: Yeah, you're right about schema, but there is a [default](https://github.com/jsx-eslint/eslint-plugin-react/pull/3805/files/40beeb52feaf2989486565a3df98f2415e1aa5d3#diff-95ccf9a24cad95dd24cdc5d59460de65a598a88ebe8c19a2ee6af25978549eb8L16) also for this rule, so that's why this code supports strings also.

Efimenko:
Honestly, not sure what to do with the "Automatic Rebase" action, didn't find any info in [CONTRIBUTING.md](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/CONTRIBUTING.md)

lib/rules/forbid-component-props.js
```diff
@@ -66,12 +71,20 @@ module.exports = {
...
+                  disallowedForPatterns: {
+                    type: 'array',
+                    uniqueItems: true,
+                    minItems: 1,
+                    items: { type: 'string' },
+                  },
                   message: { type: 'string' },
                 },
-                required: ['disallowedFor'],
+                anyOf: [
```
ljharb, line 82: should this be:
```suggestion
                oneOf: [
```
?

lib/rules/forbid-component-props.js
```diff
@@ -95,9 +113,18 @@ module.exports = {
...
```
ljharb, line 124:
```suggestion
                oneOf: [
```
this one too?

lib/rules/forbid-component-props.js
Efimenko: Not sure, because with `oneOf` you can't use `disallowedFor` and `disallowedForPatterns` simultaneously, while it is still a case. Here is one of the test cases for example: https://github.com/jsx-eslint/eslint-plugin-react/pull/3805/files#diff-0095adf2f1f3e139bb1a1666d7611671ad90ae66f67de0f5cdc7cb8884351968R825-R847

lib/rules/forbid-component-props.js
Efimenko: the same https://github.com/jsx-eslint/eslint-plugin-react/pull/3805#discussion_r1773456872

Efimenko:
> Should there also be examples of using `propNamePatterns` with one of these new options?

I've added one: https://github.com/jsx-eslint/eslint-plugin-react/pull/3805/files#diff-1cf5e390e9444274ee8064c64a72965facfa2171c5453a130f0329e57fe991adR79-R83

Also, I've added one for a more complex setup: https://github.com/jsx-eslint/eslint-plugin-react/pull/3805/files#diff-1cf5e390e9444274ee8064c64a72965facfa2171c5453a130f0329e57fe991adR106-R113 (Relates to this comment https://github.com/jsx-eslint/eslint-plugin-react/pull/3805#discussion_r1773456872)

Efimenko:
For the future, correct me if I'm wrong, but here, `rebase` is a preferable option over `merge`.

ljharb:
Yes, feel free to do whatever you like, but I'll rebase and clean it up as part of landing it.

lib/rules/forbid-component-props.js
Efimenko: @ljharb so, based on the documentation, only one can be successful at a time https://json-schema.org/draft/2020-12/json-schema-core#section-10.2.1.3-2

ljharb: sure. but "only one is required" is the intended semantics, right?

i suppose they're identical semantics, but i read "anyOf" with more optionality than "oneOf", personally ¯\\\_(ツ)_/¯

Efimenko: Yeah, semantically about "at least one required" they are identical probably, but "use several of them simultaneously" is not.

ljharb: right - and "any of" implies multiple matches are expected, "one of" implies only one match is expected

Efimenko: So, to summarize:
- `anyOf` gives the ability to use several fields like `disallowedFor` and `disallowedForPatterns` at the same time;
- `oneOf` forbids the use of several fields at the same time;

So, here we are, just discussing the difference, or we're considering the replacement with `oneOf`? A bit hard to understand 😄

ljharb: lol i'm saying that i think that the actual behavior will be identical with anyOf or oneOf here, but that `oneOf` seems more appropriate to me.

I won't hold up the PR further on this topic, though.

ljharb (approved):

lib/rules/forbid-component-props.js
ljharb: I suppose the default could be made into an array, but nbd.

lib/rules/forbid-component-props.js
Efimenko: I got it, so, actually, no.
With `oneOf`, if you pass at the same time `disallowedFor` and `disallowedForPatterns ` - it will be a validation error.
With `anyOf` - it won't.

PR #3807, opened 2024-08-27
[New] `no-string-refs`: allow this.refs in > 18.3.0

henryqdineen:
In React version 18.3.0 they released [a change to support writing to `this.refs`](https://github.com/facebook/react/pull/28867). In the [React 19 RC Upgrade Guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#run-all-react-19-codemods) they suggest running a codemod that will remove the string ref usage but still use `this.refs`, which can cause false positives for the `no-string-refs` rule. 

My PR simply tests the React version before checking the `MemberExpression`. Thanks!

### DIFF
--- a/tests/lib/rules/no-string-refs.js
+++ b/tests/lib/rules/no-string-refs.js
@@ -59,6 +59,19 @@ ruleTester.run('no-refs', rule, {
         });
       `,
     },
+    {
+      code: `
+        var Hello = createReactClass({
+          componentDidMount: function() {
+            var component = this.refs.hello;
+          },
+          render: function() {
+            return <div>Hello {this.props.name}</div>;
+          }
+        });
+      `,
+      settings: { react: { version: '18.3.0' } },
+    },
   ]),
 
   invalid: parsers.all([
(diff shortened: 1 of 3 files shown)


tests/lib/rules/no-string-refs.js
```diff
@@ -140,10 +145,28 @@ ruleTester.run('no-refs', rule, {
...
+            return <div ref={\`hello\${index}\`}>Hello {this.props.name}</div>;
+          }
+        });
+      `,
+      options: [{ noTemplateLiterals: true }],
+      settings: { react: { version: '18.3.0' } },
+      errors: [
+        { messageId: 'stringInRefDeprecated' },
+      ],
+    },
```
ljharb, line 170: shouldn't this be valid if the version is 18.3.0?

henryqdineen: Sorry for the confusion. It is invalid because of the string ref but not the `this.refs` usage. This is identical to the test case above except there is only one error instead of two.

ljharb: aha, thanks

ljharb (approved):
Bizarre decision on their part, but this seems correct. Thanks!

tests/lib/rules/no-string-refs.js
henryqdineen: I just pushed an update to add a valid test case based one of the invalid test cases but only changing the react version. I hope that helps.

PR #3809, opened 2024-08-30
[Fix] `prop-types`: false positives since upgrading typescript-eslint to v8

### LINKED ISSUE #3796: [Bug]: react/prop-types false positives since upgrading typescript-eslint to v8

### Description Overview

Since upgrading `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` to `v8.0.1`, I'm getting false positives for `react/prop-types`. It looks like this might have been an issue [before](https://github.com/jsx-eslint/eslint-plugin-react/issues/3140).

```
const Divulus = forwardRef<HTMLDivElement, PropsWithChildren>(({ children }, ref) => <div ref={ref}>{children}</div>);
Divulus.displayName = 'X';
```

`'children' is missing in props validation eslint(react/prop-types)`

`eslint`

### Expected Behavior

No error.

v7.35.0

### eslint version

v8.57.0

### node version

-

pplancq:
Hello!

I'd like to start by thanking the amazing staff at this bookshop for all their hard work.

I'm thrilled to present my first contribution, which is a correction to the false positive of `prop-types` rule that have been raised since version v8 of typescript-eslint.

Since version v8, it seems that the `typeParameters` parameter no longer exists, so I've replaced it with the `typeArguments` parameter, which is equivalent and all without causing a breaking change.

Closes #3796

### DIFF
--- a/lib/util/propTypes.js
+++ b/lib/util/propTypes.js
@@ -1044,8 +1044,10 @@ module.exports = function propTypesInstructions(context, components, utils) {
     if (
       node.parent
       && node.parent.callee
-      && node.parent.typeParameters
-      && node.parent.typeParameters.params
+      && (
+        (node.parent.typeParameters && node.parent.typeParameters.params)
+        || (node.parent.typeArguments && node.parent.typeArguments.params)
+      )
       && (
         node.parent.callee.name === 'forwardRef' || (
           node.parent.callee.object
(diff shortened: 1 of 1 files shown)


ljharb:
The change looks good, but I'd prefer not to add code to support v8 of typescript-eslint when we're not even testing on v6 yet (see #3629, which i think subsumes this PR)

pplancq:
Thanks for your reply, I was thinking more of this PR as a patch while waiting for full compatibility. 

I see that your PR has been going on for a while and I understand perfectly that some things take time. 

Now it's up to you to decide whether or not to add this modification.

ljharb:
Obsoleted by #3629.

PR #3812, opened 2024-09-01
[New] `jsx-no-literals`: add `elementOverrides` option

### LINKED ISSUE #3808: [Feature Request]: Add `allowedElements` option to `jsx-no-literals`
### What rule do you want to change?

jsx-no-literals

### How do you think the change should be implemented?

I would like to see an option to add specific JSXElements to an allow list in `jsx-no-literals`. These elements would be allowed to have literals as children. I am unsure if this should also apply the other options that this rule provides such as allowing literals in props. 

The new option could be named `allowedElements` and accept an array of allowed JSX intrinsic element names as well as named JSX element. If the named element is imported, it should use the imported name (so `Button` not `MyButton` in the case of `import { Button as MyButton } from './Button';`. If the import is a default import, it should match against the name given in the import (so `MyButton` in the case of `import MyButton from './Button';`. 

I am unsure of the best way to handle namespaced elements as I could see it going either way. Maybe check each portion of the namespaced name as well as the whole thing.

### Example
```ts
/* eslint react/jsx-no-literals: ["error", { "allowedElements": ["div"] }] */

<span>This should error</span>
<div>This should pass</div>
```

### Additional Info
Currently, we use `jsx-no-literals` to prevent `JSXText` from appearing as the child of an element and instead require developers to pass their copy in via a translation function. This is similar to one of the examples in the `jsx-no-literals` docs. However, we use `react-i18next` which has a `<Trans>` component which will automatically pull all `JSXText` as valid translatable copy. We don't encourage our devs to use this method as it currently violates `jsx-no-literals` but would like to make this method less "hacky" when it is needed.

```ts
/* eslint react/jsx-no-literals: ["error", { "allowedElements": ["Trans"] }] */

<div>This should error</div>
<Trans>This should <a href="#">pass</a></Trans>
```

### Participation

- [X] I am willing to submit a pull request to implement this change.

Pearce-Ropion:
Resolves #3808

Preface - I didn't intend to rewrite the whole rule, but I ended up needing to re-write the way the rule fetches and applies the config which resulted in the majority of the rule being reworked. 

### What is the purpose of this PR?

Adds a new option, `elementOverrides` which adds the ability to override the rule options for any named component. Specifying options within `elementOverrides` creates a new context for the rule, so the rule will only apply the new options to the specified element and its children (if
`applyToNestedElements` is `true` - see below). This means that the root rule options will not apply to the specified element.

`elementOverrides` is an object where the keys are the element names and the values are objects with the same options that this rule previously supported (`noStrings`, `allowedStrings`, `noAttributeStrings` and `ignoreProps`). `elementOverrides` also supports 2 new options: `allowElement` and `applyToNestedElements`.

- `allowElement` (default: `false`) - When `true` the rule will allow the specified element to have string literals as children, wrapped or unwrapped without warning. If the `noStrings` option is also specified, then only unwrapped literals will be allowed through. This option satisfies #3808.
- `applyToNestedElements` (default: `true`) - When `false` the rule will not apply the current options set to nested elements. This is useful when you want to apply the rule to a specific element, but not to its children.

The element name must match the following [regular expression](https://regex101.com/r/mX5rZV/1). It does not support intrinsic JSX elements (ie. `div` or `span`) and all elements that do not match the pattern will be filtered out when the config is normalized. The element name can also be a compound/namespaced/sub component (`JSXMemberExpression`) like `Modal.Button` or `React.Fragment`. If an element name like `Fragment` is specified, the override config will match on either of the following components:
```
<Fragment>text</Fragment>
<React.Fragment>text</React.Fragment>
```

The rule will also construct a simple import (or require) map specifically for renamed imports (ie `import { A as B } from 'c';`. The rule will match on the original (`imported`) specifier instead of the renamed (`local`) specifier. 

### What should reviewers focus on?

Most of the actual rule/option logic has remained unchanged. However, I re-ordered some of the node visitors for performance with regards to when `getOverrideConfig` is called since it is a relatively expensive function. Also, a lot of the visitor functions needed the current config passed in which required some changes within those too.

### How is this PR tested?

This PR adds roughly 50 tests. Each option (new and old) is individually tested for both valid and invalid cases. For each option, there are roughly 3 states that are tested:
1. Just an override
2. An override with nested elements
3. A root config and an override

All extraneous features are also tested such as imports/requires renames, the element name pattern and compound components

Based on a manual scan of the coverage report, I think all branches that need to be covered have been covered by tests. There are a few extra that exist to validate type check.

### DIFF
--- a/lib/rules/jsx-no-literals.js
+++ b/lib/rules/jsx-no-literals.js
@@ -8,6 +8,10 @@
 
 const iterFrom = require('es-iterator-helpers/Iterator.from');
 const map = require('es-iterator-helpers/Iterator.prototype.map');
+const some = require('es-iterator-helpers/Iterator.prototype.some');
+const flatMap = require('es-iterator-helpers/Iterator.prototype.flatMap');
+const fromEntries = require('object.fromentries');
+const entries = require('object.entries');
 
 const docsUrl = require('../util/docsUrl');
 const report = require('../util/report');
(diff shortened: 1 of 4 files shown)


lib/rules/jsx-no-literals.js
```diff
@@ -40,116 +167,267 @@ module.exports = {
...
+            },
           },
-        },
-        ignoreProps: {
-          type: 'boolean',
-        },
-        noAttributeStrings: {
-          type: 'boolean',
-        },
+          commonPropertiesSchema
```
Pearce-Ropion, line 193: This could probably be handled with JsonSchema's `definitions`/`$refs` properties but I couldn't figure out how to best accomplish it.

lib/rules/jsx-no-literals.js
```diff
@@ -40,116 +167,267 @@ module.exports = {
...
+        }
+      }
+    }
+
+    /**
+     * @param {ResolvedConfig} resolvedConfig
+     * @returns {boolean}
+     */
+    function shouldAllowElement(resolvedConfig) {
+      return resolvedConfig.type === 'override' && 'allowElement' in resolvedConfig && !!resolvedConfig.allowElement;
```
Pearce-Ropion, line 400: `type-check` was being picky here

lib/rules/jsx-no-literals.js
```diff
@@ -40,116 +167,267 @@ module.exports = {
...
+     * Gets an override config for a given node. For any given node, we also
+     * need to traverse the ancestor tree to determine if an ancestor's config
+     * will also apply to the current node.
+     * @param {ASTNode} node
+     * @returns {OverrideConfig | undefined}
+     */
+    function getOverrideConfig(node) {
+      if (!hasElementOverrides) return;
+
+      const allAncestorElements = getJSXElementAncestors(node);
```
Pearce-Ropion, line 374: There is an opportunity for some kind of caching/memoization here since I think `getJSXElementAncestors` can get expensive. I gave it a shot but couldn't get a viable solution.  Open to suggestions.

Pearce-Ropion:
How do I get around `eslint` 4 not doing text replacements in the unit tests?

lib/rules/jsx-no-literals.js
```diff
@@ -17,17 +21,157 @@ const getText = require('../util/eslint').getText;
...
+            ]);
+
+            return acc;
+          },
+          []
+        )
+      );
+    }
+  } else {
+    delete normalizedConfig.elementOverrides;
```
ljharb, line 169: it'd be better to not add it, rather than conditionally delete it, i think?

lib/rules/jsx-no-literals.js
ljharb: yeah, typescript sadly doesn't infer `x in obj` from a truthy `x.obj`, although it can and should

ljharb (changes requested):
Initial review.

lib/rules/jsx-no-literals.js
```diff
@@ -17,17 +21,157 @@ const getText = require('../util/eslint').getText;
...
+/**
+ * @param {boolean | undefined} value
+ * @returns {boolean}
+ */
+function valueOrTrue(value) {
+  return value !== undefined ? value : true;
 }
 
+const overridableElementPattern = '^[A-Z][\\w.]*$';
+const reOverridableElement = new RegExp(overridableElementPattern);
```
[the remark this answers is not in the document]
Pearce-Ropion: The string pattern is used in the JSON schema while the regex is used when
normalizing the config. Statically defining the regex as a literal would mean
the same regex is essentially defined twice and I generally value single source of
truth over the minor optimization of definting the regex twice.

lib/rules/jsx-no-literals.js
```diff
@@ -40,116 +184,272 @@ module.exports = {
...
+    function isViableTextNode(node, resolvedConfig) {
+      const textValues = iterFrom([trimIfString(node.raw), trimIfString(node.value)]);
+      if (some(textValues, (value) => resolvedConfig.allowedStrings.has(value))) {
+        return false;
+      }
+
+      const parent = getParentIgnoringBinaryExpressions(node);
+
+      let isStandardJSXNode = false;
+      if (typeof node.value === 'string' && !/^[\s]+$/.test(node.value) && parent.type.startsWith('JSX')) {
```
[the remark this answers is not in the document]
Pearce-Ropion: This was actually one of the few things I left alone from the original
implementation. But since I refactored pretty much everything else here, I can
do this one as well.

lib/rules/jsx-no-literals.js
Pearce-Ropion: Typescript was being a pain about this. But I can probably refactor the function
a bit to make it get along.

lib/rules/jsx-no-literals.js
```diff
@@ -17,17 +21,157 @@ const getText = require('../util/eslint').getText;
...
+const overridableElementPattern = '^[A-Z][\\w.]*$';
+const reOverridableElement = new RegExp(overridableElementPattern);
+const jsxElementTypes = new Set(['JSXElement', 'JSXFragment']);
+
 const messages = {
   invalidPropValue: 'Invalid prop value: "{{text}}"',
+  invalidPropValueInElement: 'Invalid prop value: "{{text}}" in {{element}}',
   noStringsInAttributes: 'Strings not allowed in attributes: "{{text}}"',
+  noStringsInAttributesInElement:
+    'Strings not allowed in attributes: "{{text}}" in {{element}}',
```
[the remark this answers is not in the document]
Pearce-Ropion: Whoops, I realized I had to turn off prettier fairly quickly into coding this but I guess this slipped through.

lib/rules/jsx-no-literals.js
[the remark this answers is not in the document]
ljharb: in that case, this should be `/^[A-Z][\w.]*$/`, and `reOverridableElement.pattern` should be used in the schema.

Either way, `RegExp()` is much much worse than multiple sources of truth.

ljharb (approved):

lib/rules/jsx-no-literals.js
ljharb: Fine to leave it for a follow-on.

lib/rules/jsx-no-literals.js
ljharb: maybe, but it's clearer to avoid it i think.

Pearce-Ropion:
Not sure what to do about the failing `codecov/project` check. I checked codecov while I was testing and I have covered all branches and statements and functions I believe.

ljharb:
Don’t worry about it :-)

PR #3821, opened 2024-09-12
[Fix] `no-is-mounted`: fix logic in method name check

### LINKED ISSUE #3819: [Bug]: "Do not use isMounted" linting errors after upgrading to latest `eslint-plugin-react` (v7.36.0)

### Description Overview

Before upgrading from v7.35.2 to v7.36.0, my linting tests were running with no errors, now they throw 4 errors about the usage of isMounted

`npm run lint` output:
```
  277:5   error    Do not use isMounted                      react/no-is-mounted
  281:21  error    Do not use isMounted                      react/no-is-mounted
  359:5   error    Do not use isMounted                      react/no-is-mounted
  378:5   error    Do not use isMounted                      react/no-is-mounted

✖ 40 problems (4 errors, 36 warnings)
```

We are not using isMounted in our codebase

full command used: `npm run rm && npm i --dev && npm run build && npm run lint`

I later pinned the version back to v7.35.2 and the errors were no longer present

### Expected Behavior

Before the upgrade:
```
✖ 36 problems (0 errors, 36 warnings)
```

v7.36.0

### eslint version

v8.57.0

### node version

v16.20.2

Mathias-S:
The last change to `no-is-mounted` caused the rule to error with any method name, not just with "isMounted".

Fixes #3819

### DIFF
--- a/lib/rules/no-is-mounted.js
+++ b/lib/rules/no-is-mounted.js
@@ -41,7 +41,8 @@ module.exports = {
         }
         if (
           callee.object.type !== 'ThisExpression'
-          && (!('name' in callee.property) || callee.property.name !== 'isMounted')
+          || !('name' in callee.property)
+          || callee.property.name !== 'isMounted'
         ) {
           return;
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ## Unreleased
 
+### Fixed
+* [`no-is-mounted`]: fix logic in method name check ([#3821][] @Mathias-S)
+
+[#3821]: https://github.com/jsx-eslint/eslint-plugin-react/pull/3821
+
 ## [7.36.0] - 2024.09.12
 
 ### Added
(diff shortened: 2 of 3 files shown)


lib/rules/no-is-mounted.js
```diff
@@ -41,7 +41,7 @@ module.exports = {
         }
         if (
           callee.object.type !== 'ThisExpression'
-          && (!('name' in callee.property) || callee.property.name !== 'isMounted')
+          || (!('name' in callee.property) || callee.property.name !== 'isMounted')
```
voxpelli, line 44:
```suggestion
          || !('name' in callee.property)
          || callee.property.name !== 'isMounted'
```

No need to wrap it in parentheses anymore (Also posted in https://github.com/jsx-eslint/eslint-plugin-react/pull/3731/files#r1756585084)

Mathias-S: Good point, removing the parentheses makes the check a bit more readable. I've updated the PR.

voxpelli (approved):

ljharb (approved):
thanks!

Issue #3820, opened 2024-09-12
[Bug]: If boolean props are in JSX, the react/jsx-no-literals rule in latest update(7.36.0) throws exception and breaks ESLINT


aswinsekar:
### Is there an existing issue for this?

- [X] I have searched the existing issues and my issue is unique
- [X] My issue appears in the command-line and not only in the text editor

### Description Overview

If there is a boolean props in JSX, there is an exception thrown in processing this rule and eslint broke down.

`<a isBoolean>test</a>`

[Error - 15:46:51] TypeError: Cannot read properties of null (reading 'type')
Occurred while linting <file_name>
Rule: "react/jsx-no-literals"
    at JSXAttribute (<path>/node_modules/.pnpm/eslint-plugin-react@7.36.0_eslint@8.57.0/node_modules/eslint-plugin-react/lib/rules/jsx-no-literals.js:506:44)
    at ruleErrorHandler (<path>/node_modules/.pnpm/eslint@8.57.0/node_modules/eslint/lib/linter/linter.js:1076:28)
    at <path>/node_modules/.pnpm/eslint@8.57.0/node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Array.forEach (<anonymous>)
    at Object.emit (<path>/node_modules/.pnpm/eslint@8.57.0/node_modules/eslint/lib/linter/safe-emitter.js:45:38)

Happening in both CLI & IDE

### Expected Behavior

No breakage in ESLINT setup

### eslint-plugin-react version

v7.36.0

### eslint version

v8.57.0

### node version

v18.18.2

kai-dorschner-twinsity:
Can confirm, same here

ljharb:
@kai-dorschner-twinsity can you provide the code it’s crashing on?

kai-dorschner-twinsity:
Unfortunately not. It is corporate code. Let me check if I can get a few lines out of it.

PR #3823, opened 2024-09-12
[Fix] `jsx-no-literals`: Avoid crashing on valueless boolean props

reosarevok:
b8217ed5f4104bd18b54f77ea108ec081ebbbd49 removed the `node.value` check leading to this crashing on any valueless boolean prop such as `<Component isWhatever />`

This just readds the check.

Fixes https://github.com/jsx-eslint/eslint-plugin-react/issues/3820


ljharb (approved):
Thanks!

Issue #3770, opened 2024-06-20
[Bug]: 'React' must be in scope when using JSXeslintreact/react-in-jsx-scope


vaynevayne:
### Is there an existing issue for this?

- [X] I have searched the existing issues and my issue is unique
- [X] My issue appears in the command-line and not only in the text editor

### Description Overview

The eslint rule I just created, react18, actually prompts you to introduce react at the top. This should be the default option, right?
Is this the rule intentional and I need to turn it off manually, or is there a problem with my configuration?

<img width="583" alt="image" src="https://github.com/jsx-eslint/eslint-plugin-react/assets/124666577/7d7c009d-003c-4bea-a745-d959cadc2823">

```jsx
import { fixupConfigRules } from "@eslint/compat";
import pluginJs from '@eslint/js';
import pluginReactConfig from 'eslint-plugin-react/configs/recommended.js';
import globals from 'globals';
import tseslint from 'typescript-eslint';

export default [
  { languageOptions: { globals: globals.browser } },
  pluginJs.configs.recommended,
  ...tseslint.configs.recommended,
  {
    files: ['**/*.jsx'],
    languageOptions: { parserOptions: { ecmaFeatures: { jsx: true } } },
  },
  ...fixupConfigRules(pluginReactConfig),
  {
    rules: {
      '@typescript-eslint/no-explicit-any': 'off',
    },
  },
]
```

### Expected Behavior

no error 

### eslint-plugin-react version

7.34.3

### eslint version

^9.5.0

### node version

v20.14.0

ljharb:
This is by design - you need to extend the `jsx-runtime` config, as detailed in the readme [here](https://github.com/jsx-eslint/eslint-plugin-react?tab=readme-ov-file#configuration-legacy-eslintrc-)

GabenGar:
Not sure by which design, since [getting started section](https://eslint.org/docs/latest/use/getting-started) of ESLint results in this config, so they clearly didn't get the memo:
```javascript
import globals from "globals";
import pluginJs from "@eslint/js";
import tseslint from "typescript-eslint";
import pluginReact from "eslint-plugin-react";

export default [
  {files: ["./src/**/*.{js,mjs,cjs,ts,jsx,tsx}"]},
  {languageOptions: { globals: globals.browser }},
  pluginJs.configs.recommended,
  ...tseslint.configs.recommended,
  pluginReact.configs.flat.recommended,
];
```
[...]

PR #3824, opened 2024-09-14
[readme] flat config example for react 17+

GabenGar:
Contrary to what [the comment](https://github.com/jsx-eslint/eslint-plugin-react/issues/3770#issuecomment-2179919625) says, a link to [this snippet](https://github.com/jsx-eslint/eslint-plugin-react/blob/c8917b0885094b5e4cc2a6f613f7fb6f16fe932e/index.js#L163-L176) is a red herring and adding `jsx-runtime` config to array is enough for error to go way.


ljharb:
Thanks! This advice is in fact precisely what that comment says.

Issue #3728, opened 2024-04-08
Proposal for Stringent Type Checking


y-hsgw:
Would it be feasible to enhance the strictness of type checking?

I suggest modifying the code from:

```js
module.exports = {
  //...
};
```
to:

```js
/** @type { import("eslint").Rule.RuleModule } */
const config = {
  //...
};

module.exports = config;

```

This change aims to improve type checking by specifying the type explicitly using JSDoc annotations.

I have checked existing PRs and issues but couldn't find a similar proposal being addressed. If this is already in progress, we apologize for the redundancy.

ljharb:
I'm not sure what you mean - are you talking about how the plugin code is authored?

It would be great to typescript-jsdoc annotate the entire plugin. The first form, however, can also be annotated and is superior to the second form.

y-hsgw:
> are you talking about how the plugin code is authored?

Yes, I am referring to how the plugin code is authored.

The former doesn't seem to provide type checking, so I proposed the latter.




![スクリーンショット 2024-04-08 23 23 15](https://github.com/jsx-eslint/eslint-plugin-react/assets/49516827/60edaa43-78b1-40a1-9e2f-ce0a0675034f)

Similar issue: https://github.com/microsoft/TypeScript/issues/47107

ljharb:
oof, that's terrible.

i'd want one commit that adds all the annotations, and another that refactors from `module.exports = { … }` to `const config = { … }; module.exports = config;`, so i can easily revert that commit in the future when TS fixes themselves.

PR #3825, opened 2024-09-16
Add JSDoc type annotations and Update @types/eslint Version

y-hsgw:
Addressed the following:
- Added JSDoc annotations to rules.
- Updated `@types/eslint`version
  Note: Updated because the current version does not define the type for `RuleMetaData.hasSuggestions`.
  
 Related Issue: #3728

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "@babel/plugin-syntax-do-expressions": "^7.24.7",
     "@babel/plugin-syntax-function-bind": "^7.24.7",
     "@babel/preset-react": "^7.24.7",
-    "@types/eslint": "=7.2.10",
+    "@types/eslint": "=8.56.10",
     "@types/estree": "0.0.52",
     "@types/node": "^4.9.5",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -51,7 +51,7 @@
     "@babel/plugin-syntax-do-expressions": "^7.24.7",
     "@babel/plugin-syntax-function-bind": "^7.24.7",
     "@babel/preset-react": "^7.24.7",
-    "@types/eslint": "=7.2.10",
+    "@types/eslint": "=8.56.10",
```
ljharb, line 54: this is done intentionally, since we support older eslint versions - also what about the estree types?

ljharb (changes requested):

package.json
y-hsgw: I wasn't aware that the support for older ESLint versions was intentional, and I hadn't considered the estree types either. Given this, I'll go ahead and close the PR for now.
Apologies for the confusion 🙇‍♂️

PR #3826, opened 2024-09-16
[New] `no-unstable-nested-components`: add `propNamePattern` to support custom render prop naming conventions

### LINKED ISSUE #3664: [Docs]: Add docs for `customValidators`

### Description Overview

> @JoshuaKGoldberg [suggested I should post an issue here](https://twitter.com/JoshuaKGoldberg/status/1734635311169224827) 🙃 

- `customValidators` are largely undocumented
https://github.com/search?q=repo%3Ajsx-eslint%2Feslint-plugin-react+customValidators+language%3AMarkdown&type=code&l=Markdown
- It's mentioned in the docs [here](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unstable-nested-components.md) and in other places, but I can't figure out how to use it

### Expected Behavior

A code example on how to use `customVAlidators`

n/a

### eslint version

n/a

### node version

n/a

danreeves:
I would like to adopt this rule but `render*` is not the only render prop naming convention in my codebases. Namely libraries such as [react-virtualized](https://github.com/bvaughn/react-virtualized/blob/master/docs/Table.md) have a `*Renderer` convention. 

This PR adds the ability to customize the prop name matcher.

~At the same time I cleaned up the `customValidators` documentation that appears to have been copy/paste error at implementation. I believe this would close 3664, which asks to document this non-existent option.~

### DIFF
--- a/docs/rules/no-unstable-nested-components.md
+++ b/docs/rules/no-unstable-nested-components.md
@@ -125,6 +125,7 @@ function Component() {
   {
     "allowAsProps": true | false,
     "customValidators": [] /* optional array of validators used for propTypes validation */
+    "propNamePattern": string
   }
 ]
 ...
@@ -146,6 +147,16 @@ function Component() {
     </div>
   );
 }
+```
+
+You can allow other render prop naming conventions by setting the `propNamePattern` option. By default this option is `"render*"`.
+
+For example, if `propNamePattern` is set to `"*Renderer"` the following pattern is **not** considered warnings:
+
+```jsx
+<Table
+  rowRenderer={(rowData) => <Row data={rowData} />}
+/>
 ```
 
 ## When Not To Use It
--- a/lib/rules/no-unstable-nested-components.js
+++ b/lib/rules/no-unstable-nested-components.js
@@ -5,6 +5,7 @@
 
 'use strict';
 
+const minimatch = require('minimatch');
 const Components = require('../util/Components');
 const docsUrl = require('../util/docsUrl');
 const astUtil = require('../util/ast');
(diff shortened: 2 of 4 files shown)


docs/rules/no-unstable-nested-components.md
```diff
@@ -124,7 +124,7 @@ function Component() {
   "off" | "warn" | "error",
   {
     "allowAsProps": true | false,
-    "customValidators": [] /* optional array of validators used for propTypes validation */
```
ljharb, line 127: why is this being removed here?

lib/rules/no-unstable-nested-components.js
```diff
@@ -268,22 +272,20 @@ module.exports = {
     schema: [{
       type: 'object',
       properties: {
-        customValidators: {
-          type: 'array',
-          items: {
-            type: 'string',
-          },
-        },
```
ljharb, line 276: removing something is a breaking change, and a nonstarter.

ljharb (changes requested):

docs/rules/no-unstable-nested-components.md
danreeves: Removed the cleanup and added some context to https://github.com/jsx-eslint/eslint-plugin-react/issues/3664

ljharb: oof, i see. in that case we definitely should have a separate PR to either remove it from the schema (a breaking change, unfortunately, even though it doesn't do anything) or to add the functionality (which could be a patch)

danreeves: @ljharb While we figure out what to do with the other option, what do you think about progressing this feature?

ljharb (approved):

PR #3829, opened 2024-09-19
[Fix] `no-unknown-property` support new `precedence` prop

### LINKED ISSUE #3827: [Bug]: Unknown property 'precedence' found (react/no-unknown-property)

### Description Overview

react v19 introduces a new `precedence` string prop to the [`<link>`](https://react.dev/reference/react-dom/components/link#props) and [`<style>`](https://react.dev/reference/react-dom/components/style#props) elements, but the `react/no-unknown-property` rule doesn’t yet know about it and generates linting errors when used:

```jsx
<>
    <link href="https://foo.bar" precedence="medium" rel="canonical" />
    <style href="unique-hash" precedence="anything">{` p { color: red; } `}</style>
</>
```

example error:
```
  211:39  error  Unknown property 'precedence' found  react/no-unknown-property
```

these occur when running my `lint` npm run script: `"lint": "eslint .",`

### Expected Behavior

i expect the new `precedence` attribute to not be considered an “unknown” property

v7.36.1

### eslint version

v8.57.0

### node version

v22.7.0

acusti:
Fixes #3827

### DIFF
--- a/lib/rules/no-unknown-property.js
+++ b/lib/rules/no-unknown-property.js
@@ -363,16 +363,24 @@ const REACT_ON_PROPS = [
 ];
 
 function getDOMPropertyNames(context) {
-  const ALL_DOM_PROPERTY_NAMES = DOM_PROPERTY_NAMES_TWO_WORDS.concat(DOM_PROPERTY_NAMES_ONE_WORD);
-  // this was removed in React v16.1+, see https://github.com/facebook/react/pull/10823
-  if (!testReactVersion(context, '>= 16.1.0')) {
-    return ALL_DOM_PROPERTY_NAMES.concat('allowTransparency');
-  }
-  // these were added in React v16.4.0, see https://reactjs.org/blog/2018/05/23/react-v-16-4.html and https://github.com/facebook/react/pull/12507
-  if (testReactVersion(context, '>= 16.4.0')) {
-    return ALL_DOM_PROPERTY_NAMES.concat(REACT_ON_PROPS);
-  }
-  return ALL_DOM_PROPERTY_NAMES;
+  return [].concat(
+    DOM_PROPERTY_NAMES_TWO_WORDS,
+    DOM_PROPERTY_NAMES_ONE_WORD,
+
+    testReactVersion(context, '>= 16.1.0') ? [].concat(
+      testReactVersion(context, '>= 16.4.0') ? [].concat(
+        // these were added in React v16.4.0, see https://reactjs.org/blog/2018/05/23/react-v-16-4.html and https://github.com/facebook/react/pull/12507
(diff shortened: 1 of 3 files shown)


ljharb (changes requested):
This looks great, except that we should only be supporting this property when the react version is 19+.

acusti:
@ljharb is there a way to accommodate that in the plugin, or does that just mean not adding it until there is a version of the plugin released only for react v19+?

ljharb:
@acusti there is! see the `testReactVersion` function

tests/lib/rules/no-unknown-property.js
```diff
@@ -89,6 +89,7 @@ ruleTester.run('no-unknown-property', rule, {
     { code: '<div onPointerDown={this.onDown} onPointerUp={this.onUp} />' },
     { code: '<input type="checkbox" defaultChecked={this.state.checkbox} />' },
     { code: '<div onTouchStart={this.startAnimation} onTouchEnd={this.stopAnimation} onTouchCancel={this.cancel} onTouchMove={this.move} onMouseMoveCapture={this.capture} onTouchCancelCapture={this.log} />' },
+    { code: '<link precedence="medium" href="https://foo.bar" rel="canonical" />' },
```
ljharb, line 92: shouldn't this test case only pass when the react version is set to 19+?

lib/rules/no-unknown-property.js
```diff
@@ -234,7 +234,7 @@ const DOM_PROPERTY_NAMES_ONE_WORD = [
   // OpenGraph meta tag attributes
   'property',
   // React specific attributes
-  'ref', 'key', 'children',
+  'ref', 'key', 'children', 'precedence',
```
ljharb, line 237: i think maybe this needs to be removed here
```suggestion
  'ref', 'key', 'children',
```

acusti: you were faster than me 😅

tests/lib/rules/no-unknown-property.js
acusti: definitely. i used `react: { version: '19.0.0-rc.0' }`, but if it makes more sense to do `19.0.0` even though that doesn’t match any released versions, happy to do that.

acusti:
@ljharb now that React v19 stable is released, i can update the version checks and test settings to specify `19.0.0` as the minimum version (rather than the prerelease friendly `19.0.0-0` or release-candidate-compatible `19.0.0-rc.0`). would that be preferable?

ljharb (approved):
Thanks! I rebased and refactored the function a bit. LGTM assuming tests pass.

acusti:
wonderful! thanks for reviewing, refactoring, and merging it.

PR #3732, opened 2024-04-09
Add JSDoc type annotations

y-hsgw:
Adds JSDoc annotations to rules where type changes are unnecessary, as discussed in https://github.com/jsx-eslint/eslint-plugin-react/pull/3731#discussion_r1556778045

PR #3830, opened 2024-09-23
Generate type declarations from JSDoc

### LINKED ISSUE #3797: [Feature Request]: Add TypeScript types

### Description Overview

It is possible to have the ESLint config be type checked using JSDoc comments and the `// @ts-check` annotation comment: e.g.
```js
// @ts-check

/** @type {import("eslint").Linter.Config[]} */
export default [
    // ...
]
```

But this package doesn't export types ATM. It would be nice if it did export types so it works without errors using this. Most/all other plugins I used do have types.

### Expected Behavior

For the package to have TypeScript types

v7.35.0

### eslint version

v9.8.0

### node version

v18.20.4

voxpelli:
Building upon #3732 to fix #3797 – this mimics the flow that I use in multiple places to generate type declarations from JSDoc comments – eg. in as [neostandard](https://github.com/neostandard/neostandard/blob/main/declaration.tsconfig.json) (which uses this plugin and which caused me to look into generating types here) and others that use my [types-in-js](https://github.com/voxpelli/types-in-js) oriented [tsconfig](https://github.com/voxpelli/tsconfig).

This PR consists of three commits:

1. The type generation itself – one npm script for generating them and one for cleaning them up by deleting them (reason for cleaning is not only to keep it clean, its also because TypeScript always picks a `.d.ts` file over a `.js` file of the same name, so when the types are generated the JSDoc comments are essentially ignored until one remove the generated `.d.ts` files)
2. A GitHub Actions workflow that tests the generated types against multiple TypeScript versions and TypeScript library configurations using a dummy module in `test-published-types/` – to ensure its compatible with all of them without the need to use `skipLibCheck: true` (which one should not set, see eg. https://github.com/voxpelli/tsconfig/issues/1). Since ESLint `>=9.10` has built in types it uses that for some compatibility testing as well. Some extra dependencies was needed for that as well, see https://github.com/eslint/eslint/pull/18938 And sadly those types does not support TS `3.9`, so testing matrix start at `4.0`. The workflow is [inspired by my personal one](https://github.com/voxpelli/ghatemplates/blob/cf6c2ddfd3f6ba0b79fd7f1edcecb2ceb661f269/.github/workflows/type-check.yml) which I use in eg. [`pony-cause`](https://github.com/voxpelli/pony-cause/blob/2dff2da799e3cbe609add730b8595b79ef04fa18/.github/workflows/ts-internal.yml)
3. Some tweaks to the JSDoc types, to ensure the generated types and the added workflow all passes

I have tried to adapt these additions to the style and approach of this project. I did add one `// eslint-disable-next-line valid-jsdoc` in place, I hope that's okay, else we'll work something out.


.eslintignore
```diff
@@ -0,0 +1 @@
+test-published-types
```
ljharb, line 1: this should be added in eslintrc's `ignorePatterns` instead of using an eslintignore file. it should also be added to package.json's `publishConfig.ignore` list, so we don't ever include it in the tarball.

.github/workflows/type-check.yml
```diff
@@ -0,0 +1,61 @@
...
+        ts_lib:
+          - es2015
+          - es2015,dom
+          - es2020
+          - esnext
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+          show-progress: false
```
ljharb, line 36: why are these helpful?

.github/workflows/type-check.yml
```diff
@@ -0,0 +1,61 @@
...
+        name: 'nvm install lts/* && npm install'
+        with:
+          node-version: 'lts/*'
+          skip-ls-check: true
+
+      - name: build types
+        run: npm run build-types
+
+      - name: npm install working directory
+        run: npm install --ignore-scripts
```
ljharb, line 48: why do we want to ignore scripts? this is a permissionless CI instance, there's nothing to attack

.github/workflows/type-check.yml
```diff
@@ -0,0 +1,61 @@
...
+        working-directory: test-published-types
+
+      - name: install typescript version ${{ matrix.ts_version }}
+        run: npm install --force typescript@${{ matrix.ts_version }}
```
ljharb, line 52:
```suggestion
        run: npm install --no-save typescript@${{ matrix.ts_version }}
```
we shouldn't need `--force` since dep ranges should always allow whatever version we're testing

lib/rules/jsx-no-literals.js
```diff
@@ -182,9 +182,8 @@ const elementOverrides = {
   },
 };
 
-/** @type {import('eslint').Rule.RuleModule} */
 module.exports = {
-  meta: {
+  meta: /** @type {import('eslint').Rule.RuleModule["meta"]} */ ({
```
ljharb, line 186: this seems worse, since it's checking less of the module - why is it needed?

lib/util/makeNoMethodSetStateRule.js
```diff
@@ -44,6 +44,12 @@ function shouldBeNoop(context, methodName) {
     && !testReactVersion(context, '999.999.999'); // for when the version is not specified
 }
 
+// eslint-disable-next-line valid-jsdoc
+/**
+ * @param {string} methodName Yay
+ * @param {(context: import('eslint').Rule.RuleContext) => boolean} [shouldCheckUnsafeCb] Yay
```
ljharb, line 50: "yay"?

package.json
```diff
@@ -5,6 +5,10 @@
   "description": "React specific linting rules for ESLint",
   "main": "index.js",
   "scripts": {
+    "clean-built-types": "rm -f $(find . -maxdepth 1 -type f -name '*.d.ts*') $(find lib -type f -name '*.d.ts*' ! -name 'types.d.ts')",
+    "prebuild-types": "npm run clean-built-types",
+    "build-types": "tsc -p build.tsconfig.json",
+    "prepublishOnly": "npm run build-types",
```
ljharb, line 11: this should be done in prepack, since it's for inclusion in the tarball.

ljharb (changes requested):
Interesting!

ljharb:
(re valid-jsdoc, yes, it's fine, i'm going to be ditching that rule entirely in the next semver-major and only using tsdoc)

.github/workflows/type-check.yml
voxpelli: `persist-credentials` as otherwise the token used to access the git repo will be saved in the `.git` config, which has caused security issues when people have eg zipped the entire folder or through some other way unintentionally shared that config, not knowing that they shared a token with permissions to do stuff others shouldn't be able to do. For this specific workflow it's not really needed as the permissions are explicitly set to be only for content and only for read – but I typically add it anyway in case permissions are extended in the future. I really think this should have been the default value for checkout.

`show-progress` is just to get a cleaner log – the progress log is just noise. It came with my copy and paste from my own workflow

.github/workflows/type-check.yml
voxpelli: Came with my copy and paste from my own workflow – I usually have it as best practice as permissions can change over time

I can remove it here

.github/workflows/type-check.yml
voxpelli: Good catch 👍 I had `--force` in my reusable one as else it would in some scenarios clash with the peer dependency range on my reusable tsconfig

.github/workflows/type-check.yml
ljharb: Thanks - i assume that if this is set to false, then future steps can't do git operations?

lib/rules/jsx-no-literals.js
voxpelli: There seems to be a bug in the generated type declaration otherwise.

I did contemplate whether one should maybe use the latest TS release for generating the type declarations, as bugs like that won't get fixed in `^3.9.9`

lib/util/makeNoMethodSetStateRule.js
voxpelli: That's totally by accident, must have had the focus on the wrong window and accidentally wrote that

lib/rules/jsx-no-literals.js
ljharb: I would certainly expect that - at this point i don't even care much about someone using TS < 5.

package.json
voxpelli: I'll fix 👍 I will have to focus on a client project tomorrow so I will probably not have time to look at this PR again until Wednesday earliest

.github/workflows/type-check.yml
voxpelli: Yes, I believe that to be correct

It's added to allow one to later do a `git push`, but there are better ways to do that: https://github.com/neostandard/neostandard/blob/79da5b6283d7611347cadaad7dc342e6a9bd9338/.github/workflows/sync-reusable.yml#L125

lib/rules/jsx-no-literals.js
voxpelli: Would it be possible to update the TypeScript version for the entire project? I guess it's pinned to 3.x because of the version of the typescript-eslint-parser?

.github/workflows/type-check.yml
ljharb: hm, if `gh auth setup-git` gives you the same credentials that `persist-credentials` would give, then it seems like the only protection it offers is in case someone accidentally tries to tar up a repo's `.git` directory - that seems like a very unlikely scenario.

lib/rules/jsx-no-literals.js
ljharb: Yes, that's why.

It would indeed be great to update the dev deps range to allow `^3.9.9 || ^5.6.2`, and then modify CI so it installs the appropriate typescript version as needed! That'd work well as a separate PR.

.github/workflows/type-check.yml
voxpelli: It's actually something that was discovered recently to have happened fairly widespread – and prompted GitHub to change some of their official actions to exclude hidden files by default, which they did in a minor release to try to stop the issue as soon as possible, which in turn broke a lot of CI setups that relied on hidden files being included

voxpelli: The trick with `gh auth setup-git` is that it's only available there and not anywhere else – not for any install scripts in prior steps, not for any actions in prior steps – and since I don't execute any code there the attack vectors are very slim (essentially git hooks or if somehow the built in git commands would be shimmed or something)

ljharb: right, but it's just a shell command, any part of the CI workflow can invoke it, no?

voxpelli: The token it uses is provided specifically for that step: https://github.com/neostandard/neostandard/blob/79da5b6283d7611347cadaad7dc342e6a9bd9338/.github/workflows/sync-reusable.yml#L130

lib/util/makeNoMethodSetStateRule.js
voxpelli: Actually, now I remember, I added them when I tried to figure out what `valid-jsdoc` was complaining about – as it simply says "invalid" without saying what part of the jsdoc comment it considers to be invalid (eg. `eslint-plugin-jsdoc` [requires param descriptions](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/require-param-description.md#readme) in its non-ts oriented recommended config)

package.json
voxpelli: This tab was open when I opened my computer this morning, so I squeezed it in anyhow 🙈 Most things should be fixed now

voxpelli:
@ljharb Is the "Automatic Rebase / _ / Automatic Rebase" failure something I should adress by rebasing this PR myself and force pushing? Or by eg. merging `master` into this branch? Or something you'll deal with on merge?

.github/workflows/type-check.yml
voxpelli: Here's an article on what I mentioned yesterday – didn't find it when I was on my phone: https://www.bleepingcomputer.com/news/security/github-actions-artifacts-found-leaking-auth-tokens-in-popular-projects/

voxpelli: I agree with this conclusion in the end of that article:

> The default settings for dangerous actions like 'actions/checkout' should be adjusted so credentials do not persist. Furthermore, token permissions used in workflows should be set to the least privilege necessary to limit damage in case they are exposed.

ljharb:
@voxpelli i'll deal with it before merging, you can ignore it - it fails sometimes despite a PR being rebaseable and maintainers having edit permissions ¯\\\_(ツ)_/¯

ljharb:
Is there a reason d.ts generation is preferred over manually authoring d.ts files, and tsdoc-importing definitions in comments?

voxpelli:
> Is there a reason d.ts generation is preferred over manually authoring d.ts files, and tsdoc-importing definitions in comments?

Often when people author d.ta files that don't reimport them – so they are essentially creating a parallel universe that may or may not mirror the actual code that they are shopping.

Automating it means that all of the generated types are validated against and based on the actual implementation – that way one can be sure that it never becomes out of sync 

I sometimes to author .d.ts files, but mostly for helpers that becomes much less verbose to write in TS syntax than in JSDoc, so I write them there and then import them, and that works great as well I think

I sometimes write the `index.d.ts` by hand as well, to ensure that I have full control of which top level types that are exported, and can easily export additional types that may not be exported automatically from `index.js` otherwise, but which may be needed to be referenced in callbacks or configs or such

voxpelli:
Also:

When generating it automatically from JS one can easily add the declaration source maps that I enabled in this PR, and that makes it so that when one wants to find the definition of a function in eg VS Code it will not jump to the type file, it will jump to the code file, which is nice.

We actually discussed it here today: https://x.com/voxpelli/status/1838625086494118053?s=46&t=1mQKe1AKaQ-2YwRjxDfrOg

ljharb (approved):

PR #3831, opened 2024-09-24
[New] Add suggestions to `no-unescaped-entities`

### LINKED ISSUE #3277: [FEATURE] Add suggestions for `no-unescaped-entities`
Currently, `no-unescaped-entities` only tell the problem in the description, it doesn't give suggestions so users can quickly solve the issue in their editor.

https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions

StyleShit:
Closes #3277


StyleShit:
I can't figure out why the tests are failing, seems like I'm passing the `suggestions` correctly, what am I missing here? 🤦

ljharb:
Looks pretty good, please mark ready for review once tests are passing :-)

StyleShit:
@ljharb Thanks! 🙏 

I have some issues running the tests locally
Well, actually they run, but don't fail (as opposed to the CI), so I'm kinda writing them blindly and hoping the CI will pass 😞 

Any help here would be greatly appreciated

ljharb:
You'll need to run the same commands the CI workflow is using to ensure the right versions of deps are installed; in this case i think it's that there's a lot of test cases that need to have `suggestions` added.

StyleShit:
Oh, that makes sense. Will try it, thanks!

ljharb (approved):
Thanks, looks good

PR #3835, opened 2024-09-29
[Fix] `destructuring-assignment`: fix false negative when using typeof props.a

### LINKED ISSUE #3828: [Bug]: react/destructuring-assignment inconsistent behavior

### Brief description

It appears the `react/destructuring-assignment` rule is not behaving consistently with type definitions.

### Code example
```tsx
type Props = { text: string };
export const MyComponent: React.FC<Props> = (props) => {
  return <div>{props.text}</div>;
  //           ^^^^^^^^^^
};

export const MyOtherComponent: React.FC<Props> = (props) => {
  const { text } = props;
  type MyType = typeof props.text;
  return <div>{text as MyType}</div>;
};
```

### What is happening? / What is the error?
- In `MyComponent`, linter gives a `react/destructuring-assignment` error.
- In `MyOtherComponent` linter stays silent.

To me this appears as an unexpected inconsistency.

Using the code example above;
```sh
npx eslint ./eslint-test.tsx
```

### Expected Behavior

I would expect both components to trigger the linter rule.

v7.34.3

### eslint version

v8.57.0

### node version

v18.20.2

golopot:
Fixes #3828 

The rule failed to catch usages of `typeof props.a` because it is parsed a TSQualifiedName instead of a memberExpression.

### DIFF
--- a/lib/rules/destructuring-assignment.js
+++ b/lib/rules/destructuring-assignment.js
@@ -196,12 +196,7 @@ module.exports = {
       'FunctionExpression:exit': handleStatelessComponentExit,
 
       MemberExpression(node) {
-        let scope = getScope(context, node);
-        let SFCComponent = components.get(scope.block);
-        while (!SFCComponent && scope.upper && scope.upper !== scope) {
-          SFCComponent = components.get(scope.upper.block);
-          scope = scope.upper;
-        }
+        const SFCComponent = utils.getParentStatelessComponent(node);
         if (SFCComponent) {
           handleSFCUsage(node);
         }
--- a/lib/rules/destructuring-assignment.js
+++ b/lib/rules/destructuring-assignment.js
@@ -181,6 +181,25 @@ module.exports = {
       }
     }
 
+    // valid-jsdoc cannot read function types
+    // eslint-disable-next-line valid-jsdoc
+    /**
+     * Find a parent that satisfy the given predicate
+     * @param {ASTNode} node
+     * @param {(node: ASTNode) => boolean} predicate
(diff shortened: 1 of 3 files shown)


tests/lib/rules/destructuring-assignment.js
```diff
@@ -882,10 +882,16 @@ ${'            '}
         };
       `,
       options: ['always', { destructureInSignature: 'always' }],
-      features: ['types'],
+      features: ['types', 'no-babel'],
```
ljharb, line 885: we should still be testing the pre-existing test cases against the babel parser.

lib/rules/destructuring-assignment.js
```diff
@@ -196,12 +215,7 @@ module.exports = {
       'FunctionExpression:exit': handleStatelessComponentExit,
 
       MemberExpression(node) {
-        let scope = getScope(context, node);
-        let SFCComponent = components.get(scope.block);
-        while (!SFCComponent && scope.upper && scope.upper !== scope) {
-          SFCComponent = components.get(scope.upper.block);
-          scope = scope.upper;
-        }
```
ljharb, line 204: i think we still need this; we support Flow as well as many versions of TS.

tests/lib/rules/destructuring-assignment.js
golopot: I changed the destination branch from issue-3828 to master.

ljharb (approved):
awesome, thanks! i'll get this landed in the next day or two.

Issue #3838, opened 2024-10-01
[Bug]: TS types are broken


burtek:
### Is there an existing issue for this?

- [X] I have searched the existing issues and my issue is unique
- [X] My issue appears in the command-line and not only in the text editor

### Description Overview

The types as exported in #3797 and #3836 are broken and throw TS errors while used. The package only exports named namespace of `configs`, which means you can't do any of the following:

```ts
import react from 'eslint-plugin-react';
import { config } from 'typescript-eslint';

export default config({
    plugin: { react },
    rules: {
        ...react.configs.recommended.rules,
        ...someMoreRules
    }
});
```

![image](https://github.com/user-attachments/assets/5e45ab7a-dffb-4455-8533-43875e35c6f9)

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBFApgQwMbwGZQiOByRAZwBtgA7GAWjGIFcBzcypNGPAbgChRJY4BvOKghkMwenAC+cLDnwwAnmCKoowMFSKkKHTp2FlC8QjkQBZaIgBKtYkTgBeAZK6dEAD17wAJogzJbeAMxegAKfk44KLgaBnJCAC4BBBR0KQAaSOioWyIkiOjCuAA6UpZ0YuDxQmKkYRAQRDJfb1rcwkyi6NLik0aLJBs7QiyoyU5JAEp2IA

### Expected Behavior

No TS error.

### eslint-plugin-react version

v7.37.1

### eslint version

v8.57.0

### node version

v20.9.0

### typescript version

All, tested on v5.5.2, v.5.6.2

JstnMcBrd:
This appears to be because only the `flat` configs are included in the types. You can do `reactPlugin.configs.flat.recommended`, but you can't access the legacy configs with `reactPlugin.configs.recommended`.

burtek:
Oh, that's true (though still not in line with what's actually exported). But I still can't do `plugin: { react }`

burtek:
And the `flat` config is not typescript-eslint compatible either...

![image](https://github.com/user-attachments/assets/cbdebbf5-48bc-4a17-be68-d1906dd25463)

JstnMcBrd:
Yes, I was able to add that - even the flat config types are broken. They appear to be incompatible with ESLint's types.

```ts
import type { Linter } from 'eslint';

export function reactConfig(): Linter.FlatConfig[] {
  return [
    reactPlugin.configs.flat.recommended,
[...]

PR #3840, opened 2024-10-12
[Fix] correct generated type declaration

ocavue:
Closes https://github.com/jsx-eslint/eslint-plugin-react/issues/3838

This PR fixes various issues in the generated `index.d.ts`. The full diff for `index.d.ts` can be viewed in [this link](https://github.com/ocavue/eslint-plugin-react/commit/d8be0647b10f1ae01344bc83a8254e16064280e3). The change highlight are shown below:

```diff
  rules: {
-   'react/display-name': number;
+   'react/display-name': 2;
-   'react/jsx-key': number;
+   'react/jsx-key': 2;
```

```diff
    'jsx-no-duplicate-props': import("eslint").Rule.RuleModule;
    'jsx-no-leaked-render': import("eslint").Rule.RuleModule;
-   'jsx-no-literals': {
-      meta: import("eslint").Rule.RuleMetaData;
-      create(context: any): (false & {
-      ...
-   };
+   'jsx-no-literals': import("eslint").Rule.RuleModule;
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,7 +11,7 @@ function filterRules(rules, predicate) {
 
 /**
  * @param {object} rules - rules object mapping rule name to rule module
- * @returns {Record<string, 2 | 'error'>}
+ * @returns {Record<string, SEVERITY_ERROR | 'error'>}
  */
 function configureAsError(rules) {
   return fromEntries(Object.keys(rules).map((key) => [`react/${key}`, 2]));
@@ -31,6 +31,10 @@ const plugins = [
   'react',
 ];
 
+// TODO: with TS 4.5+, inline this
+const SEVERITY_ERROR = /** @type {2} */ (2);
+const SEVERITY_OFF = /** @type {0} */ (0);
+
 const configs = {
   recommended: {
     plugins,
@@ -40,28 +44,28 @@ const configs = {
       },
     },
     rules: {
-      'react/display-name': 2,
-      'react/jsx-key': 2,
-      'react/jsx-no-comment-textnodes': 2,
-      'react/jsx-no-duplicate-props': 2,
-      'react/jsx-no-target-blank': 2,
-      'react/jsx-no-undef': 2,
-      'react/jsx-uses-react': 2,
-      'react/jsx-uses-vars': 2,
-      'react/no-children-prop': 2,
-      'react/no-danger-with-children': 2,
-      'react/no-deprecated': 2,
-      'react/no-direct-mutation-state': 2,
-      'react/no-find-dom-node': 2,
-      'react/no-is-mounted': 2,
-      'react/no-render-return-value': 2,
-      'react/no-string-refs': 2,
-      'react/no-unescaped-entities': 2,
-      'react/no-unknown-property': 2,
-      'react/no-unsafe': 0,
-      'react/prop-types': 2,
-      'react/react-in-jsx-scope': 2,
-      'react/require-render-return': 2,
+      'react/display-name': SEVERITY_ERROR,
+      'react/jsx-key': SEVERITY_ERROR,
+      'react/jsx-no-comment-textnodes': SEVERITY_ERROR,
+      'react/jsx-no-duplicate-props': SEVERITY_ERROR,
+      'react/jsx-no-target-blank': SEVERITY_ERROR,
+      'react/jsx-no-undef': SEVERITY_ERROR,
+      'react/jsx-uses-react': SEVERITY_ERROR,
+      'react/jsx-uses-vars': SEVERITY_ERROR,
+      'react/no-children-prop': SEVERITY_ERROR,
+      'react/no-danger-with-children': SEVERITY_ERROR,
+      'react/no-deprecated': SEVERITY_ERROR,
+      'react/no-direct-mutation-state': SEVERITY_ERROR,
+      'react/no-find-dom-node': SEVERITY_ERROR,
+      'react/no-is-mounted': SEVERITY_ERROR,
+      'react/no-render-return-value': SEVERITY_ERROR,
+      'react/no-string-refs': SEVERITY_ERROR,
+      'react/no-unescaped-entities': SEVERITY_ERROR,
+      'react/no-unknown-property': SEVERITY_ERROR,
+      'react/no-unsafe': SEVERITY_OFF,
+      'react/prop-types': SEVERITY_ERROR,
+      'react/react-in-jsx-scope': SEVERITY_ERROR,
+      'react/require-render-return': SEVERITY_ERROR,
     },
   },
   all: {
@@ -82,8 +86,8 @@ const configs = {
       jsxPragma: null, // for @typescript/eslint-parser
     },
     rules: {
-      'react/react-in-jsx-scope': 0,
-      'react/jsx-uses-react': 0,
+      'react/react-in-jsx-scope': SEVERITY_OFF,
+      'react/jsx-uses-react': SEVERITY_OFF,
     },
   },
 };
--- a/lib/rules/forbid-foreign-prop-types.js
+++ b/lib/rules/forbid-foreign-prop-types.js
@@ -109,7 +109,7 @@ module.exports = {
             && !ast.isAssignmentLHS(node)
             && !isAllowedAssignment(node)
           )) || (
-            // @ts-expect-error The JSXText type is not present in the estree type definitions
+            // @ts-expect-error: The JSXText type is not present in the estree type definitions
             (node.property.type === 'Literal' || node.property.type === 'JSXText')
             && 'value' in node.property
             && node.property.value === 'propTypes'
--- a/lib/rules/index.js
+++ b/lib/rules/index.js
@@ -3,7 +3,7 @@
 /* eslint global-require: 0 */
 
 /** @satisfies {Record<string, import('eslint').Rule.RuleModule>} */
-module.exports = {
+const rules = {
   'boolean-prop-naming': require('./boolean-prop-naming'),
   'button-has-type': require('./button-has-type'),
   'checked-requires-onchange-or-readonly': require('./checked-requires-onchange-or-readonly'),
@@ -108,3 +108,5 @@ module.exports = {
   'style-prop-object': require('./style-prop-object'),
   'void-dom-elements-no-children': require('./void-dom-elements-no-children'),
 };
+
+module.exports = rules;
--- a/lib/rules/jsx-fragments.js
+++ b/lib/rules/jsx-fragments.js
@@ -170,7 +170,12 @@ module.exports = {
       ImportDeclaration(node) {
         if (node.source && node.source.value === 'react') {
           node.specifiers.forEach((spec) => {
-            if ('imported' in spec && spec.imported && spec.imported.name === fragmentPragma) {
+            if (
+              'imported' in spec
+              && spec.imported
+              && 'name' in spec.imported
+              && spec.imported.name === fragmentPragma
+            ) {
               if (spec.local) {
                 fragmentNames.add(spec.local.name);
               }
--- a/lib/rules/jsx-no-literals.js
+++ b/lib/rules/jsx-no-literals.js
@@ -17,6 +17,14 @@ const docsUrl = require('../util/docsUrl');
 const report = require('../util/report');
 const getText = require('../util/eslint').getText;
 
+/** @typedef {import('eslint').Rule.RuleModule} RuleModule */
+
+/** @typedef {import('../../types/rules/jsx-no-literals').Config} Config */
+/** @typedef {import('../../types/rules/jsx-no-literals').RawConfig} RawConfig */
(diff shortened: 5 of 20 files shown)


package.json
```diff
@@ -55,12 +55,11 @@
     "@babel/plugin-syntax-do-expressions": "^7.24.7",
     "@babel/plugin-syntax-function-bind": "^7.24.7",
     "@babel/preset-react": "^7.24.7",
-    "@types/eslint": "=7.2.10",
     "@types/estree": "0.0.52",
     "@types/node": "^4.9.5",
     "@typescript-eslint/parser": "^2.34.0 || ^3.10.1 || ^4 || ^5 || ^6.20 || ^7.14.1 || ^8.4",
     "babel-eslint": "^8 || ^9 || ^10.1.0",
-    "eslint": "^3 || ^4 || ^5 || ^6 || ^7 || ^8 || ^9.7",
```
ocavue, line 63: `eslint` has built-in type declaration since version 9.10, so we don't need to install `@types/eslint` anymore.

lib/rules/index.js
```diff
@@ -3,7 +3,7 @@
 /* eslint global-require: 0 */
 
 /** @satisfies {Record<string, import('eslint').Rule.RuleModule>} */
-module.exports = {
+const rules = {
```
ocavue, line 6: `@satisfies` doesn't work for `module.exports = {` and you have to use a variable here to make it work. That's why the missing `/** @type {import('eslint').Rule.RuleModule} */` in `lib/rules/forward-ref-uses-ref.js` wasn't be found before.

index.js
```diff
@@ -27,6 +27,9 @@ const plugins = [
   'react',
 ];
 
+const SEVERITY_ERROR = /** @type {2} */ (2);
+const SEVERITY_OFF = /** @type {0} */ (0);
```
ljharb, line 31: can you elaborate on why this is needed/preferred, vs a literal number?

lib/rules/jsx-fragments.js
```diff
@@ -170,7 +170,7 @@ module.exports = {
       ImportDeclaration(node) {
         if (node.source && node.source.value === 'react') {
           node.specifiers.forEach((spec) => {
-            if ('imported' in spec && spec.imported && spec.imported.name === fragmentPragma) {
+            if ('imported' in spec && spec.imported && 'name' in spec.imported && spec.imported.name === fragmentPragma) {
```
ljharb, line 173:
```suggestion
            if (
              'imported' in spec
              && spec.imported
              && 'name' in spec.imported
              && spec.imported.name === fragmentPragma
            ) {
```

package.json
```diff
@@ -55,12 +55,11 @@
...
+    "eslint": "^3 || ^4 || ^5 || ^6 || ^7 || ^8 || ^9.10",
```
ljharb, line 62: unfortunately this is a breaking change, and we must stick to 9.7 as the lowest threshold.

ljharb (changes requested):

package.json
ocavue: There are two `^9.7` in `package.json`. One is in `peerDependencies` and another is in `devDependencies`. I only changed the one in `devDependencies`. Is it going to be a breaking change in this way?

ljharb: It means we won't be able to test on 9.7, 9.8, and 9.9, which means it will break and we won't know about it. Peer and dev dep ranges must always be identical.

ocavue: Okey let me check it back.

index.js
ocavue: In `eslint`'s type define ([link](https://github.com/eslint/eslint/blob/7dd402d33226d821a17b22c4753ce9c40fc909bd/lib/types/index.d.ts#L898)), a severity number can only be literal `0`, `1` or `2`. 

If we have `'react/display-name': number;` in `eslint-plugin-react`'s `index.d.ts`, we will get the error below:

```
Property ''react/display-name'' is incompatible with index signature.
  Type 'number' is not assignable to type 'RuleEntry<any[]> | undefined'.ts(2322) 
```

(This error was mentioned in [this comment](https://github.com/jsx-eslint/eslint-plugin-react/issues/3838#issuecomment-2386970537))

Since `RuleEntry` is `0 | 1 | 2 | "off" | "warn" | "error"` or an array. 

Once we have a literal number `'react/display-name': 2;` than TypeScript won't complent anymore.

lib/rules/jsx-no-literals.js
```diff
@@ -65,56 +65,10 @@ const commonPropertiesSchema = {
...
- * @typedef {ElementConfig & ElementOverrides} Config
- * @typedef {Config | OverrideConfig} ResolvedConfig
- */
-
 /**
  * Normalizes the element portion of the config
- * @param {RawConfig} config
- * @returns {ElementConfig}
+ * @param {import("../../types/rules/jsx-no-literals").RawConfig} config
+ * @returns {import("../../types/rules/jsx-no-literals").ElementConfig}
```
ocavue, line 71: After adding `/** @type {import('eslint').Rule.RuleModule} */` to `lib/rules/jsx-no-literals.js`, the generated `lib/rules/jsx-no-literals.d.ts` has the following content:

```ts
declare const _exports: import('eslint').Rule.RuleModule;
export = _exports;
export type RawElementConfigProperties = {
    noStrings?: boolean;
    allowedStrings?: string[];
    ignoreProps?: boolean;
    noAttributeStrings?: boolean;
};
```

TypeScript will throw error when using this .d.ts file: `An export assignment cannot be used in a module with other exported elements.ts(2309)`

Although types like `RawElementConfigProperties` are only used internally, using JSDoc `@typedef` would export them (see issue https://github.com/microsoft/TypeScript/issues/46011) and cause the error above.

To resolve the issue, I added a new `types/rules/jsx-no-literals.d.ts` file and moved all types defined with `@typedef` into it.

package.json
ocavue: I reverted this change. It seems that my PR can work anyway without touching this part.

index.js
ljharb: I guess I’m confused why TS infers number here when it’s a literal 2, but that makes sense.

ocavue: > why TS infers number here when it’s a literal 2

If TS infers 2 as a literal 2, the following code would break:

```
const obj = { a: 1, b: 2 }
obj.b = 3 // Cannot assign 3 to 2
```

TS provides a handy tool `as const` to address such issue, sadly we don't have similar thing in JSDoc.

```ts
const obj = { a: 1, b: 2 }
     // ^: { a: number; b: number }
const obj2 = { a: 1, b: 2 } as const 
     // ^: { a: readonly 1; b: readonly 2 }
```

nchevsky: > TS provides a handy tool `as const` to address such issue, sadly we don't have similar thing in JSDoc.

We actually do, since [TypeScript v4.5](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#const-assertions-and-default-type-arguments-in-jsdoc):

https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#casts

ocavue: Thank you for the correction! I tried that but it didn't work. I realized that's because this project is still using TS v3.  Since the current approach works I'd prefer not updating TS in this PR.

ljharb:
ok, i fixed the issue with index.js in master, but i've rebased this PR on top of that, since this is probably a better outcome still :-)

types/string.prototype.repeat/index.d.ts
```diff
@@ -1,3 +1,4 @@
 declare module 'string.prototype.repeat' {
-  export = typeof Function.call.bind(String.prototype.repeat);
+  function repeat(text: string, count: number): string;
```
ljharb, line 2: this means that it won't always inherit the builtin type for repeat. is this necessary?

lib/rules/jsx-no-literals.js
```diff
@@ -485,7 +447,7 @@ module.exports = {
               });
           });
       },
-    } : false, {
+    } : {}, {
```
ljharb, line 450: why this change? primitives are object-spread the same as `{}`.

types/string.prototype.repeat/index.d.ts
ocavue: The previous version of `types/string.prototype.repeat/index.d.ts` doesn't work. 

```ts
declare module 'string.prototype.repeat' {
  export = typeof Function.call.bind(String.prototype.repeat);
           // ^: The expression of an export assignment must be an identifier or qualified name in an ambient context.ts(2714)
}
```

This error didn't show up because `tsconfig.json` didn't include `types/` at all, thus `string.prototype.repeat` was been treated as `any` by TypeScript.

I'm not sure if there are better way to write this `index.d.ts`, but it definitely need some changes.

ljharb: ah, right. how about:
```suggestion
  const repeat: typeof Function.call.bind(String.prototype.repeat);
```
?

lib/rules/jsx-no-literals.js
ocavue: I don't remember why I did this change exactly, probably because `false` isn't compatible with the latest TypeScript or `@types/eslint`. 

Anyway, let me revert this change for now. We might will need to change it to `{}` in the future if we update TypeScript or `@types/eslint`.

ocavue:
@ljharb Hi. Thanks for your review. I've addressed all your previous comments.

In addition to that, I was thinking why we have issues like https://github.com/jsx-eslint/eslint-plugin-react/issues/3838 even if we have `.github/workflows/type-check.yml` to check the built types. I found there are three issues in the CI workflow and I fixed them. 

1. In `test-published-types/index.js`, to make the typescript to check the type, we need to create a variable first and export it later

```ts
// Bad
/** @type {import('eslint').Linter.Config[]} */
module.exports = [ ... ] 

// Good
/** @type {import('eslint').Linter.Config[]} */
const config = [ ... ]
module.exports = config;
```

2. In `test-published-types/package.json`, if we use `"eslint-plugin-react": "file:.."`, we would copy/link all node_modules under eslint-plugin-react to `test-published-types/node_modules/eslint-plugin-react/node_modules`. This is problematic because `test-published-types` use the built-in types from the latest `eslint`, while `eslint-plugin-react` use the deprecated `@types/eslint`. They have conflict. 
  
   I fixed this by packing `eslint-plugin-react` into a `.taz` file and install it later. By doing that, `@types/eslint`, which is a dev dependency of `eslint-plugin-react`, won't be installed under `test-published-types`.

3. There is a small issue in the eslint's type declarations:
   
```bash
$ cd test-published-types
$ npx tsc --lib es2015
   
node_modules/eslint/lib/types/index.d.ts:928:81 - error TS2574: A rest element type must be an array type.

928     type RuleSeverityAndOptions<Options extends any[] = any[]> = [RuleSeverity, ...Partial<Options>];
                                                                                    ~~~~~~~~~~~~~~~~~~~
```

I need to fix it by adding `"skipLibCheck": false` in `test-published-types/tsconfig.json`.

types/string.prototype.repeat/index.d.ts
ocavue: It doesn't work :(

```
declare module 'string.prototype.repeat' {
  const repeat: typeof Function.call.bind(String.prototype.repeat);
                                      // ^: ',' expected. ts(1005)
  export = repeat;
}
```

burtek:
Any updates on this? :eyes:

lib/rules/forbid-foreign-prop-types.js
```diff
@@ -109,7 +109,7 @@ module.exports = {
             && !ast.isAssignmentLHS(node)
             && !isAllowedAssignment(node)
           )) || (
-            // @ts-expect-error The JSXText type is not present in the estree type definitions
+            // @ts-expect-error: Literal is not a valid type
```
ljharb, line 112: O.o why is Literal not a valid type?

types/string.prototype.repeat/index.d.ts
ljharb: i can fix this later, probably by making a DT package.

ljharb (approved):

lib/rules/forbid-foreign-prop-types.js
ocavue: Sorry. It's `JSXText` that's not a valid type.

```
(node.property.type === 'Literal' || node.property.type === 'JSXText')
                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                     ^: This comparison appears to be unintentional because the 
                                        types '"ArrayExpression" | "ArrowFunctionExpression" | 
                                        "AssignmentExpression" | "AwaitExpression" | 
                                        "BinaryExpression" | "CallExpression" | "ChainExpression" | 
                                        "ClassExpression" | ... 16 more ... | "PrivateIdentifier"' 
                                        and '"JSXText"' have no overlap. 
                                        ts(2367)
```

ocavue: I have reverted this comment change.

PR #3842, opened 2024-10-18
[New] create ignorePrivate option for no-multi-comp rule

vwillyams:
Adds a new option to no-multi-comp which ignores components that are not exported.

Currently WIP:
- Thorough testing, find edge cases. `npm test` wasn't working locally so I'm hoping to rely on CI for this.
- Comparisons between exports and React components probably needs more work. Maybe there's a pre-existing lib function for this. Maybe this functionality should be moved into lib.

### DIFF
--- a/docs/rules/no-multi-comp.md
+++ b/docs/rules/no-multi-comp.md
@@ -69,6 +69,10 @@ class HelloJohn extends React.Component {
 module.exports = HelloJohn;
 ```
 
+### `exportOnly`
+
+When `true` the rule will ignore components which are not exported, which allows you to define components as long as they are only used within a private scope.
+
 ## When Not To Use It
 
 If you prefer to declare multiple components per file you can disable this rule.
--- a/docs/rules/no-multi-comp.md
+++ b/docs/rules/no-multi-comp.md
@@ -115,7 +115,7 @@ function Hello(props) {
 function HelloAgain(props) {
   return <div>Hello again {props.name}</div>;
 }
-module.exports = {Hello, HelloAgain}
+module.exports = { Hello, HelloAgain };
 ```
 
 ## When Not To Use It
--- a/docs/rules/no-multi-comp.md
+++ b/docs/rules/no-multi-comp.md
@@ -69,7 +69,7 @@ class HelloJohn extends React.Component {
 module.exports = HelloJohn;
 ```
 
-### `ignorePrivate`
+### `ignoreInternal`
 
 When `true` the rule will ignore components which are not exported, which allows you to define components that are consumed only within the same file.
 This ensures there is only one entry point for a React component without limiting the structural content of the file.
--- a/docs/rules/no-multi-comp.md
+++ b/docs/rules/no-multi-comp.md
@@ -73,6 +73,51 @@ module.exports = HelloJohn;
 
 When `true` the rule will ignore components which are not exported, which allows you to define components as long as they are only used within a private scope.
 
+Examples of **correct** code for this rule:
+
+```jsx
+export function Hello(props) {
+  return <div>Hello {props.name}</div>;
+}
+function HelloAgain(props) {
+  return <div>Hello again {props.name}</div>;
+}
+```
+
+```jsx
+function Hello(props) {
+  return <div>Hello {props.name}</div>;
+}
+class HelloJohn extends React.Component {
+  render() {
+    return <Hello name="John" />;
+  }
+}
+module.exports = HelloJohn;
+```
+
+Examples of **incorrect** code for this rule:
+
+```jsx
+export function Hello(props) {
+  return <div>Hello {props.name}</div>;
+}
+export function HelloAgain(props) {
+  return <div>Hello again {props.name}</div>;
+}
+```
+
+```jsx
+function Hello(props) {
+  return <div>Hello {props.name}</div>;
+}
+function HelloAgain(props) {
+  return <div>Hello again {props.name}</div>;
+}
+module.exports = {Hello, HelloAgain}
+```
+
+
 ## When Not To Use It
 
 If you prefer to declare multiple components per file you can disable this rule.
--- a/docs/rules/no-multi-comp.md
+++ b/docs/rules/no-multi-comp.md
@@ -69,9 +69,10 @@ class HelloJohn extends React.Component {
 module.exports = HelloJohn;
 ```
 
-### `exportOnly`
+### `ignorePrivate`
 
-When `true` the rule will ignore components which are not exported, which allows you to define components as long as they are only used within a private scope.
+When `true` the rule will ignore components which are not exported, which allows you to define components that are consumed only within the same file.
+This ensures there is only one entry point for a React component without limiting the structural content of the file.
 
 Examples of **correct** code for this rule:
 
@@ -116,7 +117,6 @@ function HelloAgain(props) {
 }
 module.exports = {Hello, HelloAgain}
 ```
-
 
 ## When Not To Use It
 
--- a/lib/rules/no-multi-comp.js
+++ b/lib/rules/no-multi-comp.js
@@ -17,6 +17,7 @@ const report = require('../util/report');
 
 const messages = {
   onlyOneComponent: 'Declare only one React component per file',
+  onlyOneExportedComponent: 'Declare only one exported React component per file',
 };
 
 /** @type {import('eslint').Rule.RuleModule} */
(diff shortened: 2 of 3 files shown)


docs/rules/no-multi-comp.md
```diff
@@ -69,6 +69,55 @@ class HelloJohn extends React.Component {
 module.exports = HelloJohn;
 ```
 
+### `ignorePrivate`
```
ljharb, line 72: "private" might not be the best name here. maybe "ignoreNonExported"?

docs/rules/no-multi-comp.md
```diff
@@ -69,6 +69,55 @@ class HelloJohn extends React.Component {
...
+```
+
+```jsx
+function Hello(props) {
+  return <div>Hello {props.name}</div>;
+}
+function HelloAgain(props) {
+  return <div>Hello again {props.name}</div>;
+}
+module.exports = {Hello, HelloAgain}
```
ljharb, line 118:
```suggestion
module.exports = { Hello, HelloAgain };
```

lib/rules/no-multi-comp.js
```diff
@@ -46,6 +51,43 @@ module.exports = {
   create: Components.detect((context, components, utils) => {
     const configuration = context.options[0] || {};
     const ignoreStateless = configuration.ignoreStateless || false;
+    const ignorePrivate = configuration.ignorePrivate || false;
+
+    const exportedComponents = new Set(); // Track exported components
+    const validIdentifiers = new Set(['ArrowFunctionExpression', 'Identifier', 'FunctionExpression']);
```
ljharb, line 57: i believe in react 19, async functions can be components as well.

lib/rules/no-multi-comp.js
```diff
@@ -46,6 +51,43 @@ module.exports = {
...
+    function getExportedComponentName(node) {
+      if (node.declaration.type === 'ClassDeclaration') {
+        return node.declaration.id.name;
+      }
+      for (const declarator of node.declaration.declarations || []) {
+        const type = declarator.init.type;
+        if (validIdentifiers.has(type)) {
+          return declarator.id.name;
+        }
+      }
```
ljharb, line 73:
```suggestion
      if (node.declaration.declarations) {
        const declarator = node.declaration.declarations.find((declarator) => validIdentifiers.has(declarator.init.type));
        if (declarator) {
          return declarator.id.name;
        }
      }
```

lib/rules/no-multi-comp.js
```diff
@@ -61,21 +103,46 @@ module.exports = {
...
+
+    const rule = {
       'Program:exit'() {
         if (components.length() <= 1) {
           return;
         }
 
         values(components.list())
           .filter((component) => !isIgnored(component))
+          .filter((component) => !isPrivate(component))
```
ljharb, line 123:
```suggestion
          .filter((component) => !isIgnored(component) && !isPrivate(component))
```

lib/rules/no-multi-comp.js
```diff
@@ -61,21 +103,46 @@ module.exports = {
...
+
+    if (ignorePrivate) {
+      rule.ExportNamedDeclaration = (node) => {
+        exportedComponents.add(getExportedComponentName(node));
+      };
+
+      rule.ExportDefaultDeclaration = (node) => {
+        exportedComponents.add(getExportedComponentName(node));
+      };
+    }
```
ljharb, line 144: instead of mutating maybe we could use Object.assign here?

lib/rules/no-multi-comp.js
vwillyams: Yeah good point, I had not thought about RSCs.

lib/rules/no-multi-comp.js
vwillyams: Yeah sure, sounds good. I'm not clear on why but I'm not super experienced in the eslint lifecycle and I can imagine that mutation may introduce issues somehow.

docs/rules/no-multi-comp.md
vwillyams: `ignoreNonExported` feels a little clunky to me but I can't think of anything better. just one of those hard problems in software.

ljharb: yeah totally :-) "ignoreInternal"?

lib/rules/no-multi-comp.js
ljharb: it probably doesn't matter much at all, but mutation generally risks slowdowns later, as opposed to creating objects all at once.

lib/rules/no-multi-comp.js
```diff
@@ -61,21 +103,46 @@ module.exports = {
...
+
+    const rule = {
       'Program:exit'() {
         if (components.length() <= 1) {
           return;
         }
 
         values(components.list())
           .filter((component) => !isIgnored(component))
```
ljharb, line 122:
```suggestion
```

vwillyams: ah yeah, the commit suggestion button resulted in some unhelpful commits. I'll resolve these manually.

docs/rules/no-multi-comp.md
vwillyams: Sounds good to me, switched over.

lib/rules/no-multi-comp.js
vwillyams: Added a little coverage for this - I'm not an RSC dev by day though so may have made some oversights.

lib/rules/no-multi-comp.js
vwillyams: Switched over.

vwillyams:
FYI I'm currently iterating on the test coverage and making sure this works in all cases. I didn't entirely expect the amount of complexity that would go into this feature! But it's been a fun learning process.

tests/lib/rules/no-multi-comp.js
```diff
@@ -22,6 +22,178 @@ const parserOptions = {
...
+  (compOne, compOneName, compTwo, compTwoName) => `
+    export default ${compOne(compOneName)}
+    export ${compTwo(compTwoName)}`, // default export
+];
+
+/**
+ * @param {ComponentExportGenerator[]} scenarioArray array of scenario generator functions which we will now convert into strings
+ * @param {ComponentGenerator[]} componentTypes array of components to generate on
+ * @param {boolean} [invalid] whether generated scenarios should expect to fail
+ * @returns {{code: string, options: object[], errors: object[]}[]}
```
ljharb: **Syntax error**: `module.exports = { ${compOneName} ${compTwoName} }` is missing a comma between the property shorthand names. Should be:
```js
module.exports = { ${compOneName}, ${compTwoName} }
```
This causes a parse error at runtime (3 test failures).

tests/lib/rules/no-multi-comp.js
```diff
@@ -22,6 +22,178 @@ const parserOptions = {
   },
 };
 
+// ------------------------------------------------------------------------------
+// Combinatorial test generation for ignoreInvalid
+// ------------------------------------------------------------------------------
+// eslint-disable-next-line valid-jsdoc
+/**
+ * @typedef {Function} ComponentGenerator
```
ljharb: The combinatorial test generation approach generates hundreds of tests, many of which pass or fail for the wrong reasons (because the rule doesn't detect the export form being tested). I'd suggest replacing this with simple, explicit test cases — like every other rule in this repo — so each test clearly documents what it's testing and can be individually debugged. The current approach makes it very hard to tell which combinations actually work vs. which silently pass by accident.

ljharb (changes requested):
I ran the tests locally — **771 out of 1545 tests fail**. The issues fall into a few categories:

1. **Crash: `TypeError: Cannot read properties of null (reading 'type')`** (105 failures) — `export { Foo }` has `node.declaration === null`, so `getExportedComponentName` crashes on `node.declaration.type`.
2. **Crash: `TypeError: Cannot read properties of null (reading 'name')`** (108 failures) — `export default function() {}` has `node.declaration.id === null`, so `getExportedComponentName` crashes on `node.declaration.id.name`. Similarly, `findComponentIdentifierFromComponent` can crash when a component has no `id` (e.g. anonymous arrow or memo-wrapped functions).
3. **Silent failures: 555 invalid tests produce no errors** — many export forms are not detected, so components that should be "exported" are treated as private and silently ignored.
4. **Parse error** (3 failures) — `module.exports = { ComponentOne() {} ComponentTwo() {} }` is missing a comma between the methods.

Detailed issues are in the inline comments below. The core problem is that `getExportedComponentName` only handles a few specific AST shapes and silently returns `undefined` for the rest, which causes false negatives instead of visible failures.

ljharb (changes requested):
added some review comments

PR #3847, opened 2024-10-30
feat: remove dependencies for Node built-ins

MichaelDeBoey:


package.json
```diff
@@ -29,24 +29,15 @@
   "homepage": "https://github.com/jsx-eslint/eslint-plugin-react",
   "bugs": "https://github.com/jsx-eslint/eslint-plugin-react/issues",
   "dependencies": {
-    "array-includes": "^3.1.8",
-    "array.prototype.findlast": "^1.2.5",
-    "array.prototype.flatmap": "^1.3.2",
     "array.prototype.tosorted": "^1.1.4",
```
MichaelDeBoey, line 32: Unfortunately `array.prototype.toSorted` is only available as of Node v20 😢

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted#browser_compatibility

Masoud9215 (approved):

PR #3849, opened 2024-11-04
[Tests] `jsx-no-script-url`: Improve tests

radu2147:
While developing [this](https://github.com/oxc-project/oxc/pull/6995), I noticed that test files for this rule don't explicitly cover the case when a non `linkAttribute`/`props` from a link component has the `"javascript:"` value. The rule seems to behave correctly for that case, so what I did in this PR is just to add a few tests for that.

### DIFF
--- a/tests/lib/rules/jsx-no-script-url.js
+++ b/tests/lib/rules/jsx-no-script-url.js
@@ -39,6 +39,10 @@ ruleTester.run('jsx-no-script-url', rule, {
     { code: '<Foo href="javascript:"></Foo>' },
     { code: '<a href />' },
     {
+      code: '<Foo other="javascript:"></Foo>',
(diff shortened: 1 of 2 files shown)


tests/lib/rules/jsx-no-script-url.js
```diff
@@ -117,10 +123,7 @@ ruleTester.run('jsx-no-script-url', rule, {
         <Bar link="javascript:"></Bar>
       </div>
     `,
-      errors: [
-        { messageId: 'noScriptURL' },
-        { messageId: 'noScriptURL' },
-      ],
+      errors: [{ messageId: 'noScriptURL' }, { messageId: 'noScriptURL' }],
```
ljharb, line 126: please revert unrelated formatting changes

ljharb (approved):

PR #3865, opened 2024-12-10
[Fix] no-unknown-property: support `onBeforeToggle`, `popoverTarget`, `popoverTargetAction` attributes

### LINKED ISSUE #3864: [Bug]: Unknown property 'onBeforeToggle' found (react/no-unknown-property)

### Description Overview

this is a follow-up to #3707. the release of React v19 stable means that support for the new `onBeforeToggle` event prop, as well as the camelCased popover attributes `popoverTarget` and `popoverTargetAction`, has now landed in the current latest stable release (they were added [in this PR](https://github.com/facebook/react/pull/27981)).

here’s an example trigger + popover:

```tsx
<>
    <button popoverTarget="foo" popoverTargetAction="show">
        Open Popover
    </button>
    <div
        id="foo"
        onBeforeToggle={event => console.log('beforeToggle from', event.oldState, 'to', event.newState)}
        popover="auto"
    >
        <h1>Popover Foo</h1>
    </div>
</>
```

example errors:
```
  68:13  error  Unknown property 'onBeforeToggle' found                                          react/no-unknown-property
  78:17  error  Unknown property 'popoverTarget' found, use 'popovertarget' instead              react/no-unknown-property
  79:17  error  Unknown property 'popoverTargetAction' found, use 'popovertargetaction' instead  react/no-unknown-property
```

these occur when running my `lint` npm run script: `"lint": "eslint .",`

### Expected Behavior

i expect the new event, `onBeforeToggle` and the camelCased attributes to not be considered “unknown” properties

v7.37.2

### eslint version

v9.16.0

### node version

v22.7.0

acusti:
Fixes #3864

### DIFF
--- a/tests/lib/rules/no-unknown-property.js
+++ b/tests/lib/rules/no-unknown-property.js
@@ -188,6 +188,15 @@ ruleTester.run('no-unknown-property', rule, {
         </div>
       `,
     },
+    {
+      code: `
+        <div>
+          <button popoverTarget="my-popover" popoverTargetAction="toggle">Open Popover</button>
+
+          <div id="my-popover" onBeforeToggle={this.onBeforeToggle} popover>Greetings, one and all!</div>
+        </div>
+      `,
+    },
   ]),
   invalid: parsers.all([
     {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ## Unreleased
 
+### Fixed
+* [`no-unknown-property`]: support `onBeforeToggle`, `popoverTarget`, `popoverTargetAction` attributes ([#3865][] @acusti)
+
+[#3865]: https://github.com/jsx-eslint/eslint-plugin-react/pull/3865
+
 ## [7.37.3] - 2024.12.23
 
 ### Fixed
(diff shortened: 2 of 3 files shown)


tests/lib/rules/no-unknown-property.js
```diff
@@ -182,11 +182,14 @@ ruleTester.run('no-unknown-property', rule, {
...
-          <button popovertarget="my-popover" popovertargetaction="toggle">Open Popover</button>
+          <button popoverTarget="my-popover" popoverTargetAction="toggle">Open Popover</button>
 
-          <div popover id="my-popover">Greetings, one and all!</div>
+          <div id="my-popover" onBeforeToggle={this.onBeforeToggle} popover>Greetings, one and all!</div>
         </div>
       `,
+      settings: {
+        react: { version: '19.0.0' },
+      },
```
ljharb, line 192: it seems like this test passed previously - instead of changing the existing one, can we add a new test?

ljharb (changes requested):

tests/lib/rules/no-unknown-property.js
acusti: @ljharb yeah that’s fair. just to confirm, however, do you want the plugin to allow both the all-lowercase version and the camelCased version of those attributes? before React added proper support in facebook/react#27981, the only way to use the attributes without getting console warning from react was to use the all-lowercase version of the attribute names. but it seems like now that the support has shipped in latest stable, it would be better to ensure that people always use the officially supported versions of the attributes, which is why i removed the all-lowercase versions of them (meaning that my changes make the original version of the test no longer pass).

i’m happy to do whichever seems best to you. there’s probably a semver consideration here also. i could even pretty easily redo the implementation so that for react >= 19, we replace the `popovertarget` and `popovertargetaction` attributes with the camel-cased version and then i can have test cases for both.

ljharb: if it already supported the all-lowercase one, then yes, it'd be semver-major to stop supporting it.

ljharb (approved):

ljharb:
Tests are now failing, it seems.

tests/lib/rules/no-unknown-property.js
acusti: @ljharb i updated the PR to add support for `onBeforeToggle`, `popoverTarget`, and `popoverTargetAction`, all as newly supported properties, and preserved the all-lowercase support and test case.

acusti:
@ljharb i added `popoverTarget` and `popoverTargetAction` to the `DOM_PROPERTIES_IGNORE_CASE` array of strings, which seems to be the way to have a DOM property that can be used both lowercased and camel-cased.

vitalyiegorov:
@ljharb Do you plan to merge/release this?

ljharb:
@vitalyiegorov yes, that's why i approved it.

PR #3879, opened 2025-01-14
fix(types): fix types of flat configs

### LINKED ISSUE #3878: [Bug]: flat configs are possibly undefined

### Description Overview

This is a follow up of https://github.com/jsx-eslint/eslint-plugin-react/issues/3874, since `flat` is `Record<string, ReactFlatConfig>`, `flat.xxx` or `flat["xxx"]` could be undefined. The fix might be explicitly list `recommended`, `all` and `jsx-runtime` as the key of the Record instead of using `string`.

Image

```javascript
import tseslint from 'typescript-eslint'
import reactPlugin from 'eslint-plugin-react'

export default tseslint.config(
  reactPlugin.configs.flat.recommended
)
```

```
Argument of type 'ReactFlatConfig | undefined' is not assignable to parameter of type 'InfiniteDepthConfigWithExtends'.
```

### Expected Behavior

No type error

v7.37.4

### eslint version

9.17.0

### node version

v22.13.0

CHC383:
Fixes #3878

### DIFF
--- a/index.js
+++ b/index.js
@@ -27,13 +27,13 @@ const deprecatedRules = filterRules(allRules, (rule) => rule.meta.deprecated);
 
 /** @type {['react']} */
 // for legacy config system
-const plugins = [
-  'react',
-];
+const plugins = ['react'];
 
 // TODO: with TS 4.5+, inline this
 const SEVERITY_ERROR = /** @type {2} */ (2);
 const SEVERITY_OFF = /** @type {0} */ (0);
+
+/** @typedef {'recommended' | 'all' | 'jsx-runtime'} AvailableFlatConfigs */
 
 const configs = {
   recommended: {
@@ -90,21 +90,19 @@ const configs = {
       'react/jsx-uses-react': SEVERITY_OFF,
     },
   },
-  flat: /** @type {Record<string, ReactFlatConfig>} */ ({
-    __proto__: null,
-  }),
+  flat: /** @type {{Record<AvailableFlatConfigs, ReactFlatConfig>}} */ ({}),
 };
 
-/** @typedef {{ plugins: { react: typeof plugin }, rules: import('eslint').Linter.RulesRecord, languageOptions: { parserOptions: import('eslint').Linter.ParserOptions } }} ReactFlatConfig */
+/** @typedef {{plugins: {react: typeof plugin}, rules: import('eslint').Linter.RulesRecord, languageOptions: {parserOptions: import('eslint').Linter.ParserOptions}}} ReactFlatConfig */
 
-/** @type {{ deprecatedRules: typeof deprecatedRules, rules: typeof allRules, configs: typeof configs & { flat: Record<string, ReactFlatConfig> }}} */
+/** @type {{deprecatedRules: typeof deprecatedRules, rules: typeof allRules, configs: typeof configs}}} */
 const plugin = {
   deprecatedRules,
   rules: allRules,
   configs,
 };
 
-Object.assign(configs.flat, {
+configs.flat = {
   recommended: {
     plugins: { react: plugin },
     rules: configs.recommended.rules,
@@ -120,6 +118,6 @@ Object.assign(configs.flat, {
     rules: configs['jsx-runtime'].rules,
     languageOptions: { parserOptions: configs['jsx-runtime'].parserOptions },
   },
-});
+};
 
 module.exports = plugin;
--- a/index.js
+++ b/index.js
@@ -90,7 +90,7 @@ const configs = {
       'react/jsx-uses-react': SEVERITY_OFF,
     },
   },
-  flat: /** @type {{Record<AvailableFlatConfigs, ReactFlatConfig>}} */ ({}),
+  flat: /** @type {{ [k in AvailableFlatConfigs]: ReactFlatConfig }} */ ({}),
 };
 
 /** @typedef {{plugins: {react: typeof plugin}, rules: import('eslint').Linter.RulesRecord, languageOptions: {parserOptions: import('eslint').Linter.ParserOptions}}} ReactFlatConfig */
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -90,14 +90,14 @@ const configs = {
       'react/jsx-uses-react': SEVERITY_OFF,
     },
   },
-  flat: /** @type {Record<string, ReactFlatConfig>} */ ({
+  flat: /** @type {Record<"recommended"|"all"|"jsx-runtime", ReactFlatConfig>} */ ({
```
ljharb, line 93:
```suggestion
  flat: /** @type {Record<'recommended' | 'all' | 'jsx-runtime', ReactFlatConfig>} */ ({
```
we should probably define this union as its own type.

CHC383: I did some refactoring on the types, but I haven't figured out a good way to type `flat` first and assign its value later. Maybe using dummy values is an option but I haven't tried to construct such a complex object yet, and type assertion on an empty object is not available in js file.

Are you open to move flat config to a separate object instead of nesting it inside the original `configs` object? However this is a breaking change and I do want to avoid it if there are other options. And just curious, what was the reason that the new flat config not being added as a new object?

ljharb: I just meant, `/** @typedef {'recommended' | 'all' | 'jsx-runtime'} AvailableFlatConfigs */` or similar.

A breaking change is not just something to avoid, it's simply not an option. Flat configs are nested by design and that will remain the case.

I'm not sure what you mean by "being added as a new object" - the "configs" key is the reasonable place to put all configs, flat or not.

CHC383: > I just meant, /** @typedef {'recommended' | 'all' | 'jsx-runtime'} AvailableFlatConfigs */ or similar.

Yeah I get that, but I got some errors while running `npm run build-types` and started to type things more explicitly than using `typeof`, now I have switched back to the minimum changes as I did at the beginning and try to make it work. So far typing the key as `AvailableFlatConfigs` make `configs` with `flat` incompatible with the [eslint `configs` type](https://github.com/eslint/eslint/blob/main/lib/types/index.d.ts#L1462) in the test-published-types step, but the package itself builds and packs without problem.

What I meant by adding a new object is that `flatConfig` itself being an individual object and can be accessed through `reactPlugin.flatConfigs`, rather than putting it inside a legacy configs object. Some plugins did it that way such as [eslint-plugin-jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/index.js#L311-L318). I had this thought as I was trying to avoid the circular type reference on the legacy config and the flat config

CHC383: Now I have a better understanding of the problem: In the [test-published-types](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/test-published-types/index.js#L9) test, `react` is supposed to be `Record<string, ESLint.Plugin>` as defined [here](https://github.com/eslint/eslint/blob/main/lib/types/index.d.ts#L1304), the [legacy configs](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/index.js#L39-L92) in `configs` fits that signature, but the new `flat` config object we put inside doesn't, this is another reason to not nested the flat config inside the legacy config object.

Maybe there is a hacky way to bypass/force this to work? Otherwise I am not sure how to type the flat config object properly while keeping the current placement, which on the one hand is a reasonable place to put configs, on the other hand doesn't comply with eslint types.

ljharb: Instead of Record, try `{ [k in AvailableFlatConfigs]: ReactFlatConfig }`?

ljharb: ESLint's own types don't dictate what we do - only its runtime behavior does. If there's something the runtime allows that the types do not, it's a bug in their types, and it should be filed as such.

CHC383: Thanks, I have updated the PR with my latest changes for future reference, it still has the typing issue but I am not sure how to proceed. There could be a few options based on our conversations so far:

1. Ask ESLint to loose their type definition for the [`plugins`](https://github.com/eslint/eslint/blob/main/lib/types/index.d.ts#L1304) in `Config` to take `Record<string, any>` so that plugins can put any object into the value. Personally I don't think this make sense from ESLint's typing and design perspective even though you explained about only the runtime behavior matters
2. Make the type of the React plugin comply with the ESLint's expectation. This could be one of the following:
    - Put legacy config into its own object and use `configs.legacy` or `configs.flat` explicitly
        ```javascript
        configs = {
            legacy: {...}
            flat: {...}
        }
        ```
    - Move flat config into a separate object
        ```javascript
        configs = {...} // legacy
        flatConfigs = {...}
        }
        ```
    However, both of them are breaking changes. Even though I don't like them as much as you do, I am not sure if putting the new flat config into the legacy config was a right decision that led to the situation today. Both [eslint-plugin-import ](https://github.com/ljharb/eslint-plugin-import) and [eslint-plugin-jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y) don't nest the flat configs in this way
3. Merge the change as is and update [test-published-types](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/test-published-types/index.js#L9) somehow to allow this. If test-published-types doesn't reflect what the actual user does, test it this way doesn't help. If it does, either 1 or 2 should happen.

Please feel free to share other potential options or take over the PR, I am just going to share a workaround in the original issue and move forward unless there is a viable option at the moment.

CHC383: To test the typing issue locally

```bash
// in eslint-plugin-react
npm pack --pack-destination /tmp/

// in eslint-plugin-react/test-published-types
// Use any TS version in https://github.com/jsx-eslint/eslint-plugin-react/actions/runs/12776433146/job/35615079535?pr=3879
npm install
npm install --no-save /tmp/eslint-plugin-react-7.37.4.tgz typescript@4.0
// Use any lib version in the action above
npx tsc --lib es2015
```

PR #3881, opened 2025-01-16
[New] `no-unused-state`: Support ignore option

fwgood:
This is adding a new property `ignore` to  ignore specific state.

### DIFF
--- a/docs/rules/no-unused-state.md
+++ b/docs/rules/no-unused-state.md
@@ -45,3 +45,16 @@ var UnusedGetInitialStateTest = createReactClass({
   }
 })
 ```
+
+## Rule Options
+
+This rule can take one argument to ignore some specific states during validation.
+
+```js
+...
+"react/no-unused-state": [<enabled>, { ignore: <ignore> }]
+...
+```
+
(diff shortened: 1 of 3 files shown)


ljharb:
Sorry this fell through the cracks. Can you elaborate a bit more on the problem this will solve?

fwgood:
> Sorry this fell through the cracks. Can you elaborate a bit more on the problem this will solve?

@ljharb I'm very glad that you can reopen this PR.

In some React frameworks that use JSX, certain specific states are defined for internal rendering purposes (such as marking rendering stages, measuring performance, etc.). There is a similar usage pattern where the state is not consumed by the developers but is consumed by the framework internally. In this case, an ignore rule is needed (the usage of links uses hooks, which is different from class-based states, but the principle is similar). [https://lynxjs.org/guide/performance/evaluating-performance.html](url)

ljharb:
I'm not familiar with Lynx but this is eslint-plugin-react - it's intended for use with a react codebase, not any jsx codebase.

Unfortunately that link doesn't explain the use case to me - but it doesn't make sense to me that you'd declare state that you don't personally set - if the framework is assigning state, it should be declaring it.

PR #3882, opened 2025-01-25
fix(types): fix types of flat configs

### LINKED ISSUE #3878: [Bug]: flat configs are possibly undefined

CHC383:
Fixes #3878

This is another attempt to fix #3878 from a different direction than https://github.com/jsx-eslint/eslint-plugin-react/pull/3879

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,9 +209,9 @@ The schema of the `settings.react` object would be identical to that of what's a
 
 This plugin exports 3 flat configs:
 
-- `flat.all`
-- `flat.recommended`
-- `flat['jsx-runtime']`
+- `flatConfigs.all`
+- `flatConfigs.recommended`
+- `flatConfigs['jsx-runtime']`
 
 The flat configs are available via the root plugin import. They will configure the plugin under the `react/` namespace and enable JSX in [`languageOptions.parserOptions`](https://eslint.org/docs/latest/use/configure/language-options#specifying-parser-options).
 
@@ -220,8 +220,8 @@ const reactPlugin = require('eslint-plugin-react');
 
 module.exports = [
   …
-  reactPlugin.configs.flat.recommended, // This is not a plugin object, but a shareable config object
-  reactPlugin.configs.flat['jsx-runtime'], // Add this if you are using React 17+
+  reactPlugin.flatConfigs.recommended, // This is not a plugin object, but a shareable config object
+  reactPlugin.flatConfigs['jsx-runtime'], // Add this if you are using React 17+
   …
 ];
 ```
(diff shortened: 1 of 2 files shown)


CHC383:
@ljharb Since #3879 is blocked, I tried to fix the typing issue from a different angle: creating a new flatConfigs object outside of the original `configs`. This eventually will become a breaking change but not at the moment, as the original `configs.flat` is still kept and only marked as deprecated through JSDoc.

If we run out of ideas on #3879, maybe we can go with this one?

ljharb:
I agree this is semver-minor, but I have no desire to go with that model. I want a single configs object with eslintrc configs, and a `flat` on it that holds all the flat configs. If the only unsolvable issue with that is that you might need to use `NonNullable<>` or `!`, I think that's acceptable (i don't think it's unsolvable, to be clear, but i'm not willing to use what i think is a worse object layout)

CHC383:
> I agree this is semver-minor, but I have no desire to go with that model. I want a single configs object with eslintrc configs, and a `flat` on it that holds all the flat configs. If the only unsolvable issue with that is that you might need to use `NonNullable<>` or `!`, I think that's acceptable (i don't think it's unsolvable, to be clear, but i'm not willing to use what i think is a worse object layout)

Yeah you are right, I just checked and found out ESLint supports `ts` config since v9.9.0, before that `NonNullable<>` or `!` are not an option, I guess it is not a big deal then compared to using `// @ts-expect-error` in the `js` config.

ljharb:
It's a fair point that they're not all the same; this, however, is what i would have preferred to do on the other two, i just didn't notice the difference.

Prefixing means you can't programmatically compare them, for example - munging string names isn't clean at all. A separate object, at least, is more reasonable, but `configs` is the only property under which i'd expect to find configs.

PR #3910, opened 2025-03-23
[Fix] `jsx-no-constructed-context-values`: detect constructed context values in React 19 `<Context>` usage

### LINKED ISSUE #3888: [Bug]: `jsx-no-constructed-context-values` doesn't detect issues when context is used as provider in react v19

### Description Overview

In React 19, you can render `<Context>` as a provider instead of `<Context.Provider>`.

However when doing so `jsx-no-constructed-context-values` doesn't detect violations.

```tsx
const MyContext = createContext();

function MyComponent(props) {
  return (<>
    {/* old style - violation detected */ }
    <MyContext.Provider value={{ hello: 'world' }} />
    
    {/* react 19 style - no violation */ }
    <MyContext value={{ hello: 'world' }} />
  </>);
}

```

v7.37.2

### eslint version

v8.57.1

### node version

v18.20.6

TildaDares:
Fixes #3888 

This PR updates the `jsx-no-constructed-context-values` rule to support React 19-style context providers using the <Context> shorthand.

### DIFF
--- a/docs/rules/jsx-no-constructed-context-values.md
+++ b/docs/rules/jsx-no-constructed-context-values.md
@@ -22,6 +22,16 @@ return (
 )
 ```
 
+```jsx
+import React from 'react';
+
+const MyContext = React.createContext();
+function Component() {
+    function foo() {}
+    return (<MyContext value={foo}></MyContext>);
+}
+```
+
 Examples of **correct** code for this rule:
 
 ```jsx
@@ -31,6 +41,11 @@ return (
         ...
     </SomeContext.Provider>
 )
+```
+
+```jsx
+const SomeContext = createContext();
+const Component = () => <SomeContext value="Some string"><SomeContext>;
 ```
 
 ## Legitimate Uses
--- a/lib/rules/jsx-no-constructed-context-values.js
+++ b/lib/rules/jsx-no-constructed-context-values.js
@@ -119,6 +119,45 @@ function isConstruction(node, callScope) {
   }
 }
 
+function isReactContext(context, node) {
+  let scope = getScope(context, node);
+  let variableScoping = null;
+  const contextName = node.name;
+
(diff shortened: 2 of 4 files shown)


ljharb (changes requested):
there's an uncovered line

TildaDares:
The codecov check is stuck, so I can't confirm if the coverage has improved. Is there a way to rerun it?

ljharb:
I had to rebase it anyways, so no worries :-) hopefully it won't be stuck this time. Thanks!

ljharb (approved):

lib/rules/jsx-no-constructed-context-values.js
```diff
@@ -119,6 +119,45 @@ function isConstruction(node, callScope) {
...
+
+  if (!variableScoping) { // Context was not found in scope
+    return false;
+  }
+
+  // Get the variable's definition
+  const def = variableScoping.defs[0];
+
+  if (!def || def.node.type !== 'VariableDeclarator') {
+    return false;
```
ljharb, line 140: looks like we need a test for this if check

TildaDares: Resolved!

ljharb: hm, seems like there's still no coverage, but maybe i rebased it out. can you take another look?

docs/rules/jsx-no-constructed-context-values.md
```diff
@@ -33,6 +43,11 @@ return (
 )
 ```
 
+```jsx
+const SomeContext = createContext();
+const Component = () => <SomeContext value="Some string"><SomeContext>;
```
jens-duttke, line 48: There's a slash missing in closing `<SomeContext>` -> `</SomeContext>`.

PR #3923, opened 2025-05-05
[Fix] `no-unknown-property`: allow `onLoad` on `body`

DerekStapleton:
They body tag supports an onload event

https://www.w3schools.com/jsref/event_onload.asp


DerekStapleton:
Another, probably better, source: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/load_event

ljharb (changes requested):
thanks! this will need some test changes

ljharb (approved):

PR #3926, opened 2025-05-22
[Fix] `display-name`: avoid false positive when React is shadowed

### LINKED ISSUE #3924: [Bug]: display-name false positive when React is shadowed

### Description Overview

`react/display-name` will report an error `Component definition is missing display name` when the `React` global import is shadowed within a function scope.

The same error does not occur for shadowed `memo` or `forwardRef` though.

#### Example

```jsx
import React, { memo, forwardRef } from 'react'

const MixedShadowed = function () {
  const memo = (cb) => cb()
  const { forwardRef } = { forwardRef: () => null }
  const [React] = [{ memo, forwardRef }]

  const Comp = memo(() => {
    return <div>shadowed</div>
  })
  const ReactMemo = React.memo(() => null)
  const ReactForward = React.forwardRef((props, ref) => {
    return `${props} ${ref}`
  })
  const OtherComp = forwardRef((props, ref) => `${props} ${ref}`)

  return [Comp, ReactMemo, ReactForward, OtherComp]
}
```

Running eslint from the command line you'll see the error:

```
  11:21  error  Component definition is missing display name  react/display-name
  12:24  error  Component definition is missing display name  react/display-name
```

### Expected Behavior

I know most people won't shadow `React`, `memo` or `forwardRef`, but I'd still expect the error to not be reported in this case.

v7.37.5

### eslint version

v9.26.0

### node version

v22.15.0

hyeonbinHur:
Fixes #3924

### DIFF
--- a/lib/rules/display-name.js
+++ b/lib/rules/display-name.js
@@ -161,6 +161,123 @@ module.exports = {
       return false;
     }
 
+    function hasVariableDeclaration(node, name) {
+      if (!node) return false;
+
+      if (node.type === 'VariableDeclaration') {
+        return node.declarations.some((decl) => {
+          if (!decl.id) return false;
+
+          // const name = ...
+          if (decl.id.type === 'Identifier' && decl.id.name === name) {
+            return true;
+          }
+
+          // const [name] = ...
+          if (decl.id.type === 'ArrayPattern') {
+            return decl.id.elements.some(
+              (el) => el && el.type === 'Identifier' && el.name === name
+            );
+          }
+
+          // const { name } = ...
+          if (decl.id.type === 'ObjectPattern') {
+            return decl.id.properties.some(
+              (prop) => prop.type === 'Property' && prop.key && prop.key.name === name
+            );
+          }
+
+          return false;
+        });
+      }
+
+      if (node.type === 'BlockStatement' && node.body) {
+        return node.body.some((stmt) => hasVariableDeclaration(stmt, name));
+      }
+
+      return false;
+    }
+
+    function isIdentifierShadowed(node, identifierName) {
+      let currentNode = node;
+
+      while (currentNode && currentNode.parent) {
+        currentNode = currentNode.parent;
+
+        if (
+          currentNode.type === 'FunctionDeclaration'
+          || currentNode.type === 'FunctionExpression'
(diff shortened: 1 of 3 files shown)


lib/rules/display-name.js
```diff
@@ -31,32 +31,36 @@ const messages = {
 module.exports = {
   meta: {
     docs: {
-      description: 'Disallow missing displayName in a React component definition',
+      description:
+        'Disallow missing displayName in a React component definition',
```
ljharb, line 35: please don't run prettier on projects that don't use it :-) and please revert all unrelated style changes.

ljharb (changes requested):
can't review til style diffs are reverted

lib/rules/display-name.js
hyeonbinHur: Sorry, I forgot to disable some style-related settings, which caused unnecessary formatting changes. I’ll remove the unnecessary style edits and commit again soon. Thanks for the review and the feedback!

ljharb: My suggestion is to keep those always disabled, and only use eslint to format files.

morganney: Except [eslint recommends to use formatters instead](https://eslint.org/blog/2023/10/deprecating-formatting-rules/#what-you-should-do-instead).

lib/rules/display-name.js
```diff
@@ -161,6 +161,89 @@ module.exports = {
...
+    }
+
+    function isIdentifierShadowed(node, identifierName) {
+      let currentNode = node;
+      while (currentNode && currentNode.parent) {
+        currentNode = currentNode.parent;
+        if (
+          currentNode.type === 'FunctionDeclaration'
+          || currentNode.type === 'FunctionExpression'
+          || currentNode.type === 'ArrowFunctionExpression'
```
morganney, line 208: Why only these scopes? For example what about `BlockStatement`:

```
{
  const memo = (cb) => cb()
  const forwardRef = (cb) => cb()
  const React = { memo, forwardRef }

  const BlockReactShadowedMemo = React.memo(() => {
    return <div>shadowed</div>
  })
}
```

hyeonbinHur: Thank you for the review. Let me take a quick look and try to find a better algorithm to cover all test cases.

PR #3928, opened 2025-05-26
[Fix]: detect missing keys in return statement with ternary operator

### LINKED ISSUE #3925: [Bug]: react/jsx-key rule is not applied when using return statement and ternary operator in iterator

### Description Overview

**Brief description**

As described in the title, `react/jsx-key` is not applied when using return statement and ternary operator in iterator.
It's working when not using return statement as shown in my below picture.

**Picture showing the current issue**

Image

**Code to reproduce error**

Here is a small code snippet to reproduce the current error:

```jsx
import { Fragment } from 'react';

const ITEMS = ['bar', 'foo'];

export default function BugIssue() {
  return (
    <Fragment>
      {ITEMS.map((item) => {
        return item === 'bar' ? <div>{item}</div> : <span>{item}</span>;
      })}
    </Fragment>
  );
}
```

### Expected Behavior

An error should be displayed when using return statement and ternary operator.

=> `Missing "key" prop for element in iterator`

v7.37.3

### eslint version

9.27.0

### node version

v22.15.0

hyeonbinHur:
Fix #3925

### DIFF
--- a/lib/rules/jsx-key.js
+++ b/lib/rules/jsx-key.js
@@ -146,7 +146,20 @@ module.exports = {
           getReturnStatements(node.body)
             .filter((returnStatement) => returnStatement && returnStatement.argument)
             .forEach((returnStatement) => {
-              checkIteratorElement(returnStatement.argument);
+              const argument = returnStatement.argument;
+
+              if (argument.type === 'ConditionalExpression') {
+                const shouldCheckNode = (n) => n && (n.type === 'JSXElement' || n.type === 'JSXFragment');
+
+                if (shouldCheckNode(argument.consequent)) {
+                  checkIteratorElement(argument.consequent);
+                }
+                if (shouldCheckNode(argument.alternate)) {
+                  checkIteratorElement(argument.alternate);
+                }
+              } else {
+                checkIteratorElement(argument);
+              }
             });
         }
       }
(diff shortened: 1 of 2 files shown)


hyeonbinHur:
Invalid Test Cases (4)
Case | Scenario | AST Patterns Covered|
|--|--|--|
1: Basic Pattern | return + ternary in map with ArrowFunction | • ArrowFunction + BlockStatement + ReturnStatement <br/>• ConditionalExpression with JSX elements <br/> • Missing key detection in ternary branches
2: Function Expression | return + ternary in map with regular function | • FunctionExpression + BlockStatement + ReturnStatement <br/> • ConditionalExpression handling across function types <br/> • Same logic applies to function() syntax
3: Array.from Method | return + ternary in Array.from iteration | • Array.from + checkFunctionsBlockStatement call <br/> • ConditionalExpression in different iteration contexts <br/> • Multiple iteration method support
4: Fragment Context | return + ternary within Fragment (Original Issue) | • Fragment + map + return + ternary pattern <br/> • Real-world user scenario reproduction <br/>• Issue #3925 exact replication

Valid Test Cases (1)
Case | Scenario | Corresponding Invalid Case|
|--|--|--|
1: Proper Keys | return + ternary with correct key props | Invalid 1 | Ensures fix doesn't break valid code with keys|

PR #3929, opened 2025-06-02
[New] `jsx-props-no-multi-spaces`: Add automatic fix for no line gap between props

justisb:
Extends automatic fix support to cover the "no line gap" half of the `jsx-props-no-multi-spaces` rule. There is already automatic fix support for the "only one space" enforcement.

Initially tagging as a `[Fix]` because it was unexpected that the rule does not fix line gaps when the rule is stated to be automatically fixable in the docs.

### DIFF
--- a/lib/rules/jsx-props-no-multi-spaces.js
+++ b/lib/rules/jsx-props-no-multi-spaces.js
@@ -80,6 +80,22 @@ module.exports = {
             prop1: getPropName(prev),
             prop2: getPropName(node),
           },
+          fix(fixer) {
+            const comments = sourceCode.getCommentsBefore ? sourceCode.getCommentsBefore(node) : [];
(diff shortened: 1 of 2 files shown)


ljharb:
@justisb ah - because you made this PR from a repo you don't own, i can't rebase it properly (you should only be making PRs from a fork under your own username).

To fix it, can you add me to https://github.com/Popmenu/eslint-plugin-react temporarily? You're free to remove me after the PR is updated.

justisb:
Thanks @ljharb. I just reopened as https://github.com/jsx-eslint/eslint-plugin-react/pull/3930 with edits from maintainers enabled.

ljharb:
oof, that leaves this PR as an eternally undeleteable ref on the repo :-/

PR #3934, opened 2025-06-12
[Docs] edited the link to the badge last commit

Olexandr88:


ljharb:
why? this will be a strictly less useful view.

Olexandr88:
> why? this will be a strictly less useful view.
In my opinion, this new link is completely consistent with this badge.
Знімок екрана 2025-06-12 о 19 00 01
I can even give an example of projects using this very link.

ljharb:
I'm sure people do link to it, but a link to the last commit is already always visible above the file view, above the readme, so it's redundant, and those projects shouldn't be using it :-)

the pulse link also lists multiple recent commits, which makes it more useful

ljharb:
The issue with that badge isn't the image, it's that runkit is down. codeql status isn't interesting, but certainly if you have another free alternative to display a badge for the status of all github actions on a branch, that'd be great.

ljharb:
Right. I think you're fundamentally not understanding the scope of the problem here :-)

The badge image is perfect. GitHub's built-in status reporting is wildly insufficient. What's required is an _external service_, that's _free_, that can summarize the status of _every status check assigned to a given branch head_. It's not a trivial fix and in the absence of a complete fix, doing _anything_ is always worse than doing nothing.

Olexandr88:
That's all I can offer
https://badgen.net/github/checks/jsx-eslint/eslint-plugin-react

Olexandr88:
Hey @ljharb I need to close this request pool?

PR #3938, opened 2025-07-09
feat(jsx-key): check for array related method calls with jsx elements/fragments

### LINKED ISSUE #3937: [Bug]: react/jsx-key does not detect usage of `Array.push` etc.

### Description Overview

Tl;dr the lint rule doesn't pick up items that are inserted into an array via push/unshift etc.

### Expected Behavior

I've seen this pattern many times and I think we could reasonably assume that when the method name matches an array insertion method we can report an error.

This isn't foolproof without type information but I think this will be valuable.

Example:

```ts
const Comp = () => {
   const items = [];
   if (someCondition) {
      // this should cause an error/warning
      items.push(<span>test</span>);
   }
   if (someOtherCondition) {
      // same
      items.push(<span>test2</span>);
   }
   return <div>{items}</div>;
}
```

7.37.5

### eslint version

9.30.1

### node version

24

SimonSchick:
WIP, pending initial feedback, slightly worried this check may be overly broad.

Things to add:
- Setting to make this behavior opt-in unless we want it to be the default.
- Optimization of esquery usage? Seems potentially very wasteful.
- Check methods for exhaustiveness, there (fill, toSpliced) being more exotic candidates

Fixes #3937

### DIFF
--- a/lib/rules/jsx-key.js
+++ b/lib/rules/jsx-key.js
@@ -264,17 +264,14 @@ module.exports = {
       },
 
       // eslint-disable-next-line no-multi-str
-      'CallExpression[callee.type="MemberExpression"][callee.property.name=/^push|unshift|splice|with$/],\
-       CallExpression[callee.type="OptionalMemberExpression"][callee.property.name=/^push|unshift|splice|with$/],\
-       OptionalCallExpression[callee.type="MemberExpression"][callee.property.name=/^push|unshift|splice|with$/],\
-       OptionalCallExpression[callee.type="OptionalMemberExpression"][callee.property.name=/^push|unshift|splice|with$/]'(node) {
-        if (node.arguments.length === 0) {
-          return;
-        }
-
+      'CallExpression[callee.type="MemberExpression"][callee.property.name=/^push|unshift|splice|with|concat$/],\
+       CallExpression[callee.type="OptionalMemberExpression"][callee.property.name=/^push|unshift|splice|with|concat$/],\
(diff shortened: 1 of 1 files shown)


ljharb:
I'm not sure it's a good idea to assume any method named "push" or "splice" is happening on an array, unless it's a TS codebase.

SimonSchick:
That was kinda my thinking too but I'm unsure how much better we can do without types.

I thought about trying to trace the symbol we are pushing to and see if it is a array declaration but that feels very much like a hack.

On the other hand I'd assume `push`, `unshift` and `concat` are the most commonly used methods for this anti-pattern and these are not exactly common method names in in relation to jsx usage so false positives should be extremely rare.

Question it if it is rare enough to proceed with this vague targetting strategy?

I did some rough searching via https://github.com/search?type=code&q=%28path%3A*.tsx+OR+path%3A*.jsx%29+%2F.%2B%5C.push%5C%28%3C.%2B%5C%29%3B%2F and found nothing that stands out, had to vary the search query a fair bit since github code search seems to limit results to 100 for some reason.

ljharb:
`concat` exists on a number of things, and i agree false positives will be rare, but nonzero is nonzero.

Adding this change under an option is smart if it has a nonzero risk of false positives, though.

SimonSchick:
I'll add an option then, and also fix the failing tests for legacy versions (I really wish those get pruned 😢 ).

SimonSchick:
Don't have the spoons for this at this time, sry.

PR #3946, opened 2025-08-19
[New]: `forbid-dom-props`: support React Strict Dom elements (`html.div`, etc)

mdjastrzebski:
## Summary

Add suport for [React Strict DOM](https://facebook.github.io/react-strict-dom/) notation of DOM nodes: `html.div`, `html.span`, `h.h1`, etc

## Details

The code is based on `forbid-component-props` rule that already supported `Namespace.Component` notation.

## Testing

Added pass/fail tests for `html.div`, `h.span`, etc


ljharb (changes requested):
This seems like something that needs to be behind an option - i assume that the lowercase only works because of the `.`? (meaning that `const { div } = html; <div />` wouldn't work?

Masoud9215 (approved):

mdjastrzebski:
@ljharb re RSD: RSD is alternative approach for React Native + Web codesharing. RN Web took RN-first approach, while RSD takes. web-first approach.

re impl: Since RSD closely resembles regular DOM elements, I though that `forget-dom-props` is the right rule to modify. Alternatively we could modify `forget-component-props` but that would require dropping requirement for first uppercase letter in the name. Wdyt?

PR #3955, opened 2025-09-22
[Fix] `no-unused-prop-types`: detect used props in nested components

### LINKED ISSUE #3524: [Bug]: False positive on `react/no-unused-prop-types` with memo + forwardRef

### Description Overview

When combining `React.memo` with `React.forwardRef`, `react/no-unused-prop-types` gives false positive.

```ts
import { Image } from "react-native";

type Props = {
  foo: string;
};

const LazyImage = React.memo<Props>(
  React.forwardRef<Image, Props>(
    ({ foo }, ref) => {
      console.log("foo", foo);
      return <Image ref={ref} />;
    }
  )
);
```

### Expected Behavior

No error

v7.32.0

### eslint version

v8.31.0

### node version

v16.17.0

avaice:
Fixes #3524

This PR fixes a bug where `react/no-unused-prop-types` was not properly detecting unused props in components wrapped with nested wrapper functions like `React.memo(React.forwardRef())`.

- Enhanced component detection logic to properly traverse nested wrapper functions when looking for propType definitions
- Added logic to find the actual component node that declares propTypes when dealing with nested wrapper patterns
- The fix ensures that prop usage is correctly detected by finding the component that actually defines the propTypes, regardless of how many wrapper functions are nested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 * [`jsx-key`]: detect missing keys in return statement with ternary operator ([#3928][] @hyeonbinHur)
 * [`jsx-key`]: detect missing keys in logical expressions ([#3986][] @yalperg)
 * [`display-name`]: avoid false positive when React is shadowed ([#3926][] @hyeonbinHur)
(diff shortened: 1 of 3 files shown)


avaice:
@ljharb 
Thanks! I rebased and added untyped test cases.

Issue #3957, opened 2025-10-03
[Bug]: no-unknown-property does not recognize onScrollEnd, onScrollEndCapture events


xfeeefeee:
### Is there an existing issue for this?

- [x] I have searched the existing issues and my issue is unique
- [x] My issue appears in the command-line and not only in the text editor

### Description Overview

Adding an onScrollEnd or onScrollEndCapture event handler is not recognized as a known property.

this triggers the warning `Unknown property 'onScrollEnd' found`
```
<div onScrollEnd={handleScrollEnd}>
```

issue appears within lib/rules/no-unknown-property.js, `DOM_PROPERTY_NAMES_TWO_WORDS`

### Expected Behavior

this should not trigger the no-unknown-property rule:

```
<div onScrollEnd={handleScrollEnd} onScrollEndCapture={handleScrollEndCapture}>
```

### eslint-plugin-react version

v7.37.5

### eslint version

v8.57.1

### node version

v22.16.0

PR #3958, opened 2025-10-03
[fix] `no-unknown-property` - add onScrollEnd and onScrollEndCapture events as known properties

xfeeefeee:
This includes onScrollEnd and onScrollEnd capture as recognized elements.

While creating #3957 I realized i could just fix it.


ljharb (changes requested):
Thanks - this needs tests.

ljharb (approved):
thanks!

PR #3972, opened 2025-12-30
Update types and RuleContext deprecations

tylermumford:
## Summary

On `RuleContext` objects, use the `filename` property instead of the deprecated `getFilename()` method. That method was marked as deprecated in ESLint v8 and is removed in ESLint v10.

## Motivation

I'm preparing a company project for ESLint v10, and I get an error that `eslint-plugin-react` uses a method that doesn't exist on the `RuleContext` object.

(Hello, this is my first PR on this repo. I hope to be helpful and respectful.)

### DIFF
--- a/lib/rules/jsx-filename-extension.js
+++ b/lib/rules/jsx-filename-extension.js
@@ -61,7 +61,7 @@ module.exports = {
   },
 
   create(context) {
-    const filename = context.getFilename();
+    const filename = context.filename;
 
     let jsxNode;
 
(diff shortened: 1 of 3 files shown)


lib/rules/jsx-filename-extension.js
```diff
@@ -61,7 +61,7 @@ module.exports = {
   },
 
   create(context) {
-    const filename = context.getFilename();
+    const filename = context.filename;
```
ljharb: it's probably safer to do this since we still support much older eslint versions?
```suggestion
    const filename = context.filename || context.getFilename();
```

ljharb (changes requested):
It'd be interesting to also add tests that run in eslint 10.

plthomasva:
I believe these changes are incorporated in #4022. Thanks for getting the ball rolling.

PR #3981, opened 2026-02-10
[New] `jsx-key`: add suggestion to insert `key={TODO}` placeholder

mglst:
## Summary

The `jsx-key` rule correctly has no auto-fix — the right key is always context-dependent and cannot be inferred automatically. However, it also had no _suggestion_, which means editors offer no quick action at all on this error. This PR adds an editor suggestion that inserts a `key={TODO}` placeholder, giving developers a one-click jump to exactly the right place.

## Behaviour

The suggestion appears on `missingIterKey` and `missingArrayKey` reports. It inserts the placeholder before the first existing attribute if any are present, otherwise after the element name:

```jsx
// before
[1, 2, 3].map((x) => <App />);
[1, 2, 3].map((x) => <App foo={x} />);

// after applying suggestion
[1, 2, 3].map((x) => <App key={TODO} />);
[1, 2, 3].map((x) => <App key={TODO} foo={x} />);
```

The `UsePrag` variants (fragment shorthand) are excluded — those need an element type change, not a key insertion. `keyBeforeSpread` and `nonUniqueKeys` are also excluded since the key already exists in those cases.

## Design: why `key={TODO}`

Several placeholder values were considered:

`key={null}` / `key={undefined}` — rejected. Both are syntactically valid JSX and hasProp() returns true once the attribute exists regardless of value, so the lint error silently disappears after applying the suggestion. A developer bulk-applying suggestions across a file would end up with no errors and no indication anything still needs fixing.

`key={}` — the ideal choice on the surface: empty JSX attribute expressions cause a TypeScript error (ts(17000): JSX attributes must only be assigned a non-empty expression), forcing the developer to act. The problem is that JSXEmptyExpression is invalid in attribute position for all standard parsers (espree, @typescript-eslint/parser, babel), so RuleTester rejects it when validating suggestion output.

`key={TODO}` — chosen. TODO is a valid identifier (parseable by all parsers, so tests work), but it is undefined, which surfaces as both an ESLint no-undef violation and a TypeScript "Cannot find name 'TODO'" error. The suggestion cannot silently pass lint or a TS build. It is also unambiguously a placeholder.


mglst:
Agreed. Instead of `false`, how about `null` or `undefined`? Both are valid keys. In conjunction with this, I suggest adding a rule warning against using `null` and `undefined` as keys if they appear as literals values. That way if a user has a component missing a required `key` prop, the fix would add it with the right placeholder, making it valid code, but they'd still get warned that they probably want to actually define a sensible key.

ljharb:
That feels very overkill. I think in order to have a suggestion here, it indeed needs to result in invalid syntax, so the user is immediately forced to deal with it.

mglst:
The hard constraint is that ESLint 9's RuleTester requires parseable suggestion output, while invalid syntax (like `key={}`) fails espree's parser. I don't see a path forward for suggestions on this rule without upstream parser changes.

I've split out the `jsx-no-nullish-key` warning into [PR #3983](https://github.com/jsx-eslint/eslint-plugin-react/pull/3983) since it has independent value.

Some details on the constraint: ESLint 9.0.0 added [suggestion output validation](https://github.com/eslint/eslint/commit/47e60f85e) (`feat!: Stricter rule test validations #17654`). `RuleTester` calls `linter.verify()` on suggestion output and asserts no fatal parse error occurs. Suggesting `key={}` causes test failures under ESLint 9 with the default espree parser:

```
A fatal parsing error occurred in suggestion fix.
Error: Parsing error: JSX attributes must only be assigned a non-empty expression
```

The restriction lives in `acorn-jsx/index.js` (`jsx_parseAttributeValue`), which espree depends on. `@typescript-eslint/parser` *does* parse `key={}` without fatal error (ts(17000) is a type diagnostic, not parse error), so TS-parser test variants would pass — but espree variants fail, and that's the default for most users.

Perhaps acorn-jsx could relax this restriction upstream, but that's outside the scope of this PR.

PR #3996, opened 2026-03-27
fix(jsx-curly-brace-presence): do not flag empty string literals in JSX children

### LINKED ISSUE #3995: [Bug]: jsx-curly-brace-presence incorrectly removes `{""}` used for whitespace preservation

### Description Overview

In some scenarios where `{""}` is used for whitespace preservation, `jsx-curly-brace-presence` incorrectly flags the braces as unnecessary, and `--fix` introduces a visual regression.

```
  return (
    <span>
      <span>The braces</span>
      {""} matter here.
    </span>
  );
}
```

Applying `"react/jsx-curly-brace-presence": ["error", { "children": "never" }]` to this block causes
```
  5:7  error  Curly braces are unnecessary here  react/jsx-curly-brace-presence
```
This is incorrect, since the `{""}` preserves the whitespace between "braces" and "matter". Additionally, `eslint --fix` leads to
```
  return (
    <span>
      <span>The braces</span>
       matter here.
    </span>
  );
}
```
which is a visual regression: "The braces matter here." becomes "The bracesmatter here.". 

### Expected Behavior

`jsx-curly-brace-presence` should not flag curly braces that are serving a purpose, and `eslint --fix` should never introduce behavior regressions.

v7.37.5

### eslint version

v9.39.4

### node version

v25.8.0

ordinary9843:
When `{""}` or `{''}` appears in JSX children and the `children` option is set to `never`, the rule reports it as an unnecessary curly brace and auto-fixes it by removing the expression entirely. This changes the JSX tree — an empty string literal in children is a distinct node from no node — and autofixes should never introduce semantic changes.

This fix adds a guard in `shouldCheckForUnnecessaryCurly` to skip empty string literals when their parent is a JSX element (children context). The behavior for props is unchanged: `bar={""}` is still flagged and correctly auto-fixed to `bar=""`, since both forms pass an identical empty string to the prop.

Fixes #3995

### DIFF
--- a/lib/rules/jsx-curly-brace-presence.js
+++ b/lib/rules/jsx-curly-brace-presence.js
@@ -369,6 +369,17 @@ module.exports = {
         return false;
       }
 
+      // Don't flag empty string literals in JSX children — removing them changes
+      // the JSX tree and can affect whitespace rendering (autofixes must be safe).
+      if (
+        jsxUtil.isJSX(parent)
+        && node.expression
+        && node.expression.type === 'Literal'
+        && node.expression.value === ''
+      ) {
+        return false;
+      }
+
       return areRuleConditionsSatisfied(parent, config, OPTION_NEVER);
     }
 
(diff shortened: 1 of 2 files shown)


ljharb:
@ordinary9843 unfortunately by deleting the fork, this PR is unrecoverable :-(

PR #4000, opened 2026-04-07
Quality: Locale-sensitive lowercasing can break module detection

tuanaiseo:
## Problem

The rule checks `require()`/`import` source names using `pragma.toLocaleLowerCase()`. Locale-sensitive casing can produce incorrect results in some locales (for example Turkish casing), causing false negatives/positives when resolving React pragma imports.

**Severity**: `medium`
**File**: `lib/util/isDestructuredFromPragmaImport.js`

## Solution

Replace `toLocaleLowerCase()` with `toLowerCase()` for stable, locale-independent module name normalization.

## Changes

- `lib/util/isDestructuredFromPragmaImport.js` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/lib/util/isDestructuredFromPragmaImport.js
+++ b/lib/util/isDestructuredFromPragmaImport.js
@@ -58,7 +58,7 @@ module.exports = function isDestructuredFromPragmaImport(context, node, variable
           && requireExpression.callee
           && requireExpression.callee.name === 'require'
           && requireExpression.arguments[0]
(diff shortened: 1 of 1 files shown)


ljharb (changes requested):
I think this might be intentional - but either way, we’d need regression tests

ljharb:
Also, please be sure any LLM usage is for your own edification and not to generate comments and code :-)

PR #4008, opened 2026-05-11
[New] `react-in-jsx-scope`, `jsx-uses-react`: automatically disable when React >= 19 is detected

jonkoops:
React 19 mandates the automatic JSX transform, so `React` no longer needs to be in scope for JSX. This makes `react-in-jsx-scope` and `jsx-uses-react` automatically become no-ops when the configured React version is >= 19.0.0.

This follows the same `return {}` early-exit pattern already used by `no-unsafe` (which gates on >= 16.3.0).

This PR does not update the documentation to discourage extending `plugin:react/jsx-runtime`, as that config could be a natural place to add future rules (e.g. enforcing named imports, see #2585). Would appreciate the maintainers' input on whether that guidance should change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 * [`no-unused-prop-types`]: detect used props in nested components ([#3955][] @avaice)
 
 ### Changed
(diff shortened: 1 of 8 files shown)


ljharb (approved):
Thanks, this seems reasonable.

PR #4009, opened 2026-05-12
[Tests] fix CI install failures in `after_install` step

### LINKED ISSUE #4007: Test matrix: ERESOLVE on (eslint 7, parser 6) cells under strict peer-deps
Three cells in the `node-minors` and `eslint-8-` matrices flake at the
`after_install` step (`npm install --no-save eslint@7 @typescript-eslint/parser@6 babel-eslint@…`)
under `NPM_CONFIG_LEGACY_PEER_DEPS=false`:

- `node-minors`: `(16, 7, 10)` — node 16, eslint 7, babel-eslint 10
- `eslint-8-`:  `(19, 7, 8, 6)` — node 19, eslint 7, babel-eslint 8, typescript-eslint 6
- `eslint-8-`:  `(19, 7, 9, 6)` — node 19, eslint 7, babel-eslint 9, typescript-eslint 6

### Failure

```
npm warn peer eslint@"^7.32.0 || ^8.2.0" from eslint-config-airbnb-base@15.0.0
npm warn peer eslint@"^7.0.0 || ^8.0.0" from @typescript-eslint/parser@6.21.0
npm ERR! code ERESOLVE
npm ERR! Cannot destructure property 'name' of 'node' as it is undefined.
```

(The `Cannot destructure …` line is npm's own crash while it tries to print
the ERESOLVE diagnostic.)

### Root cause

Both workflows set `NPM_CONFIG_LEGACY_PEER_DEPS=false` for the affected cells:

- `node-minors`: when `node-version >= 16 && eslint >= 7`
- `eslint-8-`: when `typescript-eslint >= 6`

Under strict peer-deps, the bare `npm install` resolves with the latest
eslint matching `devDependencies.eslint` (currently `^9.7`), then the
`after_install` step tries to swap it down to `eslint@7`. The combination
of `eslint-config-airbnb-base@15.0.0` (peers `eslint@^7.32 || ^8.2`),
`@typescript-eslint/parser@6` (peers `eslint@^7 || ^8`), and other transitive
peer constraints leaves npm an unsatisfiable graph.

npm's tree resolver is nondeterministic on unsatisfiable peer graphs;
depending on cache state and resolution order, the install sometimes lands
with peer warnings and sometimes throws ERESOLVE. These cells aren't
deterministically broken — they lose the coin flip often enough to keep
showing up red.

### Cross-check against #3979 (Feb 2026) and #4006

All three are intermittent. They've been seen both green and red across
the two PRs:

- `(16, 7, 10)`: green on #3979, red on early #4006 runs, green again on the latest #4006 tip
- `(19, 7, 9, 6)`: green on #3979, red on early #4006 runs, green again on the latest #4006 tip
- `(19, 7, 8, 6)`: red on #3979 and red on early #4006 runs, green on the latest #4006 tip — longest historical red streak of the three, but still not deterministic

### Resolution paths

1. **Drop strict peer-deps for these cells.** Change the `LEGACY_PEER_DEPS`
   guard in both workflows so that `eslint <= 8` paths always use
   `LEGACY_PEER_DEPS=true`. Surgical, no source changes, accepts that npm
   will warn rather than crash.
2. **Pin `eslint-config-airbnb-base`.** It's the tightest peer constraint
   in the conflict (`^7.32 || ^8.2`). Pinning to a hypothetical newer
   version with broader peers, or replacing it with
   `eslint-config-airbnb-base-typescript`, would loosen the graph.
3. **Trim the matrix.** Eslint 7 + parser 6 is the only combo failing;
   removing those cells (or only the ones still red after option 1) is
   the smallest change.

Discovered while triaging CI on #4006.

jonkoops:
Strip `eslint-doc-generator` from `package.json` before install where it causes peer dependency conflicts. It is only used for doc generation, not for running tests.

On Node < 6, its transitive dependency `pretty-format@30.4.0` introduced `npm:` aliases unsupported by npm 5. On Node 16+ with strict peer resolution, its `eslint >= 8.57.1` peer dep conflicts with the `eslint@7` installed by `after_install`, causing npm's ERESOLVE diagnostic printer to intermittently crash.

Fixes #4007

### DIFF
--- a/.github/workflows/eslint-8-.yml
+++ b/.github/workflows/eslint-8-.yml
@@ -92,6 +92,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+      - name: strip eslint-doc-generator to avoid peer conflicts
+        if: ${{ matrix.typescript-eslint >= 6 }}
+        run: echo "$(jq 'del(.devDependencies["eslint-doc-generator"])' package.json)" > package.json
       - uses: ljharb/actions/node/install@main
         name: 'nvm install ${{ matrix.node-version }} && npm install'
         with:
--- a/.github/workflows/node-minors.yml
+++ b/.github/workflows/node-minors.yml
@@ -95,6 +95,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+      - name: strip eslint-doc-generator to avoid peer conflicts
+        if: ${{ matrix.node-version < 6 || (matrix.node-version >= 16 && matrix.eslint >= 7) }}
+        run: echo "$(jq 'del(.devDependencies["eslint-doc-generator"])' package.json)" > package.json
       - uses: ljharb/actions/node/install@main
         name: 'nvm install ${{ matrix.node-version }} && npm install'
         with:


ljharb:
I'm fine with removing eslint-doc-generator in any workflow where it's a problem - but then why would --force be needed?

jonkoops:
Fair point, I originally fixed this issue partially by using `--force`, but then it still caused failures on older Node.js versions, so I proceeded applying a separate fix for that. If you don't mind just dropping the entire dependency for all conflicting parts of the matrix I will just drop the `--force`.

ljharb:
That sounds good, thanks.

ljharb (approved):

PR #4028, opened 2026-07-31
fix: preserve comments in JSX closing bracket autofix

### LINKED ISSUE #3793: [Bug]: autofix for jsx-closing-bracket-location deletes comments

### Description Overview

When using jsx-closing-bracket-location with "after-props" setting, on autofix it deletes all comments which exist before actual non-comment prop (usually it's commented props). 

For example, if run eslint --fix on this file
image

it will remove all comments before src prop
image

Same goes, of course, for html tags which have both opening and closing tags
image

comments will be removed in this case as well

I often find myself commenting some props for testing, and this deleting bothers me. To avoid it, I need to move props here and there so that there's a non-commented line before closing bracket

### Expected Behavior

I think the expected behavior should be to make last comment be commented with "/* .... */" and place closing bracket there as shown below. Technically it will be "after props", but will not remove comments which might be needed by developer

image

If maintainers find this issue suitable for fix, I'll try to fix it myself

v7.35.0

### eslint version

v8.24.0

### node version

v16.17.1

raisulchowdhury:
## Summary

Preserve comments after the final JSX attribute when `jsx-closing-bracket-location`
autofixes a closing bracket. The fixer now uses the last comment as the layout
anchor and preserves line comments safely, with regression cases for self-
closing and paired elements.

Fixes #3793.

## Validation

- focused suite: 505 passing
- full suite: 29,267 passing
- lint and type check
- `git diff --check`

## AI assistance

AI assistance was used for investigation and drafting. I reviewed the fixer,
comment-token handling, and complete snapshot/test diff, then independently
ran the focused and full suites.

### DIFF
--- a/lib/rules/jsx-closing-bracket-location.js
+++ b/lib/rules/jsx-closing-bracket-location.js
@@ -149,7 +149,8 @@ module.exports = {
         case 'after-tag':
           return tokens.tag.line === tokens.closing.line;
         case 'after-props':
-          return tokens.lastProp.lastLine === tokens.closing.line;
+          return (tokens.lastComment ? tokens.lastComment.loc.end.line : tokens.lastProp.lastLine)
+            === tokens.closing.line;
         case 'props-aligned':
         case 'tag-aligned':
         case 'line-aligned': {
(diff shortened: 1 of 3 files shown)


haimuhaimu (commented):
One preservation case is still missing when every prop is commented out. In that case `node.attributes.length === 0`, so `getTokensLocations` never records `lastComment`; `getExpectedLocation` then selects `after-tag`, whose fixer replaces everything from `node.name.range[1]` through the closing bracket.

I reproduced this on `c6338549067b0d73735423a30ddd899a8af898f7` with ESLint 9.39.5:

```jsx
<App
  // foo={bar}
/>;
```

`verifyAndFix` returns:

```jsx
<App />;
```

The non-self-closing form similarly becomes `<App ></App>;`. This silently deletes the same kind of temporarily commented prop that the PR is intended to preserve. Could we collect comments between `node.name` and `closingToken` even when there is no `JSXAttribute`/`JSXSpreadAttribute` (or decline the autofix for that case), and add both self-closing and non-empty regression cases?

haimuhaimu (approved):
Verified the follow-up on `53a2f536667acfa5ea4c7f159986b64642730061`. Re-running the original reproducer now leaves both the self-closing and non-empty comment-only elements unchanged (`fixed: false`) instead of deleting the commented prop, while the active-prop case still performs the intended safe conversion. The new regression file is also covered by the project unit-test glob. My previous concern is resolved; all four reported checks are passing. Thanks for the focused fix.

